PR #1, opened 2021-05-25
LTG-199 - Set up and deploy simple dropwizard application

JHjava:
- Set up a simple dropwizard application which has 1 endpoint which returns `hello world`. This is so that we have an application to deploy and run in the PaaS.


ci/pipeline.yaml
```diff
@@ -0,0 +1,72 @@
...
+      paths:
+        - ci/pipeline.yaml
+      branch: main
+
+  - name: di-authentication-api-upload
+    type: cf-cli
+    icon: cloud-upload
+    source:
+      api: https://api.london.cloud.service.gov.uk
+      username: ((cf-username))
```
mrwilson, line 31: We probably need to give the user access to the `build` space in PaaS

JHjava: Good shout Alex. Let me check that!

JHjava: I've given everyone access now. Cheers for that!

PR #6, opened 2021-06-03
LTG-267: create Terraform modules

dbes-gds:
## What

Refactor Terraform into modules
Groundwork for using Localstack in the local dev environment
Localstack PR will follow

## Why

Enables localstack to have a different top-level Terraform config but still reuse the bulk of the terraform code


ci/terraform/aws/api-gateway.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_api_gateway_rest_api" "di-authentication-api" {
+  name = format("di-authentication-api-%s", var.environment)
```
chrisclayson, line 2: Should probably use string interpolation here `"di-authentication-api-${var.environment`

ci/terraform/aws/api-gateway.tf
```diff
@@ -0,0 +1,3 @@
+resource "aws_api_gateway_rest_api" "di-authentication-api" {
```
chrisclayson, line 1: I'd be tempted to modularise the gateway too.

chrisclayson (approved):

PR #14, opened 2021-06-03
LTG-268: terraform and docker for localstack

dbes-gds:
## What

Add Terraform and Docker config for Localstack

## Why

Able to run Terraform against Localstack to test API Gateway and Lambda

### DIFF
--- a/docker/localstack.Dockerfile
+++ b/docker/localstack.Dockerfile
@@ -1,0 +1,7 @@
+# localstack Dockerfile
+#
+# Container used to run tasks requiring Localstack in the build pipeline.
+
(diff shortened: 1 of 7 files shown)


docker/localstack.Dockerfile
```diff
@@ -0,0 +1,7 @@
+# localstack Dockerfile
+#
+# Container used to run tasks requiring Localstack in the build pipeline.
+
+FROM localstack/localstack:0.12.11@sha256:a403cab0daa5ccfe8f9af6067f8165e9c82a0e626211b8442163300921e7e0cb
+
+COPY localstack/* /docker-entrypoint-initaws.d/
```
chrisclayson, line 7: Don't think we need this anymore.

chrisclayson (approved):

PR #16, opened 2021-06-04
LTG-266 - Create Signup handler to run on AWS lambda

JHjava:
- Create a new handler which will handle and validate a signup request.
- If successful it will return 200 and if unsucessful it will return 400.
- Add unit tests for the handler


chrisclayson:
I think we may need to discuss this architecturally ... I'm not sure we'd be passing form data directly to this endpoint.

I'd envisage that the frontend app would receive the form `POST`s, validate them, and then call the API to create the user. In which case this handler could probably take a JSON payload!?

chrisclayson (approved):
Caveated on the fact we know this will change.

PR #17, opened 2021-06-04
LTG-257 - Deploy the authorisation endpoint to AWS Lambda 

JHjava:
- Add Terraform for authorize endpoint
- Return the authorization response from the AuthorisationHandler
- Refactor the unit test to make sure we are sending the AuthCode in the response


chrisclayson (approved):
👍🏻

PR #27, opened 2021-06-07
LTG-266 - Create Lambda for /jwks.json endpoint

JHjava:
- Some client libraries require this endpoint when integrating so we need to have this run as a lambda function.
- The signing key is currently generated on the fly in the token service but eventually this will be swapped out with something more permanent
- Add terraform for jwks lambda and api gateway
- The jwks endpoint will use the well-known base path
- Amend the api_gateway_root_id to use snake case and fix the integration tests
- When constructing the lambda function_name remove any full stops. This is because we are currently using the endpoint name and the jwks endpoint is 'jwks.json' which terraform will not like.
- Rather than constructing the URL manually for the integration tests we can use the URL generated from outputs.tf

### DIFF
--- a/ci/terraform/aws/jwks.tf
+++ b/ci/terraform/aws/jwks.tf
@@ -1,0 +1,16 @@
+module "jwks" {
+  source = "../modules/endpoint-module"
+
+  endpoint_name   = "jwks.json"
+  endpoint_method = "GET"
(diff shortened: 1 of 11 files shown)


chrisclayson (approved):
👍🏻

PR #30, opened 2021-06-07
LTG-264 - migrate signup endpoint to Lambda

JHjava:
- Change payload from form post to JSON
- Remove parameter for confirmed password (this should be done by clients before calling the API).
- Refactor password validation and tests.
- Add the terraform to create the api gateway and Lambda required for the signup endpoint


chrisclayson (approved):
👍🏻

PR #31, opened 2021-06-07
LTG-284 - Migrate dynamic client endpoint to Lambda

JHjava:
- Create a new Lambda handler which will handle POST requests when a Client is registering their software
- Remove Scopes for the Client class as it is not part of the Dynamic Client Registration spec https://openid.net/specs/openid-connect-registration-1_0.html
- Add the ability to serialize and deserialize the Client class
- Write terraform for API Gateway and Lambda for Dynamic Client reg


chrisclayson (approved):
👍🏻

PR #52, opened 2021-06-14
Ltg 313 create sqs queue

JHjava:
## What?

- Create an SQS queue module to build a queue and related policies
- Add module to localstack Terraform
- Enable SQS in localstack
- Enable SQS in AWS 
- Restrict access to the SQS queue for a specific Lambda. It is currently the userexists Lambda as the Lambdas which will talk to it haven't been built yet.

## Why?

- So it can be used by the Lambda to put messages on the queue for when a user doesn't have an account. Another Lambda will read from the queue and be responsible for using Notify to send an email.


chrisclayson (approved):
👍🏻

PR #55, opened 2021-06-15
LTG-228: Read from SQS

chrisclayson:
## What?

Add a lambda function to the Terraform SQS module that will process this module.
Only this lambda has permissions to receive messages from the queue.
Create a basic Lambda function, that will simply send an e-mail verification e-mail, this will be iterated.

## Why?

We have a queue and we need something to process its messages.

### DIFF
--- a/ci/terraform/modules/sqs-queue/variables.tf
+++ b/ci/terraform/modules/sqs-queue/variables.tf
@@ -6,6 +6,60 @@ variable "environment" {
   type    = string
 }
 
-variable "principals_arns" {
+variable "sender_principal_arns" {
   type = list(string)
+}
+
+variable "lambda_name" {
+  type = string
+}
+
+variable "handler_runtime" {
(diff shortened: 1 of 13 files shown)


ci/terraform/modules/sqs-queue/variables.tf
```diff
@@ -6,6 +6,60 @@ variable "environment" {
...
+  type = string
+  description = "The id of the security for the lambda"
+}
+
+variable "subnet_id" {
+  type = list(string)
+  description = "The id of the subnets for the lambda"
+}
+
+variable "queue_arn" {
```
JHjava, line 63: Not sure we need this anymore, I can't see it being used anywhere

chrisclayson: Good spot!

JHjava (approved):

PR #58, opened 2021-06-16
LTG-313 - Make the SQS_ENDPOINT config Optional.ofNullable

JHjava:
## What?

- We only need the SQS_ENDPOINT set if we are using localstack so it is fine to be null when running in AWS.

## Why?

- So we don't get a null pointer when running in AWS


chrisclayson (approved):
Doh!

PR #60, opened 2021-06-16
LTG-313: Add VPC endpoint for SQS

chrisclayson:
## What?

- Add SQS as an endpoint to our VPC.

## Why?

Connection to SQS is timing out.

## Related PRs

#57


JHjava (approved):
🚀

PR #62, opened 2021-06-16
LTG-35: scope on auth redirect

dbes-gds:
## What?

Add scope to /authorize redirect.
Refactoring of existing integration test to reduce duplication of variables.
Integration test not provided as we don't currently have a local Redis which is making the test fail.

## Why?

So that the scope can be stored in a client session.


chrisclayson (approved):
👍🏻

PR #72, opened 2021-06-18
BAU: Improve AWS response matchers

mrwilson:
## What?

Creates a general and extendable matcher for `APIGatewayProxyResponseEvent` and adds `hasBody` as another matcher for tests.


alan-gds (approved):
Looks good to me.

Perhaps there is a way to add a `bodyContains` mapper too, to cover the various other assertions where we're still calling `result.getBody`?  It might require some messier code in the `matchesSafely` method though.

chrisclayson (approved):

mrwilson:
@alan-gds That's a good shout - we could probably switch the expected inside the class from being a `T` to being a `Matcher<T>` and have the existing matchers pass down an `is(...)` rather than the raw value.

If you want to raise a PR for that, go for it! :)

PR #104, opened 2021-06-25
LTG-374 - Retrieve email code expiry time from config

JHjava:
## What?

-  Retrieve email code expiry time from config and give it a default value of 15 minutes if the env var is not set
- We may want to make this config value more generic if we want to reuse the value for others codes as well. 

## Why?

- So it can be dynamic

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/services/ConfigurationService.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/services/ConfigurationService.java
@@ -39,6 +39,10 @@ public class ConfigurationService {
         return System.getenv("NOTIFY_API_KEY");
     }
 
+    public long getEmailCodeExpiry() {
+        return Long.parseLong(System.getenv().getOrDefault("EMAIL_CODE_EXPIRY", "900"));
+    }
+
     public String getNotificationTemplateId(NotificationType notificationType) {
         switch (notificationType) {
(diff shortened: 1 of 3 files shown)


serverless/lambda/src/main/java/uk/gov/di/services/ConfigurationService.java
```diff
@@ -39,6 +39,10 @@ public String getNotifyApiKey() {
         return System.getenv("NOTIFY_API_KEY");
     }
 
+    public long getEmailCodeExpiry() {
+        return Long.parseLong(System.getenv().getOrDefault("EMAIL_CODE_EXPIRY", "900"));
```
sanisc, line 43: are we sure we want a default here shouldn't it blow up if it is not present. Reason being if some looks at the repo they can work out the length of email code which is a possible security risk

JHjava: I could remove the default but the environment variables are currently set in this repo within the terraform anyway so it won't solve the problem of people being able to view it.

sanisc: lets chat with the team about this for the time lets leave it how the others are set

sanisc (approved):

PR #108, opened 2021-06-28
LTG-374 - Delete email code after it has been used

JHjava:
## What?

-  Delete email code after it has been used
- The code is a one time code so once it has been used successfully we should delete it from redis.
- If the code fails to delete we don't want to the user to fail as the code will disappear after a certain time period but we should be aware when it does fail to delete so log out a message for now.
- Add an integration test to ensure that a code can't be used again by a user
- Refactor the VerifyCodeIntegrationTest to reduce a lot of duplication

## Why?

- The code is a one time code so once it has been used successfully we should delete it from redis.

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/services/CodeStorageService.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/services/CodeStorageService.java
@@ -1,10 +1,14 @@
 package uk.gov.di.services;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import uk.gov.di.helpers.HashHelper;
 
 import java.util.Optional;
 
 public class CodeStorageService {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CodeStorageService.class);
 
     private final RedisConnectionService redisConnectionService;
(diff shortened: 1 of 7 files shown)


serverless/lambda/src/main/java/uk/gov/di/services/CodeStorageService.java
```diff
@@ -28,4 +32,14 @@ public void saveEmailCode(String email, String code, long codeExpiryTime) {
                 redisConnectionService.getValue(
                         EMAIL_KEY_PREFIX + HashHelper.hashSha256String(emailAddress)));
     }
+
+    public void deleteCodeForEmail(String emailAddress) {
```
mrwilson, line 36: Worth popping a unit test on this method? To check it calls through to redis.

JHjava: Yep, I'll do that. Good shout!

JHjava: I've updated and also added a check in the VerifyCodeRequestHandler happy path test that the deleteCodeForEmail method is called.

mrwilson (approved):

PR #110, opened 2021-06-28
LTG-374 - Standarise API error responses

JHjava:
## What?

- Create a ErrorResponse ENUM which will contain a list of the various error codes and a messages
- In the rare case that the ErrorResponse is unable to be desearlized, log out the message and generate a 500 response with an empty body
- Migrate the rest of the errors where we are currently returning a string from Messages.java to a ErrorResponse object
- When the code is invalid generate a 200 response with an updated state. 

## Why?

- So we have a standard way of sending error responses. 
- Use a 200 response when the code is invalid as the rest of the request was OK. Identify the code was invalid by updating the state.


mrwilson:
In principle this looks good. In regards to the JSONProcessingException, I'm less worried about serialising when we have total control of the payload - for example, where we create an `ErrorResponse` (ours) with an `ApiError` (also ours).

A way to do this in code is perhaps to have a static generator for `ErrorResponse`s that does a try-catch which in case of a serialisation error logs it out and returns a 500. This, in practice, should almost never happen and I think being able to respond to it rather than prevent it is more important.

serverless/lambda/src/main/java/uk/gov/di/helpers/ApiGatewayResponseHelper.java
```diff
@@ -3,13 +3,29 @@
...
+    }
+
+    public static <T> APIGatewayProxyResponseEvent generateApiGatewayProxyErrorResponse(
+            int statusCode, ErrorResponse errorResponse) {
+        try {
+            return generateApiGatewayProxyResponse(
+                    statusCode, new ObjectMapper().writeValueAsString(errorResponse));
+        } catch (JsonProcessingException e) {
+            LOGGER.warn("Unable to generateApiGatewayProxyErrorResponse: " + e);
+            return generateApiGatewayProxyResponse(500, "");
```
sanisc, line 27: do we want to add something to the empty string in the request or can you just pass in null?

sanisc (approved):

PR #111, opened 2021-06-28
BAU: upsize Redis

dbes-gds:
## What?

Upsize Redis.

## Why?

Seeing intermittent connection errors so seeing if this is a resource issue.


chrisclayson (approved):
👍🏻

PR #112, opened 2021-06-28
LTG-374 - Return validation errors as a standard ErrorResponse

JHjava:
## What?

- Add the relevant validation errors to the ErrorResponse so validation errors will be generated the same as all other errors
- Make the validateEmail and validatePassword methods return an Optional<ErrorResponse>
- Refactor other error responses to use the generateApiGatewayProxyErrorResponse method.
- Delete the PasswordValidation and EmailValidation enums as these errors are no in ErrorResponse

## Why?

- So we have a standard way of returning errors to the client

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
@@ -1,10 +1,8 @@
 package uk.gov.di.services;
 
-import uk.gov.di.validation.EmailValidation;
-import uk.gov.di.validation.PasswordValidation;
+import uk.gov.di.entity.ErrorResponse;
 
-import java.util.EnumSet;
-import java.util.Set;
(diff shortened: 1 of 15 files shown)


serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
```diff
@@ -1,41 +1,35 @@
...
         }
         if (!email.isBlank() && !EMAIL_REGEX.matcher(email).matches()) {
-            emailErrors.add(EmailValidation.INCORRECT_FORMAT);
+            return Optional.of(ErrorResponse.ERROR_1004);
         }
-        return emailErrors;
+        return Optional.empty();
     }
 
-    public Set<PasswordValidation> validatePassword(String password) {
```
mrwilson, line 26: Lovely!

mrwilson (approved):

PR #125, opened 2021-07-01
LTG-403 - Hook up Login lambda with DynamoDB

JHjava:
## What?

- Switch the Login lambda to use the DynamoService rather than the in memory UserService.
- Check if a user has an account before attempting to login 

## Why?

- Because we need to use a database than rely on in-memory storage 
-  So we can send back a more accurate error message


chrisclayson (approved):
👍🏻

PR #130, opened 2021-07-02
LTG-385 -  Adapt send notification lambda to handle phone number 

JHjava:
## What?

- Change the SendNotification lambda so that is a request is received with the VERIFY_PHONE_NUMBER notification type it can handle it as a phone number rather than an email. 
- Add a method to the CodeStorageService to store a hash of the phone number with a phone-number-code: prefix against a code which is generated
- Add a VERIFY_PHONE_NUMBER type to the NotificationType list so it can be used in the request when a user enters there phone number. This will allow the Lambda to differentiate between the 2 types and ensure the code is stored against the right type
- Add an optional PhoneNumber param to the SendNotificationRequest type. The send-notification lambda is used in multiple scenarios such as enter-email and enter-phone-number. We need the phone number which is entered by the user so we can use Notify to generate a message with the relevant code. Having it as an optional parameter means it is not required when the enter-email request is called.

## Why?

- Because the SendNotification Lambda will also handle requests when a user enters a phone number. We need to modify it so it can caters for multiple scenarios.


serverless/lambda/src/main/java/uk/gov/di/lambdas/SendNotificationHandler.java
```diff
@@ -114,7 +99,53 @@ public APIGatewayProxyResponseEvent handleRequest(
         }
     }
 
+    private APIGatewayProxyResponseEvent handleVerifyPhoneNumberRequest(
```
sanisc, line 102: might be possible to merge the following 2 methods

serverless/lambda/src/main/java/uk/gov/di/lambdas/SendNotificationHandler.java
```diff
@@ -114,7 +99,53 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            SendNotificationRequest sendNotificationRequest, Session session)
+            throws JsonProcessingException {
+        if (sendNotificationRequest.getPhoneNumber() == null) {
```
sanisc, line 105: do we want to validate the phone number here https://github.com/google/libphonenumber - otherwise it is going to just go onto the queue and fail x number of times till it dead letters?

JHjava: That is handled in a separate ticket so was planning on leaving that until LTG-409.

serverless/lambda/src/main/java/uk/gov/di/lambdas/SendNotificationHandler.java
JHjava: @sanisc Fair point. I've updated!

sanisc (approved):

PR #132, opened 2021-07-02
LTG-409 - Validate phone number

JHjava:
## What?

- We must check that any phone number is a valid British number before we send it over to Notify.
- We can use the https://github.com/google/libphonenumber lib to do some checking but before we use it, explicitly check that there is at least 10 characters that consist of all numbers.
- Notify strip any whitespace when validating the number.
- When a phone number is invalid we will send the appropriate Error response back.

## Why?

- To ensure that anything we put on the SQS queue is a valid number. This will ensure we only ever send valid numbers to Notify.

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
@@ -1,5 +1,7 @@
 package uk.gov.di.services;
 
+import com.google.i18n.phonenumbers.NumberParseException;
+import com.google.i18n.phonenumbers.PhoneNumberUtil;
 import uk.gov.di.entity.ErrorResponse;
 
 import java.util.Optional;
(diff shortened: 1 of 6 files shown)


serverless/lambda/src/main/java/uk/gov/di/services/ValidationService.java
```diff
@@ -32,4 +34,21 @@
         }
         return Optional.empty();
     }
+
+    public Optional<ErrorResponse> validatePhoneNumber(String phoneNumberInput) {
+        phoneNumberInput = phoneNumberInput.replaceAll("\\s+", "");
+        if ((!phoneNumberInput.matches("[0-9]+")) || (phoneNumberInput.length() < 10)) {
```
dbes-gds, line 40: <= 10 ?

JHjava: Can't some UK numbers be 10? Tracy mentioned that on the content doc

dbes-gds (dismissed):

dbes-gds (approved):

PR #143, opened 2021-07-06
GOVSI-97: Phone Code Retry Counter

chrisclayson:
## What?

- Add a `retry_counter` field to the session
- Add a max retries value to configuration
- Introduce new state for max retries reached (`PHONE_NUMBER_CODE_MAX_RETRIES_REACHED`)
- When user reaches max retries, add block to Redis with expiry (configurable)
- If user subsequently tries again, while block in place, return `PHONE_NUMBER_CODE_MAX_RETRIES_REACHED`

## Why?

We want to stop people spamming the service with login/sign-up attempts if they enter the wrong code multiple times.


JHjava (approved):
🚀

PR #153, opened 2021-07-08
GOVSI-469 - Validate that an email code can be entered wrong only 5 times

JHjava:
## What?

- Validate that an email code can be entered wrong only 5 times
- Update Redis to record whether a user is blocked or not. Redis will remove the entry after 15 minutes.
- Make the CodeMaxRetries count generic as it is currently the same for both phone code and email code.

## Why?
- So if a user attempts to use a code more than 5 times then block that user in that session from entering codes for 15 minutes.


chrisclayson (approved):
👍🏻

PR #168, opened 2021-07-15
GOVSI-445 - Add PrivateKey JWT auth on the token endpoint

JHjava:
## What?

- Add PrivateKeyAuthentication to the token endpoint and use Nimbus libraries to Verify that the PrivateKeyJWT contains the required attributes.
- Construct the Base URL in terraform
- Return ErrorResponses from Token endpoint
- Retrieve the Public Key from Dynamo to then Verify the signature of the Private Key JWT

## Why?

- Private Key JWT Client authentication is used by clients to authenticate to the authorization server when using the token endpoint
- We need the application to read the BaseURL so it can validate the audience

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/lambdas/TokenHandler.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/lambdas/TokenHandler.java
@@ -4,18 +4,9 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
-import com.nimbusds.jose.JOSEException;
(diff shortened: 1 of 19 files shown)


serverless/lambda/src/main/java/uk/gov/di/lambdas/TokenHandler.java
```diff
@@ -67,33 +69,50 @@ public TokenHandler() {
...
-        //        String email = authorizationCodeService.getEmailForCode(code);
+        boolean privateKeyJWTIsValid =
+                tokenService.validatePrivateKeyJWTSignature(
+                        client.get().getPublicKey(),
+                        privateKeyJWT,
+                        configurationService.getBaseURL().orElseThrow());
+        if (!privateKeyJWTIsValid) {
+            return generateApiGatewayProxyErrorResponse(403, ErrorResponse.ERROR_1015);
+        }
+        //        TODO: String email = authorizationCodeService.getEmailForCode(code);
```
chrisclayson, line 93: Is this comment deliberately left in?

chrisclayson (approved):
👍🏻

PR #182, opened 2021-07-21
BAU: Refactor out `RequestHelper` in integration tests

mrwilson:
## What?

Create a helper for making API requests with a sessionID.

## Why?

Removes duplication across the integration tests

One-off requests or requests with no session have not been changed.


JHjava (approved):
👍

PR #195, opened 2021-07-26
GOVSI-523 - Add validation and logic to logout endpoint

JHjava:
## What?

- Extract the session from the session cookie when it exists. If it doesn't then we just redirect the user to the default logout URI which is `https://di-authentication-frontend.london.cloudapps.digital/signed-out`
- Move the logic of the cookie parsing into its own helper 
- Validate the ID token in the request. 
- Put a placeholder in for validating the signature of the ID token. We can revisit this when we're not signing the ID token with a key generated on the fly. 
- Validate the client logout_uri
- Send the state back in the response when it is included in the request

## Why?
- This PR implements the logic from the RP initiated logout OIDC spec - https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout
- This creates an endpoint which can be called by a user whilst they are on the RP and allow them to logout by deleting the session. If we can successfully validate the logout URI in the request we will redirect the user to that location otherwise to the default logout URI on the frontend.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
@@ -74,7 +74,8 @@ public class LogoutIntegrationTest extends IntegrationTestEndpoints {
                 response.getHeaders()
                         .get("Location")
                         .contains(
-                                "https://di-auth-stub-relying-party-build.london.cloudapps.digital/"));
+                                "https://di-auth-stub-relying-party-build.london.cloudapps.digital/?state="
+                                        + "8VAVNSxHO1HwiNDhwchQKdd7eOUK3ltKfQzwPDxu9LU"));
     }
 
     private AuthorizationRequest generateAuthRequest() {
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
@@ -7,34 +7,52 @@ import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.id.Subject;
 import jakarta.ws.rs.client.Client;
 import jakarta.ws.rs.client.ClientBuilder;
+import jakarta.ws.rs.core.MultivaluedHashMap;
+import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.helpers.RedisHelper;
 import uk.gov.di.helpers.IDTokenGenerator;
 
+import java.io.IOException;
 import java.util.UUID;
 
+import static java.lang.String.format;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 public class LogoutIntegrationTest extends IntegrationTestEndpoints {
 
     private static final String LOGOUT_ENDPOINT = "/logout";
+    private static final String SET_COOKIE = "Set-Cookie";
 
     @Test
-    public void shouldReturn302AndRedirectToDefaultLogoutUri() throws JOSEException {
+    public void shouldReturn302AndRedirectToDefaultLogoutUri() throws JOSEException, IOException {
+        String sessionId = "session-id";
+        String clientSessionId = "client-session-id";
+        RedisHelper.createSession(sessionId, clientSessionId);
         RSAKey signingKey =
                 new RSAKeyGenerator(2048).keyID(UUID.randomUUID().toString()).generate();
         SignedJWT signedJWT =
                 IDTokenGenerator.generateIDToken(
                         "client-id", new Subject(), "http://localhost/issuer", signingKey);
         Client client = ClientBuilder.newClient();
+        MultivaluedMap<String, Object> headers = new MultivaluedHashMap<>();
+        headers.add(SET_COOKIE, buildCookieString(sessionId, clientSessionId));
         Response response =
                 client.target(ROOT_RESOURCE_URL + LOGOUT_ENDPOINT)
                         .queryParam("id_token_hint", signedJWT.serialize())
                         .queryParam("post_logout_redirect_uri", "http://localhost/redirect")
                         .queryParam("state", "8VAVNSxHO1HwiNDhwchQKdd7eOUK3ltKfQzwPDxu9LU")
                         .request()
+                        .headers(headers)
                         .get();
 
         assertEquals(302, response.getStatus());
+    }
+
+    private String buildCookieString(String sessionID, String clientSessionID) {
+        return format(
+                "%s=%s.%s; Max-Age=%d; %s",
+                "gs", sessionID, clientSessionID, 1800, "Secure; HttpOnly;");
     }
 }
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
@@ -15,6 +15,7 @@ import jakarta.ws.rs.core.MultivaluedHashMap;
 import jakarta.ws.rs.core.MultivaluedMap;
 import jakarta.ws.rs.core.Response;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.helpers.DynamoHelper;
 import uk.gov.di.authentication.helpers.RedisHelper;
 import uk.gov.di.helpers.IDTokenGenerator;
 
@@ -23,7 +24,9 @@ import java.net.URI;
 import java.util.UUID;
 
 import static java.lang.String.format;
+import static java.util.Collections.singletonList;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 public class LogoutIntegrationTest extends IntegrationTestEndpoints {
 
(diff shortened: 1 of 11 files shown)


serverless/lambda/src/main/java/uk/gov/di/helpers/CookieHelper.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class CookieHelper {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(CookieHelper.class);
+
+    public static final String REQUEST_COOKIE_HEADER = "Cookie";
+
+    public static Optional<SessionCookieIds> parseSessionCookie(Map<String, String> headers) {
+        if (headers == null
+                || headers.isEmpty()
```
mrwilson, line 19: We could probably change this check to just a null check, then move line 27 (`headers.getOrDefault...`) up here and have a second conditional for `cookies.isEmpty()/isBlank()`

The `getOrDefault` will step around the need for checking an empty collection, checking for the key presence, etc.

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
```diff
@@ -4,29 +4,164 @@
...
+        }
+        try {
+            String idTokenHint = queryStringParameters.get("id_token_hint");
+            SignedJWT idToken = SignedJWT.parse(idTokenHint);
+            Optional<String> audience =
+                    idToken.getJWTClaimsSet().getAudience().stream().findFirst();
+            sessionService.deleteSessionFromRedis(session.getSessionId());
+            return audience.map(
+                            t -> {
+                                final ClientRegistry clientRegistry;
```
mrwilson, line 99: Can we pull this lambda out as a method instead? It feels a bit too large to inline.

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
```diff
@@ -4,29 +4,164 @@
...
+                                return generateLogoutResponse(URI.create(logoutURI), state);
+                            })
+                    .orElse(generateDefaultLogoutResponse(state));
+        } catch (ParseException e) {
+            throw new RuntimeException();
+        }
+    }
+
+    private boolean isIDTokenSignatureValid(String idTokenHint, String sessionID) {
+        return true;
```
mrwilson, line 124: Is this unimplemented for now?

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
```diff
@@ -4,29 +4,164 @@
...
+    }
+
+    private boolean doesIDTokenExistInSession(String idTokenHint, Session session) {
+        for (Map.Entry<String, ClientSession> t : session.getClientSessions().entrySet()) {
```
mrwilson, line 128: As we only use the values off the Entry, rather than both key and value, perhaps could be refactored into something like `sessions.getClientSessions().values().stream().[...].anyMatch(..)`

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
```diff
@@ -4,29 +4,164 @@
...
+    }
+
+    private APIGatewayProxyResponseEvent generateDefaultLogoutResponse(Optional<String> state) {
+        return generateLogoutResponse(configurationService.getDefaultLogoutURI(), state);
+    }
+
+    private APIGatewayProxyResponseEvent generateLogoutResponse(
+            URI logoutUri, Optional<String> state) {
+        URIBuilder uriBuilder = new URIBuilder(logoutUri);
+        state.map(s -> uriBuilder.addParameter("state", s));
```
mrwilson, line 156:
```suggestion
        state.ifPresent(s -> uriBuilder.addParameter("state", s));
```

We don't care about the return value since we're just mutating the uriBuilder off the optional if it's there

integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
```diff
@@ -4,37 +4,92 @@
...
+        State state = new State();
+        return new AuthorizationRequest.Builder(responseType, new ClientID("test-client"))
+                .redirectionURI(URI.create("http://localhost:8080/redirect"))
+                .state(state)
+                .build();
+    }
+
+    private String buildCookieString(String sessionID, String clientSessionID) {
+        return format(
+                "%s=%s.%s; Max-Age=%d; %s",
```
mrwilson, line 92: This could perhaps be replaced with something from the standard library like:

```
var cookie = new HttpCookie("gs", "whatever-the-value-is");
cookie.setSecure(true);
cookie.setMaxAge(1800);
cookie.setHttpOnly(true);

return cookie.toString();
```

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
```diff
@@ -4,29 +4,164 @@
...
+            return audience.map(
+                            t -> {
+                                final ClientRegistry clientRegistry;
+                                try {
+                                    clientRegistry =
+                                            dynamoClientService
+                                                    .getClient(t)
+                                                    .orElseThrow(
+                                                            () -> new ClientNotFoundException(t));
+                                } catch (ClientNotFoundException e) {
```
mrwilson, line 106: We're throwing and catching immediately to rethrow, can we collapse this and throw the `RuntimeException` straight from the `orElseThrow`?

integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
chrisclayson: I was thinking similar last week when I saw how the cookie was being generated in the handler.

serverless/lambda/src/main/java/uk/gov/di/helpers/CookieHelper.java
JHjava: I've refactored this class to use the HttpCookie class to parse the cookie instead of Regex.

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
JHjava: Done

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
JHjava: Yep that is a lot neater. I have updated.

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
JHjava: Good shout. I have updated.

integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
JHjava: This has been updated as well as the CookieHelper class. I haven't touched how the cookie is being generated for now.

serverless/lambda/src/main/java/uk/gov/di/lambdas/LogoutHandler.java
JHjava: Done.

integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
```diff
@@ -88,8 +88,7 @@ private AuthorizationRequest generateAuthRequest() {
     }
 
     private String buildCookieString(String sessionID, String clientSessionID) {
-        return format(
-                "%s=%s.%s; Max-Age=%d; %s",
-                "gs", sessionID, clientSessionID, 1800, "Secure; HttpOnly;");
+        var cookie = new HttpCookie("gs", sessionID + "." + clientSessionID);
+        return cookie.toString();
```
mrwilson, line 92: Looks good. I thought we were going to keep this one as is because `cookie.toString()` doesn't add e.g. Max-Age?

serverless/lambda/src/main/java/uk/gov/di/helpers/CookieHelper.java
```diff
@@ -16,38 +15,36 @@
...
-
-        final String COOKIE_REGEX = "gs=(?<sid>[^.;]+)\\.(?<csid>[^.;]+);";
-
-        String cookies = headers.getOrDefault(REQUEST_COOKIE_HEADER, "");
-
+        String cookies = headers.get(REQUEST_COOKIE_HEADER);
         LOGGER.debug("Session Cookie: {}", cookies);
+        HttpCookie httpCookie;
+        try {
+            httpCookie = HttpCookie.parse(cookies).stream().findFirst().get();
```
mrwilson, line 26: Thinking about it, what are the chances of us getting multiple cookies on a single request? We might want a `filter(cookie -> "gs".equals(cookie.getName()))` or something

integration-tests/src/test/java/uk/gov/di/authentication/api/LogoutIntegrationTest.java
JHjava: I think this is fine to be cookie.toString as it will replicate a client sending a request to a server which wouldn't included max age

serverless/lambda/src/main/java/uk/gov/di/helpers/CookieHelper.java
JHjava: Hmm good point

mrwilson (approved):

PR #200, opened 2021-07-28
BAU - generate access token with correct claims

sanisc:
## What?

Change how the access token is generated.

## Why?
The access token should  contain certain claims which allow the token to be validated as well as provide the necessary info for the /userinfo call.


JHjava (approved):

alan-gds (approved):
LGTM

PR #211, opened 2021-07-29
BAU: Fix integration test random fails

chrisclayson:
## What?

- Ensure `pre-commit.sh` exits with the correct code so that GitHub actions reports the failures correctly.
- Update `LOGIN_URL` in `pre-commit.sh` is correct.
- Stop the Jersey client from following re-directs when 
- When running integration tests ensure that data in Dynamo and Redis is flushed by adding a `@BeforeEach` to the base class (note this will always be executed before any `@BeforeEach` methods in the sub-class ... according to the JUnit docs)
- Setup VerifyCode data correctly before running test (we weren't stubbing this data, so depending on the order tests are run in, we were getting failures).

## Why?

We were getting random failures of several integration tests, which were not being reported correctly. This PR fixes the issue with reporting the errors and, hopefully, will stop the randomness.

### DIFF
--- a/integration-tests/build.gradle
+++ b/integration-tests/build.gradle
@@ -20,6 +20,8 @@ dependencies {
             'org.eclipse.jetty:jetty-server:11.0.6',
             "software.amazon.awssdk:sqs:2.16.88",
             "org.awaitility:awaitility:4.1.0",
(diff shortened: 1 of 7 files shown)


dbes-gds (approved):
runs fine locally :)

PR #214, opened 2021-07-30
BAU - Tidy up our build.gradles

JHjava:
## What?

- Create variables for dependencies that share that same version 
- Remove a lot of stuff we don't need on the root build.gradle

## Why?

- So it is clearer

### DIFF
--- a/serverless/lambda/build.gradle
+++ b/serverless/lambda/build.gradle
@@ -9,6 +9,10 @@ repositories {
     mavenCentral()
 }
 
+def dependencyVersions = [
+    jackson_version: '2.12.4',
+]
+
 dependencies {
     implementation "com.amazonaws:aws-lambda-java-core:1.2.1",
             "com.amazonaws:aws-lambda-java-events:3.9.0",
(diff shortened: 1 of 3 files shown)


serverless/lambda/build.gradle
```diff
@@ -9,17 +9,21 @@ repositories {
...
+]
+
 dependencies {
     implementation "com.amazonaws:aws-lambda-java-core:1.2.1",
             "com.amazonaws:aws-lambda-java-events:3.9.0",
             "com.amazonaws:aws-java-sdk-dynamodb:1.12.35",
             "software.amazon.awssdk:sqs:2.16.88",
             "com.nimbusds:oauth2-oidc-sdk:9.10.2",
             "com.nimbusds:nimbus-jose-jwt:9.11.2",
-            "com.fasterxml.jackson.core:jackson-core:2.12.3",
```
mrwilson, line 19: This will upgrade these three from `2.12.3` to `2.12.4` - I assume that's probably alright, given it's a patch version increase rather than minor or major.

mrwilson (approved):

serverless/lambda/build.gradle
JHjava: Yep there's a dependabot for that so I figured I would do this whilst I was here

PR #217, opened 2021-07-30
GOVSI-495: Support 'prompt' parameter

dbes-gds:
## What?

Support 'prompt' parameter.

https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

Add test scenarios for prompt = none, login, consent and select_account
Update AuthorisationHandler to support the prompt parameters

We do not currently support consent and select_account so return an error if these parameters are supplied.

## Why?

Need to support 'prompt' as part of SSO, so that users can login to a second service without being prompted for their credentials a second time.


alan-gds (approved):
We should add an integration test to demonstrate the behaviour when the value for the prompt parameter is 'login' and the user is already authenticated.  That can be a subsequent PR though; we need to get this one in before we end up with yet more merge conflicts.

PR #218, opened 2021-07-30
GOVSI-500: Protect user journey transitions with state machine

mrwilson:
## What?

This PR adds a state-machine capability and implements transition protection on one of the user account registration endpoints.

## Why?

We want to ensure that users, whether accidentally or deliberately, cannot jump between parts of the journey and circumvent (or skip entirely) steps.

# Outstanding questions

- Is `SessionState` the right representation of the state of a user, or will it be more closely aligned with what we eventually send to the audit processing service?
- Do we need to provide another endpoint that can rely the current user state to the frontend?
- Is throwing a general error ("Invalid transition in user journey") sufficient or should it be tailored to the endpoint and previous state?

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/SignupIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/SignupIntegrationTest.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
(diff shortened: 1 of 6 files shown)


dbes-gds:
WRT your questions I think we should discuss what happens in the UI and how it gets hold of the state.  Do we want to send the user back to the place implied by the current state, or show an error page?  If the former we'll need the client to know somehow where to go.

chrisclayson (approved):

mrwilson:
We had a catch-up within the team and decided that this is good enough to built on as a first cut, the corresponding front-end implications will be implemented on that repo.

I'll be mapping out the whole state transition diagram of the application and circulating it for feedback with the team.

PR #232, opened 2021-08-02
GOVSI-651: Support multi-case cookie headers

mrwilson:
## What?

This enables `CookieHelper` to parse out cookies from either `Cookie` or `cookie` headers.

## Why?

AWS API Gateway sends lower-case headers which was breaking some flows.


sanisc (approved):
LGTM

PR #252, opened 2021-08-05
GOVSI-670: Implement state transition checks (1)

mrwilson:
## What?

Implement state transition protection for `CheckUserExistsHandler` and `VerifyCodeHandler` (email code only)

## Why?

This prevents the application jumping around the process out-of-order.

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/lambdas/VerifyCodeHandler.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/lambdas/VerifyCodeHandler.java
@@ -28,6 +28,7 @@ import static uk.gov.di.entity.SessionState.PHONE_NUMBER_CODE_MAX_RETRIES_REACHE
 import static uk.gov.di.entity.SessionState.PHONE_NUMBER_CODE_VERIFIED;
 import static uk.gov.di.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
 import static uk.gov.di.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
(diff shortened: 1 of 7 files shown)


serverless/lambda/src/main/java/uk/gov/di/lambdas/VerifyCodeHandler.java
```diff
@@ -81,20 +82,32 @@ public APIGatewayProxyResponseEvent handleRequest(
                 case VERIFY_EMAIL:
                     if (codeStorageService.isCodeBlockedForSession(
                             session.get().getEmailAddress(), session.get().getSessionId())) {
+
+                        if (isInvalidUserJourneyTransition(
+                                session.get().getState(), EMAIL_CODE_MAX_RETRIES_REACHED)) {
```
sanisc, line 87: So this will break the front end which is fine, but currently a generic status is handled for when a code is reached its max retries do we want to reimplement that?

mrwilson: As I understand, if the user legitimately reaches max retries reached (i.e. it can only come from the previous state of email code invalid) then it will "fall through" and still return the generic 200 with status code, and only drop out if it somehow jumps to max retries reached which, in practice, I don't think is possible because you need to attempt at least one code entry first?

Is that what you're asking?

sanisc: my bad i think i got confused with what the front end did i am going to close this convo.

sanisc (approved):

PR #266, opened 2021-08-10
GOVSI-277 - Create update client config endpoint

JHjava:
## What?

- Create an endpoint which can be called via /oidc/clients/{client_id} where client_id is a path parameter. They can update any number of parameters at any one time as long as they pass in a valid client_id. We will then return the entire updated client config in the response
- Change the lambda terraform to add a path_part variable.
- Add integration and unit tests

## Why?

- Clients will need to update their config. 
- The endpoint_name in each of the lambdas terraform was used for the aws_function_name and the path_part for the aws_api_gateway_resource. As we want the final path part of this lambda to be a path parameter we don't want to reuse the endpoint name for the aws_api_gateway_resource. Create a new path_part variable and configure it in each lambda.

### DIFF
--- a/ci/terraform/aws/api-gateway.tf
+++ b/ci/terraform/aws/api-gateway.tf
@@ -80,6 +80,18 @@ resource "aws_api_gateway_resource" "connect_resource" {
   path_part   = "connect"
 }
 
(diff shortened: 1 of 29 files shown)


chrisclayson (approved):
👍🏻

PR #273, opened 2021-08-11
GOVSI-461 - Validate the Authn Request on the AuthorizationHandler

JHjava:
## What?

- Send back the Errors as per the OIDC spec for when any of the validation fails.
- Move the validation into the AuthorizationService
- Add tests for the validation method

## Why?

- So we are compliant with the OIDC spec

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/services/AuthorizationService.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/services/AuthorizationService.java
@@ -2,17 +2,27 @@ package uk.gov.di.services;
 
 import com.nimbusds.oauth2.sdk.AuthorizationCode;
 import com.nimbusds.oauth2.sdk.AuthorizationRequest;
+import com.nimbusds.oauth2.sdk.ErrorObject;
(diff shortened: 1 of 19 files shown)


serverless/lambda/src/main/java/uk/gov/di/services/AuthorizationService.java
```diff
@@ -46,4 +56,42 @@ public AuthenticationSuccessResponse generateSuccessfulAuthResponse(
...
+        }
+        if (!client.get().getRedirectUrls().contains(authRequest.getRedirectionURI().toString())) {
+            LOGGER.error(
+                    "Invalid Redirect URI for Client {}. Redirect URI in request {}",
+                    client.get().getClientID(),
+                    authRequest.getRedirectionURI().toString());
+            throw new RuntimeException(
+                    format(
+                            "Invalid Redirect URI for Client %s. Redirect URI in request %s",
+                            client.get().getClientID(),
```
sanisc, line 75: I don't think we need to tell the user the client id here

sanisc (dismissed):

PR #278, opened 2021-08-12
GOVSI-463 - Use OIDC errors for OIDC endpoints

JHjava:
## What?

- Make the Token endpoint errors compliant with the OIDC spec by sending back errors stated in the spec 
- Remove custom errors we no longer need 
- Only send back the description and the code when returning OIDC errors 
- Send back an OIDC error response if any parameters are missing on the authorize request. (If the redirect_uri parameter is missing then we should throw as we shouldn't send errors to an unvalidated redirect URI)

## Why?

- To be more OIDC compliant than before

### DIFF
--- a/serverless/lambda/src/main/java/uk/gov/di/lambdas/AuthorisationHandler.java
+++ b/serverless/lambda/src/main/java/uk/gov/di/lambdas/AuthorisationHandler.java
@@ -7,7 +7,9 @@ import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent
 import com.nimbusds.oauth2.sdk.AuthorizationRequest;
 import com.nimbusds.oauth2.sdk.ErrorObject;
 import com.nimbusds.oauth2.sdk.ParseException;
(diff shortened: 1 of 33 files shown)


serverless/lambda/src/main/java/uk/gov/di/lambdas/AuthorisationHandler.java
```diff
@@ -70,28 +72,30 @@ public AuthorisationHandler() {
...
-            response.setStatusCode(400);
-            response.setBody("Cannot parse authentication request");
-
-            return response;
+            if (e.getRedirectionURI() == null) {
+                LOGGER.error("Redirect URI is missing from request");
+                throw new RuntimeException("Redirect URI is missing");
+            }
+            return generateErrorResponse(
+                    e.getRedirectionURI(), e.getState(), e.getResponseMode(), e.getErrorObject());
```
sanisc, line 87: does this not want to be e.getErrorObject().getJSONObject()?

sanisc (approved):

PR #284, opened 2021-08-13
Govsi 519 interruption screen for terms and conds update

cj1993:
## What?

Backend changes for when the interrupt screen appears and we need to recapture users accepting Terms and Conditions.


serverless/lambda/src/main/java/uk/gov/di/lambdas/UpdateProfileHandler.java
```diff
@@ -81,72 +84,99 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                            clientSession.get().getAuthRequestParams());
+                            clientId = authenticationRequest.getClientID().getValue();
+                        } catch (ParseException e) {
+                            LOGGER.info(
+                                    "Cannot retreive auth request params from client session id.");
+                            return generateApiGatewayProxyErrorResponse(
+                                    400, ErrorResponse.ERROR_1001);
+                        }
+
+                        Optional<Map<String, List<String>>> clientConsents =
```
sanisc, line 122: this doesn't look like the right object is being persisted

serverless/lambda/src/test/java/uk/gov/di/lambdas/UpdateProfileHandlerTest.java
```diff
@@ -13,6 +13,7 @@
 import com.nimbusds.openid.connect.sdk.OIDCScopeValue;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.shared.services.ConfigurationService;
```
sanisc, line 16: does there need to be an integration test for this?

sanisc (dismissed):

sanisc:
Can we squash the commits as well.

serverless/lambda/src/test/java/uk/gov/di/lambdas/UpdateProfileHandlerTest.java
dbes-gds: good to have at least one

serverless/lambda/src/main/java/uk/gov/di/lambdas/UpdateProfileHandler.java
cj1993: Yeah your right. We discussed this on Friday, this section was part of the capture consent ticket me and Dan were working on previously, the model does need updating but it is not part of the scope of this ticket and I will do the refactoring when I revisit that. It is green in the diff probably due to IntelliJ formatting code rather than it being a new addition.

serverless/lambda/src/test/java/uk/gov/di/lambdas/UpdateProfileHandlerTest.java
cj1993: Tests for the ConfigurationService or tests for the use of the ConfigurationService within UpdateProfileHandler ?

sanisc (dismissed):

dbes-gds (approved):

serverless/lambda/src/test/java/uk/gov/di/lambdas/UpdateProfileHandlerTest.java
sanisc: update handler - as most handlers will have integration tests

PR #293, opened 2021-08-16
Terraform env var for terms and conds

cj1993:
## What?

New env var into terraform to support updating Terms and Conditions


chrisclayson (approved):
👍🏻

dbes-gds (approved):

PR #297, opened 2021-08-17
Eliminate lambda cold start terraform config for provisioned concurrency

cj1993:
## What?

Terrafom config to reduce / eliminate lambda cold start by using provisioned concurrency


chrisclayson (dismissed):
👍🏻

chrisclayson (approved):

PR #298, opened 2021-08-17
GOVSI-623 - Use KMS to sign ID tokens 

JHjava:
## What?

- Add terraform to generate KMS signing key for ID tokens
-  Sign ID tokens and access tokens using KMS
- Retrieve Public Key from KMS

## Why?

- We were previously using a signing key generated on the fly. Switch to using a signing key provided by KMS.
- So clients can verify the signing of the ID tokens using the JWKS lambda 

## Further work to do 
- Split out the permissions of signing and getting the public key into separate roles so that only the Lambdas who need to sign have permission to do so 
- Add some unit tests for the new service

### DIFF
--- a/ci/terraform/aws/kms.tf
+++ b/ci/terraform/aws/kms.tf
@@ -17,10 +17,7 @@ data "aws_iam_policy_document" "kms_policy_document" {
     effect = "Allow"
 
     actions = [
-      "kms:ListKeys",
-      "kms:ListAliases",
       "kms:Sign",
-      "kms:Verify",
       "kms:GetPublicKey",
     ]
     resources = [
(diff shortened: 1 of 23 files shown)


ci/terraform/aws/kms.tf
```diff
@@ -0,0 +1,42 @@
+resource "aws_kms_key" "id_token_signing_key" {
+  description             = "KMS signing key for ID tokens"
```
mrwilson, line 2: Do we want to parametrise the name for this with the environment, or are we sharing a single signing key for all environments?

JHjava: This would give us a key for non-prod and a key for prod as they'll be in different accounts. Unless we think we need a key for each environment?

mrwilson: As I understand it, we have separate Redis and Dynamo instances for the different environment, seems to me it makes sense to keep it consistent and have separate signing keys for each environment?

JHjava: Ok if that's where things are at. It will need to be added to the alias. Let me push something up

JHjava: Should be one for each env now @mrwilson

sanisc (approved):

PR #313, opened 2021-08-18
GOVSI-623: Convert KMS signature to JWT compatible signature

chrisclayson:
## What?

Convert KMS signature to JWT compatible signature

## Why?

KMS gives a signature in a format that is incompatible with JWT.


sanisc (approved):

sanisc:
No longer needed as its fixed in another PR

PR #317, opened 2021-08-19
GOVSI-623 - Refactor Token generation KMS code

JHjava:
## What?

- Remove signing key generated on the fly
- Add a KmsConnectionService to handle all the calls to KMS 
- Move logic out of the TokenGeneratorService into the TokenService and call the KmsConnectionService to make the calls to KMS
- Change the TokenGeneratorService to use EC signatures rather than RSA 

## Why?

- It removes an unnecessary service and allows us ease of testing our logic of parsing PublicKeys and Signatures 
- Reuse shared code when possible to prevent duplication


serverless/lambda/src/main/java/uk/gov/di/services/TokenService.java
```diff
@@ -93,8 +139,31 @@ private AccessToken generateAndStoreAccessToken(
...
     }
 
-    public JWK getSigningKey() {
-        return tokenGeneratorService.getPublicKey();
+    public JWK getPublicKey() {
+        LOGGER.info("Creating GetPublicKeyRequest to retrieve PublicKey from KMS");
+        Provider bcProvider = new BouncyCastleProvider();
+        GetPublicKeyRequest getPublicKeyRequest = new GetPublicKeyRequest();
+        getPublicKeyRequest.setKeyId(configService.getTokenSigningKeyId());
+        GetPublicKeyResult publicKeyResult = kmsConnectionService.getPublicKey(getPublicKeyRequest);
```
sanisc, line 147: the above code could be moved into the kmsConnectionService.getPublicKey(getPublicKeyRequest); as the kmsconnectionserivce already has access to the configService, so it would become:

`kmsConnectionService.getPublicKey();`

JHjava: Fair point, although it would couple this method to the PublicKey of the token signing key. If we introduce other keys we would then need to add methods for each key introduced.

sanisc (approved):

PR #319, opened 2021-08-19
Back end changes to support service type for client

cj1993:
## What?

Changes to support the service type for clients (mandatory or optional) which will be used as part of the terms and conditions pages which are under development and need this data returning from the client info lambda


chrisclayson (approved):
LGTM 👍🏻

PR #331, opened 2021-08-20
Passing state back from client info

cj1993:
## What?

Client info lambda change to pass state to front end for use on terms and conditions pages


serverless/lambda/src/main/java/uk/gov/di/lambdas/ClientInfoHandler.java
```diff
@@ -79,6 +79,10 @@ public APIGatewayProxyResponseEvent handleRequest(
             Map<String, List<String>> authRequest = clientSession.get().getAuthRequestParams();
 
             String clientID = AuthenticationRequest.parse(authRequest).getClientID().getValue();
+            String state = null;
```
dbes-gds, line 82: at some point maybe we should look at the multiple calls to AuthenticationRequest.parse(authRequest)

dbes-gds (approved):

PR #343, opened 2021-08-23
GOVSI-458: modify state machine to allow multiple calls to userExists in a single session

dbes-gds:
## What?

Modify state machine to allow multiple calls to userExists in a single session.
Now able to transition from USER_NOT_FOUND to USER_NOT_FOUND again.

## Why?

User journey is changing to allow a single user to check whether accounts exist for multiple emails.  Before it was only possible to check a single email within a single session.

### DIFF
--- a/serverless/lambda/src/test/java/uk/gov/di/lambdas/CheckUserExistsHandlerTest.java
+++ b/serverless/lambda/src/test/java/uk/gov/di/lambdas/CheckUserExistsHandlerTest.java
@@ -26,6 +26,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.anyMap;
 import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.reset;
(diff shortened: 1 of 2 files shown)


serverless/lambda/src/test/java/uk/gov/di/lambdas/CheckUserExistsHandlerTest.java
```diff
@@ -45,6 +46,7 @@ public void setup() {
         handler =
                 new CheckUserExistsHandler(
                         validationService, authenticationService, sessionService);
+        reset(authenticationService);
```
mrwilson, line 49: Looks good, but what do we need the reset for?

mrwilson (approved):

serverless/lambda/src/test/java/uk/gov/di/lambdas/CheckUserExistsHandlerTest.java
dbes-gds: Was debugging the test and wasn't sure I was getting the correct response from this mock, so put this in to make sure it was reset.  Might be able to remove it, will give it a go.

PR #390, opened 2021-08-27
GOVSI-185: add authenticate endpoint to account management api

dbes-gds:
## What?

Add authenticate endpoint to account management api.

This is pretty much a clone of the oidc login endpoint with session handling removed.  It's named 'authenticate' rather than 'login' as there was an AWS/Terraform name clash with the existing login service.  Authenticate does actually reflect what the service will do.

At some point the API may need to be tweaked to add state management (and authorisation), but this is a later piece of work.

An integration test is available for use locally, but as we don't currently run the account management terraform in Github Actions this has been excluded from running in the build.

## Why?

Account management needs to check the user's password at the beginning of each flow, and needs an API to do so.

## Related PRs

https://github.com/alphagov/di-authentication-account-management/pull/48


account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/SessionState.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.accountmanagement.entity;
+
+public enum SessionState {
```
sanisc, line 3: i would strip this back as the session is controlled in the front end app and not backend

account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/AuthenticateResponse.java
```diff
@@ -0,0 +1,21 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class AuthenticateResponse extends BaseAPIResponse {
+
+    @JsonProperty("redactedPhoneNumber")
+    private String redactedPhoneNumber;
+
+    public AuthenticateResponse(
+            @JsonProperty(required = true, value = "redactedPhoneNumber")
```
sanisc, line 11: i would just return a 200 without anything else as all the information already exists in the front end session

shared/src/main/java/uk/gov/di/authentication/shared/helpers/RedactPhoneNumberHelper.java
```diff
@@ -1,4 +1,4 @@
-package uk.gov.di.authentication.frontendapi.helpers;
+package uk.gov.di.authentication.shared.helpers;
 
 public class RedactPhoneNumberHelper {
```
sanisc, line 3: this can be moved back

integration-tests/src/test/java/uk/gov/di/accountmanagement/api/AuthenticateIntegrationTest.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private static final String LOGIN_ENDPOINT = "/authenticate";
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    @Test
+    public void shouldCallLoginEndpointAndReturn200WhenLoginIsSuccessful() throws IOException {
+        String email = "joe.bloggs+3@digital.cabinet-office.gov.uk";
+        String password = "password-1";
+        String phoneNumber = "01234567890";
+        DynamoHelper.signUp(email, password);
```
sanisc, line 28: this lambda is to validate the user credentials is there a reason it is calls signup?

account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/BaseAPIResponse.java
```diff
@@ -0,0 +1,22 @@
+package uk.gov.di.accountmanagement.entity;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class BaseAPIResponse {
+
+    @JsonProperty protected SessionState sessionState;
```
sanisc, line 7: this class can be removed as the session is not on the backend

sanisc (dismissed):

integration-tests/src/test/java/uk/gov/di/accountmanagement/api/AuthenticateIntegrationTest.java
dbes-gds: that's a test helper that creates a user account, need an existing account for the test.

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
```diff
@@ -0,0 +1,79 @@
...
+            }
+            boolean hasValidCredentials =
+                    authenticationService.login(
+                            loginRequest.getEmail(), loginRequest.getPassword());
+            if (!hasValidCredentials) {
+                LOGGER.error("Invalid login credentials entered");
+                return generateApiGatewayProxyErrorResponse(401, ErrorResponse.ERROR_1008);
+            }
+            String phoneNumber =
+                    authenticationService.getPhoneNumber(loginRequest.getEmail()).orElse(null);
```
sanisc, line 63: the phone number is not needed so this can be removed

sanisc (dismissed):

sanisc (approved):

PR #412, opened 2021-09-01
Govsi 628  - Refactor authorizer terraform

JHjava:
## What?

- Delete HelloWorld Lambda and Terraform
- Refactor authorizer so it can be reused by other lambdas in account management
- By default lambdas will not use the authorizer unless `use_authorizer` is specified in the lambda config

## Why?

- Add the aws_api_gateway_authorizer to the endpoint-module and make it optional as we only want it to be used by account management.
-


ci/terraform/account-management/authorise-access-token.tf
```diff
@@ -14,10 +14,12 @@ module "authorise_access_token" {
   execution_arn             = aws_api_gateway_rest_api.di_account_management_api.execution_arn
   api_deployment_stage_name = var.environment
   lambda_zip_file           = var.lambda_zip_file
+  use_authorizer            = true
```
dbes-gds, line 17: is this needed for the authorizer itself?

PR #445, opened 2021-09-07
GOVSI-628 - Add update email message to queue and send via Notify

JHjava:
## What?

 - Add update email message to queue and send via Notify
- Create a new template ID to send a confirmation email that the email has been updated
- Create a new lambda role for access to both dynamo and SQS
- Create a new notification type for when an EMAIL has been updated

## Why?

- When an update email request has been received we need to update the email in dynamodb and then put it in on an SQS queue so it can be sent via Notify.

**TODO**
- Validate the OTP


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
```diff
@@ -0,0 +1,88 @@
...
+        LOGGER.info(
+                "Authorizer parameters received: {}", input.getRequestContext().getAuthorizer());
+        try {
+            UpdateEmailRequest updateInfoRequest =
+                    objectMapper.readValue(input.getBody(), UpdateEmailRequest.class);
+            Subject subjectFromEmail =
+                    dynamoService.getSubjectFromEmail(updateInfoRequest.getExistingEmailAddress());
+            Map<String, Object> authorizerParams = input.getRequestContext().getAuthorizer();
+
+            if (!authorizerParams.containsKey("principalId")) {
```
sanisc, line 61: there is a helper method for this that @cj1993 has added which is probably worth while using

shared/src/main/java/uk/gov/di/authentication/shared/entity/NotificationType.java
```diff
@@ -3,5 +3,6 @@
 public enum NotificationType {
     VERIFY_EMAIL,
     VERIFY_PHONE_NUMBER,
-    MFA_SMS
+    MFA_SMS,
+    EMAIL_UPDATED
 }
```
sanisc, line 8: its probably worth separating these out to use its own for account management even if they are the same

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
JHjava: Going to leave it as it is for now

shared/src/main/java/uk/gov/di/authentication/shared/entity/NotificationType.java
JHjava: Have updated

sanisc (dismissed):

sanisc (approved):

PR #448, opened 2021-09-07
GOVSI-774: Add Cloudwatch triggers for lambda warmers

chrisclayson:
## What?

- Add an, optional, CloudWatch trigger for each endpoint that triggers a warmer lambda every 5 mins

## Why?

We want keep some, if not all, lambdas "warm". This trigger will be used to trigger an execution of a warmer function.

### DIFF
--- a/ci/terraform/modules/endpoint-module/warmer.tf
+++ b/ci/terraform/modules/endpoint-module/warmer.tf
@@ -5,7 +5,7 @@ resource "aws_lambda_function" "warmer_function" {
   function_name = replace("${var.environment}-${var.endpoint_name}-lambda-warmer", ".", "")
   role          = var.warmer_lambda_role_arn
   handler       = var.warmer_handler_function_name
-  timeout       = 30
+  timeout       = 60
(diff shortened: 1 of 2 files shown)


ci/terraform/modules/endpoint-module/warmer.tf
```diff
@@ -0,0 +1,60 @@
+resource "aws_lambda_function" "warmer_function" {
+  count = var.warmer_handler_function_name == null ? 0 : 1
+
+  filename      = var.warmer_lambda_zip_file
+  function_name = replace("${var.environment}-${var.endpoint_name}-lambda-warmer", ".", "")
+  role          = var.warmer_lambda_role_arn
+  handler       = var.warmer_handler_function_name
+  timeout       = 30
```
sanisc, line 8: might want to set this to 60

chrisclayson: Good shout!

sanisc (approved):

PR #464, opened 2021-09-08
GOVSI-719 - Use Alias ID to retrieve key and configure account management with KMS

JHjava:
## What?

- Retrieve the key using the alias
- Reference the KMS key dataqsource in account management

## Why?

- By using the alias instead of the key_id, it ensures that we always retrieve the latest key. This will benefit us during key rotations as we can avoid having to restart our lambdas.
- We will need the public key in the authorizer so we can validate the signature of the access token

### DIFF
--- a/ci/terraform/account-management/api-gateway.tf
+++ b/ci/terraform/account-management/api-gateway.tf
@@ -85,6 +85,11 @@ resource "aws_lambda_function" "authorizer" {
     security_group_ids = [aws_vpc.account_management_vpc.default_security_group_id]
     subnet_ids         = aws_subnet.account_management_subnets.*.id
   }
+  environment {
+    variables = {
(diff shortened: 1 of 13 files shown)


ci/terraform/account-management/api-gateway.tf
```diff
@@ -85,6 +85,11 @@ resource "aws_lambda_function" "authorizer" {
     security_group_ids = [aws_vpc.account_management_vpc.default_security_group_id]
     subnet_ids         = aws_subnet.account_management_subnets.*.id
   }
+  environment {
+    variables = {
+      TOKEN_SIGNING_KEY_ALIAS = data.aws_kms_key.id_token_public_key.key_id
```
chrisclayson, line 90: This isn't the alias?

chrisclayson (approved):

PR #465, opened 2021-09-08
Support for pairwise subject identifiers

cj1993:
## What?

Changes to support pairwise subject identifiers

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
@@ -146,6 +146,7 @@ public class AuthCodeHandler
                                                         ResponseHeaders.LOCATION,
                                                         errorResponse.toURI().toString()));
                             }
+
                             try {
                                 if (!authorizationService.isClientRedirectUriValid(
                                         authenticationRequest.getClientID(),
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
@@ -4,11 +4,15 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.GrantType;
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.id.Subject;
+import com.nimbusds.oauth2.sdk.token.RefreshToken;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+import com.nimbusds.openid.connect.sdk.OIDCScopeValue;
 import com.nimbusds.openid.connect.sdk.OIDCTokenResponse;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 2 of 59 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -187,4 +211,44 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+
+        if (clientRegistry.getSubjectType().equalsIgnoreCase("pairwise")) {
+            if (userProfile.getClientSubjects() != null) {
+                if (!clientSubjectsContainsClientId(userProfile.getClientSubjects(), clientId)) {
+                    userProfile
+                            .getClientSubjects()
+                            .add(new ClientSubject(clientId, new Subject(), sector));
+                } else {
+                    userProfile.setClientSubjects(
+                            List.of(new ClientSubject(clientId, new Subject(), sector)));
```
chrisclayson, line 233: I don't we should be doing `new Subject()` here, we should be retrieving the existing one from the list.

chrisclayson: In fact, I don't think we need this `else` at all!?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -187,4 +211,44 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        userProfile.getEmail(), userProfile.getClientSubjects());
+            }
+        }
+    }
+
+    private boolean clientSubjectsContainsClientId(
+            List<ClientSubject> clientSubjects, String clientId) {
+        boolean present = false;
+
+        for (ClientSubject clientSubject : clientSubjects) {
```
chrisclayson, line 245: Should we be looking up on sector if client ID doesn't exist.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -209,4 +207,30 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+
+    private Subject getSubjectByEmailAndClient(String email, ClientRegistry client) {
+        UserProfile userProfile = dynamoService.getUserProfileByEmail(email);
+
+        if (client.getSubjectType().equalsIgnoreCase("public")) {
+            return new Subject(userProfile.getPublicSubjectID());
+        } else {
+            List<ClientSubject> clientSubjects = userProfile.getClientSubjects();
+            if (client.getSectorIdentifierUri() != null) {
+                for (ClientSubject clientSubject : clientSubjects) {
```
chrisclayson, line 219: Minor point, you could use the streaming API `filter(...).findFirst()` rather than a for loop here.

chrisclayson (changes requested):
A few comments that I think need addressing.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
cj1993: I don't think we can have a ClientSubject entry without a client ID will have to confirm this with @sanisc

shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
```diff
@@ -0,0 +1,38 @@
...
+            for (byte aByte : bytes) {
+                sb.append(Integer.toString((aByte & 0xff) + 0x100, 16).substring(1));
+            }
+
+            return sb.toString();
+        } catch (NoSuchAlgorithmException e) {
+            LOGGER.error("Failed to HASH", e);
+        }
+
+        return null;
```
chrisclayson, line 36: Should we throw a RuntimeException here instead of returning `null`?

chrisclayson (approved):
One small comment, but otherwise fine!

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandlerTest.java
```diff
@@ -252,4 +269,30 @@ private String buildCookieString() {
                 "%s=%s.%s; Max-Age=%d; %s",
                 "gs", SESSION_ID, CLIENT_SESSION_ID, 1800, "Secure; HttpOnly;");
     }
+
+    private UserProfile createUserProfile() {
```
sanisc, line 273: guessing these are now needed to for testing authcode handler?

sanisc (approved):

PR #478, opened 2021-09-09
GOVSI-719 - Validate AccessToken signature in UserInfo

JHjava:
## What?

- Move token validation into a separate service
- Validate AccessToken signature in UserInfo

## Why?

- The TokenService is getting heavy with a lot of stuff. Simplify that by moving out the Token validation logic into a TokenValidationService
- So only valid AccessTokens can retrieve userinfo


alan-gds (commented):
generateECKeyPair() can be removed from TokenServiceTest.  That's worth fixing, but otherwise looks good.

alan-gds (dismissed):

alan-gds (approved):

PR #479, opened 2021-09-09
BAU: add new scope for account management: 'am'

dbes-gds:
## What?

Add new scope for account management: 'am'

- Add support for custom (non-OIDC scopes).
- Expand scope validation to check for custom scopes.
- These custom scopes are private by default, meaning that it's not possible to for public users to register them against their clients using the public API.  Only we can add these scopes to clients via the database.
- Return an error is a client tries to register a private scope.
- Allow clients with private scopes to authorize.  In practice for the moment this will only be allowed for the Account Management application.

## Why?

Restrict usage of the Account Management application to our private/internal client.


JHjava (approved):
👍

PR #481, opened 2021-09-09
GOVSI-751: Add error logging in frontend lambdas

alan-gds:
## What?

Add error logging in frontend lambdas.  I missed this in #431.

I've changed some existing messages to make things more consistent across classes.  It doesn't feel worth trying to share code for that though (but I'm open to suggestions).

## Why?

Will make it easier for us to diagnose issues when the service is live.

## Related PRs

#431

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
@@ -88,7 +88,6 @@ public class UpdateProfileHandler
         String clientId;
 
         if (session.isEmpty()) {
-            LOGGER.info("Session is empty.");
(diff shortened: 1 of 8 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
```diff
@@ -88,7 +88,6 @@ public APIGatewayProxyResponseEvent handleRequest(
         String clientId;
 
         if (session.isEmpty()) {
-            LOGGER.info("Session is empty.");
```
alan-gds, line 91: Not needed as the SessionService will already have output a suitable error message to the log.

dbes-gds (approved):

chrisclayson (approved):

PR #488, opened 2021-09-10
GOVSI-719: validate access token scopes in account management custom authorizer

dbes-gds:
## What?

Validate access token scopes in account management custom authorizer.

This change will restrict access to the account management api so that only clients with the 'am' scope will be able to use it.  It should only be merged with the 'am' scope has been added to the am frontend:

https://github.com/alphagov/di-authentication-account-management/commit/c1659e598f814716debb5417be3c350c6fdcb283

## Why?

The token must contain the 'am' scope to allow access.

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthoriseAccessTokenHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthoriseAccessTokenHandler.java
@@ -2,6 +2,7 @@ package uk.gov.di.accountmanagement.lambda;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.token.AccessToken;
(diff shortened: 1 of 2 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthoriseAccessTokenHandler.java
```diff
@@ -49,17 +53,26 @@ public AuthPolicy handleRequest(TokenAuthorizerContext input, Context context) {
...
-            String subject = SignedJWT.parse(accessToken.getValue()).getJWTClaimsSet().getSubject();
             boolean isAccessTokenSignatureValid =
                     tokenValidationService.validateAccessTokenSignature(accessToken);
             if (!isAccessTokenSignatureValid) {
                 LOGGER.error("Access Token signature is not valid");
                 throw new RuntimeException("Unauthorized");
             }
             LOGGER.info("Successfully validated Access Token signature");
+
+            List<String> scopeList = claimsSet.getStringListClaim("scope");
```
JHjava, line 67: In theory it should never be null but is it worth handling any null pointer is scope or subject is not present?

dbes-gds: claimsSet.getStringListClaim() returns null for a claim that isn't there, so the null check on line 68 handles this.  Have added a null check and error for the subject.

JHjava (approved):

PR #513, opened 2021-09-14
GOVSI-718- Add Support for refresh tokens

JHjava:
## What?

- Send RefreshToken in TokenResponse.Generate, send and store a signed RefreshToken to send alongside the bearer and ID tokens back to the client. This refresh token will be able to be used to obtain a new access token.
- Allow requests into the token endpoint with the refresh_token grant type. This will request will contain the client_id, grant_type, refresh_token and scope
- Store the refresh token using the key as the following- "client_id:subjectId". We will get the client_id out of the request and the subject id from the refresh token to retrieve the refresh token from redis. The refresh token needs to be one per user, as when we generate a new access token we need to tie that access token to a user so we can get the user information using the subject_id on user_info
- Add the at_hash to the ID token which is the hash value of the access token. This binds the ID token with the access token and can be validated client side. https://openid.net/specs/openid-connect-core-1_0.html#HybridIDToken
- Support Offline Access scope so only clients that are registered to use the offline access scope will be given a refresh token. This prevents the need to return them to all clients in each token request. 

## Why?
Access tokens are short lived and need to be added to every call from the account management frontend to the account management API. If we did nothing then the Account Management frontend would either needed a special issued access token with a longer lived expiry than the rest or we would have to make them go through authorize again, neither of which is ideal. The proper OIDC way to handle this is via refresh tokens.  

We need to issue a refresh token alongside the ID token and Access Token from the Token endpoint and then store that refresh token. The refresh token has a longer expiry. We then need to adapt the Token endpoint to be able to consume RefreshTokens request where we will need to validate the request and issue the client a new refresh and access token.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
@@ -52,6 +52,7 @@ import java.util.UUID;
 import static java.util.Collections.singletonList;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
 
 public class TokenIntegrationTest extends IntegrationTestEndpoints {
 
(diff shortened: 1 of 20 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
```diff
@@ -61,7 +78,8 @@ public void shouldCallTokenResourceAndReturn200() throws JOSEException {
         RedisHelper.addAuthCodeAndCreateClientSession(
                 code, "a-client-session-id", TEST_EMAIL, generateAuthRequest().toParameters());
         Map<String, List<String>> customParams = new HashMap<>();
-        customParams.put("grant_type", Collections.singletonList("authorization_code"));
+        customParams.put(
```
dbes-gds, line 81: where is 'customParams' used? doesn't seem to get passed to the request or checked in the response.

JHjava: On line 87

integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
```diff
@@ -75,15 +93,99 @@ public void shouldCallTokenResourceAndReturn200() throws JOSEException {
...
+        Date expiryDate = Date.from(localDateTime.atZone(ZoneId.systemDefault()).toInstant());
+        JWTClaimsSet claimsSet =
+                new JWTClaimsSet.Builder()
+                        .claim("scope", scope.toStringList())
+                        .issuer("issuer-id")
+                        .expirationTime(expiryDate)
+                        .issueTime(
+                                Date.from(
+                                        LocalDateTime.now()
+                                                .atZone(ZoneId.systemDefault())
```
dbes-gds, line 173: Wondering if there's a way to keep timezones out of this and just use a timestamp? Might make checking timestamps a bit easier.  Is this possible?

sanisc (approved):

PR #517, opened 2021-09-15
BAU: shared code to retrieve user context - Session, ClientSession and ClientRegistry

dbes-gds:
## What?

Shared code to retrieve user context - Session, ClientSession and ClientRegistry.

Create a base class for handlers that need to understand user context, that is retrieve the Session, ClientSession and ClientRegistry information.

Applied to the ClientInfoHandler by way of example.

## Why?

When looking at how to implement synthetic transactions it became clear that some lambdas that don't currently have the information would need to know which client they were being used by to see if it was a test client.  The code to do this is available but duplicated in a few places, so it's better to have some shared code to do the job before rolling it out further.  This will also be useful for implementing per-client behaviours as required.

In addition, adding calls to the lambda warmer has indented handler code, so this class provides a way to separate the lambda infrastructure tasks and retrieving the context from the rest of the handler work by splitting these into two methods, with a 'handleRequestDelegate' method to do the actual work.

To rollout to additional handlers it may be necessary to make which contexts are provided configurable.


shared/src/main/java/uk/gov/di/authentication/shared/lambda/UserContextAwareHandler.java
```diff
@@ -0,0 +1,133 @@
...
+                        id -> {
+                            return clientService.getClient(id).orElse(null);
+                        })
+                .orElse(null);
+    }
+
+    protected boolean isTestClientSession(ClientSession clientSession) {
+        return clientSessionService
+                .getClientIdForClientSession(clientSession)
+                .map(
```
mrwilson, line 120: Can we use a `flatMap(clientService::getClient)` here instead to collapse the inner Optional?

dbes-gds: updated

PR #518, opened 2021-09-15
GOVSI-768: Log state transitions

alan-gds:
## What?

Log state transitions.

## Why?

This will help us trace the progress of journeys when investigating issues.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
@@ -62,18 +62,16 @@ public class StateMachine<T, A, C> {
     }
 
     public T transition(T from, A action) {
-        return states.getOrDefault(from, emptyList()).stream()
-                .filter(t -> t.getAction().equals(action))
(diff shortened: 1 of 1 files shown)


dbes-gds (approved):

shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
```diff
@@ -62,18 +62,23 @@ public StateMachine(Map<T, List<Transition<T, A, C>>> states) {
...
-                            return new InvalidStateTransitionException();
-                        })
-                .getNextState();
+        T to =
+                states.getOrDefault(from, emptyList()).stream()
+                        .filter(t -> t.getAction().equals(action))
+                        .findFirst()
+                        .orElseThrow(
+                                () -> {
+                                    LOGGER.error(
```
mrwilson, line 71: Can we make this a method reference to something like `this::logAndThrowException`?

Just a taste thing, inline lambdas push the indention in a lot.

mrwilson (approved):

PR #521, opened 2021-09-15
GOVSI-683: Implement updated T&Cs state

chrisclayson:
## What?

- Add a fluent state machine builder API
- Add a conditional path that checks whether or not the user has accepted the latest T&Cs, if not, changes to the appropriate state
- Add missing StateMachine unit tests.

## Why?

We have several conditional paths in our user journey, this PR allows this functionality in the state machine and adds one conditional state to demonstrate how this achieved.


shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
```diff
@@ -74,148 +82,144 @@ public T transition(T from, A action) {
...
+                            .collect(Collectors.toList()));
+            return stateMachineBuilder;
+        }
+
+        protected StateRuleBuilder(Builder<T, A, C> stateMachineBuilder, T state) {
+            this.stateMachineBuilder = stateMachineBuilder;
+            this.state = state;
+        }
+
+        public Builder<T, A, C> deadEnd() {
```
sanisc, line 217: could this be changed to final?

chrisclayson: `final` is a reserved word, so had to use `finalState()`

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -5,6 +5,15 @@
 
 public class ConfigurationService {
 
+    private static ConfigurationService configurationService;
+
+    public static ConfigurationService getInstance() {
```
JHjava, line 10: Is this used anywhere?

chrisclayson: Good spot ... not used any more, changed my mind on how to inject the config service.

chrisclayson: Yes ... in `StateMachine.java`, line 106 ... it's a large diff, so may not show by default.

JHjava: Ah yes that will be it

JHjava (approved):

PR #534, opened 2021-09-17
GOVSI-642: add reset password request lambda

dbes-gds:
## What?

Add reset password request lambda.

Sends a reset password email with a one-time link to allow the user to change their password.

## Why?

User needs to be able to reset their password (if they have forgotten it, for example)

## Related PRs

#533


ci/terraform/oidc/reset-password-request.tf
```diff
@@ -0,0 +1,51 @@
+module "reset-password-request" {
```
JHjava, line 1: You'll need to add the triggers and the depends_on to the oidc/api-gateway.tf

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/ResetPasswordRequestHandler.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    private APIGatewayProxyResponseEvent handleNotificationRequest(
+            String email,
+            NotificationType notificationType,
+            Session session,
+            SessionState nextState)
+            throws JsonProcessingException {
+
+        String subjectId = authenticationService.getSubjectFromEmail(email).getValue();
```
JHjava, line 159: Do we expect this to always find the user or is there a chance that a user might not exist with that email?

ci/terraform/oidc/sqs.tf
```diff
@@ -134,6 +134,9 @@ resource "aws_lambda_function" "email_sqs_lambda" {
       VERIFY_EMAIL_TEMPLATE_ID        = "b7dbb02f-941b-4d72-ad64-84cbe5d77c2e"
       VERIFY_PHONE_NUMBER_TEMPLATE_ID = "7dd388f1-e029-4fe7-92ff-18496dcb53e9"
       MFA_SMS_TEMPLATE_ID             = "7dd388f1-e029-4fe7-92ff-18496dcb53e9"
+      RESET_PASSWORD_TEMPLATE_ID      = "0aaf3ae8-1825-4528-af95-3093eb13fda0"
+      FRONTEND_BASE_URL               = var.frontend_base_url
```
chrisclayson, line 138: Doesn't this need to be `local.frontend_base_url` as we added this morning?

echo-int-test-vars.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
chrisclayson, line 1: Did you mean to commit this?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/ResetPasswordRequestHandler.java
dbes-gds: They have to have gone through checkuserexists to get here so in theory the user should exist.  It's possible if someone is running two flows at the same time and deletes the user.  Could make it return an error?

ci/terraform/oidc/sqs.tf
dbes-gds: good spot

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/ResetPasswordRequestHandler.java
JHjava: This is a very rare edge case as the user would have to delete their account between the time they enter their email and clicking the check password link therefore it we can let it blow up if the user isn't there. No action needed

ci/terraform/oidc/reset-password-request.tf
dbes-gds: added

echo-int-test-vars.sh
dbes-gds: yes, it generates the run config string you can use to run integration tests in the ide

JHjava (approved):

PR #547, opened 2021-09-20
GOVSI-644 - Keep track of how many password resets have been requested

JHjava:
## What?

- Add a reset password count to the session to keep track of how many times a user has requested to reset their password on a session. If a user has requested more than 5 password resets, then we will block the user from making anymore reset attempts for 15 minutes and return a `ERROR_1022` to the frontend.
- If a user requests another password reset once it has been blocked we will return the `ERROR_1023` to the frontend so we can display a different error message.
- To accommodate for this behaviour add a new `RESET_PASSWORD_LINK_MAX_RETRIES_REACHED` state and a `SYSTEM_HAS_SENT_RESET_PASSWORD_LINK_TOO_MANY_TIMES` action.
- Change the state machine to allow the following scenarios:
    - When `RESET_PASSWORD_LINK` has been sent allow the `SYSTEM_HAS_SENT_RESET_PASSWORD_LINK_TOO_MANY_TIMES` to change the state to `RESET_PASSWORD_LINK_MAX_RETRIES_REACHED` to accommodate for when more than 5 password resets have been requested.
    - When the `RESET_PASSWORD_LINK_SENT` then we will allow `SYSTEM_HAS_SENT_RESET_PASSWORD_LINK` so the user can generate another reset password
    - When the user is on the `RESET_PASSWORD_LINK_MAX_RETRIES_REACHED` after 15 minutes we want to allow the user to request another code and therefore allow the `SYSTEM_HAS_SENT_RESET_PASSWORD_LINK` action to move to the `RESET_PASSWORD_LINK_STATE`

## Why?

- To limit the amount of times a user can request to reset their password on a session


mrwilson:
Looks good - I notice that there's no auditing going on in this handler, but that's not due to this PR, we can schedule some work around that before go live as password resets should 100% be in the audit trail

mrwilson (approved):

PR #559, opened 2021-09-21
GOVSI-817 - Update ID token with vot and vtm claim

JHjava:
## What?

- Populate the ID token with the vot which we can obtain from the client registry.
- Populate the vtm with the trustmark uri which will point at our trustmark lambda.

## Why?

- This follows what is in the https://openid.net/specs/openid-igov-openid-connect-1_0-02.html#rfc.section.1


chrisclayson (approved):
👍🏻

PR #568, opened 2021-09-22
Temporary account locking via redis

cj1993:
## What?

When a user enters a wrong password for the fifth time, we want to lock their account for 15 minutes.


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/LoginResponse.java
```diff
@@ -10,7 +10,7 @@
     private String redactedPhoneNumber;
 
     public LoginResponse(
-            @JsonProperty(required = true, value = "redactedPhoneNumber")
+            @JsonProperty(required = false, value = "redactedPhoneNumber")
```
JHjava, line 13: You can remove `required = false` as false is the default

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -34,7 +37,8 @@
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(LoginHandler.class);
-
+    private final CodeStorageService codeStorageService;
+    private static final int MAX_INCORRECT_PASSWORD_ATTEMPTS = 5;
```
JHjava, line 41: You should get this value from config such as `configurationService.getCodeMaxRetries()`

shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
```diff
@@ -15,6 +15,7 @@
 import java.util.stream.Collectors;
 
 import static java.util.Collections.emptyList;
+import static uk.gov.di.authentication.shared.entity.SessionAction.*;
```
JHjava, line 18: Try to avoid the wildcard imports

shared/src/main/java/uk/gov/di/authentication/shared/services/RedisConnectionService.java
```diff
@@ -45,6 +45,14 @@ public void saveWithExpiry(String key, String value, long expiry) {
         }
     }
 
+    public void save(String key, String value) {
```
JHjava, line 48: Is this used?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -70,22 +79,64 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
+
+            boolean hasEnteredPasswordIncorrectBefore =
+                    codeStorageService.hasEnteredPasswordIncorrectBefore(loginRequest.getEmail());
+
             if (!hasValidCredentials) {
+                if (hasEnteredPasswordIncorrectBefore) {
+                    int count =
+                            codeStorageService.getIncorrectPasswordCount(loginRequest.getEmail());
+
+                    if (count >= MAX_INCORRECT_PASSWORD_ATTEMPTS) {
```
JHjava, line 112: Should we increase the count before this check?

shared/src/main/java/uk/gov/di/authentication/shared/services/RedisConnectionService.java
cj1993: Good spot !

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
cj1993: I have implemented this such that once we exceed the attempt limit then we stop increasing the recorded count value in redis as that would reset the expiry for the account lock back to 15 mins. I could add more code to keep track of the failed passwords past the limit without reseting the expiry in redis but that seems redundant in terms of what we are trying to achieve. Im not too precious either way but lets see what the others say.

chrisclayson (approved):
LGTM

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
JHjava: Ok that makes sense 👍

JHjava (approved):

PR #585, opened 2021-09-23
GOVSI-841 - Fix API consistency in response codes

JHjava:
## What?

- Change 200 responses with empty body to 204
- Return Session state when there is a rate limit error when requesting codes
- Add 3 new states for when the rate limit is blocked `MFA_CODE_REQUESTS_BLOCKED`, `PHONE_NUMBER_CODE_REQUESTS_BLOCKED` and `EMAIL_CODE_REQUESTS_BLOCKED`

## Why?

- 204 status codes should be used when the response has No Content but has succeeded.
- To keep responses consistent with other parts of the system


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
```diff
@@ -69,7 +69,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                                 }
                                 LOGGER.info(
                                         "User has successfully Logged in. Generating successful AuthenticateResponse");
-                                return generateApiGatewayProxyResponse(200, "");
+                                return generateApiGatewayProxyResponse(204, "");
```
chrisclayson, line 72: Could we abstract this into a new method, e.g `generateEmptySuccessApiGatewayResponse()` rather than repeat this line?

JHjava: I have updated 👍

chrisclayson (dismissed):

sanisc (approved):

PR #587, opened 2021-09-23
BAU: `VerifyCodeHandler` refactoring

mrwilson:
This is an attempt to break apart the large switch statement in `VerifyCodeHandler`.

It does this by pulling out logic into other places, primarily `ValidationService`, and removing duplication with early returns.

Feedback on the approach appreciated.


sanisc (approved):

sanisc:
LGTM

PR #588, opened 2021-09-23
BAU - Fix update endpoint

JHjava:
## What?

- It's trying to create multiple register aws_api_gateway_resource which it cannot do. So add an extra variable so it doesn't create duplicate aws_api_gateway_resource for register.

## Why?

- To fix the pipeline


chrisclayson (approved):
👍🏻

chrisclayson (approved):
👍🏻

PR #617, opened 2021-09-29
GOVSI-830: Send basic audit payload to SNS topic

alan-gds:
## What?

Send a basic audit payload to the SNS topic, rather than sending a string.

* For the moment, the payload consists only of the event name and a timestamp; we'll add other fields in a subsequent PR
* Although we're using a protobuf container called SignedAuditEvent, we're not yet actually signing it (signature will be dealt with in a later task)

## Why?

Using a structured format should help make the audit logs more searchable.

## Special exercise for reviewer

Can you spot the moment when my patience snapped and I gave up on trying to work out the 'right' way to compose a method reference and a function in Java?  I really wish GDS would let us use a modern language (Kotlin, Scala).

Open to suggestions about renaming the offending variable, or, better yet, what obscure syntax I can use to get rid of the need for it.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/matchers/AuditMessageMatcher.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/matchers/AuditMessageMatcher.java
@@ -2,7 +2,8 @@ package uk.gov.di.authentication.shared.matchers;
 
 import org.hamcrest.Description;
 import org.hamcrest.TypeSafeDiagnosingMatcher;
-import uk.gov.di.audit.AuditPayload;
+import uk.gov.di.audit.AuditPayload.AuditEvent;
+import uk.gov.di.audit.AuditPayload.SignedAuditEvent;
 
 import java.nio.charset.StandardCharsets;
 import java.util.function.Function;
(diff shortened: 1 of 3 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/matchers/AuditMessageMatcher.java
```diff
@@ -0,0 +1,70 @@
...
+    private final T expected;
+
+    private AuditMessageMatcher(
+            String name, Function<AuditPayload.AuditEvent, T> mapper, T expected) {
+        this.name = name;
+        this.expected = expected;
+
+        Function<String, AuditPayload.AuditEvent>
+                annoyingInterimVariableBecauseFPIsANightmareInJava = this::deserialiseAuditEvent;
+        this.mapper = annoyingInterimVariableBecauseFPIsANightmareInJava.andThen(mapper);
```
mrwilson, line 23: Probably the best replacement for this would be to avoid function composition entirely, for readability

```java
this.mapper = input -> mapper.apply(deserialiseAuditEvent(input));
```

alan-gds: Ohhhhhh, yeah, that makes sense.  I kind of ended up looking at it backwards, for various reasons.

alan-gds:
Happy to switch to using direct imports.  I was already on the fence about whether to do it, but I went with what IntelliJ was 'suggesting'.  I'll make a new commit with all the suggested changes.

mrwilson (approved):

PR #620, opened 2021-09-29
GOVSI-820: Trigger Uplift process from `authorize`

chrisclayson:
## What?

- On a completed journey, when auth-code issued, update the CredentialTrustLevel
- If the requested trust level is greater than the stored, update the stored to match the requested
- If the the stored trust level is null, updated the store to be the requested level
- Introduce StateMachine and UserContext to the AuthorisationHandler
- Replace some AuthorisationHandler logic with StateMachine transitions
- Reset session state to NEW rather than create new session
- Add relevant state transitions to StateMachine
- Add new condition to determine if current level of trust is below requested level of trust
- Add new condition to determine if requested level of trust is a given level
- Add new state `UPLIFT_REQUIRED_CM` to indicate an update to `Cm` is required
- Update state machine to transition to `UPLIFT_REQUIRED_CM` if the requested level is `Cm` and the current authenticated level is lower (i.e. `Cl`)

## Why?

We want to trigger MFA if the current session wasn't authenticated with MFA.


sanisc (approved):

sanisc:
Looks good - just need to test this with the different levels via the UI

sanisc (approved):

PR #636, opened 2021-10-01
GOVSI-748: create separate frontend api

dbes-gds:
## What?

Create frontend api in API Gateway, separate from the OIDC API.

Will require changes to frontend and the pipeline in order to repoint the frontend endpoints.

## Why?

Ability to isolate the two APIs.  They can then be deployed separately and have different configurations (such as access controls and rate limits).


JHjava:
Is the reset-password lambda missing from these changes?

dbes-gds:
> Is the reset-password lambda missing from these changes?

This currently lives in the frontend folder for some reason,  but it's actually an account-management lambda.  Thought I could move it as part of another PR to limit this to one thing.

ci/terraform/oidc/outputs.tf
```diff
@@ -2,19 +2,32 @@ output "base_url" {
...
+output "frontend_api_gateway_root_id" {
+  value = aws_api_gateway_rest_api.di_authentication_frontend_api.id
+}
+
 output "token_signing_key_alias" {
   value = local.id_token_signing_key_alias_name
 }
 
-output "frontend_api_key" {
+output "oidc_api_key" {
```
chrisclayson, line 21: Is this actually required?

chrisclayson (approved):

PR #642, opened 2021-10-01
GOVSI-748: create separate frontend api

dbes-gds:
## What?

Create frontend api in API Gateway, separate from the OIDC API.

Will require changes to frontend and the pipeline in order to repoint the frontend endpoints.

## Why?

Ability to isolate the two APIs. They can then be deployed separately and have different configurations (such as access controls and rate limits).


chrisclayson (dismissed):
👍🏻

chrisclayson (approved):

PR #656, opened 2021-10-04
GOVSI-843: add cookie consent field to dynamo and the client-info lambda

dbes-gds:
## What?

Add cookie consent field to dynamo and the client-info lambda.

- Create a new 'CookieConsent' attribute in the client-registry table.
- See stub clients with the new value set to 'TRUE'.
- Update the client-info lambda to return the optional 'cookie_consent' attribute.

## Why?

This flag will be used to share the user's cookie consent across frontend and account management so they only have to enter it once.  The cookie consent banner should only be displayed once when switching between apps (for the first time).

PR to update the frontend to read and act on the flag to follow.


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/ClientInfoResponse.java
```diff
@@ -25,19 +25,24 @@
     @JsonProperty("state")
     private String state;
 
+    @JsonProperty("cookie_consent")
+    private String cookieConsent;
+
     public ClientInfoResponse(
```
sanisc, line 31: i don't think this needs returning back in the response and its something internal

sanisc (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/ClientInfoResponse.java
dbes-gds: removed

PR #667, opened 2021-10-05
GOVSI-824: Cloudwatch Log encryption and retention

chrisclayson:
## What?

- Create a KMS Key for CloudWatch in the shared Terraform
- Use this key for all lambda log groups in OIDC and account management
- Add a 1 day log retention to all log groups
- Override the log retention in production to 5 days

## Why?

Implementing security best practices as advised by `checkov` ahead of ITHC

### DIFF
--- a/ci/terraform/shared/kms.tf
+++ b/ci/terraform/shared/kms.tf
@@ -132,3 +132,65 @@ resource "aws_iam_role_policy_attachment" "attach_audit_signing_key_policy_dynam
   role       = aws_iam_role.dynamo_sqs_lambda_iam_role.name
   policy_arn = aws_iam_policy.audit_signing_key_lambda_kms_signing_policy[0].arn
 }
+
+# Cloudwatch Log Encryption
+data "aws_partition" "current" {}
+
+data "aws_region" "current" {}
+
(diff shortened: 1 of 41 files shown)


sanisc (approved):

ci/terraform/shared/kms.tf
```diff
@@ -131,4 +131,66 @@ resource "aws_iam_role_policy_attachment" "attach_audit_signing_key_policy_dynam
...
+resource "aws_kms_key" "cloudwatch_log_encryption" {
+  description             = "KMS key for Cloudwatch logs"
+  deletion_window_in_days = 30
+  enable_key_rotation     = true
+  policy                  = data.aws_iam_policy_document.cloudwatch.json
+
+  tags = local.default_tags
+}
+
+resource "aws_kms_alias" "cloudwatch_key_alias" {
```
mrwilson, line 193: Is there a particular reason we're creating an alias if we're passing the ARN to cloudwatch?

chrisclayson: Good spot, it came from the sample, but you're right, it's not used, I'll remove.

mrwilson (approved):

PR #668, opened 2021-10-05
GOVSI-843: Add facility to share cookie consent across clients

dbes-gds:
## What?

Add facility to share cookie consent across clients

Uses a new internal-only database flag in the client-registry 'CookieConsent'.  This flag can only be set by the database schema migration scripts in the build pipeline, and cannot be set by external clients using the client registration api.  It is only available for our use to share existing consent across the internal clients.

When the flag is set any existing consent should be shared across the internal applications.  This is done by the 'auth-code' endpoint appending additional parameters to the redirect uri when redirecting:

- cookie_consent=accept - set consent to accept in the target client.
- cookie_consent=reject  - set consent to reject in the target client.
- cookie_consent=not-engaged - either use existing consent or show the cookie banner 

## Why?

The user's cookie consent should be shared across frontend and account management so they only have to provide it once, rather than see the cookie banner several times for what is supposed to look like a single application.

A PR to use the parameters in the frontend applications will follow.

### DIFF
--- a/ci/terraform/shared/stub-rp-clients.tf
+++ b/ci/terraform/shared/stub-rp-clients.tf
@@ -69,5 +69,8 @@ resource "aws_dynamodb_table_item" "stub_rp_client" {
     SubjectType = {
       S = "public"
     }
+    CookieConsentShared = {
+      N = "1"
+    }
   })
 }
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
@@ -29,6 +29,9 @@ import uk.gov.di.authentication.shared.services.SessionService;
 import uk.gov.di.authentication.shared.state.StateMachine;
 import uk.gov.di.authentication.shared.state.UserContext;
 
+import java.net.HttpCookie;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.List;
 import java.util.Map;
 import java.util.NoSuchElementException;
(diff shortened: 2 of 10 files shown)


ci/terraform/shared/stub-rp-clients.tf
```diff
@@ -69,5 +69,8 @@ resource "aws_dynamodb_table_item" "stub_rp_client" {
     SubjectType = {
       S = "public"
     }
+    CookieConsent = {
+      S = "TRUE"
```
sanisc, line 73: should be N = "1"

shared/src/main/java/uk/gov/di/authentication/shared/entity/ClientRegistry.java
```diff
@@ -19,6 +19,7 @@
     private String serviceType;
     private String sectorIdentifierUri;
     private String subjectType;
+    private String cookieConsent;
```
JHjava, line 22: Should this be a boolean?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -190,16 +187,78 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                                            Map.of(
-                                                    ResponseHeaders.LOCATION,
-                                                    authenticationResponse.toURI().toString()));
+
+                            try {
+                                String clientRegistryCookieConsentValue =
+                                        authorizationService.getClientCookieConsentValue(
+                                                authenticationRequest.getClientID());
+                                String cookieConsentParamValue =
+                                        getCookieConsentParamValue(
```
JHjava, line 196: Should we only ever do this check and pass the COOKIE_CONSENT_PARAM_NAME  in the auth response if the CookieConsent is true? and if it is false we can just ignore and continue?

ci/terraform/shared/stub-rp-clients.tf
dbes-gds: updated

shared/src/main/java/uk/gov/di/authentication/shared/entity/ClientRegistry.java
dbes-gds: now converted to boolean

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
dbes-gds: updated to only add the params when CookieConsent is true

JHjava (approved):

PR #686, opened 2021-10-06
GOVSI-865: Create skeleton for audit-processor deployment

mrwilson:
## What?

A "Hello, world" configuration for the audit processors listening to the event stream.


ci/terraform/audit-processors/variables.tf
```diff
@@ -0,0 +1,27 @@
...
+  description = "The name of the AWS role to assume, leave blank when running locally"
+  type        = string
+}
+
+variable "environment" {
+  type = string
+}
+
+variable "aws_region" {
+  default = "eu-west-2"
```
dannyhayes-gds, line 12: I don't have much experience with terraform so it might be fine, but does the type need to be listed for this variable?

PR #692, opened 2021-10-06
GOVSI-892 - Check if an account is locked before processing login att…empt

ghost:
## What?

- If an account is locked we shouldn't bother checking if the credentials the user has entered are valid. We should fail fast and return the state back to the frontend before processing anything else. This is done by email and not session
- Refactor some of the CodeStorageService methods so they can be reused and add tests.
- Add an additional test for the LoginHandler to capture the above scenario

## Why?

- Because a user shouldn't be able to progress to the next screen if their account is locked.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
@@ -72,7 +72,6 @@ public class LoginHandler extends BaseFrontendHandler<LoginRequest>
             LoginRequest request,
             UserContext userContext) {
         LOGGER.info("Request received to the LoginHandler");
-
         try {
             LoginRequest loginRequest = objectMapper.readValue(input.getBody(), LoginRequest.class);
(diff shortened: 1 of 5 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -83,10 +82,22 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
+                        stateMachine.transition(
+                                userContext.getSession().getState(),
+                                USER_ENTERED_INVALID_PASSWORD_TOO_MANY_TIMES,
+                                userContext);
+                sessionService.save(userContext.getSession().setState(nextState));
+                return generateApiGatewayProxyResponse(
+                        200, new LoginResponse(null, userContext.getSession().getState()));
+            }
+
+            if (incorrectPasswordCount == 0 && currentState.equals(ACCOUNT_TEMPORARILY_LOCKED)) {
```
sanisc, line 100: should this be a note
!currentState.equals(ACCOUNT_TEMPORARILY_LOCKED) ?

sanisc (approved):

PR #709, opened 2021-10-08
GOVSI-380: Add subject ID to audit payload

alan-gds:
## What?

Add subject ID to audit payload

## Why?

We want to include this field in the audit log.


alan-gds:
Nope, I've done this wrong!

PR #970, opened 2021-11-03
BAU: Refactor `RedisConnectionService`

chrisclayson:
## What?

- Remove uses of `e.printStackTrace()`
- Throw exception from Redis ops rather than returning a default value
- Add missing integration tests to improve coverage
- Refactor to use common pattern for each Redis operation

## Why?

Sonar analysis show the use `e.printStackTrace()` as a "hotspot" to be addressed.
Returning a default value may hide problems
Test coverage is good.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/RedisConnectionService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/RedisConnectionService.java
@@ -17,6 +17,7 @@ import static io.lettuce.core.support.ConnectionPoolSupport.createGenericObjectP
 public class RedisConnectionService implements AutoCloseable {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(RedisConnectionService.class);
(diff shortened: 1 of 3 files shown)


JHjava (dismissed):

shared/src/main/java/uk/gov/di/authentication/shared/services/RedisConnectionService.java
```diff
@@ -40,63 +43,48 @@ public RedisConnectionService(ConfigurationService configurationService) {
...
+    }
+
+    public String popValue(final String key) {
+        return executeCommand(
+                commands -> {
+                    commands.multi();
+                    commands.get(key);
+                    commands.del(key);
+                    TransactionResult result = commands.exec();
+                    return result.get(0);
```
mrwilson, line 82: We could probably refactor this to use `commands.getdel(key)`?

https://redis.io/commands/getdel/

PR #980, opened 2021-11-04
GOVSI-1064 - Extend state logging

JHjava:
## What?

- Add sessionID to state transitions and invalid state transitions 
- Remove logging the InvalidStateTransition in the Lambda 

## Why?

- By adding the sessionID we can have a better ID of the users behaviour 
- We are currently logging the InvalidStateTransition twice and we should just rely on the StateMachine to log this error

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.shared.state;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import uk.gov.di.authentication.shared.entity.Session;
 import uk.gov.di.authentication.shared.entity.SessionAction;
 import uk.gov.di.authentication.shared.entity.SessionState;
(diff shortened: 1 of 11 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
```diff
@@ -107,17 +107,28 @@ private T transition(T from, A action, Optional<C> context) {
                         > 1) {
             throw handleNoTransitionContext(from, action);
         }
+        final String sessionId;
+        if (context.isPresent() && context.get() instanceof UserContext) {
+            sessionId = ((UserContext) context.get()).getSession().getSessionId();
+        } else {
+            sessionId = null;
+        }
```
mrwilson, line 115: Perhaps something like:
```
var sessionId =  context
  .filter(c -> c instanceof UserContext)
  .map(UserContext.class::cast)
  .map(UserContext::getSession)
  .map(Session::getSessionId)
  .orElse(null);
```

JHjava: A lot nice 👍  Have updated

sanisc (approved):

PR #1007, opened 2021-11-09
GOVSI-1065: Default state transitions

chrisclayson:
## What?

- Store a list of transitions that can occur at any state (normally backward steps), not yet populated
- Add builder for default transitions
- Alter the `transition()` method to look at the the `anyStateTransition` list for a matching action/condition before throwing the normal exception.

## Why?

We have a number of transitions in the state machine that can occur at any state, by allowing a default transition we can keep the code more succinct!


shared/src/main/java/uk/gov/di/authentication/shared/state/StateMachine.java
```diff
@@ -7,12 +7,7 @@
 import uk.gov.di.authentication.shared.entity.SessionState;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
+import java.util.*;
```
sanisc, line 10: should this be a * i think somewhere in the gds way it says no star imports?

chrisclayson: Well spotted ... commit pushed!

sanisc (approved):

PR #1014, opened 2021-11-10
GOVSI-1107: Exclude transition errors from alerting

chrisclayson:
## What?

- Exclude error that start with the phrase `Session attempted invalid transition from` from triggering alerts in the 2nd line channel.

## Why?

These errors are tied with a known, non-critical, issue and causing a lot of noise, when we have addresses the core problems then we will remove the exclusion.


sanisc (approved):

sanisc:
This is temporary until we address the known issues.

PR #1023, opened 2021-11-12
GOVSI-1055: Create new base roles for `oidc` deployment

mrwilson:
Step 1 to break dependency on shared roles


chrisclayson (approved):
👍🏻

PR #1036, opened 2021-11-15
GOVSI-1055: Use new roles in API lambdas

mrwilson:
This PR uses the roles created in a previous PR as the execution role in API lambdas


JHjava:
@mrwilson Any value in testing this in sandpit or should it be OK?

dbes-gds (approved):

mrwilson:
For posterity - tested in sandpit and no obvious failures.

PR #1046, opened 2021-11-16
GOVSI-1105: Convert Update Profile integration tests to new pattern

chrisclayson:
## What?

- Use new test pattern for Update Profile handler integration tests
- Add new base `constructFrontendHeaders` method to build common headers (need to retrofit this to other tests)
- Rename original `constructHeaders` method to `constructOidcHeaders` reflecting its use with OIDC API

## Why?

We're converting the pattern for all integration tests

## Related PRs

#1026 
#1027 
#1031 
#1037 and more


chrisclayson:
Superceded

PR #1062, opened 2021-11-18
Revert "GOVSI-1105: Convert User Exists integration tests to new pattern"

chrisclayson:
Reverts alphagov/di-authentication-api#1048


chrisclayson:
Not required

PR #1093, opened 2021-11-23
GOVSI-1143 - Pass persistent-session-id to the audit service

JHjava:
## What?

- Create helper method to extract di-persistent-session-id from cookie
- When the persistent-session-id is sent as a header to an endpoint, extract it and pass it to the audit
- When the persistent-session-id is passed in the cookie (EG Auth Code) extract it from the cookie and pass it to audit

## Why?
When we receive a request via a 302 redirect, we want to extract the di-persistent-session-id from the cookie if it exists. If it doesn't exist then we will set it to 'unknown'. This is so that we have the capability to audit it
- The audit does not actually do anything with the persistent-session-id yet but this work is required before we are able to do that


mrwilson:
This is looking good - adding a new field to the protobuf schema and regenerating the java classes needs to be its own deploy so we don't accidentally clash versions in production.

I think we merge this, make and merge the schema change, and then make a final merge to add the fields to the audit object 👍🏻

chrisclayson (approved):

PR #1094, opened 2021-11-23
GOVSI-1114: Add SQS extension to manage queue dependency

chrisclayson:
## What?

- Add a new SqsQueueExtension to create queue dependency before running tests
- Ensure each test class gets its own queue
- Clear queue before each test
- Refactor integration tests for handlers that send to notifications queue to assert on queue contents rather than use the NotifyStub to wait for notify requests.

## Why?

## Related PRs


JHjava (approved):
👍

PR #1102, opened 2021-11-24
BAU - - Check that current password is not null

JHjava:
## What and Why?

- Before we check that the users new password is not the same as the old password, we should check that the old password is not null. If it is null, we know that it is a migrated user which is resetting their password.


chrisclayson (approved):
👍🏻

dbes-gds (approved):

PR #1116, opened 2021-11-26
GOVSI-1114: SSM extension

chrisclayson:
## What?

- Create an SSM Parameter Store extension to manage parameter store dependencies
- Use the extension to create resources required by the ConfigurationService

## Why?

Improves test coverage by using Localstack to provide Parameter Store values to integration tests.

### DIFF
--- a/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/ParameterStoreExtension.java
+++ b/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/ParameterStoreExtension.java
@@ -1,0 +1,47 @@
+package uk.gov.di.authentication.sharedtest.extensions;
+
+import com.amazonaws.client.builder.AwsClientBuilder;
+import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagement;
+import com.amazonaws.services.simplesystemsmanagement.AWSSimpleSystemsManagementClient;
+import com.amazonaws.services.simplesystemsmanagement.model.PutParameterRequest;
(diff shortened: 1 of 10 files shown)


shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/ParameterStoreExtension.java
```diff
@@ -0,0 +1,47 @@
...
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+import java.util.Map;
+
+import static com.amazonaws.services.simplesystemsmanagement.model.ParameterType.SecureString;
+
+public class ParameterStoreExtension implements BeforeAllCallback {
+
+    protected static final String REGION = System.getenv().getOrDefault("AWS_REGION", "eu-west-2");
+    protected static final String LOCALSTACK_ENDPOINT =
```
mrwilson, line 17: I think these two lines (and the creation of the `AwsClientBuilder.EndpointConfiguration`) are duplicated across all our extensions.

Can we pull out a `localStackEndpointConfiguration()` static method to share among them?

chrisclayson: They are, I was going to do that in a separate change. But I'll add it to this this one.

chrisclayson: Commit pushed!

mrwilson (approved):

PR #1118, opened 2021-11-26
Bump Spotless to 6.0.0

jamietanna:
As part of the upgrade, all projects now need the `repositories`
directive.

In this case, all but the top-level `build.gradle` supported this, so we
can add it to align with all the other projects.


JHjava:
Closing this has this dependency was updated on the original dependabot PR.

PR #1125, opened 2021-11-29
GOVSI-1141: log session id in account management handlers

dbes-gds:
## What?

Log session id in account management handlers.
The session id is passed as a header to each api call from the account management frontend.
Include in audit events.

## Why?

Ability to view a single session across both frontend and account management in the logs and audit trail.

## Related PRs

https://github.com/alphagov/di-authentication-account-management/pull/249

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
@@ -13,13 +13,16 @@ import uk.gov.di.accountmanagement.entity.AuthenticateRequest;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.PersistentIdHelper;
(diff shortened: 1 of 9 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
```diff
@@ -56,7 +58,12 @@ public APIGatewayProxyResponseEvent handleRequest(
         return isWarming(input)
                 .orElseGet(
                         () -> {
-                            LOGGER.info("Request received to the AuthenticateHandler");
+                            String sessionId =
+                                    RequestHeaderHelper.getHeaderValueOrElse(
+                                            input.getHeaders(), SESSION_ID_HEADER, "");
+                            LOGGER.info(
```
JHjava, line 64: To save you from appending it to each log line, you could use the `attachSessionIdToLogs`.

dbes-gds: Looks like a good idea, don't have the actual session in AM, but could add a similar method that just adds the session id?

JHjava (approved):

PR #1135, opened 2021-11-30
GOVSI-1106: log updated session-id in authorisation handler

dbes-gds:
## What?

Log updated session-id in authorisation handler.

The authorisation is where the session-id actually gets created so the session-id can change within the lambda itself.  Change session-id logging so that the session-id logged by the logging context is the session-id most recently allocated.

## Why?

Old session-id were being logged when new ones had been created, so there was a difference between what was in the log messages and what was in the session-id field.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/LogLineHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/LogLineHelper.java
@@ -12,4 +12,11 @@ public class LogLineHelper {
     public static void attachSessionIdToLogs(String sessionId) {
         ThreadContext.put("sessionId", sessionId);
     }
+
+    public static void updateAttachedSessionIdToLogs(String sessionId) {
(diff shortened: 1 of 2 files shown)


mrwilson (approved):

shared/src/main/java/uk/gov/di/authentication/shared/helpers/LogLineHelper.java
```diff
@@ -12,4 +12,11 @@ public static void attachSessionIdToLogs(Session session) {
     public static void attachSessionIdToLogs(String sessionId) {
         ThreadContext.put("sessionId", sessionId);
     }
+
+    public static void updateAttachedSessionIdToLogs(String sessionId) {
+        if (ThreadContext.containsKey("sessionId")) {
+            ThreadContext.remove("sessionId");
```
jamietanna, line 18: Is there any reason we're doing it this way instead of i.e. using the MDC?

PR #1138, opened 2021-11-30
BAU: GitHub action services vs Docker Compose

chrisclayson:
## What?

- Use Github action services instead of Docker Compose

## Why?

Compare and contrast build speeds


chrisclayson:
Superseded by #1172

PR #1164, opened 2021-12-06
BAU: Build with `-Werror` and `-Xlint:deprecation`, then fix resulting problems.

mrwilson:
Does what it says on the tin, to tighten up our code.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -107,6 +107,14 @@ subprojects {
     }
 }
 
+allprojects {
+    gradle.projectsEvaluated {
+        tasks.withType(JavaCompile) {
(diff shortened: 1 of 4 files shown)


dbes-gds (dismissed):

JHjava:
Had a look at the failing tests and they seem to be caused by a bug with AWS KMS api. 

Similar issue raised - https://github.com/aws/aws-sdk/issues/125

mrwilson:
@JHjava I've resolved it for now by suppressing the warnings (as it's only test code) - I'm looking at a proper fix in the meantime.

JHjava (approved):

PR #1165, opened 2021-12-06
GOVSI-1165: add IPV authorisation handler

dbes-gds:
## What?

Add IPV authorisation handler.

Create a new endpoint for the frontend to call which redirects to IPV /authorize by building an authorization request, sending the appropriate redirect back to the frontend.

The lambda has been created in a new java module called 'ipv-api' but is effectively a frontend api call.  'BaseFrontendHandler' has been moved to shared so it can be used in the new handler.

The endpoint is only deployed to build for the time being pending proving.

## Why?

If a service requests a level of confidence requiring ID then we should redirect to the IPV service once the user is authenticated.

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVAuthorisationHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVAuthorisationHandler.java
@@ -1,0 +1,128 @@
+package uk.gov.di.authentication.ipv.lambda;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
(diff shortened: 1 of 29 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVAuthorisationHandler.java
```diff
@@ -0,0 +1,128 @@
...
+                    userContext.getSession().getSessionId(),
+                    userContext
+                            .getClient()
+                            .map(ClientRegistry::getClientID)
+                            .orElse(AuditService.UNKNOWN),
+                    AuditService.UNKNOWN,
+                    request.getEmail(),
+                    IpAddressHelper.extractIpAddress(input),
+                    AuditService.UNKNOWN,
+                    PersistentIdHelper.extractPersistentIdFromHeaders(input.getHeaders()));
```
JHjava, line 111: Would this be extracted via the cookie rather than the headers?

JHjava (approved):
One non blocker comment. 👍

PR #1172, opened 2021-12-07
GOVSI-1158: Optimise Github Actions

chrisclayson:
## What

- Revert alphagov/di-authentication-api#1149
- Parallelise unit and integration tests into separate jobs and cache resulting reports
- Add build job and cache build
- Use GitHub actions service containers to save on Docker Compose build/up time

## Why

The problem requiring #1149 has been resolved.

Other changes required to keep total test time at around 5 mins (if not under).

### DIFF
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -74,6 +74,12 @@ jobs:
           name: account-management-integration-test-reports
           path: account-management-integration-tests/build/reports/tests/test/
           retention-days: 5
+      - name: Run SonarCloud Analysis
+        if: ${{ github.actor != 'dependabot[bot]' }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
+        run: ./gradlew sonarqube
       - name: Upload Docker Container Logs
         uses: actions/upload-artifact@v2
         if: failure()
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -30,7 +30,8 @@ jobs:
       - name: Run Spotless
         run: ./gradlew --no-daemon spotlessCheck
 
-  run-tests:
+
+  build:
     runs-on: ubuntu-latest
     steps:
       - name: Check out repository code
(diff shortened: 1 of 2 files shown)


.github/workflows/pre-merge-checks.yml
```diff
@@ -30,8 +30,42 @@ jobs:
...
+        uses: actions/cache@v2
+        with:
+          path: |
+            .gradle/
+            */build/
+            !*/build/reports
+            !*/build/jacoco
+          key: ${{ runner.os }}-build-${{ github.sha }}
+      - name: Run Build
+        run: ./gradlew --parallel build -x test -x account-management-integration-tests:test -x spotlessApply -x spotlessCheck
```
jamietanna, line 63: Why are we removing spotless?

jamietanna: Have we looked at setting up Gradle parallelisation in the `gradle.properties`, as then it works locally, too?:

```
org.gradle.parallel=true
```

Also maybe JUnit5 parallelisation too, if not already done?

(via https://www.jvt.me/posts/2021/03/11/gradle-speed-parallel/)

.github/workflows/pre-merge-checks.yml
```diff
@@ -30,8 +30,42 @@ jobs:
...
+          distribution: 'adopt'
+      - name: Cache Gradle packages
+        uses: actions/cache@v2
+        with:
+          path: |
+            ~/.gradle/caches
+            ~/.gradle/wrapper
+          key: ${{ runner.os }}-gradle-${{ hashFiles('**/*.gradle*', '**/gradle-wrapper.properties') }}
+          restore-keys: |
+            ${{ runner.os }}-gradle-
```
jamietanna, line 52: is this missing something at the end?

.github/workflows/pre-merge-checks.yml
chrisclayson: We're not, the spotless check happens in the "style-checks" job, we just exclude from the build and test jobs to save ourselves some seconds.

.github/workflows/pre-merge-checks.yml
chrisclayson: No, this is by design. I believe this was, originally, lifted straight from the docs (https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows), it does partial matching to pick up a similar cache if the one specified in `key` doesn't exist.

jamietanna: Ah nice, TIL!

chrisclayson: Though, actually, just spotted in the docs, we can simplify this as the `setup-java` step has it built it.

.github/workflows/pre-merge-checks.yml
chrisclayson: Sadly, on the `org.gradle.parallel=true` thing, I did consider it, but this could cause us some issues with the integration tests (currently). When running locally, using the `pre-commit.sh` it uses the `--parallel` switch for the build and unit tests, but not for integration tests.

However, I believe its all a little moot as I believe we only get one CPU in GitHub actions, so it can't even run the unit tests in parallel, I actually meant to take that switch out.

jamietanna: Hmm that's a shame. Hopefully we can solve that Integration Test issue so we can re-enable parallelisation everywhere - I've suffered in the past of trying to unpick it after many months of the codebase implicitly assuming that tests execute sequentially

chrisclayson: I was wrong, its actually 2-cpu workers, I leave the flag there for the mo.

sanisc (approved):

PR #1174, opened 2021-12-07
GOVSI-1145 - Validate cookie values

JHjava:
## What?

- If the cookie_consent value in Authorize is not valid then don't pass the cookie_consent parameter to the frontend
- If the cookie_consent value in AuthCode is not valid then set the cookie_consent parameter to 'not-engaged'
- Tidy up AuthCode tests and remove ConfigurationService from AuthCode

## Why?

- To ensure we are checking cookie_consent values at the first opportunity and only passing only values that are valid

### DIFF
--- a/shared/src/test/java/uk/gov/di/authentication/shared/services/AuthorizationServiceTest.java
+++ b/shared/src/test/java/uk/gov/di/authentication/shared/services/AuthorizationServiceTest.java
@@ -16,6 +16,9 @@ import org.hamcrest.MatcherAssert;
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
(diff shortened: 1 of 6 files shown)


shared/src/test/java/uk/gov/di/authentication/shared/services/AuthorizationServiceTest.java
```diff
@@ -444,6 +447,28 @@ void shouldGetPersistentCookieIdFromExistingCookie() {
         assertEquals(persistentSessionId, "some-persistent-id");
     }
 
+    @ParameterizedTest
```
chrisclayson, line 450: Could simplify these into a single test and make the `MethodSource` pass the expected value too.

JHjava: Good suggestion. This has now been updated.

chrisclayson (approved):

PR #1188, opened 2021-12-09
GOVSI-1166 - Create service to parse and validate AuthorisationResponse from IPV

JHjava:
## What?

- Check the response contains both the state and code query parameters. Eventually we will be carrying out more checks on the state param.
- Check the if the response contains an error. If so, we'll need to treat this differently. Probably by returning it to the frontend.
- Add additional tests and integration tests

## Why?

- To ensure that the response meets the validation requirements

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -4,14 +4,17 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.ErrorObject;
 import org.apache.http.client.utils.URIBuilder;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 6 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -36,6 +48,19 @@ public APIGatewayProxyResponseEvent handleRequest(
                 .orElseGet(
                         () -> {
                             LOG.info("Request received to IPVCallbackHandler");
+                            Optional<ErrorObject> errorObject =
+                                    responseService.validateResponse(
+                                            input.getQueryStringParameters());
+                            if (errorObject.isPresent()) {
+                                LOG.error("Error in IPV AuthorisationResponse");
```
jamietanna, line 55: Should we log the error that's present?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -36,6 +48,19 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                // TODO - Do something with this error object
+                                return new APIGatewayProxyResponseEvent()
+                                        .withStatusCode(302)
+                                        .withHeaders(
+                                                Map.of(
+                                                        ResponseHeaders.LOCATION,
+                                                        buildRedirectUri()));
```
jamietanna, line 62: should this include information about what's up? Or is that part of the `TODO`?

chrisclayson (approved):

PR #1215, opened 2021-12-17
AUTH-1253: log aws request id and psid to logging context

dbes-gds:
## What?

Log aws request id and psid to logging context.

## Why?

Use the aws request id to exclude smoke tester traffic from queries and reports.

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -46,6 +46,9 @@ import static uk.gov.di.authentication.oidc.entity.RequestParameters.GA;
 import static uk.gov.di.authentication.shared.entity.SessionAction.USER_HAS_STARTED_A_NEW_JOURNEY;
 import static uk.gov.di.authentication.shared.entity.SessionAction.USER_HAS_STARTED_A_NEW_JOURNEY_WITH_LOGIN_REQUIRED;
 import static uk.gov.di.authentication.shared.entity.SessionState.INTERRUPT_STATES;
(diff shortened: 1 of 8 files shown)


jamietanna (commented):
Is this something we want to add tests for?

Also, is there any reason we're using this means over `MDC`?

dbes-gds:
@jamietanna will add some tests.  MDC is just a wrapper around ThreadContext so I don't think makes much difference.

mrwilson:
@jamietanna MDC is an slf4j thing, as I understand, which we aren't using. ThreadContext is the equivalent log4j2 concept.

> Log4j 2 continues with the idea of the MDC and the NDC but merges them into a single Thread Context. The Thread Context Map is the equivalent of the MDC and the Thread Context Stack is the equivalent of the NDC.

https://logging.apache.org/log4j/2.x/manual/thread-context.html

@dbes-gds The `CaptureLoggingExtension` should have all the magic needed to do assertions against logging.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -112,6 +115,8 @@ public APIGatewayProxyResponseEvent handleRequest(
                                     ipAddress,
                                     AuditService.UNKNOWN,
                                     persistentSessionId);
+                            attachLogFieldToLogs(PERSISTENT_SESSION_ID, persistentSessionId);
```
mrwilson, line 118: We need to add `.flatMap(InputSanitiser::sanitiseBase64)` to the method `getExistingOrCreateNewPersistentSessionId`. We did this in `PersistentIdHelper`.

Maybe there's some refactoring to do here.

JHjava (approved):

PR #1268, opened 2022-01-07
AUTH-1228: Switch account management lambdas to new security groups

chrisclayson:
## What?

- Switch the SQS Notify lambda to the "egress allowed" group
- Switch all other lambda to "VPC only" group

## Why?

We wish to restrict egress from the VPC to only those lambdas that need. We previous created the security groups in 
#1255, so we need to switch over.

## Related PRs

#1255 
#1250 (this was the same change for the OIDC server VPC)

### DIFF
--- a/ci/terraform/account-management/api-gateway.tf
+++ b/ci/terraform/account-management/api-gateway.tf
@@ -85,7 +85,7 @@ resource "aws_lambda_function" "authorizer" {
   timeout          = 30
   memory_size      = 2048
   vpc_config {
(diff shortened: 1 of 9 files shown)


mrwilson (approved):

alan-gds (commented):
Do we also need this change to be made in ci/terraform/account-management/redis.tf?

chrisclayson:
> Do we also need this change to be made in ci/terraform/account-management/redis.tf?

Not at this stage that is in next PR #1269.

alan-gds (approved):
Excellent, have a tick then.

PR #1344, opened 2022-01-26
AUTh-1315 - Create SPOTResponse lambda and new Dynamo table

JHjava:
## What?

- Create a SPOTResponse lambda which will read from an SNS topic, deserialize the message into the SPOTResponse object and write it to a Dynamo table. 
- Create a new Dynamo table named `spot-credential` which will hold the credential returned from SPOT. It will be stored against the Subject identifier, which we can extract from the access token (in a separate Lambda) to retrieve the credential. 
- Create a new Lambda role specific to the SPOTResponse lambda, so that Dynamo write permissions are limited to this lambda. We will set up a new role for the lambda required to read from this table. 
- The Dynamo table will be created in all environments but the lambda will only be created in environments where `ipv_api_enabled` is set to true. 

## Why?

- So we can retrieve information from SPOT, process it and write it to Dynamo. This data in Dynamo will then be read by a new Lambda (not yet created) which will access it using the subject identifier in the given access token.

## Other comments

- I have set the `prevent_destroy` on the new dynamo table for false for now in case changes are required. 
- I will need to set an expiration on the data in the SPOT table. 
- The SNS topic does not yet exist which is why no further permissions have been configured with this lambda.

### DIFF
--- a/ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandlerTest.java
+++ b/ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandlerTest.java
@@ -1,0 +1,51 @@
+package uk.gov.di.authentication.ipv.lambda;
+
+import com.amazonaws.services.lambda.runtime.events.SNSEvent;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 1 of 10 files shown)


ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandlerTest.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.Optional;
+
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+
+class SPOTResponseHandlerTest {
+
+    @RegisterExtension
+    public final CaptureLoggingExtension logging =
```
chrisclayson, line 23: Is this used anywhere?

chrisclayson (approved):

PR #1351, opened 2022-01-27
AUTH-1208: Pull dependencies from Artifactory instance rather than Maven Central

mrwilson:
We want to pull dependencies from the artifactory cache rather than directly from Maven central or gradle


alan-gds (approved):
LGTM

PR #1358, opened 2022-01-28
BAU: Refactoring 202220128 (`NotificationType`)

mrwilson:
Each `NotificationType` constant knows its template environment variable, and corresponding value of the lookup. We don't need a separate method to look those up.

This pushes that knowledge into the `NotificationType` enum and adds a `TemplateAware` interface to make testing interactions simpler by removing the need for a `System.getenv()` call.


JHjava (approved):
👍  This could be done on the AM Notification too

mrwilson:
@JHjava Good shout, I'll raise a future PR for that

PR #1367, opened 2022-02-01
AUTH-1281: Add customer support link for Notify templates

alan-gds:
## What?

Pass the `customer-support-link` personalisation variable to more of our Notify templates.

## Why?

We are going to update several templates to include this link at the bottom, so that customers know how to give us feedback.

PR #1368, opened 2022-02-02
AUTH-1281: create new configuration for the contact-us route

dbes-gds:
## What?

Create new configuration for the contact-us route.
Update NotificationHandler to send the contact-us-link to the account create, email confirmation and password reset templates in frontend api.

## Why?

The contact-us route is separate from the support route and goes to a different page.
Three templates need to be updated to link to the contact-us page.

## Related PRs

#1367


alan-gds (approved):
One small quibble but basically this is all fine. 👍

PR #1395, opened 2022-02-09
GOVSI-1063 - Remove unused dynamo policies 

JHjava:
## What?

- Remove unused dynamo policies 
- Give lambda roles a unique name

## Why?

 - We heading towards separate roles for our lambdas. This is handled in lambda-roles.tf and therefore the policies in dynamo.tf are not used and can be removed


ci/terraform/shared/dynamodb.tf
```diff
@@ -166,58 +166,3 @@ resource "aws_dynamodb_table" "spot_credential_table" {
...
-      aws_dynamodb_table.user_credentials_table.arn,
-      aws_dynamodb_table.user_profile_table.arn,
-      "${aws_dynamodb_table.user_profile_table.arn}/index/*",
-      "${aws_dynamodb_table.user_credentials_table.arn}/index/*",
-      aws_dynamodb_table.client_registry_table.arn,
-    ]
-  }
-}
-
-resource "aws_iam_policy" "lambda_dynamo_policy" {
```
chrisclayson, line 198: This is still used in `data-transfer.tf` line 97.

chrisclayson (approved):

PR #1396, opened 2022-02-09
AUTH-1174 - Create separate lambda role for the Client Registry lambdas

JHjava:
## What?

- Create a 2 new Lambda role which will be used by the 2 client registration lambdas UpdateClientConfigHandler and ClientRegistrationHandler. These Lambdas only require the KMS signing policy and Lambda SNS policy as part of the audit and also a new dynamo_client_registry_write_policy.
- Create a new policy called dynamo_client_registry_write_policy which will only give the UpdateClientConfigHandler and ClientRegistrationHandler permissions to the Client Registry table. Before the had access to the user tables which they did not need.
- Move Dynamo policies into a separate file as it was getting too big
- Remove localstack condition on IAM policies

## Why?
- To give lambas permissions to the things they need

### DIFF
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -1,0 +1,135 @@
+data "aws_dynamodb_table" "user_credentials_table" {
+  name = "${var.environment}-user-credentials"
+}
+
+data "aws_dynamodb_table" "user_profile_table" {
+  name = "${var.environment}-user-profile"
+}
+
+data "aws_dynamodb_table" "client_registry_table" {
+  name = "${var.environment}-client-registry"
+}
+
+data "aws_dynamodb_table" "spot_credential_table" {
+  name = "${var.environment}-spot-credential"
+}
+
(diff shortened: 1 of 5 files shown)


ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -0,0 +1,135 @@
...
+data "aws_dynamodb_table" "client_registry_table" {
+  name = "${var.environment}-client-registry"
+}
+
+data "aws_dynamodb_table" "spot_credential_table" {
+  name = "${var.environment}-spot-credential"
+}
+
+data "aws_iam_policy_document" "dynamo_access_policy_document" {
+  count = var.use_localstack ? 0 : 1
```
chrisclayson, line 18: Personally, I would ditch all of these now, we don't need them as we don't regularly deploy to localstack and its OK to deploy to localstack anyway (they just get ignored).

JHjava: Good idea. I have updated @chrisclayson

PR #1397, opened 2022-02-09
AUTH-1174 - Create new Lambda role for authorize

JHjava:
## What?

- Limit the permissions given to the authorize lambda but creating a new role specific to it.
- Create new policies for read access to the client registration and user tables. The majority of the lambdas only need READ access.

## Why?

- Authorize had a lot of permissions which it didn't require. Ensure that we only give it the permissions that the Authorize lambda requires.

### DIFF
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -41,6 +41,28 @@ data "aws_iam_policy_document" "dynamo_access_policy_document" {
   }
 }
 
+data "aws_iam_policy_document" "dynamo_user_read_policy_document" {
+  statement {
+    sid    = "AllowAccessToDynamoTables"
+    effect = "Allow"
+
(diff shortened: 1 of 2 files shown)


ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -110,6 +149,22 @@ resource "aws_iam_policy" "dynamo_client_registry_access_policy" {
   policy = data.aws_iam_policy_document.dynamo_client_registration_policy_document.json
 }
 
+resource "aws_iam_policy" "dynamo_client_registry_read_access_policy" {
+  name_prefix = "dynamo-client-registry-read-policy"
+  path        = "/${var.environment}/oidc-default/"
```
mrwilson, line 154: Should this be `oidc-authorize` rather than `oidc-default`?

chrisclayson (dismissed):

PR #1405, opened 2022-02-10
BAU: Return optional UserProfile from AuthenticationService

alan-gds:
## What?

First part of an incremental refactoring to make the return type of AuthenticationService#getUserProfileByEmail an Optional rather than returning nulls.  At this stage we are only refactoring cases where nullability was already being taken into account.

## Why?

The original method returns null if a user does not exist for the given email address.  This shouldn't happen in the ordinary flow but there are corner cases where it could occur if requests are sent outside of the ordinary frontend flow.  It's better to reflect the possibility of a user profile not being found by making the return type an Optional; we do have to handle its potential absence in some cases.

## Note

The original version of the method ought to be deprecated, but as we are no longer allowed to have unresolved warnings, marking it as deprecated is too awkward to be worthwhile. :(


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/RemoveAccountHandler.java
```diff
@@ -88,17 +89,17 @@ public APIGatewayProxyResponseEvent handleRequest(
 
                                 String email = removeAccountRequest.getEmail();
 
-                                UserProfile userProfile =
-                                        authenticationService.getUserProfileByEmail(email);
+                                Optional<UserProfile> userProfile =
+                                        authenticationService.getUserProfileByEmailMaybe(email);
```
dbes-gds, line 93: could use .map then orElseThrow to remove the get() on optional?

dbes-gds (approved):

PR #1419, opened 2022-02-14
AUTH-1357: Publish product-level metrics to CloudWatch

mrwilson:
## What?

This PR adds the capability for a handler, through `CloudWatchMetricsService`, to publish metrics into CloudWatch for analysis and dashboards

## Why?

We get a lot of platform/infrastructure-level metrics in CloudWatch for "free" but anything that's specific to our domain needs to be pushed by us.

### DIFF
--- a/frontend-api/build.gradle
+++ b/frontend-api/build.gradle
@@ -13,9 +13,10 @@ dependencies {
             configurations.jackson,
             configurations.nimbus,
             configurations.sqs,
+            configurations.cloudwatch,
             configurations.bouncycastle,
-            project(":shared"),
-            "software.amazon.awssdk:cloudwatch:2.17.128"
+            project(":shared")
+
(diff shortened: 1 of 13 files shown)


frontend-api/build.gradle
```diff
@@ -14,7 +14,8 @@ dependencies {
             configurations.nimbus,
             configurations.sqs,
             configurations.bouncycastle,
-            project(":shared")
+            project(":shared"),
+            "software.amazon.awssdk:cloudwatch:2.17.128"
```
JHjava, line 18: This could be managed at the top level build.gradle and brought in similar to others. It also shares a SDK with SQS so that version number could be put in a dependency version so there's a single place where it is upgraded.

JHjava (commented):
Change looks good. One minor comment about managing the new dependency.

JHjava (approved):
👍

PR #1443, opened 2022-02-17
AUTH-1293: Code signing for API warmers

chrisclayson:
## What?

- Add AWS code signing configuration to each of the API warmers lambda functions
- Change the way in which the deployment job determines the filename for the lambda ZIP file, when using AWS Signer the filename is not consistent. It will be the only ZIP file in the release directory though.
- Move the `code_signing_config_arn` directive in the endpoint lambda
to the same position in the resource as all other lambdas (just after
the S3 location), keep things consistent
- Add code signing profile to SPoT response lambda as this was missed.

## Why?

We are moving away from Github releases and GPG signing to S3 and AWS signer

## Related PRs

https://github.com/alphagov/di-infrastructure/pull/184


alan-gds (approved):
Looks good!

PR #1446, opened 2022-02-22
GOVSI-1063 - API changes to support state machine in frontend

JHjava:
## What?

- Make changes to the OIDC and Frontend API to support the move of the state machine to the frontend
- Remove the state transitions and the state machine itself
- Where we return error states in the response, replace that with an ERROR response

## Why?

- We can give the user a better experience by moving the state machine to the frontend and can prevent a lot of the invalid state transition exceptions. 
- We can reduce the amount of code in the backend and simplify it

## Related PRs
https://github.com/alphagov/di-authentication-frontend/pull/507


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SendNotificationHandler.java
```diff
@@ -218,100 +182,80 @@ private APIGatewayProxyResponseEvent handleNotificationRequest(
...
             UserContext userContext, NotificationType notificationType)
             throws ClientNotFoundException {
         if (configurationService.isTestClientsEnabled()) {
             LOG.warn("TestClients are ENABLED");
         } else {
-            return false;
+            return true;
         }
         String emailAddress = userContext.getSession().getEmailAddress();
         return userContext
```
dbes-gds, line 261: need to negate this return value

sanisc (approved):

PR #1460, opened 2022-02-23
AUTH-1169: Add integration tests for SendOtpNotificationHandler

alan-gds:
## What?

Add integration tests for SendOtpNotificationHandler

## Why?

Adds missing test coverage.


account-management-integration-tests/build.gradle
```diff
@@ -16,6 +16,7 @@ dependencies {
             configurations.lambda,
             configurations.lettuce,
             "com.fasterxml.jackson.datatype:jackson-datatype-jsr310:${dependencyVersions.jackson_version}",
+            "com.googlecode.libphonenumber:libphonenumber:8.12.42",
```
mrwilson, line 19: I think we use this import in a number of places, we should probably pull it up into `build.gradle` as a configuration like lettuce

alan-gds: I thought about this at the time but forgot to come back to it; now done.

JHjava (approved):

PR #1479, opened 2022-02-28
AUTH-1397: use dynamo transactions for remove account

dbes-gds:
## What?

Use dynamo transactions for remove account.

- Make remove account atomic using the dynamo db transactions api.
- Some refactoring of schema data to simplify the DynamoService and make using the transactions api easier.

## Why?

Removing an account means removing an item from two separate tables.  If one remove fails then the database is left in an incorrect state and the application fails when the user tries to login.

## Related

https://github.com/alphagov/di-authentication-api/pull/1431

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBSchemaHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBSchemaHelper.java
@@ -1,0 +1,61 @@
+package uk.gov.di.authentication.shared.dynamodb;
+
+import com.amazonaws.services.dynamodbv2.AmazonDynamoDB;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapper;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBMapperConfig;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+import com.amazonaws.services.dynamodbv2.model.Delete;
+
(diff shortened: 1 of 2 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBSchemaHelper.java
```diff
@@ -0,0 +1,61 @@
...
+        this.dynamoDB = dynamoDB;
+        this.environment = environment;
+    }
+
+    public enum Table {
+        USER_CREDENTIALS_TABLE("user-credentials", "Email"),
+        USER_PROFILE_TABLE("user-profile", "Email");
+
+        private String tableName;
+        private String key;
```
JHjava, line 26: These can both be `final`

JHjava: We could be explicit with the key as `primaryKey`

JHjava (commented):
Couple of minor comments/suggestions but overall looks good 👍

shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBSchemaHelper.java
dbes-gds: updated

JHjava (dismissed):

JHjava (approved):

PR #1530, opened 2022-03-15
AUTH-1439: Add new notification type to reset password with a code

chrisclayson:
## What?

- Add `RESET_PASSWORD_WITH_CODE` to `NotificationType` enum
- Handle new notification type in frontend API `NotificationHandler` lambda.
- Add the `RESET_PASSWORD_WITH_CODE_TEMPLATE_ID` key into the template map with a default value of the template ID from Notify test account
- Add same key into `production-overrides.tfvars` with the template ID from the production Notify account

## Why?

Which are changing the password reset flow to use a code rather than link that is sent.


chrisclayson:
Closing, new PR to raised.

PR #1537, opened 2022-03-16
AUTH-1397: use DynamoDB transactions for update email

dbes-gds:
## What?

Use DynamoDB transactions for update email.

## Why?

Use transactions to maintain database integrity given that deletes and creates are required in two separate tables for the operation to succeed.

DynamoDB mappers only support transactions on a single table if table overrides are in use.  This means using a different api which requires converting fields to AttributeValue in order to use transactions.

## Related PRs

#1479

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
@@ -1,0 +1,7 @@
+package uk.gov.di.authentication.shared.dynamodb;
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+
+public interface DynamoDBItem {
+    public java.util.Map<String, AttributeValue> toItem();
+}
--- a/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
@@ -3,5 +3,5 @@ package uk.gov.di.authentication.shared.dynamodb;
 import com.amazonaws.services.dynamodbv2.model.AttributeValue;
 
 public interface DynamoDBItem {
-    public java.util.Map<String, AttributeValue> toItem();
+    java.util.Map<String, AttributeValue> toItem();
 }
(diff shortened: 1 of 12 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.di.authentication.shared.dynamodb;
+
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+
+public interface DynamoDBItem {
+    public java.util.Map<String, AttributeValue> toItem();
```
JHjava, line 6: `public` is not required here.

shared/src/main/java/uk/gov/di/authentication/shared/entity/UserProfile.java
```diff
@@ -160,4 +181,39 @@ public UserProfile setSalt(byte[] salt) {
...
+        if (getCreated() != null)
+            attributes.put(ATTRIBUTE_CREATED, new AttributeValue(getCreated()));
+        if (getUpdated() != null)
+            attributes.put(ATTRIBUTE_UPDATED, new AttributeValue(getUpdated()));
+        if (getTermsAndConditions() != null)
+            attributes.put(
+                    ATTRIBUTE_TERMS_AND_CONDITIONS, getTermsAndConditions().toAttributeValue());
+        if (getClientConsent() != null) {
+            Collection<AttributeValue> consents = new ArrayList<>();
+            getClientConsent().stream().forEach(c -> consents.add(c.toAttributeValue()));
```
JHjava, line 208: You don't need the `stream` here so it can be ` getClientConsent().forEach(c -> consents.add(c.toAttributeValue()));`

JHjava (commented):
Looks good. Couple of minor comments

shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/DynamoDBItem.java
dbes-gds: updated

shared/src/main/java/uk/gov/di/authentication/shared/entity/UserProfile.java
dbes-gds: updated

JHjava (approved):

PR #1586, opened 2022-03-24
AUTH-1347 - Send the necessary information in SPOT payload

JHjava:
## What?

- Based on what we know now and what is documented in this RFC https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0013-spot-request-response-structures.md, form the SPOT request payload.

## Why?

- So we can send the relevant information to SPOT

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTClaims.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTClaims.java
@@ -1,0 +1,27 @@
+package uk.gov.di.authentication.ipv.entity;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class SPOTClaims {
+
+    @JsonProperty(value = "vot")
+    private String vot;
+
+    @JsonProperty(value = "vtm")
+    private String vtm;
+
+    public SPOTClaims(String vot, String vtm) {
+        this.vot = vot;
+        this.vtm = vtm;
+    }
+
+    public SPOTClaims() {}
+
+    public String getVot() {
+        return vot;
+    }
+
+    public String getVtm() {
+        return vtm;
+    }
+}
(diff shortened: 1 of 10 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTClaims.java
```diff
@@ -0,0 +1,27 @@
...
+
+public class SPOTClaims {
+
+    @JsonProperty(value = "vot")
+    private String vot;
+
+    @JsonProperty(value = "vtm")
+    private String vtm;
+
+    public SPOTClaims(
```
chrisclayson, line 13: Not sure you need this constructor anymore ... if the `ObjectMapper` instance is instantiated with `ObjectMapperFactory.getInstance()` it is configured correctly to just work from the annotation on the private fields.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTRequest.java
```diff
@@ -4,24 +4,51 @@
...
+    @JsonProperty(value = "in_salt")
+    private byte[] salt;
+
+    @JsonProperty(value = "out_sub")
+    private String sub;
+
+    @JsonProperty(value = "log_ids")
+    private LogIds logIds;
 
     public SPOTRequest(
```
chrisclayson, line 22: As above!

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTClaims.java
JHjava: As discussed, keeping the constructor but remove the JsonProperty annotations from it and then adding a default constructor which will be used for desearlizing in the tests.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/SPOTRequest.java
JHjava: Address above

chrisclayson (approved):

PR #1598, opened 2022-03-28
AUTH-105 -  Change SpotResponse lambda to read from SQS 

JHjava:
## What?

- We will be receiving SPOT responses via an SQS queue. This changes our SPOTResponseHandler to listen to an SQS queue instead of an SNS topic.
- We are currently assuming only one credential will be present when the status is OK.
- Use UTC for PrivateKeyJWT in integration test
 
## Why?

- Change the structure of the SPOT response based on the RFC - https://github.com/alphagov/digital-identity-architecture/blob/main/rfc/0013-spot-request-response-structures.md
- The token integration test was failing as the clocks moved forward on Sunday. We need to be explicit and use UTC

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandler.java
@@ -39,7 +39,7 @@ public class SPOTResponseHandler implements RequestHandler<SQSEvent, Object> {
         for (SQSMessage msg : event.getRecords()) {
             try {
                 var spotResponse = objectMapper.readValue(msg.getBody(), SPOTResponse.class);
-                if (!spotResponse.getStatus().equals(SPOTStatus.OK.toString())) {
+                if (spotResponse.getStatus() != SPOTStatus.OK) {
                     LOG.warn(
                             "SPOTResponse Status is not OK. Actual Status: {}",
                             spotResponse.getStatus());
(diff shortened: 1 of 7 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/SPOTResponseHandler.java
```diff
@@ -33,29 +35,30 @@ public SPOTResponseHandler(DynamoSpotService dynamoSpotService) {
...
-        try {
-            return Optional.of(objectMapper.readValue(message, SPOTResponse.class));
-        } catch (JsonProcessingException e) {
-            LOG.error("Unable to deserialize SPOT response");
-            return Optional.empty();
+    public Object handleRequest(SQSEvent event, Context context) {
+        for (SQSMessage msg : event.getRecords()) {
+            try {
+                var spotResponse = objectMapper.readValue(msg.getBody(), SPOTResponse.class);
+                if (!spotResponse.getStatus().equals(SPOTStatus.OK.toString())) {
```
mrwilson, line 42: Rather than comparing the string values, can we deserialise the status directly into a `SPOTStatus1?

Then we can do `if (spotResponse.getStatus() != SPOTStatus.OK) {...}`

JHjava: Thanks @mrwilson. I have amended.

mrwilson (approved):

PR #1623, opened 2022-03-31
AUT-111: Correct the redirect in IPV handler

chrisclayson:
## What?

- The redirect handler is using the frontend API as the base URL rather than the OIDC API, when re-directing to the `auth-code` endpoint. 

## Why?

The `auth-code` endpoint is on the OIDC API.


chrisclayson:
Closing as this is wrong solution

PR #1638, opened 2022-04-05
AUT-254: Add backchannelLogoutUri to ClientRegistry

alan-gds:
## What?

Add backchannelLogoutUri to ClientRegistry, as an optional field.

## Why?

This means we can store a registered logout URL for RPs, to support backchannel logouts.

## Please check

* Capitalisation - are we okay with backchannel being one word in the field name?  Might be a typo in the Jira ticket.
* Should the URL be sanitised?  I notice in ClientRegistrationHandler we sanitise some but not all URLs that are sent to the ClientService.


mrwilson:
Cheers Alan

On (1) the OIDC spec calls it "Back-Channel" so `back_channel` is probably the better representation? I don't have a strong feeling if we keep it though

On (2) I think yes, we should validate URIs that RPs are "sending" to us

mrwilson (approved):

PR #1644, opened 2022-04-07
AUT-190 - Add support for request_uri param based on feature flag

JHjava:
## What?

- Add support for use of the request_uri parameter -  https://openid.net/specs/openid-connect-core-1_0.html#RequestObject
- Add a feature flag for use of the request_uri parameter. This will default to false to start with in all environments and return the REQUEST_URI_NOT_SUPPORTED error if it is used in the meantime
- Add additional field to client registry so RPs can specify to use the request_uri
- Create a new lambda which will be invoked from Authorize. This lambda will make a GET request to the request_uri location, parse the JWT and validate the signature by using the same key registered for validating the private key jwt. 
- Add additional integration tests. 

## Why?

- This is subject to change but we want to allow an RP to securely make an auth request to authorize. 
- We opted for the request_uri over the request param so we didn't risk being restricted URI size limitations


integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
```diff
@@ -57,9 +60,13 @@
     private static final KeyPair KEY_PAIR = KeyPairHelper.GENERATE_RSA_KEY_PAIR();
     public static final String DUMMY_CLIENT_SESSION_ID = "456";
 
+    @RegisterExtension
+    public static final RequestURILambdaStubExtension requestURILambdaStub =
+            new RequestURILambdaStubExtension(6001);
```
chrisclayson, line 65: We don't have to do this now, but to avoid any port conflict, you could not assign a port here and just use a dynamically assigned one. You'd have to override the configuration service like we do here: https://github.com/alphagov/di-authentication-api/blob/8361ab0613f1bc9bfd7ecde026c03999348de72e/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/NotifyIntegrationTest.java#L23

integration-tests/build.gradle
```diff
@@ -34,6 +34,8 @@ test {
     jacoco {
     }
     environment "AUDIT_SIGNING_KEY_ALIAS", "alias/local-audit-payload-signing-key-alias"
+    environment "AUTHORIZE_REQUEST_LAMBDA_ARN", "arn:authorize-request:eu-west-2:6546546465"
```
chrisclayson, line 37: See comment about overriding config service, it might be better than hard-coding values in here.

shared/src/main/java/uk/gov/di/authentication/shared/entity/UpdateClientConfigRequest.java
```diff
@@ -30,6 +30,9 @@
     @JsonProperty("service_type")
     private String serviceType;
 
+    @JsonProperty("request_uris")
+    private List<String> requestUris;
```
chrisclayson, line 34: Would `List<URI>` be more appropriate?

chrisclayson (dismissed):

integration-tests/build.gradle
JHjava: Updated

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
JHjava: Good shout. I have updated

shared/src/main/java/uk/gov/di/authentication/shared/entity/UpdateClientConfigRequest.java
JHjava: There's a good argument for that in both this class and the ClientRegistrationRequest. I kept it consistent with how we have list other attributes that could contain URIs. Maybe there's a refactoring needed here but leaving it out of scope for this change

chrisclayson (approved):

PR #1647, opened 2022-04-07
AUT-233 - Add claims to ClientRegistry

JHjava:
## What?

- Add claims to ClientRegistry
- Make the backchannel uri optional when validating it 

## Why?

- So RPs are only issued claims that they have registered to receive.

### DIFF
--- a/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationResponse.java
+++ b/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationResponse.java
@@ -39,6 +39,9 @@ public class ClientRegistrationResponse {
     @JsonProperty("service_type")
     private final String serviceType;
 
+    @JsonProperty("claims")
+    private final List<String> claims;
+
     public ClientRegistrationResponse(
             @JsonProperty(required = true, value = "client_name") String clientName,
             @JsonProperty(required = true, value = "client_id") String clientId,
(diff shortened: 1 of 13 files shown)


client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationResponse.java
```diff
@@ -48,7 +51,8 @@ public ClientRegistrationResponse(
             @JsonProperty(value = "post_logout_redirect_uris") List<String> postLogoutRedirectUris,
             @JsonProperty(value = "back_channel_logout_uri") String backChannelLogoutUri,
             @JsonProperty(required = true, value = "service_type") String serviceType,
-            @JsonProperty(required = true, value = "subject_type") String subjectType) {
+            @JsonProperty(required = true, value = "subject_type") String subjectType,
+            @JsonProperty(required = true, value = "claims") List<String> claims) {
```
sanisc, line 55: does this need to be mandatory ? what about non id services

sanisc (changes requested):

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationResponse.java
JHjava: Fixed

sanisc (approved):

PR #1672, opened 2022-04-13
AUT-245 - Use IPV sector identifier for IPV auth Subject ID

JHjava:
## What?

- Use a placeholder IPV domain as the sector identifier. This will need to be updated with the real value when it is known.
- Create a new method in the ClientSubjectHelper which can take in the Sector Identifier in directly so it can calculate the pairwise subject

## Why?

- Because we need to be using the IPV sector identifier instead of the RP's

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
@@ -32,6 +32,17 @@ public class ClientSubjectHelper {
         }
     }
 
+    public static Subject getSubjectWithSectorIdentifier(
+            UserProfile userProfile,
+            String sectorIdentifierURI,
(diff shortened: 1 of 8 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
```diff
@@ -68,11 +79,17 @@ static String returnHost(String uri) {
         }
     }
 
-    public static String calculatePairwiseIdentifier(String subjectID, String sector, byte[] salt) {
+    //    public static String calculatePairwiseIdentifier(String subjectID, String sectorURI,
```
dbes-gds, line 82: some comments

JHjava: Ooops

dbes-gds (approved):

PR #1722, opened 2022-04-26
AUT-241: Improve logout payloads

mrwilson:
Set `exp` claim for two minutes into the future
Set explicit `typ` header for logout requests


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/BackChannelLogoutRequestHandler.java
```diff
@@ -81,6 +84,7 @@ public JWTClaimsSet generateClaims(BackChannelLogoutMessage inputEvent) {
                 .jwtID(UUID.randomUUID().toString())
                 .audience(inputEvent.getClientId())
                 .subject(inputEvent.getSubjectId())
+                .expirationTime(Date.from(clock.instant().plus(2, ChronoUnit.MINUTES)))
```
JHjava, line 87: Is it better to use the `NowHelper.java` here

mrwilson: I thought that, but it's difficult to stub that out. Maybe that's a refactoring smell itself. 🤔

I'll timebox a bit to see if I can make the NowHelper more testable

JHjava: Yes, good point. I'm happy to approve if this is something we want to come back to.

JHjava (approved):

PR #1736, opened 2022-04-27
AUT-298: Deployment of document app callback

chrisclayson:
## What?

- Add the callback endpoint to the OIDC API
- Add new policies to allow reading/writing of the credential table
- Add new parameter to store PEM of CRI signing certificate

## Why?

We have implemented a callback lambda, lets now publish to an API endpoint.

## Related PRs

#1723 
#1729


JHjava:
I missed this earlier and this will need it too is the addition for the method and integration trigger to the `
resource "aws_api_gateway_deployment" "deployment" {
  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
`

chrisclayson:
> 

Good spot ... commit pushed.

JHjava (approved):

PR #1756, opened 2022-05-03
Update well known endpoint

sanisc:
Update well known endpoint.


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/WellknownHandler.java
```diff
@@ -104,7 +104,7 @@ public APIGatewayProxyResponseEvent handleRequest(
                                                 JWSAlgorithm.HS512));
                                 providerMetadata.setServiceDocsURI(
                                         new URI(
-                                                "https://auth-tech-docs.london.cloudapps.digital/"));
+                                                "https://oidc.account.gov.uk/.well-known/openid-configuration"));
```
chrisclayson, line 107: We should probably put this in config.

Also, I think this is the wrong URL.

mrwilson (approved):

PR #1779, opened 2022-05-05
AUT-345: Add XRay Segment annotations

chrisclayson:
## What?

- Re-add the lines to associate annotations to the segment
- Add a main segment at the beginning of the `TokenHandler` lambda `handleRequest` method

## Why?

A previous attempt to add the annotations failed as it seems you cannot add annotations to a Lambda facade segment, so we now explicitly create a segment for the lambda.


mrwilson:
I wonder if for readability we pull out almost everything in `handleRequest` to another method, then handleRequest becomes

```
startSegment("handleRequest");
try {
  doTheThing();
} finally {
  endSegment();
}
```

What do you think?

mrwilson (approved):

PR #1867, opened 2022-05-18
AUT-363: Make the client_id parameter optional in the Token service

dbes-gds:
## What?

Make the client_id parameter optional in the Token service.

- If client_id not supplied then retrieve it from the private key JWT.
- If client_id is supplied then validate that the client id matches the value in the private key JWT.

## Why?

Our clients are 'authenticated' so according to the spec client_id is an optional parameter:

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
@@ -147,16 +147,24 @@ public class TokenHandler
                             }
 
                             Map<String, String> requestBody = parseRequestBody(input.getBody());
-                            String clientID = requestBody.get("client_id");
-                            attachLogFieldToLogs(CLIENT_ID, clientID);
-                            addAnnotation("client_id", clientID);
                             addAnnotation("grant_type", requestBody.get("grant_type"));
 
+                            String clientID;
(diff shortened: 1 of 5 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -147,16 +147,24 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(
...
-                            String clientID = requestBody.get("client_id");
-                            attachLogFieldToLogs(CLIENT_ID, clientID);
-                            addAnnotation("client_id", clientID);
                             addAnnotation("grant_type", requestBody.get("grant_type"));
 
+                            String clientID;
                             ClientRegistry client;
                             try {
+                                if (requestBody.get("client_id") != null) {
+                                    clientID = requestBody.get("client_id");
```
JHjava, line 156: we could remove this and retrieve it from the secure jwt every time?

dbes-gds: Yes to be in an another PR after seeing who's using it.

JHjava (approved):

PR #1872, opened 2022-05-18
AUT-363: ignore client_id parameter in token

dbes-gds:
## What?

Ignore client_id parameter in token.

- This is an optional parameter, completing the switch over from when it was mandatory.
- Remove our own validation and processing of the parameter, always get the value from the privateJWT.
- Nimbus will still validate that this value matches the one in the privateJWT if it is provided.

## Why?

Our clients are 'authenticated' so according to the spec client_id is an optional parameter:

https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.3

## Related PRs

#1867


chrisclayson (dismissed):

JHjava:
@dbes-gds are we still reliant on client_id on line 376?

 `requestBody.get("client_id");`

JHjava (approved):

PR #1941, opened 2022-05-30
AUT-356: Migrate OIDC, IPV, DocApp and client registry APIs to GSON

chrisclayson:
## What?

- Add annotations to entity classes
- Change handlers to use `SerializationService` rather than Jackson
- Update integration tests to serialise requests using GSON rather than Jackson
- Update AM API entities to include GSON annotations (as well as Jackson)

## Why?

We are migrating from Jackson to GSON for performance benefits

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/entity/DocAppAuthorisationResponse.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/entity/DocAppAuthorisationResponse.java
@@ -1,10 +1,14 @@
 package uk.gov.di.authentication.app.entity;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
 
 public class DocAppAuthorisationResponse {
 
     @JsonProperty("redirectUri")
+    @SerializedName("redirectUri")
+    @Expose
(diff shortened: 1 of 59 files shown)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/entity/DocAppAuthorisationResponse.java
```diff
@@ -1,10 +1,14 @@
 package uk.gov.di.authentication.app.entity;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
 
 public class DocAppAuthorisationResponse {
 
     @JsonProperty("redirectUri")
```
JHjava, line 9: Do we still need this?

chrisclayson: Possibly not, there were a few places that I left both in because of the way the tests were run.

I'll be raising another PR later to totally remove Jackson from the project, I'll pick these remnants up when I do that.

JHjava: 👍 cool, that makes sense.

PR #1952, opened 2022-05-31
AUT-356: Replace Hibernate Validator

chrisclayson:
## What?

Replace Hibernate Validator with a custom, simple, validator.

## Why?

There is some, significant, time being taken by the validator. Hibernate providers all sorts of clever functionality that we don't need, we only require simple `null` checks.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/validation/RequiredFieldValidator.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/validation/RequiredFieldValidator.java
@@ -1,0 +1,43 @@
+package uk.gov.di.authentication.shared.validation;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+import java.util.ArrayList;
+import java.util.List;
+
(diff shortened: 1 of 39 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/validation/RequiredFieldValidator.java
```diff
@@ -0,0 +1,43 @@
...
+            clazz = clazz.getSuperclass();
+        }
+        return violations;
+    }
+
+    private List<String> validateRequiredFields(Object object, Class clazz) {
+        List<String> violations = new ArrayList<>();
+        for (var field : clazz.getDeclaredFields()) {
+            try {
+                field.setAccessible(true);
```
mrwilson, line 33: We should probably check whether the field is accessible before setting this, so that we can set it back as false in a `} finally {}`?

chrisclayson: Good shout!

mrwilson (approved):

PR #1970, opened 2022-06-06
BAU - Create custom adapter for Nimbus Subject object 

JHjava:
## What?

- Write a custom adapter for searlization of the Subject in the ClientSession. 
- Add an asserts to an existing test
- Remove debugging logging

## Why? 

- We no longer use Jackson so we need to manually write a custom adapter now that we use GSON to be able to searlize the Subject as expected.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/ClientSession.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/ClientSession.java
@@ -1,10 +1,7 @@
 package uk.gov.di.authentication.shared.entity;
 
 import com.google.gson.annotations.Expose;
-import com.google.gson.annotations.JsonAdapter;
 import com.nimbusds.oauth2.sdk.id.Subject;
-import uk.gov.di.authentication.shared.serialization.LocalDateTimeAdapter;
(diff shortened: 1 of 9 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/ClientSession.java
```diff
@@ -21,7 +22,9 @@ public class ClientSession {
 
     @Expose private VectorOfTrust effectiveVectorOfTrust;
 
-    @Expose private Subject docAppSubjectId;
+    @JsonAdapter(SubjectAdapter.class)
```
chrisclayson, line 25: I'd be tempted to globally register the adapter in the `SerializationService` (see L32 for how we did `StateAdapter`) ... this is behaviour we want anywhere GSON encounters a `Subject` rather than just this specific class.

JHjava: I've added it to the `SerializationService` but without this here, tests will fail as the `docAppSubjectId` will sometimes be null. Not sure how to get around this otherwise.

PR #2000, opened 2022-06-09
Aut 30/add roles to common passwords dynamodb

DTaylorGDS:
## What?

Add read permissions for new common passwords Dynamo table for three lambdas, which will use this to validate whether or not passwords is a common password and produce an error accordingly.

## Why?

So that password validation can consider common words

## Related PRs

https://github.com/alphagov/di-authentication-api/pull/1990

### DIFF
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -34,7 +34,7 @@ data "aws_iam_policy_document" "dynamo_user_write_policy_document" {
     ]
     resources = [
       data.aws_dynamodb_table.user_credentials_table.arn,
-      data.aws_dynamodb_table.common_passwords_table.arn,
+      data.aws_dynamodb_table.user_profile_table.arn,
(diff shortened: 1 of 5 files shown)


ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -30,7 +34,7 @@ data "aws_iam_policy_document" "dynamo_user_write_policy_document" {
     ]
     resources = [
       data.aws_dynamodb_table.user_credentials_table.arn,
-      data.aws_dynamodb_table.user_profile_table.arn,
+      data.aws_dynamodb_table.common_passwords_table.arn,
```
JHjava, line 37: Is this change intentional?

JHjava (commented):
PR looks great. Just one question.

ci/terraform/oidc/dynamo-policies.tf
DTaylorGDS: Thanks for pointing out - it was accidental on my part.

JHjava (approved):

PR #2001, opened 2022-06-09
BAU: Measure sent SMS by country code

mrwilson:
This publishes a metric `SendingSms` broken down by environment and country code, so we can compare "sent SMS" to successfully registered accounts.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/PhoneNumberHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/PhoneNumberHelper.java
@@ -20,6 +20,16 @@ public class PhoneNumberHelper {
         }
     }
 
+    public static String getCountry(String phoneNumber) {
+        try {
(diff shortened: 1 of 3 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/PhoneNumberHelper.java
```diff
@@ -20,6 +20,16 @@ public static String formatPhoneNumber(String phoneNumber) {
         }
     }
 
+    public static String getCountry(String phoneNumber) {
+        try {
+            return Integer.toString(
```
dbes-gds, line 25: isn't country code a string?

mrwilson: Unfortunately not, `getCountryCode()` returns an int

sanisc (approved):

PR #2009, opened 2022-06-10
AUT-430: Store alternative mfa in the database

dbes-gds:
## What?

Store alternative mfa in the database.

- Draft data model for storing a list of mfa methods in the database.

## Why?

We will need to store a secret key for a registered authentication app in order to validate OTP.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/MFAMethod.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/MFAMethod.java
@@ -1,0 +1,104 @@
+package uk.gov.di.authentication.shared.entity;
+
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBAttribute;
+import com.amazonaws.services.dynamodbv2.datamodeling.DynamoDBDocument;
+import com.amazonaws.services.dynamodbv2.model.AttributeValue;
+
+import java.util.Map;
+import java.util.Objects;
+
+@DynamoDBDocument
+public class MFAMethod {
+
+    public static final String ATTRIBUTE_MFA_METHOD_TYPE = "MfaMethodType";
+    public static final String ATTRIBUTE_CREDENTIAL_VALUE = "CredentialValue";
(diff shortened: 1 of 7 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
```diff
@@ -296,6 +298,19 @@ public Optional<String> getPhoneNumber(String email) {
                         .getPhoneNumber());
     }
 
+    @Override
+    public void updateMFAMethod(
+            String email, MFAMethodType mfaMethodType, boolean enabled, String credentialValue) {
+        String dateTime = LocalDateTime.now().toString();
```
JHjava, line 304: Using `NowHelper.now()` might be better

shared/src/main/java/uk/gov/di/authentication/shared/entity/MFAMethod.java
```diff
@@ -0,0 +1,104 @@
...
+
+@DynamoDBDocument
+public class MFAMethod {
+
+    public static final String ATTRIBUTE_MFA_METHOD_TYPE = "MfaMethodType";
+    public static final String ATTRIBUTE_CREDENTIAL_VALUE = "CredentialValue";
+    public static final String ATTRIBUTE_ENABLED = "Enabled";
+    public static final String ATTRIBUTE_UPDATED = "Updated";
+
+    private String mfaMethodType;
```
JHjava, line 18: Any reason we can't set this directly as the `MFAMethodType` type?

JHjava (commented):
Couple of small points but other than that it looks good

shared/src/main/java/uk/gov/di/authentication/shared/entity/MFAMethod.java
dbes-gds: Going to look at this later on as there a couple of ways to do it, need to see what the best one is.  It will always be a String in the database anyway.

PR #2046, opened 2022-06-21
AUT-429: Verify MFA code lambda (new lambda)

DTaylorGDS:
## What?

- This is the foundation for a new lambda that will be able to check any recognised MFA code - this PR covers both infrastructure as code and application logic
- This PR implements only a single type of MFA - Auth App Code - but SMS should already be possible to build in on the same structure
- This code should not affect anything currently live - and will require further testing once the front-end is available to test against it.

## Why?

We are introducing Auth App MFA to the service and there is currently no way to check if a code generated by an app is valid (based on a known, stored secret).


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    @Override
+    public APIGatewayProxyResponseEvent handleRequestWithUserContext(
+            APIGatewayProxyRequestEvent input,
+            Context context,
+            VerifyMfaCodeRequest codeRequest,
+            UserContext userContext) {
+
+        attachSessionIdToLogs(userContext.getSession());
```
JHjava, line 95: This can be removed. The `BaseFrontendHandler` will do this for you

ci/terraform/oidc/verify_mfa_code.tf
```diff
@@ -0,0 +1,86 @@
...
+    LAMBDA_MIN_CONCURRENCY = var.lambda_min_concurrency
+  }
+
+  use_localstack = var.use_localstack
+
+  depends_on = [
+    aws_api_gateway_rest_api.di_authentication_frontend_api,
+    aws_api_gateway_resource.connect_resource,
+    aws_api_gateway_resource.wellknown_resource,
+    aws_sqs_queue.email_queue,
```
JHjava, line 84: These lambdas doesn't have any dependencies on these so they can be removed.

shared/src/main/java/uk/gov/di/authentication/shared/validation/AuthAppCodeValidator.java
```diff
@@ -0,0 +1,174 @@
...
+                dynamoService,
+                maxRetries);
+        this.dynamoService = dynamoService;
+        this.userContext = userContext;
+        this.WINDOW_TIME = configurationService.getAuthAppCodeWindowLength();
+        this.ALLOWED_WINDOWS = configurationService.getAuthAppCodeAllowedWindows();
+    }
+
+    @Override
+    public Optional<ErrorResponse> codeValidationErrors(String code) {
```
JHjava, line 48: I would rename this to something like `validateCode`

shared-test/src/main/java/uk/gov/di/authentication/sharedtest/helper/AuthAppStub.java
```diff
@@ -0,0 +1,68 @@
...
+import javax.crypto.Mac;
+import javax.crypto.spec.SecretKeySpec;
+
+import java.util.concurrent.TimeUnit;
+
+public class AuthAppStub {
+    private final int CODE_DIGITS = 6;
+    private final long TIME_WINDOW_IN_MILLISECONDS = TimeUnit.SECONDS.toMillis(30);
+
+    // secret must parse as valid base32, or else an exception will be thrown by the library
```
JHjava, line 15: Do we need this comment? Could we catch any exception and then log a message out instead?

shared-test/src/main/java/uk/gov/di/authentication/sharedtest/helper/AuthAppStub.java
```diff
@@ -0,0 +1,68 @@
...
+                truncatedHash <<= 8;
+                truncatedHash |= (hash[offset + i] & 0xFF);
+            }
+
+            truncatedHash &= 0x7FFFFFFF;
+            truncatedHash %= (int) Math.pow(10, CODE_DIGITS);
+
+            return (int) truncatedHash;
+        } catch (Exception ex) {
+            // Zero return simulates InvalidKeyException or NoSuchAlgorithmException for test usage.
```
JHjava, line 55: A log message here might be more useful than a comment.

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
```diff
@@ -320,6 +320,21 @@ public void updateMFAMethod(
...
+        String dateTime = NowHelper.toTimestampString(NowHelper.now());
+        var userCredentials =
+                userCredentialsMapper.load(UserCredentials.class, email.toLowerCase(Locale.ROOT));
+        var mfaMethod =
+                userCredentials.getMfaMethods().stream()
+                        .filter(
+                                method ->
+                                        method.getMfaMethodType().equals(mfaMethodType.getValue()))
+                        .findFirst();
+        mfaMethod.ifPresent(method -> method.setMethodVerified(true));
```
JHjava, line 334: This could just fail quitely if in the unlikely event no mfaMethod is found and therefore it will not set verified to true. But the lambda will still return a 204. 

I would assume it is present as you're already checking it is in the AuthCodeAppvalidator and throw if it isn't.

JHjava (commented):
Looks good. Left a few comments, shout if you have any questions. 

Is asking for the `MFAMethodType` just future proofing things? As we already have the VerifyCode lambda to validate codes sent to Phones and emails?

JHjava (approved):

PR #2052, opened 2022-06-22
AUT-435: Replay events mechanism

chrisclayson:
## What?

- Add a lambda to replay fraud messages that are placed in a specific S3 bucket
- Add a script that will copy a given list of audit events from audit cold storage to a given S3 bucket (where they can be picked up for replay)
- Add terraform to create the replay bucket
- Add terraform to create the lambda function and execution role
- Add terraform to trigger lambda when objects created in the replay bucket
- DO NOT ENABLE Splunk log subscription.

## Why?

We had an issue where some fraud events were dropped from logging and we need to replay them. This PR gives us a mechanism to pickup the missing events from the S3 "cold storage" and copy them to a different bucket where they can be picked up to replay.

The missing events have been identified using an Athena query. The query gives the S3 URLs for the individual events, these can be fed into the shell script (included in this PR) to copy those events to the replay bucket.

Not enabling the Splunk log subscription, gives the ability to dry run this process before enabling to do the real transfer.

### DIFF
--- a/audit-processors/src/main/java/uk/gov/di/authentication/audit/lambda/CounterFraudAuditReplayLambda.java
+++ b/audit-processors/src/main/java/uk/gov/di/authentication/audit/lambda/CounterFraudAuditReplayLambda.java
@@ -1,0 +1,108 @@
+package uk.gov.di.authentication.audit.lambda;
+
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.S3Event;
+import com.amazonaws.services.s3.AmazonS3;
+import com.amazonaws.services.s3.AmazonS3Client;
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.util.JsonFormat;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ObjectMessage;
+import uk.gov.di.audit.AuditPayload.AuditEvent;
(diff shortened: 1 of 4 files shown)


ci/terraform/audit-processors/counter_fraud_replay.tf
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+  kms_key_arn = local.lambda_env_vars_encryption_kms_key_arn
+
+  runtime = "java11"
+
+  tags = local.default_tags
+}
+
+resource "aws_lambda_permission" "sns_can_execute_subscriber_fraud_realtime_logging_replay_lambda" {
```
mrwilson, line 97: Probably worth removing the reference to SNS?

audit-processors/src/main/java/uk/gov/di/authentication/audit/lambda/CounterFraudAuditReplayLambda.java
```diff
@@ -0,0 +1,108 @@
...
+    public Void handleRequest(S3Event input, Context context) {
+
+        input.getRecords()
+                .forEach(
+                        s3 -> {
+                            var key = s3.getS3().getObject().getKey();
+                            var bucket = s3.getS3().getBucket().getName();
+
+                            LOG.info("Processing file {}", key);
+                            var content = client.getObjectAsString(bucket, key);
```
mrwilson, line 54: We're assuming that each file contains 1 event here, I'm not sure if that's 100% true?

ci/terraform/audit-processors/counter_fraud_replay.tf
chrisclayson: Done (and rebased).

mrwilson (approved):

PR #2062, opened 2022-06-23
BAU: Refactor security headers and add to API

mrwilson:
- BAU: Create enum for security headers
- BAU: Generate security headers map from enum
- BAU: Use builder-esque pattern to create responses rather than setters
- BAU: Attach default headers to existing APIs that don't use helper

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ApiGatewayResponseHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ApiGatewayResponseHelper.java
@@ -12,11 +12,12 @@ import uk.gov.di.authentication.shared.services.SerializationService;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 public class ApiGatewayResponseHelper {
 
-    enum SecurityHeaders {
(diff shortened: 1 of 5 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/ApiGatewayResponseHelper.java
```diff
@@ -66,32 +78,40 @@ public static APIGatewayProxyResponseEvent generateEmptySuccessApiGatewayRespons
...
+
+    public static APIGatewayProxyResponseEvent generateApiGatewayProxyResponse(
+            int statusCode,
+            String body,
+            Map<String, String> headers,
+            Map<String, List<String>> multiValueHeaders) {
+
+        var allHeaders = SecurityHeaders.headers();
+
+        Optional.ofNullable(headers).ifPresent(allHeaders::putAll);
```
dbes-gds, line 103: Does this only add the security headers if there are other headers?  Presumably you always want the security headers even if there are no other headers?

mrwilson: If there are other headers, then it adds them to the existing `allHeaders` variable, so it will always have those base headers.

dbes-gds (approved):

PR #2083, opened 2022-06-27
AUT-475: Create a custom `resource-id` scope for non-urn format subject ID

mrwilson:
This is needed for a spike into a service which does not (yet) support URN-format subject identifiers

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
@@ -306,7 +306,7 @@ public class TokenService {
             idTokenClaims.setClaim("vot", vot);
         }
         idTokenClaims.setClaim("vtm", trustMarkUri.toString());
-        if (requiresResourceId) {
+        if (requiresResourceId && configService.isResourceIdScopeEnabled()) {
             idTokenClaims.setClaim(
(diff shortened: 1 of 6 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
```diff
@@ -300,6 +306,11 @@ private SignedJWT generateIDToken(
             idTokenClaims.setClaim("vot", vot);
         }
         idTokenClaims.setClaim("vtm", trustMarkUri.toString());
+        if (requiresResourceId && configService.isResourceIdScopeEnabled()) {
+            idTokenClaims.setClaim(
+                    CustomScopeValue.RESOURCE_ID.getValue(), subject.getValue().split(":")[4]);
+        }
```
JHjava, line 312: I think we should do this in userinfo by checking the scopes in the access token?

JHjava (commented):
I think we should return this info in `userinfo`.

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
mrwilson: I checked with them in the initial discussion because my preference was user-info too, and it needs to be in the identity token for the software they're using

JHjava (approved):

PR #2104, opened 2022-06-30
AUT-437: Add common passwords transfer lambda log group to Terraform

DTaylorGDS:
## What?

- Add common passwords transfer lambda log group to Terraform
- Add variable for retention - default 1, but 5 days for production
- Add shared.tf to consume outputs of shared state bucket, including CloudWatch encryption key arn

## Why?

- To allow log groups to be created in each environment, so that logs can be viewed in CloudWatch for the new common passwords S3->Dynamo data transfer lambda

## Related PRs

https://github.com/alphagov/di-authentication-api/pull/2089 - base lambda including Terraform
https://github.com/alphagov/di-authentication-api/pull/2097 - addition of some logging variables (as placeholders at that point to match the input arguments of the deploy-utils-build task)

### DIFF
--- a/ci/terraform/utils/common_passwords_update_lambda.tf
+++ b/ci/terraform/utils/common_passwords_update_lambda.tf
@@ -48,3 +48,29 @@ resource "aws_lambda_permission" "common_passwords_dynamo_update_lambda_invoke_p
   principal     = "s3.amazonaws.com"
   source_arn    = "arn:aws:s3:::${aws_s3_bucket.common_passwords.id}"
 }
+
+resource "aws_cloudwatch_log_group" "common_passwords_update_lambda_log_group" {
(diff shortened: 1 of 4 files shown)


ci/terraform/utils/sqs.tf
```diff
@@ -0,0 +1,25 @@
...
+  retention_in_days = var.cloudwatch_log_retention
+
+  tags = local.default_tags
+
+  depends_on = [
+    aws_lambda_function.common_passwords_dynamo_update_lambda
+  ]
+}
+
+resource "aws_cloudwatch_log_subscription_filter" "sqs_lambda_log_subscription" {
```
chrisclayson, line 15: And this one!

chrisclayson (approved):

PR #2107, opened 2022-06-30
BAU - Update Processing Identity Lambda to return additional status

JHjava:
## What?

- Add an attribute to the session which counts how many attempts there has been at calling the ProcessingIdentity lamda. If it's the first attempt and there's nothing in dynamo, then we assume that the user has got into a weird state and we will return the NO_ENTRY status back to the frontend.
- This count will be reset when the user hits authorize again, regardless of whether they have an existing session.
- Add a cloudwatch metric counter to track the different statuses returned from the ProcessingIdentity lambda
## Why?

- We currently return ERROR when there is nothing in the Identity Credentials table with the given subject ID. This may be because we received an error from SPOT and then went on to delete the credential. It could also happen if there was nothing ever in the Identity Credentials table for that subject ID, which can happen if the user does not complete the IPV journey, presses back and gets into a weird state. We want to be able to differentiate these 2 different use cases.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/ProcessingIdentityIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/ProcessingIdentityIntegrationTest.java
@@ -33,7 +33,7 @@ import static java.util.Collections.singletonList;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static uk.gov.di.authentication.api.AuthorisationIntegrationTest.configurationService;
 import static uk.gov.di.authentication.shared.helpers.ClientSubjectHelper.calculatePairwiseIdentifier;
-import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasBody;
+import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasJsonBody;
 import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasStatus;
 
 public class ProcessingIdentityIntegrationTest extends ApiGatewayHandlerIntegrationTest {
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/ProcessingIdentityIntegrationTest.java
```diff
@@ -138,7 +139,34 @@ void shouldReturnStatusOfERRORWhenEntryIsNotInDatabase() throws Json.JsonExcepti
...
+        var response =
+                makeRequest(
+                        Optional.of(format("{ \"email\": \"%s\"}", TEST_EMAIL_ADDRESS)),
+                        headers,
+                        Map.of());
+
+        assertThat(response, hasStatus(200));
+        assertThat(
+                response,
+                hasBody(
```
mrwilson, line 163: Very minor thing - we have a `hasJsonBody` matcher that takes an object that we could use here instead.

mrwilson (dismissed):

ci/terraform/oidc/processing-identity.tf
```diff
@@ -12,6 +12,7 @@ module "ipv_processing_identity_role" {
     aws_iam_policy.lambda_sns_policy.arn,
     aws_iam_policy.pepper_parameter_policy.arn,
     aws_iam_policy.redis_parameter_policy.arn,
+    aws_iam_policy.cloudwatch_metrics_putdata_policy.arn,
```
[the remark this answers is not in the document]
JHjava: Removed

integration-tests/src/test/java/uk/gov/di/authentication/api/ProcessingIdentityIntegrationTest.java
JHjava: Good shout. Updated.

mrwilson (approved):

PR #2109, opened 2022-07-01
AUT-277: Add a robots.txt to the OIDC API - Build only

dbes-gds:
## What?

Add a robots.txt to the OIDC API - Build only.

## Why?

As this is a public API it's being scraped by robots which creates unecessary noise in the logs.
Adding a robots.txt to keep them away.

### DIFF
--- a/ci/terraform/oidc/api-gateway.tf
+++ b/ci/terraform/oidc/api-gateway.tf
@@ -133,6 +133,7 @@ resource "aws_api_gateway_deployment" "deployment" {
       module.ipv-capacity.method_trigger_value,
       var.doc_app_api_enabled ? module.doc-app-callback[0].integration_trigger_value : null,
       var.doc_app_api_enabled ? module.doc-app-callback[0].method_trigger_value : null,
+      var.use_robots_txt ? aws_api_gateway_integration_response.robots_txt_integration_response[0] : null,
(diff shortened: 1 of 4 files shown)


ci/terraform/oidc/api-gateway.tf
```diff
@@ -133,6 +133,7 @@ resource "aws_api_gateway_deployment" "deployment" {
       module.ipv-capacity.method_trigger_value,
       var.doc_app_api_enabled ? module.doc-app-callback[0].integration_trigger_value : null,
       var.doc_app_api_enabled ? module.doc-app-callback[0].method_trigger_value : null,
+      var.use_robots_txt ? aws_api_gateway_integration_response.robots_txt_integration_response : null,
```
mrwilson, line 136: I think the response will need a `[0]` on it

mrwilson: Because there's a count

mrwilson (changes requested):
Just the count change

ci/terraform/oidc/api-gateway.tf
dbes-gds: thanks updated

mrwilson (approved):

PR #2127, opened 2022-07-11
BAU - Add additional logging to doc app callback

JHjava:
## What?

- Add additional logging to doc app callback

## Why?

- To make it easier to debug the journey

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
@@ -26,11 +26,14 @@ import uk.gov.di.authentication.shared.services.SerializationService;
 import uk.gov.di.authentication.shared.services.SessionService;
 
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
 import static uk.gov.di.authentication.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
+import static uk.gov.di.authentication.shared.helpers.LogLineHelper.LogFieldName.CLIENT_ID;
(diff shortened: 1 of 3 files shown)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -100,27 +103,38 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
                             try {
                                 var sessionCookiesIds =
                                         CookieHelper.parseSessionCookie(input.getHeaders())
-                                                .orElseThrow();
+                                                .orElse(null);
```
DTaylorGDS, line 106: Would it be possible to keep your orElseThrow() and then do the logging and exception as part of an inline lambda? Would save you separately checking nulls with `if` unless I've misunderstood (quite possible)

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -100,27 +103,38 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
...
+                                    LOG.error("No session cookie present");
+                                    throw new RuntimeException("No session cookie present");
+                                }
                                 var session =
                                         sessionService
                                                 .readSessionFromRedis(
                                                         sessionCookiesIds.getSessionId())
-                                                .orElseThrow();
+                                                .orElse(null);
+                                if (Objects.isNull(session)) {
```
DTaylorGDS, line 116: Same here - so either my previous comment was sensible, or ignore here too...

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
JHjava: @DTaylorGDS  I moved it away from the `orElseThrow` as I wanted to be able to differenciate between the cookie not being present and the session not being present and log different error messages for each scenario

DTaylorGDS (approved):

PR #2131, opened 2022-07-12
AUT-537: Add generic pre-validation of supplied string auth app code to validity check

DTaylorGDS:
## What?
- Add check that supplied auth app code string - transmitted from front end - which must be parsed to int as part of validation algorithm, is neither empty not too long
- This avoids `NumberFormatException` when in these cases the supplied code must by definition be invalid, as it is unprocessable by our algorithm, so simply returning a validity check result of `false` achieves the desired effect (Note: NumberFormatException was thrown silently until now for long numbers because `Integer.ParseInt` cannot handle any number larger than 2^31 - 1)
- Have also revisited the relevant unit test, which I only noticed when trying to run it with an example number that it was actually not setting up an invalid code correctly (it was giving the correct error response, but for the wrong reason - `secret not found`). In doing so I realised that the setup for a correct code was already doing everything needed, so reused that function and deleted the deficient function (`setUpInvalidAuthCode()`)

## Why?

The exception meant that the front end was receiving ERROR 1001, and going on to show 500 status code, when this was not at all reflective of what was going wrong.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/validation/AuthAppCodeValidator.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/validation/AuthAppCodeValidator.java
@@ -96,6 +96,10 @@ public class AuthAppCodeValidator extends MfaCodeValidator {
     }
 
     public boolean isCodeValid(String code, String secret) {
+        if (code.isEmpty() || code.length() > 6) {
+            return false;
+        }
+
         int codeToCheck = Integer.parseInt(code);
 
         if (secret == null) {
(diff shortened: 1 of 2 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/validation/AuthAppCodeValidator.java
```diff
@@ -96,6 +96,10 @@ public Optional<String> getMfaCredentialValue() {
     }
 
     public boolean isCodeValid(String code, String secret) {
+        if (code.isEmpty() || code.length() > 9) {
```
dbes-gds, line 99: sticking to 6 digits? so could check > 6 ?

DTaylorGDS: thanks - changed

dbes-gds (approved):

PR #2177, opened 2022-07-25
BAU: Differentiate test journeys in metrics

mrwilson:
- BAU: Expose `isTestJourney` method to determine if journey is "real"
- BAU: Add `IsTest` dimension to `SignIn` metric to differentiate test journey

We want to be able to exclude test journeys from some of our dashboards

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AuthorizationService.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AuthorizationService.java
@@ -28,6 +28,7 @@ import java.net.URI;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import static java.lang.String.format;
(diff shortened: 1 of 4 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AuthorizationService.java
```diff
@@ -230,4 +231,18 @@ private boolean areClaimsValid(OIDCClaimsRequest claimsRequest, ClientRegistry c
...
+
+    public boolean isTestJourney(ClientID clientID, String emailAddress)
+            throws ClientNotFoundException {
+        var client =
+                dynamoClientService
+                        .getClient(clientID.toString())
+                        .orElseThrow(() -> new ClientNotFoundException(clientID.toString()));
+
+        return Optional.ofNullable(client)
+                .map(ClientRegistry::getTestClientEmailAllowlist)
```
dbes-gds, line 243: are we using this for smoke tests?

mrwilson: We will adapt this for smoke tests right now, ahead of further changes to how we run them in future as discussed on Slack.

dbes-gds (approved):

PR #2190, opened 2022-08-01
AUT-561: Limit OTP/MFA code retry by email not browser session

DTaylorGDS:
## What?

- Increment and save MFA code attempts against email address (persistent) rather than session (temporary) in Redis 
- Make counter clear itself after 15 minutes (aligned to 'lifespan' of OTP codes themselves)
- Remove simple counter of retries currently stored within session object in Redis
- Refactor `CodeStorageService` to take in `ConfigurationService` rather than `RedisService(ConfigurationService)` in its constructor, in order to remove Redis implementation detail from handlers, and to make accessing other config parameters in future easier e.g. if we want to set the MFA counter's 'time to live' via config rather than the current hard-coding of `900L`.

Note: Whilst not directly part of the functionality, also in this PR is a change to Sandpit default var of how long a code block lasts, so that we only need to wait 30 seconds in manual tests rather than the default 15 minutes until the email address is 'released' from the block. I used this whilst testing this ticket and think it is a generally useful default in Sandpit.

## Why?

- So that it is not possible to gain unlimited attempts at brute forcing MFA code by starting a new browser session after `MAX_WRONG_ATTEMPTS - 1` attempts

### DIFF
--- a/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/HandlerIntegrationTest.java
+++ b/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/HandlerIntegrationTest.java
@@ -28,6 +28,10 @@ import java.util.Optional;
 import static org.mockito.Mockito.mock;
 
 public abstract class HandlerIntegrationTest<Q, S> {
+    private static final String REDIS_HOST = "localhost";
+    private static final int REDIS_PORT = 6379;
+    private static final String REDIS_PASSWORD = null;
+    private static final boolean DOES_REDIS_USE_TLS = false;
     protected static final String LOCAL_ENDPOINT_FORMAT =
             "http://localhost:45678/restapis/%s/local/_user_request_";
     protected static final String LOCAL_API_GATEWAY_ID =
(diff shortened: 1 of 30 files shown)


shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/HandlerIntegrationTest.java
```diff
@@ -196,6 +202,26 @@ public String getEmailQueueUri() {
             return notificationQueue.getQueueUrl();
         }
 
+        @Override
```
chrisclayson, line 205: I'm unsure why we need both these overrides and the above changes to the `ParameterStoreExtension`?

chrisclayson (approved):
👍🏻

DTaylorGDS:
> Looks good ... just wondering what will happen for an "in flight" journey when this is deployed? Will it result in retry counter going back to `0` for that journey or will something more severe happen?

It should mean that any existing counter saved against the session is ignored (and it will in any case disappear in a relatively short time - 15 mins). The counter will then start at one the first time an attempt is made to increment it. Before that point, any attempt to read the counter will return null from Redis which will return zero from the 'read' function.

shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/HandlerIntegrationTest.java
DTaylorGDS: Thanks. We worked on this together and have refactored the code so that it no longer requires the overrides.

PR #2206, opened 2022-08-08
AUT-583: Add IPV Callback Exception where cookie not parsed

DTaylorGDS:
## What?

- Add IPV Callback Exception where cookie not parsed

## Why?

- In order to allow cookie not parsed exceptions to result in frontend redirect (via rethrowing them as `IpvCallbackException`)

## Related PRs

- Amends https://github.com/alphagov/di-authentication-api/pull/2202


DTaylorGDS:
Closing as discussion with @chrisclayson confirmed my suspicion that this is not a nice way to do things. Instead, will make CookieHelper method an instance method, which will be easier to mock.

PR #2213, opened 2022-08-09
AUT-584: Grant KMS permissions to lambda and enable in staging

mrwilson:
- AUT-584: Allow lambdas to encrypt the payload on the queue
- AUT-584: Enable TxMA audit in staging

### DIFF
--- a/ci/terraform/modules/txma-audit-queue/policy.tf
+++ b/ci/terraform/modules/txma-audit-queue/policy.tf
@@ -8,6 +8,18 @@ data "aws_iam_policy_document" "txma_audit_queue_access_policy_document" {
     actions   = ["sqs:SendMessage", ]
     resources = [aws_sqs_queue.txma_audit_queue.arn]
   }
+
+  statement {
+    effect = "Allow"
+    sid    = "AllowAccessToKeyForEncryptingPayloads"
+    actions = [
+      "kms:GenerateDataKey",
+      "kms:Decrypt"
+    ]
+    resources = [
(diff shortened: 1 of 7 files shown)


ci/terraform/modules/txma-audit-queue/policy.tf
```diff
@@ -8,6 +8,18 @@ data "aws_iam_policy_document" "txma_audit_queue_access_policy_document" {
     actions   = ["sqs:SendMessage", ]
     resources = [aws_sqs_queue.txma_audit_queue.arn]
   }
+
+  statement {
+    effect = "Allow"
+    sid    = "AllowAccessToKeyForEncryptingPayloads"
+    actions = [
+      "kms:GenerateDataKey",
```
chrisclayson, line 16: Do we definitely need this permission? For SPOT the "kms:Descrypt" permission  was sufficient.

chrisclayson (approved):

ci/terraform/modules/txma-audit-queue/policy.tf
mrwilson: I was following the guidance from https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-key-management.html (Configure KMS permissions for producers) and before I did this I got a "does not have kms:GenerateDataKey permissions" error in the logs.

mrwilson: I'd definitely be interested in reduce permissions if possible, I wonder what we're doing in SPOT that's different.

PR #2231, opened 2022-08-16
AUT-607-validate-otp: Implement new sms validator to validate otp cod…

ayoshebby:
## What?

Create new SMS validator for validating error codes

## Why?

A new validator is required to capture error codes when user attempts to their update phone number in the accounts manager. The validator will return error codes when OTP code is blocked for session, request has been exceeded and invalid OTP code. These different error codes will the then determine the UI page/route to be displayed from the PR below.

## Related PRs

[AUT-607: Build UI pages to represent different error codes](https://github.com/alphagov/di-authentication-account-management/pull/561)


JHjava (commented):
Looks good. Couple of small changes to make and then I would make it a PR!

PR #2271, opened 2022-08-22
AUT-661: Bulk update of account verified flag

chrisclayson:
## What?

- Add new Lambda that sets the `accountVerified` flag in the user profile table if the user either has a verified phone number or a verified MFA method
- The lambda will be manually triggered in the console
- The lambda should scan the user profile table and retrieve records in batches (size is passed as a parameter to lambda)
- The lambda should write in batches of 25 (this is the maximum allowed size by the DynamoDB API)

## Why?

We will be using the `accountVerified` flag as the basis for a new metric and we need to back fill it on existing accounts

## Related PRs

#2232 
#2233


utils/src/main/java/uk/gov/di/authentication/utils/lambda/AccountVerifierLambda.java
```diff
@@ -0,0 +1,127 @@
...
+        if (!updates.isEmpty()) submit(updates);
+
+        return null;
+    }
+
+    private TransactWriteItem updateForItem(Map<String, AttributeValue> item) {
+        return new TransactWriteItem()
+                .withUpdate(
+                        new Update()
+                                .withTableName("sandpit-user-profile")
```
dbes-gds, line 85: per environment names?

utils/src/main/java/uk/gov/di/authentication/utils/lambda/AccountVerifierLambda.java
```diff
@@ -0,0 +1,127 @@
...
+                                .withExpressionAttributeValues(
+                                        Map.of(
+                                                ":accountVerified",
+                                                new AttributeValue().withN(TRUE_N))));
+    }
+
+    private ScanResult getRecords(int batchSize, Map<String, AttributeValue> lastEvaluatedKey) {
+        var request =
+                new ScanRequest()
+                        .withTableName("sandpit-user-profile")
```
dbes-gds, line 97: per account name?

chrisclayson: Commit pushed

utils/src/main/java/uk/gov/di/authentication/utils/lambda/AccountVerifierLambda.java
chrisclayson: Commit pushed

dbes-gds (approved):

PR #2304, opened 2022-08-31
AUT-639: Groundwork to choose Notify templates by language

dbes-gds:
## What?

Groundwork to choose Notify templates by language.

Put in place the infrastructure to be able to choose which Notify template to use by language, but do not make use of it.  Able to demonstrate that the new functionilty will work, and that the existing functionality is unchanged.
A subsequent PR will add the Welsh template ids to the right environment variables, and tweak the code to vary the template by language.

- Expect a new header in the frontend api endpoints containing the user language called 'User-Language'.
- Read the value of this new header in 'BaseFrontendHandler' to make it available in the UserContext, so available to all endpoints that use the base class.
- English is set to the default in all cases where the header is not 'cy' or 'CY', including when the header is missing.
- The header needs to be added to the frontend api calls which trigger notifications.  In the meantime no behaviour changes.
- Add a 'userLanguage' field to 'NotifyRequest', which is the message put on the notify queue that triggers a message send to Notify.
- Update all tests to populate and check for this field in the message.
- In the 'NotificationHandler' and 'NotificationService' that reads messages off the queue and sends them to Notify, only log the value in the 'userLanguage' but do not change the existing behaviour.  Only the existing English templates will be used.

## Why?

Notify messages should be sent either in English or Welsh depending on the user language.
The backend apis do not receive cookies from the frontend, so the language needs to be added as a header in the apis.  This will be done in subsequent frontend PRs.
The user can potentially change their language at any point after the session has been created, so we always need to transmit the current value of the client lng cookie when making api calls that will send Notify messages.  This means the language cannot just be set once when a session is created, and why we can't rely on getting the user language from the session.

## Related PRs

#2292

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/NotificationType.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/NotificationType.java
@@ -1,32 +1,43 @@
 package uk.gov.di.authentication.shared.entity;
 
+import uk.gov.di.authentication.shared.helpers.LocaleHelper.SupportedLanguage;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 
 public enum NotificationType implements TemplateAware {
     VERIFY_EMAIL(
             "VERIFY_EMAIL_TEMPLATE_ID",
-            Map.of(Locale.forLanguageTag("cy"), "VERIFY_EMAIL_TEMPLATE_ID_CY")),
-    VERIFY_PHONE_NUMBER("VERIFY_PHONE_NUMBER_TEMPLATE_ID"),
-    MFA_SMS("MFA_SMS_TEMPLATE_ID"),
-    RESET_PASSWORD("RESET_PASSWORD_TEMPLATE_ID"),
-    PASSWORD_RESET_CONFIRMATION("PASSWORD_RESET_CONFIRMATION_TEMPLATE_ID"),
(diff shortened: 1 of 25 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/NotificationType.java
```diff
@@ -1,21 +1,60 @@
...
+        this(templateName);
+        this.languageSpecificTemplates = languageSpecificTemplates;
+    }
+
     public String getTemplateId() {
         return System.getenv(templateName);
     }
+
+    public String getTemplateId(SupportedLanguage language) {
+        return configurationService.getNotifyTemplateId(getTemplateName(language));
```
chrisclayson, line 50: Do we want to return the ID for the English version of the template if, for whatever reason, the language specific template ID can't be found? Belt and braces!

dbes-gds: Yes good call, it does actually do that at the moment if you look at NotificationType:

   String getTemplateName(SupportedLanguage language) {
        if (languageSpecificTemplates.containsKey(language)) {
            return languageSpecificTemplates.get(language);
        } else {
            return templateName;
        }
    }

If it doesn't find the language in languageSpecificTemplates (which contains Welsh only) then it goes back to the default.

Until the next PR this won't actually be in use, it's currently still using the existing getTemplateId()

chrisclayson (approved):

PR #2326, opened 2022-09-06
AUT-708: Add client session id to audit payloads 

ayoshebby:
## What?

Add client session id to audit payloads and implement across all lambdas which have access to client session id.

## Why?

We pass this value (client session id) up to IPV Core and the Document Checking App but don’t publish it to our own audit log.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
@@ -39,7 +39,7 @@ public class AuditService {
 
     public void submitAuditEvent(
             AuditableEvent event,
-            String requestId,
+            String clientSessionId,
             String sessionId,
             String clientId,
             String subjectId,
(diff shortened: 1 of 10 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
```diff
@@ -39,7 +39,7 @@ public AuditService(ConfigurationService configurationService) {
 
     public void submitAuditEvent(
             AuditableEvent event,
-            String requestId,
+            String clientSessionId,
```
dbes-gds, line 42: Were we not using the requestId?
Are txma expecting the clientSessionId to come through like this?

ayoshebby: Yes the request Id was redundant. This is now replaced with client session id and captured in Txma's `withGovukSigninJourneyId`. I will need to do a follow up with Txma to ensure that's pulling through.

dbes-gds (approved):

PR #2337, opened 2022-09-09
AUT-708-part-2: Add client session id to audit payloads

ayoshebby:
## What?

Add client session id to audit payloads for lambdas in frontend-api

- Add `clientSessionId` to `UserContext` which is used in majority of the handlers
- Refactor `BaseFrontendHandler` to set the client session id in the user context in order for all handlers to have access to the `clientSessionId`

## Why?

We pass this value (client session id) up to IPV Core and the Document Checking App but don’t publish it to our own audit log.

## Related PRs

[Part 1](https://github.com/alphagov/di-authentication-api/pull/2336)
[Authorise Handler](https://github.com/alphagov/di-authentication-api/pull/2338)

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -141,9 +141,9 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
                     session,
                     mfaMethodType,
                     input,
-                    context,
                     userContext,
-                    isRegistration);
+                    isRegistration,
+                    clientSessionId);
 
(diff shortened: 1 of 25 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -208,9 +208,9 @@ private void processCodeSession(
             Session session,
             MFAMethodType mfaMethodType,
             APIGatewayProxyRequestEvent input,
-            Context context,
             UserContext userContext,
-            boolean isRegistration) {
+            boolean isRegistration,
+            String clientSessionId) {
```
chrisclayson, line 213: Do you need to add this as a parameter if you've already got the `userContext` object above?

chrisclayson (approved):

PR #2344, opened 2022-09-13
AUT-488: Support EC key signing

DTaylorGDS:
## What?

- Allow RPs to use algorithms other than RSA to sign JWTs using elliptic-curve cryptography.

Note: Although it appears that many files have changed, quite a few of these were a single refactoring of `KeyPairHelper.GENERATE_RSA_KEY_PAIR()` method name to use more conventional camel-case.

## Why?

- Aligns to ADR: https://github.com/alphagov/digital-identity-architecture/blob/e023ab11812afabb68f7cf5d524db9b50c33cc7f/adr/0030-kms-key-selection.md


DTaylorGDS:
Closing as will be superseded by alternative approach

PR #2363, opened 2022-09-21
AUT-389-backend-changes: Return password max retries error code

ayoshebby:
## What?

- Add password max retries to account management during authentication
- Return error code 1028 when account is temporarily locked after 5 unsuccessful attempts

## Why?

- Currently users are able to retry sign in even when their account is locked
- Users should be redirected to a `/account-locked` page instead of login page when they try to sign in again

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/domain/AccountManagementAuditableEvent.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/domain/AccountManagementAuditableEvent.java
@@ -8,7 +8,9 @@ public enum AccountManagementAuditableEvent implements AuditableEvent {
     UPDATE_PHONE_NUMBER,
     ACCOUNT_MANAGEMENT_AUTHENTICATE,
     DELETE_ACCOUNT,
-    SEND_OTP;
+    SEND_OTP,
+    ACCOUNT_TEMPORARILY_LOCKED,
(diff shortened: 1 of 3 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/domain/AccountManagementAuditableEvent.java
```diff
@@ -8,7 +8,9 @@ public enum AccountManagementAuditableEvent implements AuditableEvent {
     UPDATE_PHONE_NUMBER,
     ACCOUNT_MANAGEMENT_AUTHENTICATE,
     DELETE_ACCOUNT,
-    SEND_OTP;
+    SEND_OTP,
+    ACCOUNT_TEMPORARILY_LOCKED,
```
DTaylorGDS, line 12: Were these audit events agreed with someone? Was just looking at frontend-api audit events and can't see an analogous one: https://github.com/alphagov/di-authentication-api/blob/d6d39776743a5be555ea45dd59f0a6e389cf2ea4/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/domain/FrontendAuditableEvent.java

Not to say this is wrong, but maybe worth double checking what events to send if not done already

DTaylorGDS (approved):

PR #2377, opened 2022-09-27
AUT-692: Add Client Name as a dimension for sign in metrics

dbes-gds:
## What?

Add Client Name as a dimension for sign in metrics.

Covers metrics for both account creation and account reuse.

## Why?

Allow reporting on sign ins and sign ups by client name in Grafana (rather then the client id which is less readable)

## Related PRs

#2355


JY0280 (approved):

JHjava:
Is there a risk that the ClientName might change?

PR #2451, opened 2022-10-25
AUT-802 - Return the MfaMethodType from the StartResponse

JHjava:
## What?

- Return the MfaMethodType from the StartResponse

## Why?

- The frontend uplift journey depends on what MFA type the user is configured with. The frontend does not currently know this so we need to return this information from start.
- Currently the frontend directs the user to the /uplift path when a user is configured with the Auth App MFA method, this is incorrect

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
@@ -14,19 +14,25 @@ import uk.gov.di.authentication.shared.conditions.IdentityHelper;
 import uk.gov.di.authentication.shared.conditions.UpliftHelper;
 import uk.gov.di.authentication.shared.entity.ClientRegistry;
 import uk.gov.di.authentication.shared.entity.ClientSession;
+import uk.gov.di.authentication.shared.entity.MFAMethod;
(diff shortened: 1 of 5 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -126,6 +138,23 @@ public UserStartInfo buildUserStartInfo(
                             clientRegistry.isIdentityVerificationSupported(),
                             identityEnabled);
         }
+        if (userContext.getUserProfile().filter(UserProfile::isPhoneNumberVerified).isPresent()) {
+            mfaMethodType = MFAMethodType.SMS;
+        } else if (userContext.getUserCredentials().stream()
+                .filter(t -> Objects.nonNull(t.getMfaMethods()) && !t.getMfaMethods().isEmpty())
+                .anyMatch(
```
mrwilson, line 145: We've got two nested `anyMatch` calls here, can we simplify this condition?

JHjava: That's much nicer!

mrwilson (approved):

PR #2484, opened 2022-11-04
ACI-115: Use CheckUserExists lambda to return user account status

ayoshebby:
## What?

1. Refactor `CheckUserExistsHandler` to include check for  when user account is locked/unlocked
2. Return a new error code `1045 - User account locked`
3. Revert changes from [PR](https://github.com/alphagov/di-authentication-api/pull/2363) as the changes required applies to global Sign In rather than Account Management module only.

## Why?

This is required in the Sign-In journey when a user tries to enter their password after their account has been locked, the page needs to redirect to a new account locked page detailing why the user cannot sign in

## Related PRs

[Backend changes](https://github.com/alphagov/di-authentication-frontend/pull/826)

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
@@ -16,7 +16,6 @@ import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.serialization.Json.JsonException;
 import uk.gov.di.authentication.shared.services.AuditService;
 import uk.gov.di.authentication.shared.services.AuthenticationService;
-import uk.gov.di.authentication.shared.services.CodeStorageService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 import uk.gov.di.authentication.shared.services.DynamoService;
 import uk.gov.di.authentication.shared.services.SerializationService;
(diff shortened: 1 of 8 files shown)


two-first-names (approved):

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
```diff
@@ -219,6 +224,39 @@ void shouldReturn400IfEmailAddressIsInvalid() {
...
+                Map.of(
+                        "Session-Id",
+                        session.getSessionId(),
+                        CLIENT_SESSION_ID_HEADER,
+                        CLIENT_SESSION_ID));
+        event.setRequestContext(contextWithSourceIp("123.123.123.123"));
+        APIGatewayProxyResponseEvent result = handler.handleRequest(event, context);
+
+        assertEquals(400, result.getStatusCode());
+
```
JHjava, line 246: Could try something like and also add an assert to check the expected error response
```suggestion
         assertThat(result, hasStatus(400));
        assertThat(result, hasJsonBody(ErrorResponse.ERROR_1045));
```

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
```diff
@@ -78,26 +68,6 @@ public APIGatewayProxyResponseEvent authenticateRequestHandler(
         try {
             AuthenticateRequest loginRequest =
                     objectMapper.readValue(input.getBody(), AuthenticateRequest.class);
-            var persistentSessionId =
```
JHjava, line 81: How comes the account management changes are being reverted? Could add the reason for the revert to the commit and PR description

JHjava (commented):
PR looks good. One question about the revert and a suggestion on the unit test. It might also benefit from an additional integration test in the `UserExistsIntegrationTest`

JHjava (approved):

PR #2520, opened 2022-11-18
ACI-19: Add isPasswordResetRequired info to login handler response

DTaylorGDS:
## What?

- Add isPasswordResetRequired info to login handler response
- If true, then the frontend can route to the password reset page

## Why?

- Pre-requisite for front end changes to allow user to be prompted to reset password when their existing password is detected as a common password.
- If this PR is approved I intend to use the new field in the frontend [here](https://github.com/alphagov/di-authentication-frontend/blob/66e206c80ee8b4d74530b3bc6d87082caa172d05/src/components/enter-password/enter-password-controller.ts#L92) by adding it to the state machine context, and from there saying that the next page if `resetPasswordRequired` is `true` on a `USER_JOURNEY_EVENTS.CREDENTIALS_VALIDATED` event should be to send the user to the normal Reset Password page, from where they would follow the state flow exactly as if they had manually clicked through from 'I've forgotten my password' to get a reset.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/LoginResponse.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/LoginResponse.java
@@ -36,6 +36,11 @@ public class LoginResponse {
     @Required
     private boolean consentRequired;
 
+    @SerializedName(value = "passwordChangeRequired")
+    @Expose
+    @Required
+    private boolean passwordChangeRequired;
+
     public LoginResponse() {}
 
     public LoginResponse(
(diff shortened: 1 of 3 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/LoginResponse.java
```diff
@@ -36,6 +36,11 @@ public class LoginResponse {
     @Required
     private boolean consentRequired;
 
+    @SerializedName(value = "passwordChangeRequired")
+    @Expose
+    @Required
```
dbes-gds, line 41: As required will need a frontend change at exactly the same time?

dbes-gds: ok as it's the response it shouldn't make a difference

dbes-gds (approved):

PR #2527, opened 2022-11-22
AUT-796 - Rename IPV Sector URI config name

JHjava:
## What?

- Rename IPV Sector URI to Internal Sector URI 

## Why?

- We will eventually use the Internal Sector URI for multiple internal functions. Therefore we need to give it a more generic name.


chrisclayson (approved):
Looks sensible 👍🏻

PR #2540, opened 2022-11-28
BAU - Retry Doc App Token and UserInfo request if the first request fails

JHjava:
## What?

- Retry Doc App Token and UserInfo request if the first request fails

## Why?

- There is currently an issue on the Doc App API where 3 of their la,bdas start timing out after 3 seconds having logged nothing. This appears to be an issue with the AWS lambda environment rather than anything in the Doc App API.
- For an immediate workaround, retry the calls to the Doc App userinfo and token endpoints if the first request is unsuccessful.

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DocAppCriService.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DocAppCriService.java
@@ -88,7 +88,15 @@ public class DocAppCriService {
     public TokenResponse sendTokenRequest(TokenRequest tokenRequest) {
         try {
             LOG.info("Sending TokenRequest");
-            return TokenResponse.parse(tokenRequest.toHTTPRequest().send());
+            int count = 0;
+            int maxTries = 2;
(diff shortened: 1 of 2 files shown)


chrisclayson (dismissed):

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DocAppCriService.java
```diff
@@ -88,7 +88,15 @@ public TokenRequest constructTokenRequest(String authCode) {
     public TokenResponse sendTokenRequest(TokenRequest tokenRequest) {
         try {
             LOG.info("Sending TokenRequest");
-            return TokenResponse.parse(tokenRequest.toHTTPRequest().send());
+            int count = 0;
+            int maxTries = 2;
+            TokenResponse tokenResponse;
+            do {
+                if (count == 1) LOG.warn("Retrying DocApp token request");
```
chrisclayson, line 95: I wonder if `(count > 0)` might be better, but its fine.

chrisclayson (approved):

PR #2571, opened 2022-12-07
AUT-804: Create 2x bulk test users lambdas

DTaylorGDS:
## What?

- Add 2x lambda and 1x S3 bucket (to support one of the lambdas)
- The create lambda allows bulk creation of test users by uploading a CSV file into a new bucket called `${env}-bulk-test-user-bucket`
- The create lambda then reads through that file and uploads to both `UserProfile` and `UserCredentials` tables
- Any user created this way will have a flag indicating they are a test user (new GSI created on both tables for this)
- The delete lambda is intended to be triggered manually only (input is any String, it's not used). This lambda uses the new GSIs to find test users via the new flag, and then deletes them all

## Why?

Requirement to bulk upload users given by test colleagues.


ci/terraform/utils/bulk_test_user_delete_dynamo_access.tf
```diff
@@ -0,0 +1,57 @@
...
+  statement {
+    sid    = "AllowAccessToDynamoTables"
+    effect = "Allow"
+
+    actions = [
+      "dynamodb:BatchGet*",
+      "dynamodb:DescribeTable",
+      "dynamodb:Get*",
+      "dynamodb:Query",
+      "dynamodb:Scan",
```
chrisclayson, line 11: I'm not convinced we need `Scan` as we we'll `Query`ing on the index.

ci/terraform/utils/bulk_test_user_create_dynamo_access.tf
```diff
@@ -0,0 +1,57 @@
...
+  statement {
+    sid    = "AllowAccessToDynamoTables"
+    effect = "Allow"
+
+    actions = [
+      "dynamodb:BatchGet*",
+      "dynamodb:DescribeTable",
+      "dynamodb:Get*",
+      "dynamodb:Query",
+      "dynamodb:Scan",
```
chrisclayson, line 40: I'm not convinced we need `Scan` as we we'll `Query`ing on the index.

ci/terraform/utils/bulk_test_user_delete_dynamo_access.tf
DTaylorGDS: Turns out we did, per your parallel comment on Slack, so keeping in (will resolve the other comment too which is the same lambda)

utils/src/test/resources/test_users_unit_test.txt
```diff
@@ -0,0 +1,76 @@
+Email,Password,Phone2FA,PhoneNumber,AuthApp2FA,AuthAppSecret
```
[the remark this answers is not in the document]
DTaylorGDS: That's a very sensible idea. Will make a note for a follow-up PR

ci/terraform/utils/bulk_test_user_create_lambda.tf
```diff
@@ -0,0 +1,79 @@
...
+  role_name   = "bulk_test_user_create_lambda_role"
+
+  policies_to_attach = [
+    aws_iam_policy.user_profile_dynamo_write_access.arn,
+    aws_iam_policy.user_credentials_dynamo_write_access.arn,
+    aws_iam_policy.bulk_test_user_s3_read_access.arn,
+  ]
+}
+
+resource "aws_lambda_function" "bulk_test_user_create_lambda" {
```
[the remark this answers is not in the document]
DTaylorGDS: Thanks for additional guidance over Slack. Have added these lines:

```
  tracing_config {
    mode = "Active"
  }
```

chrisclayson (approved):

PR #2574, opened 2022-12-07
AUT-915 - Allowing fixed OTP for reset password and remove reset password with a link logic

JY0280:
## What?

- Support fixed otp for the reset password flow 
- Remove reset password with a link logic

## Why?

- To allow the running/debugging of the pertaining reset password AC test scenario.
- We no longer support retting a password with a link so remove the logic to simplify our code


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/ResetPasswordHandler.java
```diff
@@ -163,6 +172,10 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                     PersistentIdHelper.extractPersistentIdFromHeaders(input.getHeaders()));
         } catch (JsonException e) {
             return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
+        } catch (ClientNotFoundException e) {
+            e.printStackTrace();
```
JHjava, line 176: This line can go

JHjava (commented):
A few suggestions. I think it could also benefit from an additional unit test or 2.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/ResetPasswordRequestHandler.java
```diff
@@ -108,40 +110,31 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
-                    userContext.getSession().getSessionId(),
-                    userContext
-                            .getClient()
-                            .map(ClientRegistry::getClientID)
-                            .orElse(AuditService.UNKNOWN),
-                    AuditService.UNKNOWN,
-                    request.getEmail(),
-                    IpAddressHelper.extractIpAddress(input),
-                    AuditService.UNKNOWN,
-                    persistentSessionId);
```
dbes-gds, line 113: Only thing about moving the audit event is that you won't get an audit event if 'validatePasswordResetCount' generates an error, so audit wouldn't know if a user tries to reset their password 5 times and fails

JHjava: Resolved

dbes-gds (approved):

PR #2624, opened 2022-12-19
AUT-783: Test client allowlist to support regex email matching

dbes-gds:
## What?

Test client allowlist to support regex email matching.

Adds facility to add a regex to the test client email allowlist which can match against multiple email addresses. 

## Why?

Added primarily to support performance testing where a large number of existing accounts need to be able to login.
Avoids loading multiple email addresses to the allowlist for performance test runs.

## Related PRs

#2571 
#2615 

Line up with this PR before merging #2574


JHjava (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/helpers/TestClientHelper.java
```diff
@@ -27,15 +29,24 @@ public static boolean isTestClientWithAllowedEmail(
...
         LOG.info(
                 "Is request from a test client with a test client email address: {}",
                 isTestClientWithAllowedEmail);
 
         return isTestClientWithAllowedEmail;
     }
+
+    public static boolean emailMatchesAllowlist(String emailAddress, List<String> regexAllowList) {
+        for (String regex : regexAllowList) {
+            if (Pattern.compile(regex).matcher(emailAddress).find()) {
```
mrwilson, line 46:
```suggestion
            if (Pattern.matches(regex, emailAddress)) {
```

mrwilson: It also might be worth putting a try {} catch {} here in case the regex is malformed else this will crash out.

mrwilson (approved):

DTaylorGDS (approved):

ayoshebby (approved):

PR #2700, opened 2023-01-23
AUT-963 - Send back custom error to RP when it is received from DCMAW

JHjava:
## What?

- When the customDocAppClaim is enabled and we have received the custom error response from DCMAW, then we want to forward on that error directly to the RP.
- This is to mitigate the issue where the user does not have a valid session due to cross browser issues. Therefore we will not have the clientID handy, so retrieve that from parameter store. We can then use the clientID to get the redirectURI.
- Normally we would use the redirectURI in the auth request but as that is not available, we will get it from the clientRegistry. There could be multiple redirectURIs configured so we will just get the first one in this case.

## Why?

- The users were not being redirected back to the RP because of cross browser issues so this is to mitigate that by identifying which users have been impacted and redirecting them back to the RP with an error.

## TODO 
- Confirm with DCMAW which custom error they will be sending back and make the changes to this draft PR 
- Confirm that there is only ever 1 redirect URI on record with this RP

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
@@ -5,7 +5,9 @@ import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
 import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.auth.verifier.InvalidClientException;
 import com.nimbusds.oauth2.sdk.http.HTTPRequest;
 import com.nimbusds.openid.connect.sdk.AuthenticationErrorResponse;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
@@ -24,15 +26,18 @@ import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.services.AuditService;
 import uk.gov.di.authentication.shared.services.ClientSessionService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
+import uk.gov.di.authentication.shared.services.DynamoClientService;
 import uk.gov.di.authentication.shared.services.JwksService;
 import uk.gov.di.authentication.shared.services.KmsConnectionService;
 import uk.gov.di.authentication.shared.services.RedisConnectionService;
 import uk.gov.di.authentication.shared.services.SerializationService;
 import uk.gov.di.authentication.shared.services.SessionService;
 
+import java.net.URI;
 import java.util.Map;
 
 import static com.nimbusds.oauth2.sdk.http.HTTPRequest.Method.POST;
+import static java.lang.String.format;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
 import static uk.gov.di.authentication.shared.helpers.ConstructUriHelper.buildURI;
 import static uk.gov.di.authentication.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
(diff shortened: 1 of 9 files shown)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -111,6 +119,32 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
...
+                                                new InvalidClientException(
+                                                        format(
+                                                                "No client found with clientId: %s",
+                                                                clientId)));
+                var redirectUrl = clientRegistry.getRedirectUrls().get(0);
+                //        TODO - Pass error received from DCMAW back to RP and ensure additional
+                // query params are included
+                var errorResponse =
+                        new AuthenticationErrorResponse(
+                                URI.create(redirectUrl),
```
galund, line 138: I'm pretty confident that we don't want to use "temporarily unavailable", let's review this Thursday :-)

galund (changes requested):

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
JHjava: We don't. That's why there is a TODO to update it when the expected error is known

galund: We have an [RFC update](https://github.com/alphagov/digital-identity-architecture/pull/158) that I'd appreciate your and @SenthilKKumar eyes on. This RFC really only covers credential issuers and IPV Core, so potentially we should have a similar(ish) update in the OpenID Connect profile RFC too?

LMK if you'd like to discuss

JHjava: @galund mind having a look? Error has been updated for to reflect the RFC. There will be a slightly different implementation with IPV to be able to get the session from the state.

galund: So the goal (as far as I'm concerned) of our conversations was to ensure Orchestration didn't have to do something particularly different for DCMAW vs IPV Core. Sounds like we need a rerun of the conversation the other day as it feels like we haven't arrived in that place.

I thought we agreed that the trigger will be the `access_denied` and you should *not* be parsing the `error_description` looking for a special "missing context" string.

Instead, when you receive the OAuth error, there are two cases depending whether there is a current active valid session in Orchestration:

* if there is a session, the user cancelled out normally.
* if there is no session, then the user may be able to pick up their journey if they log in again and come back.

We can return the same error in both cases to HMRC ie `access_denied`, and we should use the same `state` that was passed in on the original request. Note that `state` is mandatory, so I don't think we can get away with not looking up the session, in that second case?

HMRC should then perform similar logic:

* if there is a session, the user cancelled out and they should find another way to prove the user's identity (or whatever they do next)... but they might also want to *offer* the user to try again with the app, because people change their minds.
* if there is no session, they will need to sign the user in with SCP again, and *then* they should offer the user to try again with the App (and in this case, they are likely - we hope - to succeed).

SenthilKKumar: Currently there is a technical challenge, Orchestration will be able to identify appropriate RP session using `state` only when the RP sessions are managed using DynamoDB instead of Redis. 

cc: @mrwilson

galund: Ah, so maybe we can't fix this for anyone but HMRC just yet anyway. Still it'd be nice to be clear with HMRC that what they should do depends on whether they have a current session, rather than on magic words in the error description (assuming that really is what we agreed the other day, and I think you had to miss that Senthil but if you and Alex and anyone else wants to chat, do shout.)

JHjava: Senthil and I spoke. So we are not dependent on the move to Dynamo, we are going to store the client session id against the state in Redis before we hand off to DCMAW. This is outlined in PR https://github.com/alphagov/di-authentication-api/pull/2793 which supersedes this one

JHjava:
Closing this PR. It has been superseded by PR - https://github.com/alphagov/di-authentication-api/pull/2793

PR #2709, opened 2023-01-25
AUT-834: WAF rules exception for the smoke test client

dbes-gds:
## What?

WAF rules exception for the smoke test client.

This PR overrides a specific WAF rule for the smoke test client only.  The client id is added to SSM in the smoketest pipeline and retrieved when creating the OIDC api.

## Why?

The smoke test microclient runs on localhost in the smoke test, however the WAF will block authorize requests containing a redircect_uri to localhost, so the smoke test will not run.

## Related PRs

https://github.com/alphagov/di-authentication-smoke-tests/pull/69
https://github.com/alphagov/di-authentication-smoke-tests/pull/66

### DIFF
--- a/ci/terraform/oidc/api-gateway.tf
+++ b/ci/terraform/oidc/api-gateway.tf
@@ -289,10 +289,6 @@ module "dashboard" {
   use_localstack   = var.use_localstack
 }
 
-data "aws_ssm_parameter" "smoke_test_client_id" {
-  name = "${var.environment}-smoke-test-client-id"
-}
-
 resource "aws_wafv2_web_acl" "wafregional_web_acl_oidc_api" {
   count = var.use_localstack ? 0 : 1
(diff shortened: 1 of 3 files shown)


ci/terraform/oidc/api-gateway.tf
```diff
@@ -289,6 +289,10 @@ module "dashboard" {
   use_localstack   = var.use_localstack
 }
 
+data "aws_ssm_parameter" "smoke_test_client_id" {
+  name = "${var.environment}-smoke-test-client-id"
+}
+
```
JHjava, line 295: Maybe put this in `ssm.tf`. Are you looking to add the other allowed client ids later?

JHjava (commented):
This looks good. We can update the AWS provider version for the other modules after so they are in sync.

JHjava (approved):

JHjava (approved):

PR #2786, opened 2023-02-23
AUT-923: Only put otp in the test bucket for certain message types

dbes-gds:
## What?

Only put otp in the test bucket for certain message types.

## Why?

When running smoke tests (and test client tests) we put the otp in a bucket for the test client to retrieve.  However, not all Notify message types send an OTP, so there is no need to put the otp in the bucket for those messages.

Fixes an error when the create account smoke test runs.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/NotificationHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/NotificationHandler.java
@@ -10,6 +10,7 @@ import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
+import uk.gov.di.authentication.shared.entity.NotificationType;
 import uk.gov.di.authentication.shared.entity.NotifyRequest;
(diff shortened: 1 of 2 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/NotificationHandler.java
```diff
@@ -182,11 +186,18 @@ private String buildContactUsUrl(String referer) {
...
+    void writeTestClientOtpToS3(NotificationType notificationType, String otp, String destination) {
         Boolean isNotifyDestination =
                 configurationService.getNotifyTestDestinations().contains(destination);
-        if (isNotifyDestination) {
-            LOG.info("Notify Test Destination used in request. Writing to S3 bucket");
+        Boolean isOTPNotificationType =
+                VERIFY_EMAIL.equals(notificationType)
+                        || VERIFY_PHONE_NUMBER.equals(notificationType)
+                        || MFA_SMS.equals(notificationType)
+                        || RESET_PASSWORD_WITH_CODE.equals(notificationType);
```
JHjava, line 196: Minor point but something like this might be more readable - 
```suggestion
        var isOTPNotificationType = List.of(VERIFY_EMAIL, MFA_SMS, VERIFY_PHONE_NUMBER, RESET_PASSWORD_WITH_CODE).contains(notificationType);
```

JHjava (approved):

PR #2788, opened 2023-02-24
BAU: Accept FIXED_LINE_OR_MOBILE numbers

dbes-gds:
## What?

Accept FIXED_LINE_OR_MOBILE numbers.

## Why?

Following IPN switch-on we are seeing traffic from countries where number ranges can either be fixed or mobile.  Add support for these numbers so that we accept them through phone number validation.

https://groups.google.com/g/libphonenumber-discuss/c/flHAmI5-1sw

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ValidationHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ValidationHelper.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.shared.helpers;
 
 import com.google.i18n.phonenumbers.NumberParseException;
 import com.google.i18n.phonenumbers.PhoneNumberUtil;
+import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberType;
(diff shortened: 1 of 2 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/ValidationHelper.java
```diff
@@ -84,6 +86,10 @@ public static Optional<ErrorResponse> validatePhoneNumber(
         }
     }
 
+    static boolean isAcceptedPhoneNumberType(PhoneNumberType phoneNumberType) {
+        return MOBILE.equals(phoneNumberType) || FIXED_LINE_OR_MOBILE.equals(phoneNumberType);
+    }
+
```
JHjava, line 92: I wonder if we should log out here the phone number type which might help identifying any issues with users not receiving SMS?

dbes-gds: have added

JHjava (approved):

PR #2793, opened 2023-02-27
AUT-963 - Perform lookup on state param when no session cookie is present

JHjava:
## What?

- Save the clientSessionID against the STATE param before redirecting the user to DCMAW
- Redirect error back to RP when no session cookie is present by looking up the users client session ID using the STATE param in the error response back from DCMAW 
- The clientSessionID stored against the STATE param will only be used in the case of a error received back from DCMAW

## Why?

- When a user has no session cookie, we need to be able to determine the users client session so we can find the RPs initial authentication request and navigate the user back to the RP. To be able to achieve this we need to store the clientSessionID against the state, so when no session cookie is present we can perform a look up using the state and be able to send a error response back to the RP
- There are some cases where a user is redirected back to Orchestration where there isn't a session present for the user. This can be a result of a cross-browser issue when a mobile user completes their app journey and is handled back to the web. Is the browser opened by the app isn't the same browser which the user started their journey, then the user will not have a session cookie present and Orchestration will display an error to the user.
- To return the users back to the RP when no session cookie is present, we will do a lookup in our session store using the state param to retrieve a client session ID. This client session ID will then be used to retrieve the RP's initial authentication request which will contain the RPs redirect URI and state param. This will allow orchestration to redirect an `access_denied` error back to the RP with the RPs state param from the initial authentication request. The RP can then take appropriate action.


galund (approved):
Don't take this as gospel but I'm seeing all the logic I expected to see here, and similar will apply with IPV Core as soon as we've figured out how to explain to RPs the change we're making (shout if I can help with that).

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -287,4 +289,51 @@ private APIGatewayProxyResponseEvent redirectToFrontendErrorPage() {
...
+    private APIGatewayProxyResponseEvent generateNoSessionErrorResponse(
+            Map<String, String> queryStringParameters)
+            throws ParseException, DocAppCallbackException {
+        LOG.info(
+                "Attempting to generate error response using state. CustomDocAppClaimEnabled: {}",
+                configurationService.isCustomDocAppClaimEnabled());
+        if (configurationService.isCustomDocAppClaimEnabled()
+                && queryStringParameters.containsKey("error")
+                && queryStringParameters.get("error").equals(OAuth2Error.ACCESS_DENIED.getCode())
+                && queryStringParameters.containsKey("state")) {
```
mrwilson, line 302: I'd suggest pulling this check into a separate method for readability.

JHjava: Good shout. I have updated

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -287,4 +289,55 @@ private APIGatewayProxyResponseEvent redirectToFrontendErrorPage() {
...
+            throw new DocAppCallbackException(
+                    "Session Cookie not present and access_denied or state param missing from error response");
+        }
+    }
+
+    private boolean isAccessDeniedErrorAndStatePresent(Map<String, String> queryStringParameters) {
+        return configurationService.isCustomDocAppClaimEnabled()
+                && queryStringParameters.containsKey("error")
+                && queryStringParameters.get("error").equals(OAuth2Error.ACCESS_DENIED.getCode())
+                && queryStringParameters.containsKey("state");
```
mrwilson, line 341: A very minor thing - I reckon we should put another condition on the end of this chain which is "state param is non-empty".

I _think_ that it's possible (since the callback happens in browser and urls can be changed) that it could be an empty string (`&state=&...`) which we would want to fail on?

Not a blocker though, if we roll this into a future PR.

mrwilson (approved):

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
JHjava: 👍 Good point. Agreed to resolve in a future PR.

PR #2810, opened 2023-03-03
AUT-1125 - Create shared class for no session orchestration logic

JHjava:
## What?

- The NoSessionOrchestrationService is called in the scenario where no session cookie is found for the user once the user returns from either DCMAW or IPV.
- The main method in this class is the `generateNoSessionOrchestrationEntity` which checks if the query parameters it receives contains an `access_denied` error response along with a `state` parameter. It will also check that this feature is enabled as to start with it will not be enabled across all environments to allow testing.
- If the access_denied error response and state parameter are both present, it will perform a lookup in Redis using the state to retrieve a client session id. It will use that client session id to perform an additional lookup in Redis to retrieve the client session of the user. The client session is required to be able to retrieve the RPs initial authentication request which will contain the RP redirect URI and RP state param, which are both required for sending the access_denied error response back to the RP.
- The generateNoSessionOrchestrationEntity will finally return a NoSessionEntity object, which will contain everything required for the lambda to return the user back to the RP with the access_denied error response.

## Why?

- The logic currently in the DocAppCallback lambda will be very similar for the IpvCallback lambda. Therefore it makes sense to pull this out into a shared class to reduce code duplication and have a single point for the logic.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/NoSessionOrchestrationService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/NoSessionOrchestrationService.java
@@ -1,0 +1,101 @@
+package uk.gov.di.authentication.shared.services;
+
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
+import com.nimbusds.oauth2.sdk.id.State;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.shared.entity.NoSessionEntity;
+import uk.gov.di.authentication.shared.exceptions.NoSessionException;
+
+import java.util.Map;
+import java.util.Optional;
+
+import static java.lang.String.format;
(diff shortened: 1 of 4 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/NoSessionOrchestrationService.java
```diff
@@ -0,0 +1,101 @@
...
+                    "Session Cookie not present and access_denied or state param missing from error response. NoSessionResponseEnabled: {}",
+                    noSessionResponseEnabled);
+            throw new NoSessionException(
+                    format(
+                            "Session Cookie not present and access_denied or state param missing from error response. NoSessionResponseEnabled: %s",
+                            noSessionResponseEnabled));
+        }
+    }
+
+    public void storeClientSessionIdAgainstState(String clientSessionId, State state) {
```
dbes-gds, line 79: When will this get called?

JHjava: It will get called during either the IPV or DCMAW handoff

dbes-gds (approved):

PR #2860, opened 2023-03-23
Only use lock if lock table is specified

two-first-names:
## What?

Only use lock if lock table is specified

## Why?

Prod was trying to use the lock table, despite none being specified.


ci/tasks/deploy-test-services.yml
```diff
@@ -10,6 +10,7 @@ params:
   DEPLOYER_ROLE_ARN: ((deployer-role-arn-non-prod))
   DEPLOY_ENVIRONMENT: build
   STATE_BUCKET: digital-identity-dev-tfstate
+  STATE_LOCKING_TABLE: digital-identity-dev-tfstate-locking
```
dbes-gds, line 13: this task has it set?

two-first-names: That's a "whoops I didn't save it in my last PR" 😅

PR #2896, opened 2023-04-06
Terraform will error instead of destroy zones

two-first-names:
## What?

Terraform will error instead of destroy zones

## Why?

This will ensure hosted zones don't accidentally get deleted, as the name server values will be hardcoded in the root zone.


gtvj (approved):
Hi @two-first-names - I'm no expert in Terraform but have looked at the docs and this looks good to me.

PR #2926, opened 2023-04-19
AUT-1209: Switch auth app validator to use independent counter

DTaylorGDS:
## What?
- Auth app validator will increment a counter in code storage service cache which is specific to its MFA Type
- SMS will continue to use a generic (non-SMS-specific) prefix until the validation method can be switched out from using the static validation helper to using `MfACodeValidator`'s class methods
- This means that in the `VerfifyMfaCodeHandler`, we still need to delete the counter for both the MFA specific and generic prefixes, see: `blockCodeForSessionAndResetCountIfBlockDoesNotExist()` method for where this happens
- Note: The business logic change applies only to auth apps. SMS will continue to use the 'old' style prefix for now because it uses an older validator helper rather than the methods of the parent class whilst validating OTP correctness. This is deemed okay because the bug identified only really affects account creation and users who attempt to use auth apps first before switching to SMS. Additionally, since there are only two types, the 'generic' key without the compound MFA type modifier will effectively belong only to SMS. 

## Why?

- Fixes a bug whereby a user could be blocked from entering OTPs after entering 5 incorrect Auth App OTPs followed by a single incorrect SMS OTP during sign up, because the cache was storing both types of incorrect code entry under the same prefix/key in the cache

## Related PRs
A previous PR filled the cache with the new prefix (doubled up with updating the counter using the old prefix too) so that the switch over of checking if incorrect attempts have been exceeded will not allow more than the current number of attempts by combining limits across both prefixes: https://github.com/alphagov/di-authentication-api/pull/2924

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -195,6 +195,7 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
             APIGatewayProxyRequestEvent input,
             UserContext userContext,
             boolean isRegistration) {
+        String emailAddress = session.getEmailAddress();
 
         var auditableEvent =
                 errorResponse
@@ -204,20 +205,18 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
         if (isRegistration
                 && errorResponse.isEmpty()
                 && mfaMethodType.equals(MFAMethodType.AUTH_APP)) {
-            authenticationService.setAccountVerified(session.getEmailAddress());
-            authenticationService.setMFAMethodVerifiedTrue(
-                    session.getEmailAddress(), mfaMethodType);
+            authenticationService.setAccountVerified(emailAddress);
+            authenticationService.setMFAMethodVerifiedTrue(emailAddress, mfaMethodType);
         } else if (isRegistration
                 && errorResponse.isEmpty()
                 && mfaMethodType.equals(MFAMethodType.SMS)) {
-            authenticationService.updatePhoneNumberAndAccountVerifiedStatus(
-                    session.getEmailAddress(), true);
+            authenticationService.updatePhoneNumberAndAccountVerifiedStatus(emailAddress, true);
         }
 
         if (errorResponse
                 .map(t -> List.of(ErrorResponse.ERROR_1034, ErrorResponse.ERROR_1042).contains(t))
                 .orElse(false)) {
-            blockCodeForSessionAndResetCountIfBlockDoesNotExist(session, mfaMethodType);
+            blockCodeForSessionAndResetCountIfBlockDoesNotExist(emailAddress, mfaMethodType);
         }
 
         auditService.submitAuditEvent(
@@ -229,7 +228,7 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
                         .map(ClientRegistry::getClientID)
                         .orElse(AuditService.UNKNOWN),
                 session.getInternalCommonSubjectIdentifier(),
-                session.getEmailAddress(),
+                emailAddress,
                 IpAddressHelper.extractIpAddress(input),
                 AuditService.UNKNOWN,
                 extractPersistentIdFromHeaders(input.getHeaders()),
@@ -237,8 +236,7 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
     }
 
     private void blockCodeForSessionAndResetCountIfBlockDoesNotExist(
-            Session session, MFAMethodType mfaMethodType) {
-        String emailAddress = session.getEmailAddress();
+            String emailAddress, MFAMethodType mfaMethodType) {
 
         if (codeStorageService.isBlockedForEmail(emailAddress, CODE_BLOCKED_KEY_PREFIX)) {
             return;
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -238,15 +238,21 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
 
     private void blockCodeForSessionAndResetCountIfBlockDoesNotExist(
             Session session, MFAMethodType mfaMethodType) {
-        if (!codeStorageService.isBlockedForEmail(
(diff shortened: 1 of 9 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/CodeStorageService.java
```diff
@@ -50,6 +52,16 @@ public int getIncorrectMfaCodeAttemptsCount(String email) {
...
+    public int getIncorrectMfaCodeAttemptsCount(String email, MFAMethodType mfaMethodType) {
+        Optional<String> count =
+                Optional.ofNullable(
+                        redisConnectionService.getValue(
+                                MULTIPLE_INCORRECT_MFA_CODES_KEY_PREFIX
+                                        + mfaMethodType.getValue()
+                                        + HashHelper.hashSha256String(email)));
+        return count.map(Integer::parseInt).orElse(0);
+    }
     // TODO: remove this transitional method when cache reflects MFA-type-specific prefixing i.e.
```
JHjava, line 65: Same as above. Can we also remove it for the `deleteIncorrectMfaCodeAttemptsCount`

JHjava (commented):
A few minor comments. I think it could also do with an Integration test to give us an extra degree of confidence that the fix works.

shared/src/main/java/uk/gov/di/authentication/shared/validation/MfaCodeValidator.java
```diff
@@ -26,22 +26,17 @@ boolean isCodeBlockedForSession() {
...
-        codeStorageService.increaseIncorrectMfaCodeAttemptsCount(
-                emailAddress); // TODO: remove this transitional method call when cache reflects
-        // only following line
         codeStorageService.increaseIncorrectMfaCodeAttemptsCount(emailAddress, mfaMethodType);
     }
 
     void resetCodeIncorrectEntryCount(MFAMethodType mfaMethodType) {
-        codeStorageService.deleteIncorrectMfaCodeAttemptsCount(
-                emailAddress); // TODO: remove this transitional method call when cache reflects
-        // only following line
```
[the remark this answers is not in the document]
DTaylorGDS: In as much as our changes apply only to SMS, which will be using a generic prefix in the validator when incrementing, do we not want to keep this reference to make sure we also perform the 'generic' delete? Thinking about it, maybe it should look something like this?:

```
    private void blockCodeForSessionAndResetCountIfBlockDoesNotExist(
            Session session, MFAMethodType mfaMethodType) {
        if (!codeStorageService.isBlockedForEmail(
                session.getEmailAddress(), CODE_BLOCKED_KEY_PREFIX)) {
            codeStorageService.saveBlockedForEmail(
                    session.getEmailAddress(),
                    CODE_BLOCKED_KEY_PREFIX,
                    configurationService.getBlockedEmailDuration());
             if(mfaMethodType === MFAMethodType.SMS){
                codeStorageService.deleteIncorrectMfaCodeAttemptsCount(session.getEmailAddress());
            } else {
                codeStorageService.deleteIncorrectMfaCodeAttemptsCount(
                    session.getEmailAddress(), mfaMethodType);
                    }
        }
    }
```

shared/src/main/java/uk/gov/di/authentication/shared/services/CodeStorageService.java
```diff
@@ -42,6 +42,8 @@ public CodeStorageService(RedisConnectionService redisConnectionService) {
         this.redisConnectionService = redisConnectionService;
     }
 
+    // TODO: remove this transitional method when cache reflects MFA-type-specific prefixing i.e.
+    // method call with MFAMethodType argument
```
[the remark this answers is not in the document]
DTaylorGDS: Sure :+1:

shared/src/main/java/uk/gov/di/authentication/shared/services/CodeStorageService.java
DTaylorGDS: Sure :+1:

shared/src/main/java/uk/gov/di/authentication/shared/validation/MfaCodeValidator.java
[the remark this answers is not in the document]
JHjava: Yes, great point. Completely missed that!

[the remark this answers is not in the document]
DTaylorGDS: Cool, thanks. This change now adopted in latest commit.

integration-tests/src/test/java/uk/gov/di/authentication/api/VerifyMfaCodeIntegrationTest.java
```diff
@@ -281,9 +281,34 @@ void whenAuthAppCodeRetriesLimitExceededForSignInBlockEmailAndReturn400()
...
-        assertThat(userStore.isAuthAppVerified(EMAIL_ADDRESS), equalTo(true));
+        assertEquals(0, redis.getMfaCodeAttemptsCount(EMAIL_ADDRESS, MFAMethodType.AUTH_APP));
+        assertTrue(redis.isBlockedMfaCodesForEmail(EMAIL_ADDRESS));
+        assertTrue(userStore.isAccountVerified(EMAIL_ADDRESS));
+        assertTrue(userStore.isAuthAppVerified(EMAIL_ADDRESS));
+    }
+
+    @Test
+    void
+            whenAuthCodeRetriesLimitReachedButEmailAddressNotBlockedAllowSmsAttemptAndReturn400WithoutBlockingFurtherRetries()
```
JHjava, line 292: The test looks good but could be renamed as there is no auth code retry limit during sign up. It's about ensuring that the auth code retry count doesn't impact the SMS retries.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -238,15 +238,21 @@ private void processCodeSession(
 
     private void blockCodeForSessionAndResetCountIfBlockDoesNotExist(
             Session session, MFAMethodType mfaMethodType) {
```
JHjava, line 240: Aware this already existed but we could pass emailAddress directly into this method.

JHjava (approved):
A couple of minor comment but the change looks good overall

integration-tests/src/test/java/uk/gov/di/authentication/api/VerifyMfaCodeIntegrationTest.java
DTaylorGDS: I'd forgotten that, but good point; changing now

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
DTaylorGDS: Also a good point; changing now

JHjava (dismissed):

JHjava (approved):

PR #2929, opened 2023-04-20
AUT-1210: Make MFA code block specific per MFA type

DTaylorGDS:
## What?
- Currently there is one prefix used when cacheing the block against an email address
- This is shared between SMS and auth app
- We will change this to a compound prefix which references the MFA type

## Why?
- This will mean a block will apply only to a single type
- e.g. if SMS is blocked, auth app will not be blocked


DTaylorGDS:
Got out of sync with main branch and change was so trivial it was faster to replicate on a new branch

PR #2931, opened 2023-04-20
AUT-1207 - Update the MFA Method for registration requests in the VerifyMfaCodeHandler

JHjava:
## What?

- Update the users phone number or auth app secret in the VerifyMfaCodeHandler for registration requests
- Use the authAppSecret in the VerifyMfaCodeRequest to validate the code for AUTH_APP registration requests 
- We can leave the updating of the user profile in update-profile there for now as the call to the `/verify-mfa-code` endpoint will just overwrite it if it exists
- Validate Auth App secret in VerifyMfaCodeHandler
- Add audit events for when the profile is updated in VerifyMfaCodeHandler

## Why?

- This is required so that we can retain the users existing MFA method until their new one has been confirmed during the account recovery journey.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -125,6 +125,10 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
                     mfaCodeValidator.validateCode(
                             codeRequest.getCode(), codeRequest.getProfileInformation());
 
+            if (errorResponse.filter(ErrorResponse.ERROR_1041::equals).isPresent()) {
+                return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1041);
+            }
+
             processCodeSession(errorResponse, session, input, userContext, codeRequest);
 
             sessionService.save(session);
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -39,6 +39,7 @@ import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent
 import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent.CODE_VERIFIED;
 import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent.INVALID_CODE_SENT;
 import static uk.gov.di.authentication.shared.entity.LevelOfConfidence.NONE;
+import static uk.gov.di.authentication.shared.entity.MFAMethodType.AUTH_APP;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
 import static uk.gov.di.authentication.shared.helpers.PersistentIdHelper.extractPersistentIdFromHeaders;
 import static uk.gov.di.authentication.shared.services.AuditService.MetadataPair.pair;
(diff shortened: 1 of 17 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/validation/MfaCodeValidator.java
```diff
@@ -40,5 +40,5 @@ void resetCodeIncorrectEntryCount(MFAMethodType mfaMethodType) {
         codeStorageService.deleteIncorrectMfaCodeAttemptsCount(emailAddress, mfaMethodType);
     }
 
-    public abstract Optional<ErrorResponse> validateCode(String code);
+    public abstract Optional<ErrorResponse> validateCode(String code, String profileInformation);
```
DTaylorGDS, line 43: Two little questions here:

1) I wonder if there is an alternative name that would be more descriptive of the purpose? I guess I think of this as something like `optionalSecretOverride`. The drawback with calling it `profileInformation` is that in the although the context is clear where you're using it, in other lambda (verify code) if we start using the same abstract class, that variable name will be less meaningful

2) I also wonder if there is any reason not to just overload the method so that for validators where this field will never make sense, we can just ignore it and continue to call `validateCode(String code)` i.e. keep a second method that just does something like this:

```
    Optional<ErrorResponse> validateCode(String code){
        return validateCode(code, null);
    }
```

It's not so much that this makes any functional difference; but maybe if we then only called validate with the extra parameter for auth app registration scenario, it could make the handler logic that calls `validateCode` more explicit. But that's probably just personal preference/way I'd read the code, so may not seem better for other people.

DTaylorGDS (dismissed):

shared/src/main/java/uk/gov/di/authentication/shared/validation/MfaCodeValidator.java
DTaylorGDS: Even I don't think 2 makes sense after a little reflection....!

JHjava: Agreed. I've tried to make it clearer by creating an abstract class for the Code Request. Let me know what you think

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -120,10 +122,13 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                 return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1002);
             }
 
-            var errorResponse = mfaCodeValidator.validateCode(codeRequest.getCode());
+            var errorResponse = mfaCodeValidator.validateCode(codeRequest);
 
-            processCodeSession(
-                    errorResponse, session, mfaMethodType, input, userContext, isRegistration);
+            if (errorResponse.filter(t -> t.equals(ErrorResponse.ERROR_1041)).isPresent()) {
```
mrwilson, line 127: Small suggestion to use a method reference rather than a lambda
```suggestion
            if (errorResponse.filter(ErrorResponse.ERROR_1041::equals).isPresent()) {
```

JHjava: Good shout 👍  I'll update this.

JHjava: Done

DTaylorGDS (approved):
For what it's worth, this seems smart to me, and personally find this more explicit now in terms of interpreting the purpose of the code 👍

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -191,49 +196,60 @@ private FrontendAuditableEvent errorResponseAsFrontendAuditableEvent(
...
-        } else if (isRegistration
-                && errorResponse.isEmpty()
-                && mfaMethodType.equals(MFAMethodType.SMS)) {
-            authenticationService.updatePhoneNumberAndAccountVerifiedStatus(emailAddress, true);
-            authenticationService.setMFAMethodEnabled(emailAddress, MFAMethodType.AUTH_APP, false);
+        submitAuditEvent(
+                auditableEvent, session, userContext, input, codeRequest.getMfaMethodType());
+
+        if (codeRequest.isRegistration() && errorResponse.isEmpty()) {
+            switch (codeRequest.getMfaMethodType()) {
```
DTaylorGDS, line 213: Sonar doesn't like the switch. I don't really have an opinion

PR #2941, opened 2023-04-25
AUT-1207 - Port phoneNumber validation to send notification

JHjava:
## What?

- Port phoneNumber validation to send notification

## Why?

- We will soon stop calling update-profile for phone numbers but we don't want to lose the backend validation.


ayoshebby (approved):

DTaylorGDS (approved):
Looks good to me

PR #2944, opened 2023-04-26
AUT-1207 - Update userprofile in a one dynamo call

JHjava:
## What?
 
- Update userprofile in a one dynamo call from VerifyMfaCode

## Why?

- Rather than making multiple calls to update the phone number and then update the verify status, we can update it all in a single call


DTaylorGDS (approved):
Nice

PR #2962, opened 2023-05-03
AUT-1242-part-1: Introduce JourneyType parameter in VerifyMfaCodeRequest

ayoshebby:
## What?

Part 1 of AUT-1242 introduces `JourneyType` parameter in the `VerifyMfaCodeRequest` object for journey type values of:
-  Registration
- Account Recovery
- Sign In

## Why?

- So that the VerifyMfaCode lambda is aware of what journey is being parsed in the request
- Break the work required for AUT-1242 into multiple PRs


JHjava (commented):
Change looks good. One minor point would be update where you're using `JourneyType.UNKNOWN` in the UnitTests for a sign in flow to `JourneyType.SIGN_IN`

JHjava (approved):

PR #2984, opened 2023-05-11
AUT-1242 -  Refactor MFA CodeProcessors ahead of AccountRecovery changes

JHjava:
## What?

This PR contains no changes of behaviour and is more a refactoring to make it easy to add support when processing AccountRecovery MFA code requests.

**Key changes in this PR**
- Remove isRegistration from the VerifyMfaCodeRequest as it is no longer being sent from the frontend and is therefore redundant 
- Move JourneyType to the abstract CodeRequest class
- Use Dynamo Transactional write requests when updating multiple tables 
- Extend the usage of the CodeProcessor classes by adding an `processSuccessfulCodeRequest` method
- Move classes from shared to frontend-api 

## Why?

- We can make use of the existing CodeProcessor classes to take care of the logic when processing MfaCodeRequests. This improves readability of the VerifyMfaCodeHandler and makes it easier to follow. 
- The VerifyMfaCode classes are only used by the frontend-api and it makes sense to keep them there. This will slim down the size of shared.
- Dynamo transactional requests will ensure that either the whole dynamo requests completes or does not complete. Either way, we can be more confident the database will not be left in an inconsistent state.


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/AuthAppCodeProcessor.java
```diff
@@ -84,6 +93,21 @@ public Optional<ErrorResponse> validateCode(CodeRequest codeRequest) {
         return Optional.empty();
     }
 
+    @Override
+    public void processSuccessfulCodeRequest(String ipAddress, String persistentSessionId) {
+        if (codeRequest.getJourneyType().equals(JourneyType.REGISTRATION)) {
```
JHjava, line 98: This will be eventually turned into a switch statement. AccountRecovery journey type will remove the old method and add the new one whilst SignIn journey type will remove the block if one is present

ayoshebby (approved):

PR #3025, opened 2023-05-24
BAU: Log smoke test otp and phone number in integration only

dbes-gds:
## What?

Log smoke test otp and phone number in integration only.

## Why?

Create account smoke tests are failing intermittently due to the test otp being incorrect.  Adding additional logging to help diagnose the issue.


JHjava (changes requested):
You need to declare the `ENVIRONMENT` environment variable in sqs.tf

PR #3090, opened 2023-06-19
AUT-1348: Switch OTP request block check to new prefix

DTaylorGDS:
## What?
- A previous commit introduced a new prefix style for OTP request blocking
- Historically, there was a single OTP code request block per session
- The new approach subdivides this by request type
- For example, a block for email OTP in registration does not mean the session is blocked for an SMS OTP request
- This commit builds on that change, by switching over business logic where the blocking status is checked
- It will now use the 'new' (compound) prefix and not the old prefix

## Why?
- Avoids 'cross pollution' between different types of code request block

## Related PRs
- New prefix style introduced here: https://github.com/alphagov/di-authentication-api/pull/3086


JHjava (commented):
Could be worth some additional unit and/or integration tests to verify that blocks on other Journey and MFA type don't impact another Journey and MFA type.

JHjava (approved):

PR #3099, opened 2023-06-20
AUT-1349: Add MFAMethodType to CodeRequestType

DTaylorGDS:
## What?
- CodeRequestType was keyed off NotificationType
- However, this is not universal
- All handlers either have direct access to an MFAMethodType or can infer it from the NotificationType
- The reverse is not true
- Making this change will therefore allow certain handlers e.g. VerifyMfaCodeHandler to use the getCodeRequestType static method

## Why?
- This will in turn allow all handers to obtain CodeRequestTypes for the purpose of computing Redis prefixes on a per-OTP basis
- By using the same CodeRequestType class as the basis, this will be cleaner and more consistent than if we had two different bases - one for NotificationType and one for MFAMethodType
- Making this change will allow us to make changes to Redis prefixes for OTP/MFA attempts (which happens in the VerifyMfaCodeHandler amongst others)

## Related PRs
- Changes some of the underlying mechanics (but should not change the actual prefixes in Redis or business behaviour) from: https://github.com/alphagov/di-authentication-api/pull/3090 and https://github.com/alphagov/di-authentication-api/pull/3082
- Supersedes an abandoned PR here: https://github.com/alphagov/di-authentication-api/pull/3094

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/CodeRequestType.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/CodeRequestType.java
@@ -4,12 +4,11 @@ import java.util.HashMap;
 import java.util.Map;
 
 public enum CodeRequestType {
-    EMAIL_REGISTRATION(NotificationType.VERIFY_EMAIL, JourneyType.REGISTRATION),
-    EMAIL_ACCOUNT_RECOVERY(
-            NotificationType.VERIFY_CHANGE_HOW_GET_SECURITY_CODES, JourneyType.ACCOUNT_RECOVERY),
-    SMS_ACCOUNT_RECOVERY(NotificationType.VERIFY_PHONE_NUMBER, JourneyType.ACCOUNT_RECOVERY),
-    SMS_REGISTRATION(NotificationType.VERIFY_PHONE_NUMBER, JourneyType.REGISTRATION),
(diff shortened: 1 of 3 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/CodeRequestType.java
```diff
@@ -64,12 +67,12 @@ public boolean equals(Object obj) {
...
             }
             CodeRequestTypeKey other = (CodeRequestTypeKey) obj;
-            return notificationType == other.notificationType && journeyType == other.journeyType;
+            return mfaMethodType == other.mfaMethodType && journeyType == other.journeyType;
         }
 
         @Override
         public int hashCode() {
-            return 31 * notificationType.hashCode() + journeyType.hashCode();
+            return 31 * mfaMethodType.hashCode() + journeyType.hashCode();
```
JHjava, line 75: Because we're changing the key in the CodeRequestType, will that make any requests made under the old key for the CodeRequestType redundant?

JHjava (approved):

shared/src/main/java/uk/gov/di/authentication/shared/entity/CodeRequestType.java
JHjava: Dom and Ayo have walked me through this and believe there will be no impact.

PR #3118, opened 2023-06-27
AUT-1371: IPVCallbackHandler redirects to a separate error page when no session

dbes-gds:
## What?

IPVCallbackHandler redirects to a separate error page when there is no session (the session has timed-out).

To be depolyed after the corresponding frontend PR which creates the new error page.

## Why?

The new page has been added as part of the work to reduce session timeout from 2hrs to 1hr.  Due to the decrease they may be people who land back in authentication having completed IPV after an hour when the session has expired.  The new screen is to provide more information in this scenario.

## Related PRs

https://github.com/alphagov/di-authentication-frontend/pull/1081

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IPVCallbackNoSessionException.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IPVCallbackNoSessionException.java
@@ -1,0 +1,9 @@
+package uk.gov.di.authentication.ipv.entity;
+
+import uk.gov.di.authentication.shared.exceptions.NoSessionException;
+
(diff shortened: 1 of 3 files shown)


JHjava:
Do we also need to change the error page for when the `NoSessionException` is caught?

dbes-gds:
> Do we also need to change the error page for when the `NoSessionException` is caught?

Have updated the 'catch' block to catch 'NoSessionException'.  'IPVCallbackNoSessionException' is a subclass of 'NoSessionException' so this will handle both exceptions, the compiler wouldn't let me be explicit and put both exceptions in the catch block.

JHjava (approved):

PR #3146, opened 2023-07-10
AUT-1403: Upshift code block max retries so that user is blocked on the 6th time of requesting too many Email OTP codes 

ayoshebby:
## What?

- In the account recovery journey, a user is blocked on the 5th attempt of requesting an email OTP, they should be blocked on the 6th attempt instead.
- The default code_max_retries is set to 5
- Initiate code block when user's codeRequestCount is greater than code_max_retries

## Why?

Ensure users are given enough attempts to have a go at requesting Email OTP during account recovery journey


JHjava:
@ayoshebby  I think this PR can be closed now

PR #3170, opened 2023-07-19
AUT-1421-Part-1: Implement new code blocked prefix for Password Reset journey

ayoshebby:
## What?

- Implement new code attempts block prefix for Password Reset journey in the Verify Code handler
- This is a 2 part PR, the first part is to store the new code attempts blocked prefix for Password Reset journey along with the old code blocked prefix in order to keep up with the Redis cache
- The second PR will be to remove the old prefix once the Redis cache is up to date

## Why?

In the Frontend, when a user starts a new session after being blocked from a previous session and they attempt to reset their password, an EMAIL OTP code is being sent to the user even though they are blocked and the user is given another attempt to enter the code again.


JHjava (commented):
I've left one question. 

Please can you also add the ticket number and a small description of the change to the commit.

JHjava (approved):

PR #3186, opened 2023-07-26
AUT-1251 - Add the Orchestration client_id and Auth audience to config

JHjava:
## What?

- Add the Orchestration client_id and Auth audience to config
- Set these values for staging only. Currently the Orch/Auth split API is not enabled in any environment

## Why?

- The client_id populates the client_id and issuer in the secure auth request from Orchestration to Authentication. These values will be populated from config
- The audience populated the audience claim in the secure auth request from Orchestration to Authentication. This value will be populated from config

### DIFF
--- a/ci/terraform/oidc/authorize.tf
+++ b/ci/terraform/oidc/authorize.tf
@@ -26,6 +26,7 @@ module "authorize" {
   environment     = var.environment
 
   handler_environment_variables = {
+    AUTH_AUDIENCE                        = var.auth_audience
     DOMAIN_NAME                          = local.service_domain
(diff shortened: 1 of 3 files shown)


ayoshebby (approved):

isaac-GDS (commented):
Noticed that we have `clientId` in the claims and the auth request. They both come from difference sources. Could they use the same source or are they different ?

JHjava:
@isaac-GDS Good spot! I fixed that in this earlier PR - https://github.com/alphagov/di-authentication-api/pull/3208/files

PR #3196, opened 2023-08-02
Remove references to localstack

two-first-names:
## What?

Removes localstack, as it is no longer used.

## Why?

This will simplify the terraform and make security checks work properly.


ethanmills:
replaced by https://github.com/govuk-one-login/authentication-api/pull/3966

PR #3197, opened 2023-08-02
AUT-1492 - Pass additional claims to Authentication

ghost:
## What?

- Filter on scopes to send to Authentication as we only want to send standard OIDC scopes to Authentication so Orchestration will filter out any other scopes that Authentication does not care about.
- Add additional claims to secure auth request to Authentication

## Why?

- So Authentication knows what to return to Orchestration

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -14,8 +14,10 @@ import com.nimbusds.oauth2.sdk.id.ClientID;
 import com.nimbusds.oauth2.sdk.id.State;
 import com.nimbusds.openid.connect.sdk.AuthenticationErrorResponse;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+import com.nimbusds.openid.connect.sdk.OIDCClaimsRequest;
 import com.nimbusds.openid.connect.sdk.OIDCError;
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -369,10 +370,12 @@ private APIGatewayProxyResponseEvent redirect(
                                             authenticationRequest.getScope()))
                             .claim(
                                     "redirect_uri",
-                                    configurationService.getOrchestrationRedirectUri())
-                            .build();
+                                    configurationService.getOrchestrationRedirectUri());
 
-            var encryptedJWT = authorizationService.getSignedAndEncryptedJWT(claims);
+            var claimsSetRequest =
```
BeckaL, line 375: Are there tests that we can add for these changes?

BeckaL (commented):
Just having a look through to try and get some context, while here I wondered if there were some tests that could be added.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
JHjava: I don't think there are tests for the Orch/Auth request when the feature flag is enabled. 

I can look to fill the test gaps as a separate PR

ayoshebby (approved):

PR #3207, opened 2023-08-04
AUT-1491: Create Authentication AuthCode lambda 

ayoshebby:
## What?

- Create a new handler 'AuthorizationAuthCodeHandler'
- This is the lambda that will generate an OAuth authorization response to be passed back to Orchestration which will consist of:
- `code` the authorization code generated in the in the newly created lambda
- `state` present in the initial request `AuthCodeRequest`
- This lambda will use the newly created `DynamoAuthCodeService` which will save a new auth code amongst other fields in Dynamo
- The changes also includes the necessary terraform infrastructure to ensure the handler is connected to a corresponding lambda function

## Why?
- Part of auth/orch split. Authentication will need to generate an OAuth authorization response to Orchestration

## Related PRs
[Auth Code Store](https://github.com/alphagov/di-authentication-api/pull/3194)

### DIFF
--- a/ci/terraform/oidc/build-overrides.tfvars
+++ b/ci/terraform/oidc/build-overrides.tfvars
@@ -11,6 +11,7 @@ spot_enabled                        = false
 language_cy_enabled                 = true
 internal_sector_uri                 = "https://identity.build.account.gov.uk"
 extended_feature_flags_enabled      = true
+support_auth_orch_split             = false
 custom_doc_app_claim_enabled        = true
 ipv_no_session_response_enabled     = true
 doc_app_cri_data_v2_endpoint        = "credentials/issue"
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java
@@ -1,0 +1,68 @@
+package uk.gov.di.authentication.frontendapi.entity;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import uk.gov.di.authentication.shared.validation.Required;
+
+public class AuthCodeRequest {
+
+    @SerializedName("redirect-uri")
+    @Expose
+    @Required
+    private String redirectUri;
+
+    @SerializedName("state")
+    @Expose
+    @Required
+    private String state;
+
+    @SerializedName("requestedScopeClaims")
+    @Expose
+    @Required
+    private String requestedScopeClaims;
+
+    @SerializedName("email")
+    @Expose
+    @Required
+    private String email;
+
+    public AuthCodeRequest(
+            String redirectUri, String state, String requestedScopeClaims, String email) {
+        this.redirectUri = redirectUri;
+        this.state = state;
+        this.requestedScopeClaims = requestedScopeClaims;
+        this.email = email;
+    }
+
+    public String getRedirectUri() {
+        return redirectUri;
+    }
+
+    public void setRedirectUri(String redirectUri) {
+        this.redirectUri = redirectUri;
+    }
+
+    public String getState() {
+        return state;
+    }
+
+    public void setState(String state) {
+        this.state = state;
+    }
+
+    public String getRequestedScopeClaims() {
+        return requestedScopeClaims;
+    }
+
+    public void setRequestedScopeClaims(String requestedScopeClaims) {
+        this.requestedScopeClaims = requestedScopeClaims;
+    }
+
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+}
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
@@ -1,0 +1,94 @@
+package uk.gov.di.authentication.frontendapi.lambda;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.AuthorizationCode;
+import com.nimbusds.oauth2.sdk.AuthorizationSuccessResponse;
+import com.nimbusds.oauth2.sdk.id.State;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.frontendapi.entity.AuthCodeRequest;
+import uk.gov.di.authentication.shared.entity.ErrorResponse;
+import uk.gov.di.authentication.shared.lambda.BaseFrontendHandler;
+import uk.gov.di.authentication.shared.serialization.Json.JsonException;
+import uk.gov.di.authentication.shared.services.AuthenticationService;
+import uk.gov.di.authentication.shared.services.ClientService;
+import uk.gov.di.authentication.shared.services.ClientSessionService;
+import uk.gov.di.authentication.shared.services.ConfigurationService;
+import uk.gov.di.authentication.shared.services.DynamoAuthCodeService;
+import uk.gov.di.authentication.shared.services.SessionService;
+import uk.gov.di.authentication.shared.state.UserContext;
+
+import java.net.URI;
+
+import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
+import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
+import static uk.gov.di.authentication.shared.helpers.LogLineHelper.LogFieldName.AWS_REQUEST_ID;
+import static uk.gov.di.authentication.shared.helpers.LogLineHelper.attachLogFieldToLogs;
+
+public class AuthenticationAuthCodeHandler extends BaseFrontendHandler<AuthCodeRequest> {
+
+    private static final Logger LOG = LogManager.getLogger(AuthenticationAuthCodeHandler.class);
+
+    private final DynamoAuthCodeService dynamoAuthCodeService;
+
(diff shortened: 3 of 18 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
```diff
@@ -0,0 +1,134 @@
...
+                            authCodeRequest.getClientId(), userProfile.get().getSubjectID());
+
+            dynamoAuthCodeService.saveAuthCode(
+                    userProfile.get().getSubjectID(),
+                    authorisationCode,
+                    authCodeRequest.getRequestedScopeClaims(),
+                    false);
+
+            var authorizationResponse =
+                    new AuthCodeResponse(authorisationCode, authCodeRequest.getState());
```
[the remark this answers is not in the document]
ayoshebby: Think it's worth keeping the `state` assuming the frontend ui will do some validation comparing the state to ensure the response is coming from the expected source.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+                "auth-code::" + getClass().getSimpleName(),
+                () -> authCodeRequestHandler(input, context));
+    }
+
+    private APIGatewayProxyResponseEvent authCodeRequestHandler(
+            APIGatewayProxyRequestEvent input, Context context) throws Json.JsonException {
+        attachLogFieldToLogs(AWS_REQUEST_ID, context.getAwsRequestId());
+
+        if (configurationService.isAuthCodeHandlerEnabled()) {
+            var authCodeRequest = gson.fromJson(input.getBody(), AuthCodeRequest.class);
```
DTaylorGDS, line 75: This may be off-base, but we do have an existing Serialisation service class https://github.com/alphagov/di-authentication-api/blob/main/shared/src/main/java/uk/gov/di/authentication/shared/services/SerializationService.java#L23

It also works off Gson, but I wonder if it would give you some freebies in terms of the validation you're doing in this method? 

I _think_ as long as your `AuthCodeRequest` is correctly annoted with `@Required` and `@Expose` and that sort of thing, you might find it would throw you a Json Exception with the missing fields identified?

shared/src/main/java/uk/gov/di/authentication/shared/entity/ErrorResponse.java
```diff
@@ -59,7 +59,14 @@ public enum ErrorResponse {
             "System is blocked from sending any email verification codes for changing how to receive security codes"),
     ERROR_1048(
             1048,
-            "User entered invalid email verification code for changing how to receive security codes too many times");
+            "User entered invalid email verification code for changing how to receive security codes too many times"),
+    ERROR_1049(1049, "Redirect URI is empty"),
```
DTaylorGDS, line 63: This sort of goes with my other comment about using the serialisation service. If you did, you might lose the granularity of all of these error codes...but not sure that matters (I personally don't think so, but putting it out there as a comment for you and others to weigh in on)

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java
```diff
@@ -0,0 +1,68 @@
...
+    @Expose
+    @Required
+    private String redirectUri;
+
+    @SerializedName("state")
+    @Expose
+    @Required
+    private String state;
+
+    @SerializedName("requestedScopeClaims")
```
DTaylorGDS, line 19: Sorry to nitpick, but there was some discussion around the nature of what would be held in this field. For example, at first we weren't sure if we could hold a scope like `email` and have that unpacked into claims (being `email` and `email_verified`).Same for phone. After discussing with @JHjava , he said the authorize handler _would_ do this unpacking in future (I have made that assumption in the recent PR for the userinfo handler). Although there's no functional effect, I wonder if we might just call this (and any similar references) `claims` and not bring scopes into it?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+                LOG.info(
+                        "Error message: Email from session does not have a user profile required to extract Subject ID");
+                return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1054);
+            }
+
+            var authorisationCode = new AuthorizationCode();
+            dynamoAuthCodeService.saveAuthCode(
+                    userProfile.get().getSubjectID(),
+                    authorisationCode.getValue(),
+                    authCodeRequest.getRequestedScopeClaims(),
```
DTaylorGDS, line 95: It's not wrong to have this as a string imho, but @dbes-gds's earlier PR for the next step in this was storing claims as List<String>. Again, no functional problem, but might be neater to keep it uniform?: https://github.com/alphagov/di-authentication-api/pull/3217/files#diff-86eab8a351a5d00cf8271124a49c05d31378a08c3f7b2ad5f2411efd0dbb6316R15

DTaylorGDS: (just for context, because I know I mentioned the scope/claim naming before, the link I just gave you still says scopes, but that was overwritten here in a later commit: https://github.com/alphagov/di-authentication-api/commit/6381f9a2e66cf04c9e6e1968200d0d1d25e34510#diff-86eab8a351a5d00cf8271124a49c05d31378a08c3f7b2ad5f2411efd0dbb6316R15)

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
```diff
@@ -0,0 +1,128 @@
...
+public class AuthenticationAuthCodeHandler
+        implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+
+    private static final Logger LOG = LogManager.getLogger(AuthenticationAuthCodeHandler.class);
+
+    private final Gson gson;
+    private final DynamoAuthCodeService dynamoAuthCodeService;
+    private final ConfigurationService configurationService;
+    private final DynamoService dynamoService;
+    private final AuthorisationCodeService authorisationCodeService;
```
DTaylorGDS, line 39: Doesn't look like you're using this?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java
ayoshebby: As the change is out of scope from what was initially required in the ticket, I will create another ticket and convert the field to List of string (claims) instead.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
ayoshebby: Same as above, the change will be moved to another PR

shared/src/main/java/uk/gov/di/authentication/shared/entity/ErrorResponse.java
ayoshebby: Agreed, since using the serialisation service class which provides out of the box validation, it sought of makes this redundant

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeRequest.java
ayoshebby: I've created a subtask to rename that field: https://govukverify.atlassian.net/browse/AUT-1592

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -144,6 +144,10 @@ public boolean isBulkUserEmailEmailSendingEnabled() {
                 .equals("true");
     }
 
+    public boolean isAuthCodeHandlerEnabled() {
```
DTaylorGDS, line 147: Do we want to feature flag this independently of the wider auth orch split? We have this further up:

```
public boolean isAuthOrchSplitEnabled() {
        return System.getenv().getOrDefault("SUPPORT_AUTH_ORCH_SPLIT", "false").equals("true");
    }
```

To me it feels more of a faff than it's worth to toggle all the lambdas separately, but if you disagree and/or already discussed with James, fair enough

ci/terraform/oidc/build-overrides.tfvars
```diff
@@ -11,6 +11,7 @@ spot_enabled                        = false
 language_cy_enabled                 = true
 internal_sector_uri                 = "https://identity.build.account.gov.uk"
 extended_feature_flags_enabled      = true
+support_auth_orch_split             = true
```
DTaylorGDS, line 14: I know I suggested using the same flag, so apologies if this is confusing. Is it important to have this turned on in Build now? The reason I ask is because the feature flag changes the behaviour of the authorize endpoint in a way that would break the acceptance tests right now; so if you do need it toggled on in Build, I guess maybe it does need a different flag after all.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandler.java
```diff
@@ -0,0 +1,97 @@
...
+                    LOG.info(
+                            "Error message: Email from session does not have a user profile required to extract Subject ID");
+                    return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1049);
+                }
+
+                var authorisationCode = new AuthorizationCode();
+                dynamoAuthCodeService.saveAuthCode(
+                        userProfile.get().getSubjectID(),
+                        authorisationCode.getValue(),
+                        authCodeRequest.getRequestedScopeClaims(),
```
DTaylorGDS, line 81: I think you should be able to get this off the existing `request` being passed as an argument

DTaylorGDS (approved):

PR #3208, opened 2023-08-04
AUT-1492 -  Add integration tests and fix bugs for secure auth request to Authentication 

JHjava:
## What?

- Ensure scope is sent as a string in secure auth request
- Ensure claims are actually added to the secure auth request to authentication
- Remove duplicate Orchestration client_id
- Create integration test for the secure Auth request to Authentication
- Adding additional logging to assist with debugging
- Add additional checks to unit tests

## Why?

- There were a few bugs in our secure auth request
- Increase testing coverage by validating the contents of what we are sending to Authentication in the secure Auth request

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/OrchestrationToAuthenticationAuthorizeIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/OrchestrationToAuthenticationAuthorizeIntegrationTest.java
@@ -1,0 +1,387 @@
+package uk.gov.di.authentication.api;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.crypto.RSADecrypter;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.oauth2.sdk.AuthorizationRequest;
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.ResponseType;
+import com.nimbusds.oauth2.sdk.Scope;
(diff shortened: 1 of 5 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/OrchestrationToAuthenticationAuthorizeIntegrationTest.java
```diff
@@ -0,0 +1,385 @@
...
+        var signedJWTResponse = decryptJWT((EncryptedJWT) encryptedRequestObject);
+
+        validateStandardClaimsInJar(signedJWTResponse);
+        assertThat(
+                Objects.nonNull(signedJWTResponse.getJWTClaimsSet().getClaim("claim")),
+                equalTo(true));
+        var claimsRequest =
+                OIDCClaimsRequest.parse(
+                        (String) signedJWTResponse.getJWTClaimsSet().getClaim("claim"));
+        assertThat(
```
BeckaL, line 213: Obviously this is a bit of a new domain for me, but I'm not entirely clear from reading some of these tests what is different in each of the assertions, and what the important thing that's changing is. For example, in this case the test is about the GOVUK_ACCOUNT, but without context it's not clear to me which of these assertions is about that (is it the legacy_subject_id?) and therefore what's the important bit of this test.

I wonder if it could be made a bit clearer by a) abstracting away a bit of the noise in the tests, and b) if I'm correct about this being the important line, maybe extracting the legacy_subject_id as a variable (`govukAccountClaim`?)

BeckaL (commented):
Generally looks good to me, just a comment about maybe making the tests a bit clearer to read

integration-tests/src/test/java/uk/gov/di/authentication/api/OrchestrationToAuthenticationAuthorizeIntegrationTest.java
JHjava: @BeckaL that's fair. I've pushed up a change to make it clearer

BeckaL (approved):
LGTM!

PR #3211, opened 2023-08-07
Aut 1252/create orch callback handler

DTaylorGDS:
**Note to reviewer:** There are quite a few Sonar-flagged code smells that have been left 'as is'. This is something @billrdunn and I paired on, and we discussed all of them, but might be worth talking through them with whoever reviews this PR to ensure we are meeting code quality standards.

## What?

- Add auth callback handler
- This is the lambda that will be called by the authentication frontend (i.e. sign-in/account creation journey) at the end of its authentication flow
- Authentication will supply a code via this callback
- At this point, the lambda in this PR will need to exchange that code for an access token and use that token to obtain a userinfo response
- This will then be saved in Dynamo, and will be used by Orchestration to serve userinfo requests from Relying Parties
- Note: The userinfo response may vary in terms of which fields/content it has - authentication's userinfo lambda (not yet created) will dynamically 'decide' what to populate based on RPs requests scopes/claims

## What isn't included?

- Terraform/infrastructure
- Ability to handle IPV (identity-enabled) journeys
- Redirect to orchestration frontend error page rather than the existing (authentication) frontend error page
- Some degree of cleanup that will need to occur by the end of the auth/orch split, but for various reasons cannot happen yet:
    - Uplifting level of trust (this will be tracked in auth not orch)
    - Tracking MFA method (this will not be available to orch)
    -  Possibly an additional audit event, which corresponds to `ORCH_AUTHENTICATION_SUCCESS`; this was asked for in the Google docs that describes the API phase of the split, but there may be some further questions to answer around this (it may be that `AUTH_CALLBACK_RESPONSE_RECEIVED` is what this is and could be renamed)

## Why?
- Part of auth/orch split. Orchestration currently needs to access User Profile table to serve UserInfo endpoint reponses, but this data source will not be available after the split. This step puts the required information behind HTTP calls, rather than direct Dynamo access.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.api;
 
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.oauth2.sdk.AuthorizationCode;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ResponseType;
 import com.nimbusds.oauth2.sdk.Scope;
 import com.nimbusds.oauth2.sdk.id.ClientID;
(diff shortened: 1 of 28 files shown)


JHjava (commented):
I've left a couple of comments. I think in the future there is an opportunity for a refactoring to incorporate the logic for IPV/Doc App and this flow into a shared class

JHjava (approved):

PR #3219, opened 2023-08-09
Aut 1441/bulk email table

BeckaL:
## What?

Adds bulk email users table and interface for retrieving and modifying data within it.

Includes:
* Terraform to create a new table.
* Test extension to create and test the table locally.
* Service with data access methods.
* Integration test to test the service

## Why?

As part of sending a bulk email to update on terms and conditions, we need a new DynamoDB table to store all subject ids and their status (whether we've sent the email already or not).

### DIFF
--- a/ci/terraform/shared/dynamodb.tf
+++ b/ci/terraform/shared/dynamodb.tf
@@ -367,3 +367,47 @@ resource "aws_dynamodb_table" "auth_code_store" {
 
   tags = local.default_tags
 }
+
+resource "aws_dynamodb_table" "bulk_email_users" {
+  name         = "${var.environment}-bulk-email-users"
+  count        = contains(["build", "sandpit"], var.environment) ? 1 : 0
+  billing_mode = var.provision_dynamo ? "PROVISIONED" : "PAY_PER_REQUEST"
+
+  hash_key = "SubjectID"
+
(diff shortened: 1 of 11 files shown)


ci/terraform/shared/dynamodb.tf
```diff
@@ -366,4 +366,48 @@ resource "aws_dynamodb_table" "auth_code_store" {
...
+    type = "S"
+  }
+
+  point_in_time_recovery {
+    enabled = true
+  }
+
+  server_side_encryption {
+    enabled     = true
+    kms_key_arn = aws_kms_key.auth_code_store_signing_key.arn
```
JHjava, line 397: We will want to create a new key for this

ci/terraform/utils/bulk_user_email_dynamo_access.tf
```diff
@@ -0,0 +1,33 @@
+data "aws_dynamodb_table" "bulk_email_users" {
+  name = "${var.environment}-bulk-email-users"
```
JHjava, line 2: Do we need to add a count on this similar to in `dynamodb.tf`?

If we do, then we also need to update the terraform resources which reference this resource to use [0]

dbes-gds:
Added new kms key and counts for dynamo resources, tested in sandpit.

JHjava (approved):

PR #3225, opened 2023-08-14
AUT-1443: Add bulk email sending lambda

dbes-gds:
## What?

Add bulk email sending lambda

Lambda triggered by a cloudwatch scheduled event which reads the contents of the bulk-user-email table, retrieves the email corresponding to the subject id, sends an email to the user, then updates the status for the user email in the table.

The table is read and the emails are sent in batches with a configurable size.  After a batch has been processed sending can be paused (sleeped) for a configurable period in order to throttle email sending if required.

Includes unit and integration tests for the handler.

'getUserProfileFromSubject' in the DynamoService has been updated to throw an exception when a user is not found by subject id.

This PR only adds the code for the lambda handler and tests it.  Infrastructure to deploy the handler will follow.

## Why?

We want to be able to bulk email a specific template to a controlled audience of users.

## Related PRs

#3219

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
@@ -23,7 +23,6 @@ import uk.gov.di.authentication.shared.entity.TermsAndConditions;
 import uk.gov.di.authentication.shared.entity.User;
 import uk.gov.di.authentication.shared.entity.UserCredentials;
 import uk.gov.di.authentication.shared.entity.UserProfile;
-import uk.gov.di.authentication.shared.exceptions.UserNotFoundBySubjectIdRuntimeException;
 import uk.gov.di.authentication.shared.helpers.Argon2EncoderHelper;
 import uk.gov.di.authentication.shared.helpers.Argon2MatcherHelper;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
@@ -533,13 +532,23 @@ public class DynamoService implements AuthenticationService {
                 QueryEnhancedRequest.builder().consistentRead(false).queryConditional(q).build();
         Optional<UserProfile> userProfile =
                 subjectIDIndex.query(queryEnhancedRequest).stream()
-                        .findFirst()
-                        .flatMap(page -> page.items().stream().findFirst());
+                        .limit(1)
+                        .map(t -> t.items().get(0))
+                        .findFirst();
         if (userProfile.isEmpty()) {
-            throw new UserNotFoundBySubjectIdRuntimeException(
-                    "No user profile found with query search");
+            throw new RuntimeException("No userCredentials found with query search");
         }
         return userProfile.get();
+    }
+
+    public Optional<UserProfile> getOptionalUserProfileFromSubject(String subject) {
+        QueryConditional q =
+                QueryConditional.keyEqualTo(Key.builder().partitionValue(subject).build());
+        QueryEnhancedRequest queryEnhancedRequest =
+                QueryEnhancedRequest.builder().consistentRead(false).queryConditional(q).build();
+        return dynamoUserProfileTable.index("SubjectIDIndex").query(queryEnhancedRequest).stream()
+                .findFirst()
+                .flatMap(page -> page.items().stream().findFirst());
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
```diff
@@ -0,0 +1,142 @@
...
+
+    private final BulkEmailUsersService bulkEmailUsersService;
+
+    private final DynamoService dynamoService;
+
+    private final NotificationService notificationService;
+
+    private final ConfigurationService configurationService;
+
+    protected final Json objectMapper = SerializationService.getInstance();
```
JHjava, line 40: Is this used?

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
```diff
@@ -0,0 +1,142 @@
...
+            } catch (InterruptedException e) {
+                LOG.warn("Thread sleep for bulk user email batch pause interrupted.");
+            }
+        } while (!userSubjectIdBatch.isEmpty() && batchCounter < bulkUserEmailMaxBatchCount);
+
+        return null;
+    }
+
+    private void sendNotifyEmail(String email) throws NotificationClientException {
+        notificationService.sendEmail(
```
JHjava, line 128: Worth a log line in this method?

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
```diff
@@ -0,0 +1,142 @@
...
+                            "Bulk user email batch pausing for: {}ms",
+                            bulkUserEmailBatchPauseDuration);
+                    Thread.sleep(bulkUserEmailBatchPauseDuration);
+                    LOG.info("Bulk user email batch pause complete.");
+                }
+            } catch (InterruptedException e) {
+                LOG.warn("Thread sleep for bulk user email batch pause interrupted.");
+            }
+        } while (!userSubjectIdBatch.isEmpty() && batchCounter < bulkUserEmailMaxBatchCount);
+
```
JHjava, line 123: Worth a log line here to identify when a batch has completed?

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
```diff
@@ -0,0 +1,142 @@
...
+    @Override
+    public Void handleRequest(ScheduledEvent event, Context context) {
+
+        LOG.info("Bulk User Email Send has been triggered.");
+        final int bulkUserEmailBatchQueryLimit =
+                configurationService.getBulkUserEmailBatchQueryLimit();
+        final int bulkUserEmailMaxBatchCount = configurationService.getBulkUserEmailMaxBatchCount();
+        final long bulkUserEmailBatchPauseDuration =
+                configurationService.getBulkUserEmailBatchPauseDuration();
```
JHjava, line 80: Do you think it's worth logging out these values at the start of the handler?

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
```diff
@@ -532,11 +533,11 @@ public UserProfile getUserProfileFromSubject(String subject) {
...
                 subjectIDIndex.query(queryEnhancedRequest).stream()
-                        .limit(1)
-                        .map(t -> t.items().get(0))
-                        .findFirst();
+                        .findFirst()
+                        .flatMap(page -> page.items().stream().findFirst());
         if (userProfile.isEmpty()) {
-            throw new RuntimeException("No userCredentials found with query search");
+            throw new UserNotFoundBySubjectIdRuntimeException(
+                    "No user profile found with query search");
```
JHjava, line 540: Might be worth duplicating this method and making the exception a checked exception to enforce it gets handled and to avoid impacting anything else that already calls it

JHjava (commented):
Looks good overall. Few minor comments.

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
dbes-gds: separate method created

JHjava (approved):

PR #3246, opened 2023-08-18
AUT-1567: Bulk user email audience loader lambda

dbes-gds:
## What?

Bulk user email audience loader lambda.

A new lambda function that when triggered queries the user-profile table to retrieve the required audience for an email campaign, then loads those users in to the bulk-user-email table.

Includes both code and infrastructure, will deploy only to sandpit and build.

This is the first iteration of the loader, subsequent versions will refine the query to return a more specific audience based on a user's terms and conditions version, and potentially the ability to segment loads based on the account creation date so that loads can be split up into several runs rather than one.

This initial version will be used to do some performance testing to see how quickly the loads can take place, then subsequent versions will add improvements and batching if required.

## Why?

We need to create the audience for an email campaign before sending the emails.

## Related PRs

#3227 
#3225

### DIFF
--- a/ci/terraform/utils/variables.tf
+++ b/ci/terraform/utils/variables.tf
@@ -111,7 +111,17 @@ variable "bulk_user_email_send_schedule_enabled" {
 
 variable "bulk_user_email_send_schedule_expression" {
   type    = string
-  default = "cron(0 15 ? * MON-FRI *)"
+  default = "cron(0 15 ? * MON-FRI 2049)"
+}
+
+variable "bulk_user_email_audience_loader_schedule_enabled" {
+  type    = bool
+  default = false
+}
+
+variable "bulk_user_email_audience_loader_schedule_expression" {
+  type    = string
+  default = "cron(0 13 ? * MON-FRI 2049)"
 }
 
 variable "txma_account_id" {
--- a/ci/terraform/utils/variables.tf
+++ b/ci/terraform/utils/variables.tf
@@ -99,6 +99,11 @@ variable "bulk_user_email_batch_pause_duration" {
   default = null
 }
 
+variable "bulk_user_email_max_audience_load_user_count" {
+  type    = number
+  default = null
+}
+
 variable "bulk_user_email_email_sending_enabled" {
   type    = string
   default = "false"
--- a/ci/terraform/utils/variables.tf
+++ b/ci/terraform/utils/variables.tf
@@ -115,8 +115,9 @@ variable "bulk_user_email_send_schedule_enabled" {
 }
 
 variable "bulk_user_email_send_schedule_expression" {
-  type    = string
-  default = "cron(0 15 ? * MON-FRI 2049)"
+  type        = string
+  description = "Run at 15:00 every Friday in 2049.  Designed not to trigger, replace with desired expression."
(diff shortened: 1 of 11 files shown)


ci/terraform/utils/variables.tf
```diff
@@ -111,7 +116,17 @@ variable "bulk_user_email_send_schedule_enabled" {
...
+}
+
+variable "bulk_user_email_audience_loader_schedule_enabled" {
+  type    = bool
+  default = false
+}
+
+variable "bulk_user_email_audience_loader_schedule_expression" {
+  type    = string
+  default = "cron(0 13 ? * MON-FRI 2049)"
```
JHjava, line 129: Could be worth adding a description here and on like 118 to summarize the expression

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailAudienceLoaderScheduledEventHandler.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public Void handleRequest(ScheduledEvent event, Context context) {
+        LOG.info("Bulk User Email audience load triggered.");
+
+        final long bulkUserEmailMaxAudienceLoadUserCount =
+                configurationService.getBulkUserEmailMaxAudienceLoadUserCount();
+
+        itemCounter = 0;
+        dynamoService
+                .getBulkUserEmailAudienceStream()
```
JHjava, line 56: I think we only want to add users who didn't accept a certain version of terms and conditions

utils/src/test/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailAudienceLoaderScheduledEventHandlerTest.java
```diff
@@ -0,0 +1,120 @@
...
+import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.shared.entity.BulkEmailStatus;
+import uk.gov.di.authentication.shared.entity.UserProfile;
+import uk.gov.di.authentication.shared.services.BulkEmailUsersService;
+import uk.gov.di.authentication.shared.services.ConfigurationService;
+import uk.gov.di.authentication.shared.services.DynamoService;
+
+import java.util.List;
+
+import static org.mockito.Mockito.*;
```
JHjava, line 15: Wildcard import

JHjava (commented):
A couple of comments and there are also a couple of code smells

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailAudienceLoaderScheduledEventHandler.java
dbes-gds: agreed that's coming in the next version

ci/terraform/utils/variables.tf
dbes-gds: designed not to trigger and must be overridden, have added a comment, event is not enabled by default

JHjava (approved):

PR #3289, opened 2023-08-31
AUT-1622: Pass Google Analytics parameter to auth frontend via `/authorize` endpoint

DTaylorGDS:
## What?
- Pass a Google Analytics parameter to the authentication frontend via OIDC/orchestration `/authorize` endpoint

## Why?

- This is a Google Analytics value that needs to be picked up in the authentication frontend
- It originates in an email sent by the Face-to-Face service to let users know that their identity has been processed
- From there the IPV Return Relying Party passes this query param alongside its 'normal' authorize request
- In this commit, I have added the ability to pass through the value to the auth frontend
- A separate commit will be needed in the frontend repository to do anything with that value
- It should never change any system behaviour and is purely for analytics

## Related PRs
- Frontend PR to follow
- A change has also previously been made in IPV Return RP to add this parameter to their `/authorize` request


BeckaL (dismissed):
LGTM!

mrwilson (changes requested):
Just parking this here while I open a conversation.

JHjava (approved):

PR #3292, opened 2023-09-01
AUT-1468: Flip audience load query to include rather than exclude specific terms and conditions versions

dbes-gds:
## What?

Flip audience load query to include rather than exclude specific terms and conditions versions.
Update default versions for loading in each environment.

## Why?

Avoids the risk of a negative query loading unexpected data.  For example '<> 1.0' would load everything that is not version 1.0, including unexpected data, whereas '= 1.0' is guaranteed to only load what is expected.


utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
```diff
@@ -90,7 +90,7 @@ public Void handleRequest(ScheduledEvent event, Context context) {
         final long bulkUserEmailBatchPauseDuration =
                 configurationService.getBulkUserEmailBatchPauseDuration();
         final List<String> bulkUserEmailExcludedTermsAndConditions =
-                configurationService.getBulkUserEmailExcludedTermsAndConditions();
+                configurationService.getBulkUserEmailIncludedTermsAndConditions();
```
BeckaL, line 93: I think we need to change more than this here, since the logic in the handler is still excluding these (e.g. the logic in `sendEmailIfRequiredAndUpdateStatus` will need changing, and some tests will need updating too

BeckaL (commented):
Generally looks good, just think a bit of the logic in the sender handler that I added earlier today needs tweaking

utils/src/main/java/uk/gov/di/authentication/utils/lambda/BulkUserEmailSenderScheduledEventHandler.java
dbes-gds: thanks updated

BeckaL (approved):
Looks sensible to me, thanks!

PR #3294, opened 2023-09-01
AUT-1474: Deploy bulk email components to production

dbes-gds:
## What?

Add default prod config.
Flick switch to deploy to prod.
Upgrade the lambdas from java11 to java 17.

## Why?

Bulk email load and first batch send is coming soon.
Going to run a subset of the user load in production for proving purposes before the full load.

java17 seems to offer performance improvements vs java11


BeckaL (approved):
This all looks sensible to me 👍

PR #3300, opened 2023-09-04
AUT-1468: Inject SystemService to email sender lambda

dbes-gds:
## What?

Inject SystemService to email sender lambda.

## Why?

Fixes NPE when running the lambda as the SystemService is not currently set.


BeckaL (approved):
👍

PR #3304, opened 2023-09-04
AUT-1468: Add 'CreatedAt' timestamp when adding row to bulk-email-users table

dbes-gds:
## What?

Add 'CreatedAt' timestamp when adding row to bulk-email-users table.

## Why?

Adds ability to see when each row was loaded.
Makes it easier to differentiate if user loading takes place over several sessions.

## Related PRs

#3293


BeckaL (dismissed):
All looks sensible to me 👍

PR #3306, opened 2023-09-04
AUT-1468: Set BULK_USER_EMAIL_INCLUDED_TERMS_AND_CONDITIONS in email sender lambda

dbes-gds:
## What?

Set BULK_USER_EMAIL_INCLUDED_TERMS_AND_CONDITIONS in email sender lambda.

## Why?

Otherwise the config will always be empty...

This configuration is used to determine whether a user has already accepted a version of the terms and conditions that would exclude them from receiving an email.


BeckaL (approved):
LGTM

PR #3329, opened 2023-09-11
AUT-1618: Set access token store isUsed true when token used to obtai…

DTaylorGDS:
…n Auth->Orch UserInfo

- The 'isUsed' status is one of the validation checks in the UserInfo handler
- We need to set it to true once a successful token->userinfo exchange has been made so that userinfo cannot be issued a second time
- e.g. to a malicious user who has intercepted the access token in some way

## What?

Please include a summary of the change.

## Why?

## Related PRs


JHjava (commented):
Should we also check if it has been used at the time of the request and if it has been used then log something? Somewhere within here - https://github.com/alphagov/di-authentication-api/blob/main/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java#L96-L108

DTaylorGDS:
> https://github.com/alphagov/di-authentication-api/blob/main/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java#L96-L108

We check it here: https://github.com/alphagov/di-authentication-api/blob/main/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java#L125

Is that okay?

JHjava (approved):

PR #3334, opened 2023-09-13
AUT-1498: Create service to initiate IPV journey

isaac-GDS:
## What?

Part one of AUT-1498. Move out the logic from IPVAuthorisationHandler lambda into a component that can be called from the AuthenticationCallbackHandler to trigger the IPV journey from Orchestration.

Not all logic copied since the service is called in orchestration now.

## Question?

Regarding a code smell: Method has a lot of parameters, is there anyway of simplifying this ?

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationService.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationService.java
@@ -1,0 +1,149 @@
+package uk.gov.di.authentication.oidc.services;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.AuthorizationRequest;
+import com.nimbusds.oauth2.sdk.ResponseType;
+import com.nimbusds.oauth2.sdk.id.ClientID;
+import com.nimbusds.oauth2.sdk.id.State;
+import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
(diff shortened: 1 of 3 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationService.java
```diff
@@ -0,0 +1,149 @@
...
+        this.auditService = auditService;
+        this.authorisationService = authorisationService;
+        this.cloudwatchMetricsService = cloudwatchMetricsService;
+    }
+
+    public APIGatewayProxyResponseEvent sendRequestToIPV(
+            APIGatewayProxyRequestEvent input,
+            AuthenticationRequest authRequest,
+            UserProfile userProfile,
+            Session session,
```
BeckaL, line 64: Thoughts on how to reduce the number of parameters here would be to wrap some of them up into a class (or really just a record in java 17) containing all the user-y stuff from here, and then realised this [already exists](https://github.com/alphagov/di-authentication-api/blob/4df4c3c72d1c661dd481c287b3cb062a6d7a7ffe/shared/src/main/java/uk/gov/di/authentication/shared/state/UserContext.java#L12) and is used in the existing handler. Would that be appropriate to use here?

isaac-GDS: Did think about using this as it was the way to bring them in before. However, was reluctant to use it as the authCallback lambda has already extracted out the information we need. We would need to manipulate our current data into `UserContext` and then when in our service do some manipulation to get it back to our original data. Example of this is `pairwiseSubject`.

BeckaL: Fair enough :-)

PR #3337, opened 2023-09-15
AUT-1249: Add authentication token endpoint

DTaylorGDS:
## What?
- Add authentication token endpoint

## Why?
- This is the endpoint used by orchestrator callback lambda to obtain a token in exchange for a code it receives from auth frontend
- The orchestrator can then use that token to obtain userinfo, which is temporarily stored in an orch datastore
- From that datastore, it can serve userinfo requests to relying parties
- This token endpoint is therefore an important part of the flow which allows orch to handle userinfo responses without direct recourse to user information tables on the auth side

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/validation/PrivateKeyJwtClientAuthValidator.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/validation/PrivateKeyJwtClientAuthValidator.java
@@ -1,7 +1,6 @@
 package uk.gov.di.authentication.shared.validation;
 
 import com.nimbusds.jose.JOSEException;
-import com.nimbusds.jose.JWSHeader;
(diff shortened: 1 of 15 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/validation/PrivateKeyJwtClientAuthValidator.java
```diff
@@ -135,35 +125,4 @@ private boolean hasPrivateKeyJwtExpired(SignedJWT signedJWT) {
         }
         return false;
     }
-
-    private ClientCredentialsSelector<?> generateClientCredentialsSelector(String publicKey) {
```
BeckaL, line 139: nice, I like that we can get rid of this all!

PR #3348, opened 2023-09-20
AUT-1689: Add rp-sector-uri and is-new-account params in AuthCodeStore table

ayoshebby:
## What?
Add two new params to the AuthCodeStore table `SectorIdentifier` (rp-sector-uri) and `IsNewAccount` as these parameters  are required to be reflected in the AuthCodeStore table parsed from `/userinfo` endpoint
- `SectorIdentifier` (rp-sector-uri): value already being passed as a claim from the OIDC API under the field name rp_sector_host
- Modify AuthCodeRequest to parse the new params
- Save the new params into the table via DynamoAuthCodeService

## Why?

For the userinfo ticket `AUT-1250` there were a couple of fields were not reflected in anything passed by frontend to /orch-auth-code and therefore not stored in the auth code store


DTaylorGDS (approved):
Looks good to me

PR #3359, opened 2023-09-21
AUT-1472: Configure batches of emails to be sent between 11:00 - 12:00 BST on 23/09/2023

billrdunn:
**Only merge once Notify incident is confirmed as fully resolved**


BeckaL (approved):
LGTM

PR #3360, opened 2023-09-21
AUT-1479: schedule next batch of emails

BeckaL:
Another batch of runs scheduled for 1 - 2pm today (BST)


billrdunn (approved):

BeckaL:
Don't actually need this now, we were running the cron job for 2 hours anyway (turns out that in aws cron land, 10-11 has a duration of two hours)

PR #3361, opened 2023-09-21
AUT-1472: Send two extra batches of emails

billrdunn:
## What?

Modify cron expression to send two batches of emails - 15:30 and 15:36 on 21/09/2023

## Why?

Previous run of sending emails ended 5,000 emails (two batches) short of 50,000 target. This was due to not all jobs being run.

## Related PRs

[Previous scheduled run](https://github.com/alphagov/di-authentication-api/pull/3359)


BeckaL (approved):
We've looked at https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-cron-expressions.html and this should be a valid aws cron expression

PR #3362, opened 2023-09-21
AUT-1707: Refactor AuthCodeHandler response to contain redirect URI

ayoshebby:
## What?

- Prior to the change, the handler returned the `authorization` response object
- The response now returns a json body with "location" denoting the redirect uri with appropriate query params

## Why?

To behave similarly to the existing auth code handler, which returns the `authentication` response as a URI with the query params already constructed.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeResponse.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeResponse.java
@@ -1,28 +1,6 @@
 package uk.gov.di.authentication.frontendapi.entity;
 
-public class AuthCodeResponse {
+import com.google.gson.annotations.Expose;
+import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
 
-    private String code;
-    private String state;
-
-    public AuthCodeResponse(String code, String state) {
-        this.code = code;
-        this.state = state;
-    }
-
-    public String getCode() {
-        return code;
-    }
-
-    public void setCode(String code) {
-        this.code = code;
-    }
-
-    public String getState() {
-        return state;
-    }
-
-    public void setState(String state) {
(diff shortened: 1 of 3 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AuthenticationAuthCodeHandlerTest.java
```diff
@@ -160,17 +183,15 @@ void shouldReturn200AndSaveNewAuthCodeRequest() throws Json.JsonException {
                         anyString(),
                         eq(false));
         assertThat(result, hasStatus(200));
-        var authorizationResponse = new AuthCodeResponse(TEST_AUTHORIZATION_CODE, TEST_STATE);
-        assertThat(result, hasBody(objectMapper.writeValueAsString(authorizationResponse)));
+        assertTrue(new JSONObject(result.getBody()).has("location"));
```
BeckaL, line 186: Here, is it worth explicitly asserting on the full body of the response. So asserting that it is { "location": "some-redirect-uri?with=params" }? It's nice as a bit of documentation as to what's returned by the endpoint, and eliminates the possibility that e.g. location is null

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AuthCodeResponse.java
```diff
@@ -1,28 +1,6 @@
...
-
-    public String getState() {
-        return state;
-    }
-
-    public void setState(String state) {
-        this.state = state;
-    }
-}
+public record AuthCodeResponse(@Expose @Required String location) {}
```
BeckaL, line 6: THIS IS SO MUCH NICER!!! 🥳

BeckaL (approved):
LGTM

PR #3392, opened 2023-09-29
AUT-1460: Decouple Doc App callback from Authentication

ayoshebby:
## What?

- Save new auth code then generate authorisation response
- Redirect back to RP withough going through Authentication when decouple feature flag is enabled
- Add audit event for new auth code issued

## Why?

RP's no longer need to redirect to frontend and perform authentication i.e no token exchange required

## Related PRs

[Feature flag](https://github.com/alphagov/di-authentication-api/pull/3387)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/domain/DocAppAuditableEvent.java
```diff
@@ -9,7 +9,8 @@ public enum DocAppAuditableEvent implements AuditableEvent {
     DOC_APP_SUCCESSFUL_TOKEN_RESPONSE_RECEIVED,
     DOC_APP_UNSUCCESSFUL_TOKEN_RESPONSE_RECEIVED,
     DOC_APP_SUCCESSFUL_CREDENTIAL_RESPONSE_RECEIVED,
-    DOC_APP_UNSUCCESSFUL_CREDENTIAL_RESPONSE_RECEIVED;
+    DOC_APP_UNSUCCESSFUL_CREDENTIAL_RESPONSE_RECEIVED,
+    AUTH_CODE_ISSUED;
```
DTaylorGDS, line 13: Do you think we could reuse `OrchestrationAuditableEvent.AUTH_CODE_ISSUED`? I can see some arguments in both direction, but if Doc App journey is orchestration-only, no auth involvement, it doesn't seem false to me to use that category, and it comes with the advantage that it keeps a single event for what is one of our most 'high profile' audit logs (because people read it as basically == 'login') and less risk of accidental divergence when we move to a true canonical 'login' event, as I think is planned

ayoshebby: The only problem I had is `OrchestrationAuditableEvent` class exists in the `oidc-api` project whilst doc app is in the `doc-checking-app-api` project. I would need to add a dependency which I thought wasn't necessary for one small addition to use the enum

DTaylorGDS: Cool. That's fair enough 👍

DTaylorGDS (approved):

PR #3394, opened 2023-10-02
Aut 1458/decouple doc app handoff

billrdunn:
## What?

Please include a summary of the change.

## Why?

## Related PRs


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -201,22 +241,123 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
+
+        if (existingSession.isEmpty()) {
+            updateAttachedSessionIdToLogs(session.getSessionId());
+            LOG.info("Created session");
+        } else {
+            var oldSessionId = session.getSessionId();
+            sessionService.updateSessionId(session);
+            updateAttachedSessionIdToLogs(session.getSessionId());
+            LOG.info("Updated session id from {} - new", oldSessionId);
+        }
```
JHjava, line 282: This looks to be duplicated in here and the `handleAuthJourney` method. Wonder if it's worth putting this before either the `handleDocAppJourney` or the `handleAuthJourney` methods are invoked to avoid this duplication

ci/terraform/oidc/doc-app-authorize.tf
```diff
@@ -34,7 +34,12 @@ module "doc-app-authorize" {
     DOC_APP_JWKS_URL                   = var.doc_app_jwks_endpoint
     DOC_APP_ENCRYPTION_KEY_ID          = var.doc_app_encryption_key_id
     DOC_APP_TOKEN_SIGNING_KEY_ALIAS    = local.doc_app_auth_key_alias_name
+    DOC_APP_API_ENABLED                = var.doc_app_api_enabled
+    DOC_APP_DOMAIN                     = var.doc_app_domain
+    DOC_APP_DECOUPLE_ENABLED           = var.doc_app_decouple_enabled
     INTERNAl_SECTOR_URI                = var.internal_sector_uri
+    CUSTOM_DOC_APP_CLAIM_ENABLED       = var.custom_doc_app_claim_enabled
+    SUPPORT_AUTH_ORCH_SPLIT            = var.support_auth_orch_split
```
JHjava, line 42: We need to add the Doc App variables to the `authorize.tf` instead of this file as it's the `AuthorisationHandler.java` that we are adding the functionality too

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -201,22 +241,123 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
+                        configurationService.isCustomDocAppClaimEnabled(),
+                        configurationService.getDocAppDomain());
+        LOG.info("Doc app request received");
+
+        var clientSession =
+                clientSessionService.generateClientSession(
+                        authRequest.toParameters(),
+                        LocalDateTime.now(),
+                        orchestrationAuthorizationService.getEffectiveVectorOfTrust(authRequest),
+                        clientName);
```
JHjava, line 298: Similar to the above comment, we do this in both the `handleDocAppJourney` and `handleAuthJourney` methods so wonder if we could move this before these methods are invoked to avoid duplication

JHjava (changes requested):
I've left a few comments. Let me know if you want to discuss any of it. 

It could also do with an additional Integration test or 2 in `AuthorisationIntegrationTest`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -173,430 +198,625 @@ public void setUp() {
...
+                    new ProxyRequestContext()
+                            .withIdentity(new RequestIdentity().withSourceIp("123.123.123.123")));
+
+            APIGatewayProxyResponseEvent response = makeHandlerRequest(event);
+
+            assertThat(response, hasStatus(302));
+            assertEquals(
+                    "https://localhost:8080?error=invalid_request&error_description=Invalid+request%3A+Missing+response_type+parameter&state="
+                            + STATE.getValue(),
+                    response.getHeaders().get(ResponseHeaders.LOCATION));
```
BeckaL, line 557: Not sure where exactly this exists in this diff, but we should make sure the doc app requests contain only the following query params:

client id
request
response_mode (or response_type, can't remember exactly)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -180,7 +219,12 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
             LOG.warn("No query string parameters are present in the Authentication request", e);
             throw new RuntimeException(
                     "No query string parameters are present in the Authentication request", e);
+        } catch (ClientNotFoundException e) {
+            LOG.warn("Invalid client or client not found in Client Registry");
+            return generateApiGatewayProxyResponse(
+                    400, OAuth2Error.INVALID_CLIENT.toJSONObject().toJSONString());
```
JHjava, line 225: Because this Lambda has been called via the browser instead of an API call from the frontend, the error response would need to be a 302 and contain the location. 

However as the client is invalid then as part of OAuth we should not return the user back to the RP but instead throw an exception. https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1

JHjava (changes requested):
Spotted that the `authorize.tf` is missing this policy `aws_iam_policy.doc_app_auth_kms_policy.arn` and left one other comment. Other than that, I really like the refactoring's to reduce code duplication and the use of nested tests.

JHjava (approved):

PR #3435, opened 2023-10-17
ATO-21: Refactor request validation

billrdunn:
## What?

Refactoring

## Still to do
- Create more specific exceptions (similar to `ClientRegistryValidationException`) instead of throwing generic `RuntimeException`


ethanmills:
Do we need this change to be behind a feature flag?

PR #3443, opened 2023-10-18
Ato 81/modify authorize terraform to allow post requests

gn-GDS:
## What?

Allow post requests to authorize endpoint

## Why?

Standard authorise request uses a GET request and has a string limit. 
We currently don’t know what our current string limit is or if it’s a problem. 
This will remove risk of hitting query string limits with the full encrypted and signed HMRC payload

## Related PRs

[Parent Ticket](https://govukverify.atlassian.net/jira/software/c/projects/ATO/boards/390?selectedIssue=ATO-22)

### DIFF
--- a/ci/terraform/account-management/authenticate.tf
+++ b/ci/terraform/account-management/authenticate.tf
@@ -17,7 +17,7 @@ module "authenticate" {
 
   endpoint_name   = "authenticate"
   path_part       = "authenticate"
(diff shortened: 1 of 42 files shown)


ethanmills:
Moved this to draft until tf upgrade is done

ethanmills (approved):
Let's disable pipelines when merging this

PR #3449, opened 2023-10-19
AUT-1767: Add Auth_Code_Issued Audit Event to AuthCallBack Lambda

LazarAlexandru-Constantin:
## What?

Add Auth_Code_Issued Audit Event to AuthCallBack Lambda

## Why?

There is an audit event which is triggered when we generate an auth code for an RP - [https://github.com/alphagov/di-authentication-api/blob/main/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java#L2[…]01](https://github.com/alphagov/di-authentication-api/blob/main/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java#L287-L301)

We will also need this audit event on the AuthenticationCallbackHandler as the place where we will generate Auth Codes for RPs would have moved once we enable the split, so not to replicate it would mean the new journey is not fully equivalent in audit terms.

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -18,6 +18,7 @@ import com.nimbusds.openid.connect.sdk.claims.UserInfo;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.ipv.services.IPVAuthorisationService;
+import uk.gov.di.authentication.oidc.domain.OidcAuditableEvent;
 import uk.gov.di.authentication.oidc.domain.OrchestrationAuditableEvent;
 import uk.gov.di.authentication.oidc.exceptions.AuthenticationCallbackException;
 import uk.gov.di.authentication.oidc.services.AuthenticationAuthorizationService;
@@ -31,6 +32,7 @@ import uk.gov.di.authentication.shared.entity.VectorOfTrust;
 import uk.gov.di.authentication.shared.exceptions.UnsuccessfulCredentialResponseException;
 import uk.gov.di.authentication.shared.helpers.ConstructUriHelper;
 import uk.gov.di.authentication.shared.helpers.CookieHelper;
+import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.PersistentIdHelper;
 import uk.gov.di.authentication.shared.services.AuditService;
 import uk.gov.di.authentication.shared.services.AuthorisationCodeService;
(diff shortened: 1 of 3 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -340,6 +342,21 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        clientId,
+                        userInfo.getSubject().getValue(),
+                        Objects.isNull(userSession.getEmailAddress())
+                                ? AuditService.UNKNOWN
+                                : userSession.getEmailAddress(),
+                        IpAddressHelper.extractIpAddress(input),
+                        Objects.isNull(userInfo.getPhoneNumber())
+                                ? AuditService.UNKNOWN
+                                : userInfo.getPhoneNumber(),
+                        persistentSessionId);
```
DTaylorGDS, line 358: Could you just remind me how come we dropped the metadata pairs? from the event as it currently exists in the Auth Code handler?: https://github.com/alphagov/di-authentication-api/blob/cf8b052285b36ae1a6605b30444f31d8e6d3e29b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java#L299

I'm not sure what they are used for, if anything, but they should be available as part of `userInfo.getClaim(<name>)` based on what the userinfo response will return: https://github.com/alphagov/di-authentication-api/blob/cf8b052285b36ae1a6605b30444f31d8e6d3e29b/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java#L49

As you can see, `rp_pairwise_id` and `new_account` will always be part of that response. Internal subject ID will also be present, but this will be `userInfo.getSubject().getValue()` because it's set here: https://github.com/alphagov/di-authentication-api/blob/cf8b052285b36ae1a6605b30444f31d8e6d3e29b/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java#L41

Sorry, I know we discussed this before, and I can't remember if you explained why we don't have these pairs any more, but in general, I think we should be trying to preserve the existing 'shape' of the audit event

LazarAlexandru-Constantin:
Hi Dom, the reason why the metadata pairs were dropped was because all of the other audit events in the CallbackHandler were submitted without them, so this way we would've had all of the events submitted with the same parameters.
But, upon further inspection, it would be better to have them, like you said above. The OIDC events are submitted with the metadata pairs, so I will update this PR to contain them.

LazarAlexandru-Constantin:
Hello everybody, as you can see in the AuthenticationCallbackHandler file, in the audit event, the internalSubjectId metadata pair always has the value AuditService.UNKNOWN. 

This was done due to the fact that this information should not reach Orchestration at all, it should only be available in Auth. The internalSubjectId is used to generate the rpPairwiseId, and only the rpPairwiseId should be sent to orch (according to the new orch and auth responsibilities and data access restrictions). 

We have decided to always set the value of the internal id to UNKNOWN in order to not modify the structure of the Audit event, as it may break existing TXMA configurations, as they expect to receive the "internalSubjectId" metadata pair. 

We have not modified the AuthenticationCodeHandler as it will not be used at all after the auth-orch split will be done, and all of the requests will be done by the updated AuthenticationCallbackHandler. 

If it later required, after the split, the "internalSubjectId" may be removed from the audit event, if requested by TXMA, as it will always have the Audit.UNKOWN value.

dbes-gds (approved):

PR #3450, opened 2023-10-19
ATO-86: update terraform version

gn-GDS:
## What?

Update CICD Terraform to 1.1.9

## Why?

- We are running quite an old version of terraform.
- There are some features in the newer version of terraform that are required for other tickets.


ethanmills:
@gn-GDS could you write a couple of words either here or on the ticket to describe the sandbox testing that has been done please?

dbes-gds (approved):

PR #3453, opened 2023-10-19
AUT-1774: Add timestamp and extend di-persistent-session-id cookie

ayoshebby:
## What?

- Increase the max age of the `di-persistent-session-id` to 18 months rolling max-age
- For new cookies, append the timestamp to the end of the cookie value by using a double dash --
- For existing cookies, check whether it has a timestamp suffixed onto the cookie value, if not update the value by appending the timestamp

## Why?

So that the max-age for `di-persistent-session-id` is 18 months rather than 13

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
@@ -61,6 +61,7 @@ import uk.gov.di.authentication.shared.entity.CredentialTrustLevel;
 import uk.gov.di.authentication.shared.entity.ResponseHeaders;
 import uk.gov.di.authentication.shared.entity.Session;
 import uk.gov.di.authentication.shared.entity.VectorOfTrust;
+import uk.gov.di.authentication.shared.helpers.CookieHelper;
 import uk.gov.di.authentication.shared.helpers.DocAppSubjectIdHelper;
 import uk.gov.di.authentication.shared.services.AuditService;
 import uk.gov.di.authentication.shared.services.ClientService;
@@ -85,6 +86,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import static java.lang.String.format;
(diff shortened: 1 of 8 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -254,6 +259,17 @@ void shouldRedirectToLoginWhenUserHasNoExistingSession() {
                             pair("client-name", RP_CLIENT_NAME));
         }
 
+        private String generateExpectedPersistentCookieString() {
+            // Use LocalDateTime.now() to generate the dynamic part of the cookie value
+            var now = LocalDateTime.now();
```
JHjava, line 264: We need this to be UTC so we could use the `NowHelper` class

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
```diff
@@ -160,4 +161,8 @@ public static String buildCookieString(
                 "%s=%s; Max-Age=%d; Domain=%s; %s",
                 cookieName, cookieValue, maxAge, domain, attributes);
     }
+
+    public static String appendTimestampToCookieValue(String cookieValue) {
+        return cookieValue + "--" + LocalDateTime.now();
```
JHjava, line 166: We need this date to be in UTC so we could use the `NowHelper.java`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -254,6 +259,17 @@ void shouldRedirectToLoginWhenUserHasNoExistingSession() {
                             pair("client-name", RP_CLIENT_NAME));
         }
 
+        private String generateExpectedPersistentCookieString() {
+            // Use LocalDateTime.now() to generate the dynamic part of the cookie value
```
dbes-gds, line 263: mind removing all the comments?

JHjava (approved):

BeckaL:
Change looks good to me, could you avoid merging main into a feature branch as per the [GDS way](https://gds-way.cloudapps.digital/standards/source-code/working-with-git.html#branching-merging-conventions) - you can undo this and rebase instead with a bit of git wrangling, let me know if you'd like any help

BeckaL (approved):
Approving on the basis that James has previously approved and the changes since were small 👍

PR #3455, opened 2023-10-20
ATO-86: Update terraform version in CI and minimum to 1.3.1

CarlyG55:
## What?

Update terraform version in CI and minimum version to 1.3.1

## Why?

The terraform version in our deployment pipeline is 1.3.1, but this does not match the terraform version in our CI. There are features in newer versions of terraform that we cannot use as the CI version is currently 1.0.4.

## Related PRs
https://github.com/alphagov/di-authentication-api/pull/3443


ethanmills:
I'm happy with these changes, I'm just waiting to get access to some other bits to understand which of the ci/tasks files are still used and if this is a safe change to make to them

ethanmills (approved):

PR #3459, opened 2023-10-20
AUT-1467: Bulk Email switch for delivery receipts lambda

dbes-gds:
## What?

Bulk Email switch for delivery receipts lambda

Adds a switch to turn bulk email on and off in the NotifyCallbackHandler.

This PR should be merged before attempting to deploy #3431 again.

## Why?

This is so the lambda does not try to describe the user-profile table when the bulk email deployment switch is off (deploy_bulk_email_users_count).  If the deployment switch is off the NotifyCallbackHandler does not have the iam rights to access the user-profile table, so the lambda fails to start and throws errors.

## Related PRs

#3431

### DIFF
--- a/delivery-receipts-api/src/test/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandlerTest.java
+++ b/delivery-receipts-api/src/test/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandlerTest.java
@@ -44,6 +44,7 @@ class NotifyCallbackHandlerTest {
     private static final String BEARER_TOKEN = "1244656456457657566345";
     private static final String ENVIRONMENT = "test";
     private NotifyCallbackHandler handler;
+
     private final Context context = mock(Context.class);
     private final ConfigurationService configurationService = mock(ConfigurationService.class);
     private final DynamoService dynamoService = mock(DynamoService.class);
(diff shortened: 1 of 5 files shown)


delivery-receipts-api/src/test/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandlerTest.java
```diff
@@ -157,27 +158,34 @@ void shouldUpdateBulkEmailDeliveryReceiptsStatusForTermsAndConditionsEmailType()
...
-        handler.handleRequest(event, context);
+        handlerBulkEmailOn.handleRequest(event, context);
 
         verify(dynamoService).getUserProfileByEmailMaybe(email);
 
         verify(bulkEmailUsersService).updateDeliveryReceiptStatus(subjectId, "delivered");
     }
 
     @Test
-    void shouldNotUpdateBulkEmailDeliveryReceiptsStatusForTermsAndConditionsEmailType()
```
BeckaL, line 175: I think we should keep a test like this - I think the name of this test is actually wrong (it should be "should not update Bulk Email Delivery Receipts Status for **Email Updated** type") but I think this is a valid test to keep in for when the switch is *on*, and we should just add your switch off test in on top of this

BeckaL (approved):
LGTM

BeckaL (approved):

PR #3463, opened 2023-10-23
ATO-109: Remove no longer needed moved blocks from endpoints

CarlyG55:
## What?

Removing the moved blocks that were implemented as part of https://github.com/alphagov/di-authentication-api/pull/3443

## Why?

After deployment to all environments, moved blocks can be removed as they are no longer necessary
## Related PRs
https://github.com/alphagov/di-authentication-api/pull/3443


ethanmills (approved):
It would be good to disable the pipelines again and just check that this is a noop in build if that's not too much of a pain

PR #3467, opened 2023-10-24
Updated govuk one org Auth team as codeowner

pskushwaha1:
## What?
Updated govuk one org Auth team as codeowner

## Why?
We are moving repo to new org


fpmrqs (dismissed):

pskushwaha1:
not Required

PR #3484, opened 2023-10-27
AUT-1826: added CMK encryption

alhcomer:
## What?

Added CMK encryption to the following resources in dynamodb.tf:

- account_modifiers_table 
- commons_passwords_table 
- doc_app_credential_table 
- identity_credentials_table 
- client_registry_table
- user_profile_table
- user_credentials_table

Encryption keys were declared for each resource in kms.tf

## Why?

Lack of CMK encryption primarily impacts the confidentiality and integrity of affected Dynamo
databases. The use of CMK keys allows for more control over the cluster data encryption/decryption process.


ci/terraform/shared/dynamodb.tf
```diff
@@ -196,7 +199,8 @@ resource "aws_dynamodb_table" "identity_credentials_table" {
   }
 
   server_side_encryption {
-    enabled = !var.use_localstack
+    enabled     = !var.use_localstack
```
JHjava, line 202: I know this isn't part of your change but as we don't use localstack anymore then these can just be set to true

ci/terraform/shared/kms.tf
```diff
@@ -449,4 +449,172 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
...
+        ],
+        Resource = "*"
+      }
+    ]
+  })
+  tags = local.default_tags
+}
+
+resource "aws_kms_key" "identity_credentials_table_encryption_key" {
+  description              = "KMS encryption key for client registry table in DynamoDB"
```
JHjava, line 527: This description needs to be updated as it is referencing the wrong table

ci/terraform/shared/kms.tf
```diff
@@ -449,4 +449,172 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
...
+        ],
+        Resource = "*"
+      }
+    ]
+  })
+  tags = local.default_tags
+}
+
+resource "aws_kms_key" "doc_app_credential_table_encryption_key" {
+  description              = "KMS encryption key for client registry table in DynamoDB"
```
JHjava, line 551: This description needs to be updated as it is referencing the wrong table

ci/terraform/shared/kms.tf
```diff
@@ -449,4 +449,172 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
...
+        ],
+        Resource = "*"
+      }
+    ]
+  })
+  tags = local.default_tags
+}
+
+resource "aws_kms_key" "common_passwords_table_encryption_key" {
+  description              = "KMS encryption key for client registry table in DynamoDB"
```
JHjava, line 575: This description needs to be updated as it is referencing the wrong table

ci/terraform/shared/kms.tf
```diff
@@ -449,4 +449,172 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
...
+        ],
+        Resource = "*"
+      }
+    ]
+  })
+  tags = local.default_tags
+}
+
+resource "aws_kms_key" "account_modifiers_table_encryption_key" {
+  description              = "KMS encryption key for client registry table in DynamoDB"
```
JHjava, line 599: This description needs to be updated as it is referencing the wrong table

ethanmills:
Do we need to give something permission to use these keys? I think none of our code will be able to read from these tables with this change.

ci/terraform/shared/kms.tf
```diff
@@ -449,4 +449,172 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
 resource "aws_kms_alias" "authentication_callback_userinfo_encryption_key_alias" {
   name          = "alias/${var.environment}-authentication-callback-userinfo-encryption-key-alias"
   target_key_id = aws_kms_key.authentication_callback_userinfo_encryption_key.key_id
+}
```
dbes-gds, line 452: Looks like there's an issue with the lambdas not being able to decrypt.  Need to make some addtions.

PR #3490, opened 2023-10-27
ATO-107: Remove required annotation from claims in AuthCodeRequest

CarlyG55:
## What?

Removing 'required' annotation from claims in AuthCodeRequest

## Why?

These claims can be empty, if no scopes are requested. This causes the login flow to error after the user enters an auth code, since it validate the AuthCodeRequest object and expects claims to be required. This code is only currently used in sandpit, as it is behind the auth/orch split flag. Other instances claims do not have the required flag.


CarlyG55:
One of the integration tests is currently failing, but I think the test is incorrect - having no scopes is not an invalid state

CarlyG55:
I've now removed the integration test - shout if you think this is wrong!

mrwilson (approved):

PR #3491, opened 2023-10-27
AUT-1908: Update urls in the emails and SMS

ayoshebby:
## What?
Change the urls that take users to the contact form from the emails and SMSs they receive when they create a [GOV.UK](http://gov.uk/) One Login, sign in, or reset their password or 2FA method (account recovery).

- Remove "referer"
- Modify contact us route from "contact-us" to "contact-gov-uk-one-login"

## Why?

So that users go to the new ‘triage’ page rather than directly to the contact form.


BeckaL (approved):
All looks OK to me

PR #3507, opened 2023-11-01
ATO-121: Change JAR required logic

ethanmills:
## What?

- unrevert changes for ATO-21
- Change JAR required check to test scopes in client registry
 
## Why?

Previously this was checking the scopes inthe auth request. Change this to check the scopes in the client registry. This is because the acceptance tests do not send the doc check scope. It is unclear if the real requests contain this scope.

## Related PRs

#3475 
#3504


ethanmills:
I'll once again pause the pipeline and test this out on Build

mrwilson (approved):

CarlyG55 (approved):

PR #3516, opened 2023-11-02
BAU: Set production lambdas to 1.5GB mem

ethanmills:
Dynatrace integration requires this as a minimum

### DIFF
--- a/ci/terraform/account-management/production-overrides.tfvars
+++ b/ci/terraform/account-management/production-overrides.tfvars
@@ -13,7 +13,7 @@ cloudwatch_log_retention = 5
 
 performance_tuning = {
   authorizer = {
-    memory          = 1024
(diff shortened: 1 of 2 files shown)


two-first-names (approved):

ci/terraform/account-management/production-overrides.tfvars
CarlyG55, line 1: When I implemented this on LEP, I think it was suggested that you add an extra 1.5GB of memory for the lambda to run with - I might have been misinterpreting the docs/how it works though

ethanmills: We don't have more than 1.5GB on any of our lambdas and as far as I'm aware haven't seen any performance impacts

CarlyG55 (approved):

PR #3519, opened 2023-11-03
BAU: Skip Sonar checks on merge group

ethanmills:
This should speed up the merge queues, and we don't need this check at this stage

### DIFF
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -202,14 +202,17 @@ jobs:
       - run-integration-tests
     steps:
       - name: Check out repository code
+        if: github.event_name != 'merge_group'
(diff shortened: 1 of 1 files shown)


.github/workflows/pre-merge-checks.yml
```diff
@@ -202,14 +202,17 @@ jobs:
       - run-integration-tests
```
CarlyG55, line 202: Is there a reason we haven't got the if statement just for the job?

ethanmills: It's a required check, so need it to pass

CarlyG55: Ah sure, sounds good

CarlyG55 (approved):

PR #3530, opened 2023-11-06
ATO-146: Create identity progress handler

CarlyG55:
## What?

This PR creates the identity progress handler for the orchestration frontend to use to check if SPOT has returned identity credentials whilst the spinner page is showing. This PR create the `BaseOrchestrationHandler` that AuthCodeHandler will extend in the future once this lambda is under the remit of Orchestration. There is also some minor refactoring to move `AuthenticationUserInfo` and `AuthenticationUserInfoStorageService` to the shared folder. This move is implemented in the first commit for ease of reviewing.

## Why?

The existing `ProcessingIdentityHandler` requires the email to be posted, however the Orchestration frontend does not hold state, so does not have this.

## Related PRs
https://github.com/govuk-one-login/orchestration-frontend/pull/8


ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
```diff
@@ -0,0 +1,57 @@
...
+import java.net.URI;
+
+public class IdentityProgressResponse {
+
+    @SerializedName("status")
+    @Expose
+    @Required
+    private IdentityProgressStatus status;
+
+    @SerializedName("client-name")
```
CarlyG55, line 17: There doesn't seem to be a consistent casing for serialized name. I've chosen the same casing as AuthCodeHandler, as that's the other orchestration handler we'll have

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
```diff
@@ -0,0 +1,57 @@
...
+    @Expose
+    @Required
+    private String clientName;
+
+    @SerializedName("redirect-uri")
+    @Expose
+    @Required
+    private URI redirectUri;
```
CarlyG55, line 25: Similarly, there isn't consistency between if we should return uri/state as strings or their given types - I'm happy to return strings if people think that's the correct option

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
ethanmills, line 1: Is this a BaseOrchestrationHandler or a BaseOrchestationFrontendHandler?

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        this.configurationService = configurationService;
+        this.sessionService = new SessionService(configurationService);
+        this.clientSessionService = new ClientSessionService(configurationService);
+    }
+
+    @Override
+    public APIGatewayProxyResponseEvent handleRequest(
+            APIGatewayProxyRequestEvent input, Context context) {
+        return segmentedFunctionCall(
+                "ipv-api::" + getClass().getSimpleName(),
```
ethanmills, line 56: Is this always the right segment?

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        userSessionBuilder.withClientSession(clientSession.orElse(null));
+
+        var clientID =
+                clientSession
+                        .map(ClientSession::getAuthRequestParams)
+                        .map(t -> t.get(CLIENT_ID))
+                        .flatMap(v -> v.stream().findFirst());
+        attachLogFieldToLogs(LogLineHelper.LogFieldName.CLIENT_ID, clientID.orElse(UNKNOWN));
+        userSessionBuilder.withClientId(clientID.orElse(null));
+
```
ethanmills, line 95: Would be good to have the persistent session ID stuff here too, but not fail if it's not present

shared/src/main/java/uk/gov/di/authentication/shared/state/UserSession.java
ethanmills, line 1: Might be worth calling this OrchestrationUserSession for now to make it clear who's concern it is

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
```diff
@@ -0,0 +1,57 @@
...
+
+    @SerializedName("state")
+    @Expose
+    @Required
+    private State state;
+
+    public IdentityProgressResponse() {}
```
ethanmills, line 32: Why do we need the no-args constructor here?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
ethanmills, line 1: Could this be a record, or would that break the serialisation?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
ethanmills: Happy with this. An approval test is a good way to check the serialisation if that's something you're concerned about

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
CarlyG55: I wasn't really sure if `AuthCodeHandler` is considered frontend or not

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
CarlyG55: Yeah, this is one I was a little unsure of. I guess it's going to depend on where this is implemented, and will change when we move to our own repo. I'm not really sure what we should replace if with though

shared/src/main/java/uk/gov/di/authentication/shared/lambda/BaseOrchestrationHandler.java
CarlyG55: I think elsewhere with the persistent session we just add to logs - is that what you mean?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
CarlyG55: I think this is just me not having that much java experience, and having seen other examples in the code so thinking I needed it 😅

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
CarlyG55: I *think* it should be alright https://github.com/google/gson/releases/tag/gson-parent-2.10

ipv-api/src/main/java/uk/gov/di/authentication/ipv/entity/IdentityProgressResponse.java
CarlyG55: Added one to check the above as well

ethanmills (approved):

PR #3531, opened 2023-11-06
ATO-100: Remove isDocAppApiEnabled flag

billrdunn:
## What?

- Remove `DOC_APP_API_ENABLED` environment variable.
- Remove `ConfigurationService.isDocAppApiEnabled` method.

## Why?

- Value is always `true` so env variable and method are redundant.


ethanmills:
I don't actually know enough about our deployment pipeline to know if this is safe or not. Do you know if the env var gets removed at the same time as the code is updated, or might it happen before?

billrdunn:
Closing 

> I don't actually know enough about our deployment pipeline to know if this is safe or not. Do you know if the env var gets removed at the same time as the code is updated, or might it happen before?
> 
> If you're not sure it might be safest to split this change into two: one to remove from the java files, then a terraform followup.

Closing to split PR into two as per @ethanmills's suggestion above.

PR #3534, opened 2023-11-07
DO NOT MERGE: AUT-1826: added CMK encryption to doc app table

alhcomer:
## What?

Implemented CMK encryption/decryption capability to doc_app_credential table and all lambdas that require access to the table.

## Why?

The Fidus ITHC recommended adding CMK encryption to dynamodb tables. This is to improve confidentiality and integrity of dynamodb tables and improve control over encryption/decryption processes.


dbes-gds:
Need a zero downtime approach before this can be deployed.

PR #3535, opened 2023-11-07
ATO-178: Remove isDocAppApiEnabled from ConfigurationService

billrdunn:
## What?
Remove `isDocAppApiEnabled()` from `ConfigurationService`.

## Why?
Value is always `true` so method is redundant.

## Related PRs
https://github.com/govuk-one-login/authentication-api/pull/3536


ethanmills (dismissed):

integration-tests/src/test/java/uk/gov/di/authentication/api/JwksIntegrationTest.java
```diff
@@ -42,9 +30,7 @@ void shouldReturn200And2KeysWhenDocAppIsEnabled() throws ParseException {
 
     private static class JwksTestConfigurationService extends IntegrationTestConfigurationService {
```
ethanmills, line 31: probably don't need this class any more do we?

ethanmills (approved):

PR #3562, opened 2023-11-13
ATO-194: Pass split enabled flag to correct lambda

ethanmills:
## What?

Pass the the auth/orch split flag to the oidc userinfo lambda rather than the auth external one

## Why?

This is the lambda which uses it


ethanmills:
> Is the flag correct for all other lambdas?

We set it in more more lambda where it's not needed, but it's passed everywhere it's used

PR #3563, opened 2023-11-13
ATO-193: Set the Orch redirect url in all environments

ethanmills:
## What?

Pass through the correct redirect url for the orch->auth OAuth flow. The currently deployed code just has `orchestration-redirect` as a default, so the redirect goes to the `signin.` domain, which is not correct.

## Related PRs
Draft until https://github.com/govuk-one-login/authentication-frontend/pull/1217 goes in

### DIFF
--- a/ci/terraform/oidc/build.tfvars
+++ b/ci/terraform/oidc/build.tfvars
@@ -29,4 +29,5 @@ AZUx4RCDu+VWAZpPi1NaF5XWvkFNFwH+MyLkATh90UEJDe+ayKW6AXFcRQ==
 -----END PUBLIC KEY-----
 EOT
 
-orch_client_id = "orchestrationAuth"
+orch_client_id    = "orchestrationAuth"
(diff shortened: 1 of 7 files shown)


ci/terraform/oidc/build.tfvars
```diff
@@ -29,4 +29,5 @@ AZUx4RCDu+VWAZpPi1NaF5XWvkFNFwH+MyLkATh90UEJDe+ayKW6AXFcRQ==
 -----END PUBLIC KEY-----
 EOT
 
-orch_client_id = "orchestrationAuth"
\ No newline at end of file
+orch_client_id    = "orchestrationAuth"
```
billrdunn, line 32: There's a few changes which are just formatting. Separate into another commit?

ethanmills: This is a result of the line below being added and tf fmt being run. It aligns the equals signs, so these are related changed and need to  be in the same commit

hannah-leong (dismissed):

hannah-leong (approved):

PR #3566, opened 2023-11-14
ATO-181: Integrate API Gateway with Orchestration frontend Network Load Balancer

billrdunn:
## What?
Integrate existing API Gateway with Orch frontend Network Load Balancer (NLB)

## Why?
- API Gateway does not support integration of private Application Load Balancer (ALB)
- NLB and VPC Link are used instead, which sit in between API Gateway and ALB.

## Related PRs
[ATO-180](https://github.com/govuk-one-login/authentication-api/pull/3566) adds the ALB and NLB to the Orchestration frontend

[ATO-180]: https://govukverify.atlassian.net/browse/ATO-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ci/terraform/oidc/api-gateway.tf
+++ b/ci/terraform/oidc/api-gateway.tf
@@ -576,3 +576,47 @@ EOF
     aws_api_gateway_method.robots_txt_method,
   ]
 }
+
+resource "aws_api_gateway_resource" "orch_frontend_resource" {
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
+  parent_id   = aws_api_gateway_rest_api.di_authentication_api.root_resource_id
+  path_part   = "orch-frontend"
+}
+
+resource "aws_api_gateway_method" "orch_frontend_method" {
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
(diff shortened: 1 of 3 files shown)


ci/terraform/oidc/api-gateway.tf
```diff
@@ -576,3 +576,47 @@ EOF
     aws_api_gateway_method.robots_txt_method,
   ]
 }
+
+resource "aws_api_gateway_resource" "orch_frontend_resource" {
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
+  parent_id   = aws_api_gateway_rest_api.di_authentication_api.root_resource_id
+  path_part   = "orch-frontend"
```
ethanmills, line 583: We're probably going to need to think about our pathnames and routing on the frontend, unless this strips out this path segment?

billrdunn: The path segment does get stripped

ethanmills (dismissed):

ethanmills (approved):

PR #3574, opened 2023-11-16
BAU: Remove audit processors

ethanmills:
## What?
This is the first of four changes. It removes the audit processors stack, which consists of three lambdas and an S3 bucket, plus supporting resources.

- Tear down the audit processors stack (this change)
- Remove all terraform, Java code and github actions reference to the stack
- Remove SNS from Java code
- Destroy the events SNS topic

## Why?

These resources have been unused since we moved to TxMA. The storage bucket has since been emptied.


CarlyG55 (commented):
How does this work in terms of tearing down the infrastructure? If we just delete all the files, will the infrastructure just stay there? Do we run a terraform destroy?

ethanmills:
Good catch. I've reduced the scope of this PR to just tear down the infra - the deploy code and tf backend are retained here. They'll go next

dbes-gds (approved):

PR #3575, opened 2023-11-16
BAU: Remove Java dependency on SNS

ethanmills:
## What?

Change 2/3.

This removes the dependency on SNS, the localstack SNS resource, and the SNS test extensions.

## Why?

All application code which uses this has long since been removed. The test resources were being set up, but are not used by any test.

## Related PRs

#3574


ethanmills:
Not totally sure, but it also looks like this takes about a minute off the integration suite runtime

PR #3589, opened 2023-11-21
ATO-202: Accept list in vtr

CarlyG55:
## What?

Accept a list of values in the vtr parameter for request object

## Why?

At the moment, the vtr expects a string, and throws an error if you pass in a list (the expected format)

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
@@ -13,6 +13,7 @@ import org.apache.logging.log4j.Logger;
 
 import java.net.URI;
 import java.text.ParseException;
+import java.util.List;
 import java.util.Objects;
 
 public class RequestObjectToAuthRequestHelper {
@@ -41,8 +42,15 @@ public class RequestObjectToAuthRequestHelper {
                             .requestObject(authRequest.getRequestObject());
 
             if (Objects.nonNull(jwtClaimsSet.getClaim("vtr"))) {
-                builder.customParameter("vtr", (String) jwtClaimsSet.getClaim("vtr"));
-            }
+                if (jwtClaimsSet.getClaim("vtr") instanceof String vtr) {
+                    builder.customParameter("vtr", vtr);
+                } else if (jwtClaimsSet.getClaim("vtr") instanceof List<?> vtrList && vtrList.stream().allMatch(vtr -> vtr instanceof String)) {
+                    builder.customParameter("vtr", String.format("[\"%s\"]", String.join("\",\"", jwtClaimsSet.getStringArrayClaim("vtr"))));
+                } else {
+                    LOG.warn("Cannot parse Vectors of Trust");
+                    throw new RuntimeException("Cannot parse Vectors of Trust");
+                }
+             }
             if (Objects.nonNull(jwtClaimsSet.getClaim("ui_locales"))) {
                 try {
                     String uiLocales = (String) jwtClaimsSet.getClaim("ui_locales");
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
@@ -1,5 +1,6 @@
 package uk.gov.di.authentication.oidc.helpers;
 
+import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.jwt.SignedJWT;
 import com.nimbusds.langtag.LangTagUtils;
 import com.nimbusds.oauth2.sdk.ResponseType;
@@ -10,6 +11,8 @@ import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
 import com.nimbusds.openid.connect.sdk.Nonce;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.shared.serialization.Json;
+import uk.gov.di.authentication.shared.services.SerializationService;
 
 import java.net.URI;
 import java.text.ParseException;
@@ -17,6 +20,7 @@ import java.util.List;
 import java.util.Objects;
 
 public class RequestObjectToAuthRequestHelper {
+    private static final Json objectMapper = SerializationService.getInstance();
 
     private static final Logger LOG = LogManager.getLogger(RequestObjectToAuthRequestHelper.class);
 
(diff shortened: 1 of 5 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -41,7 +42,19 @@ public static AuthenticationRequest transform(AuthenticationRequest authRequest)
                             .requestObject(authRequest.getRequestObject());
 
             if (Objects.nonNull(jwtClaimsSet.getClaim("vtr"))) {
-                builder.customParameter("vtr", (String) jwtClaimsSet.getClaim("vtr"));
+                if (jwtClaimsSet.getClaim("vtr") instanceof String vtr) {
```
billrdunn, line 45: Could define `var vtrClaim = jwtClaimsSet.getClaim("vtr")` at the top of the method

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -214,10 +214,28 @@ private static boolean isSignatureValid(SignedJWT signedJWT, String publicKey) {
...
+                            redirectURI,
+                            new ErrorObject(
+                                    OAuth2Error.INVALID_REQUEST_CODE, "Request vtr not valid"));
+                }
+            } else {
+                authRequestVtr = emptyList();
+            }
             var vectorOfTrust = VectorOfTrust.parseFromAuthRequestAttribute(authRequestVtr);
             if (vectorOfTrust.containsLevelOfConfidence()
                     && !ipvCapacityService.isIPVCapacityAvailable()) {
```
billrdunn, line 241: Should there be a log message for this error response?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -214,10 +214,28 @@ private static boolean isSignatureValid(SignedJWT signedJWT, String publicKey) {
...
-                            : List.of(jwtClaimsSet.getClaim(VTR_PARAM).toString());
+            if (Objects.nonNull(jwtClaimsSet.getClaim("vtr"))) {
+                if (jwtClaimsSet.getClaim("vtr") instanceof String vtr) {
+                    authRequestVtr = List.of(vtr);
+                } else if (jwtClaimsSet.getClaim("vtr") instanceof List<?> vtrList
+                        && vtrList.stream().allMatch(vtr -> vtr instanceof String)) {
+                    authRequestVtr =
+                            List.of(
+                                    String.format(
+                                            "[\"%s\"]",
```
billrdunn, line 225: This formatting feels a bit odd - is it required to satisfy `VectorOfTrust.parseFromAuthRequestAttribute()`? 
If so would it be better to create a new parsing method in `VectorOfTrust`?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -41,7 +42,19 @@ public static AuthenticationRequest transform(AuthenticationRequest authRequest)
                             .requestObject(authRequest.getRequestObject());
 
             if (Objects.nonNull(jwtClaimsSet.getClaim("vtr"))) {
```
billrdunn, line 44: I think that this could be refactored to remove a layer of nesting:

```
if (vtrClaim instanceof String) { ... }
else if (vtrClaim instanceof List && all elements strings) { ... }
else if (vtrClaim nonNull) { return error}
else { // set to empty list }
```

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -231,6 +249,11 @@ private Optional<AuthRequestError> validateVtr(JWTClaimsSet jwtClaimsSet, URI re
             return errorResponse(
                     redirectURI,
                     new ErrorObject(OAuth2Error.INVALID_REQUEST_CODE, "Request vtr not valid"));
+        } catch (ParseException e) {
```
billrdunn, line 252: Could catch both in one block as they return the same thing, ie.
`catch (IllegalArgumentException | ParseException e)`

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -214,10 +214,28 @@ private static boolean isSignatureValid(SignedJWT signedJWT, String publicKey) {
...
+                    LOG.error("vtr in AuthRequest is not valid. Could not be parsed");
+                    return errorResponse(
+                            redirectURI,
+                            new ErrorObject(
+                                    OAuth2Error.INVALID_REQUEST_CODE, "Request vtr not valid"));
+                }
+            } else {
+                authRequestVtr = emptyList();
+            }
             var vectorOfTrust = VectorOfTrust.parseFromAuthRequestAttribute(authRequestVtr);
```
billrdunn, line 239: I think this is the only call that throws `IllegalArgumentException` so the `try` block could be moved down to here

billrdunn:
Can you squash the spotlessApply and integration test fix commits into other commits?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
CarlyG55: Yeah it's required to get it into the format to keep it consistent with the query parameter approach. I didn't realise how ugly it looked after spotless had its way with it - although I'd argue that the parsing method should maybe be in `RequestObjectToAuthRequestHelper`, as it's due to the formatting of the `RequestObject` that we have to do this

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
CarlyG55: I didn't go with this as there's more information in the `IllegalArgumentException` log that we don't have at the `ParseException` stage, and I thought it might make it unclear for people reading the logs

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
CarlyG55: I'm not sure - this is something we configure to turn off, so although we return an error response, it's more that it's because we don't want it on in build. Thoughts @ethanmills?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
CarlyG55: In that case we'd need two try blocks for the two different exception blocks - not sure if that's preferable or not 🤷

oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
CarlyG55: We don't actually do anything if this is null though? We just skip all these steps

CarlyG55: Oh okay actually I think I misunderstood a bit - we just don't need the else in your block above, because nothing happens in that case. I can see what you mean, but I'm still not entirely convinced - I think the logic reads a bit more clearly this way - in that if it's null, we don't need to anything, then if it's not null, we act based off of type, rather than combining the logic

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
ethanmills: Yeah, let's log. It makes debugging so much easier

ethanmills: Happy to log, but not in this change. It's unrelated

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
ethanmills: Can we use objectMapper.writeAsString or similar to just serialise this?

ethanmills (approved):

PR #3591, opened 2023-11-21
INCIDEN-542: Fix persistent session IDs

DTaylorGDS:
Note to reviewer: Please review this PR especially carefully. This has been deployed to Sandpit at the time of writing, so it should be possible to empirically compare behaviours between Sandpit and other environments. Please also treat the code logic with particular scepticism, as we are introducing changes on top of changes, and if we get this wrong, it will become very complex indeed to apply a fix that works for all possible users.

## What?
- Change the formatting behaviour around how we set `persistent-session-id`

## Why?
- Amends changes introduced in 75a10df4376397d5a454b87b5cee689e13a71e20
- For a time, users have ended up with:
1. Double timestamps in the session ID
2. Losing the existing UUID-style IDs from the beginning of the existing session ID

- The changes in this commit attempt to resolve that for a number of possible starting points:
1. No existing persistent-session-id
2. An 'old style' persistent-session-id which could look like `s1mZGVHhkM-bU-kTBrjVwXuFUW0`
3. A corrupted persistent-session-id, which could look like `--1700558480962--1700558480963--1700558480964` (seemingly up to a maximum of three timestamps) so something like `--1700558480962` is also possible

Scenario 3 applies only to users who already had a `persistent-session-id` and who used the site between the point of the amended PR taking effect and whenever this PR takes effect.

## Related PRs
- Amends: https://github.com/govuk-one-login/authentication-api/pull/3453

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
@@ -11,11 +11,11 @@ import java.util.Optional;
 import java.util.stream.Stream;
 
 import static java.lang.String.format;
+import static uk.gov.di.authentication.shared.helpers.PersistentIdHelper.isValidPersistentSessionCookieWithDoubleDashedTimestamp;
 
 public class CookieHelper {
 
     private static final Logger LOG = LogManager.getLogger(CookieHelper.class);
-
     public static final String REQUEST_COOKIE_HEADER = "Cookie";
     public static final String RESPONSE_COOKIE_HEADER = "Set-Cookie";
     public static final String PERSISTENT_COOKIE_NAME = "di-persistent-session-id";
@@ -123,12 +123,11 @@ public class CookieHelper {
         // we should just generate a new one to get back on track. It will be possible to remove
         // this 13 months after it is first merged(== the expiry time of that cookie at the point of
         // issue in November 2023) i.e. 26/11/2024
-        String VALID_PERSISTENT_SESSION_ID_FORMAT_REGEX = "[A-Za-z0-9-_]{27}--\\d{13}";
-        if (!persistentId.matches(VALID_PERSISTENT_SESSION_ID_FORMAT_REGEX)) {
-            return Optional.empty();
+        if (isValidPersistentSessionCookieWithDoubleDashedTimestamp(persistentId)) {
+            return Optional.of(persistentId);
         }
 
-        return Optional.of(persistentId);
+        return Optional.empty();
     }
 
     private static Optional<HttpCookie> parseStringToHttpCookie(String cookie) {
@@ -173,11 +172,5 @@ public class CookieHelper {
 
     public static String appendTimestampToCookieValue(String cookieValue) {
         return cookieValue + "--" + NowHelper.now().getTime();
-    }
-
-    public static boolean isValidPersistentSessionCookieWithDoubleDashedTimestamp(
-            String cookieValue) {
-        String persistentSessionIdPattern = "[A-Za-z0-9-_]{27}--\\d{13}";
-        return cookieValue.matches(persistentSessionIdPattern);
     }
 }
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
@@ -121,8 +121,8 @@ public class CookieHelper {
         // Temporary measure, as commit 75a10df4376397d5a454b87b5cee689e13a71e20 introduced a bug
         // whereby persistent session ID could end up as --<timestamp>--<timestamp>. In these cases
         // we should just generate a new one to get back on track. It will be possible to remove
-        // this 13 months after it is first merged(== the expiry time of that cookie at the point of
-        // issue in November 2023) i.e. 26/11/2024
+        // this 18 months after it is first merged(== the expiry time of that cookie at the point of
+        // issue in November 2023) i.e. 26/05/2025
         if (isValidPersistentSessionCookieWithDoubleDashedTimestamp(persistentId)) {
             return Optional.of(persistentId);
         }
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
@@ -11,6 +11,7 @@ import java.util.Optional;
 import java.util.stream.Stream;
 
 import static java.lang.String.format;
+import static uk.gov.di.authentication.shared.helpers.PersistentIdHelper.isOldPersistentSessionCookieWithoutTimestamp;
 import static uk.gov.di.authentication.shared.helpers.PersistentIdHelper.isValidPersistentSessionCookieWithDoubleDashedTimestamp;
 
 public class CookieHelper {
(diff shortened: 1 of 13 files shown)


ayoshebby (dismissed):

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
```diff
@@ -166,7 +175,7 @@ public static String appendTimestampToCookieValue(String cookieValue) {
     }
 
     public static boolean isValidCookieWithDoubleDashedTimestamp(String cookieValue) {
-        String sessionIdPattern = ".*--\\d+";
+        String sessionIdPattern = ".+--\\d+[^-]*";
```
ethanmills, line 178: Would it be correct to have `^.+--\\d+$` ? I'm struggling with the intent of the `[^-]*` above a little

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
```diff
@@ -118,6 +118,15 @@ public static Optional<String> parsePersistentCookie(Map<String, String> headers
...
 
+        // Temporary measure, as commit 75a10df4376397d5a454b87b5cee689e13a71e20 introduced a bug
+        // whereby persistent session ID could end up as --<timestamp>--<timestamp>. In these cases
+        // we should just generate a new one to get back on track. It will be possible to remove
+        // this 13 months after it is first merged(== the expiry time of that cookie at the point of
+        // issue in November 2023)
+        if (persistentId.startsWith("--")) {
+            return Optional.empty();
+        }
+
```
ethanmills, line 129: Since the ID generator just returns random b64url encoded bytes, I think it's probably possible for the ID part of this to start with `--`. Perhaps we ought to add a check to the ID generator to prevent that?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/PersistentIdHelper.java
```diff
@@ -27,18 +26,21 @@ public static String extractPersistentIdFromCookieHeader(Map<String, String> hea
...
-        } else {
-            return appendTimestampToCookieValue(IdGenerator.generate());
+        // due to change in persistent session ID format in November 2023, this could be in the old
+        // format <ID>
+        // or new format <ID>--<timestamp> e.g. U8gFw6gLpuWWHyOvj21CGphM60--1700505785611
+        var parsedOrGeneratedCookie =
+                CookieHelper.parsePersistentCookie(headers)
+                        .orElse(appendTimestampToCookieValue(IdGenerator.generate()));
+
+        if (parsedOrGeneratedCookie.contains("--")) {
```
ethanmills, line 36: This assumes a raw ID without timestamp can never contain `--`. Is that true?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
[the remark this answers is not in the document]
DTaylorGDS: Yes

ethanmills:
I think we should be a little tighter on the pattern matching here where we can be. Something like `[A-Za-z0-9-_]{28}--\d{14}`

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
DTaylorGDS: It's to check that we haven't generated a double timestamp like some of what we've seen 'in the wild' e.g. `--1700558480962--1700558480963` as a persistent-session-id. That's the intent. As to whether it's necessary, up to you?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
```diff
@@ -118,6 +118,15 @@ public static Optional<String> parsePersistentCookie(Map<String, String> headers
         }
         final String persistentId = cookieValues[0];
 
+        // Temporary measure, as commit 75a10df4376397d5a454b87b5cee689e13a71e20 introduced a bug
+        // whereby persistent session ID could end up as --<timestamp>--<timestamp>. In these cases
+        // we should just generate a new one to get back on track. It will be possible to remove
+        // this 13 months after it is first merged(== the expiry time of that cookie at the point of
```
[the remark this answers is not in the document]
DTaylorGDS: Yes, fair point, thanks

shared/src/main/java/uk/gov/di/authentication/shared/helpers/PersistentIdHelper.java
```diff
@@ -27,18 +26,21 @@ public static String extractPersistentIdFromCookieHeader(Map<String, String> hea
...
+            return parsedOrGeneratedCookie;
         }
+
+        String sanitisedCookie =
+                InputSanitiser.sanitiseBase64(parsedOrGeneratedCookie)
+                        .orElseThrow(() -> new RuntimeException(parsedOrGeneratedCookie));
```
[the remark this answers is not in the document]
DTaylorGDS: I've added a little more to it, whilst keeping it as a generic RuntimeException and basically still keeping it a bit minimalist. Is this okay for you or would you rather I e.g. throw a custom exception?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
DTaylorGDS: That's very true. Does the new proposed code handle this, where instead of checking `startsWith` we apply a fuller regex pattern match?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/PersistentIdHelper.java
DTaylorGDS: No indeed. That's a faulty assumption. Does this now work for you?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
DTaylorGDS: I've gone with the same regex as elsewhere where you'd recommended that - this broke some tests since we commonly used something like `a-persistent-session-id` as a test value, which doesn't match the format requirement, but no bad thing to make our tests use a 'proper' ID?

DTaylorGDS:
> I think we should be a little tighter on the pattern matching here where we can be. Something like `[A-Za-z0-9-_]{28}--\d{14}` (assuming I've got my character counts right and the strings are fixed length etc.)

Fair point, completely taken. Have made some changes to this effect at the specific parts of the code where you pointed this out

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
ethanmills: This regex would still match a double timestamp. The leading `.+` would match the first timestamp, and the `--\\d+` would match the second.

CarlyG55 (dismissed):

shared/src/main/java/uk/gov/di/authentication/shared/helpers/CookieHelper.java
DTaylorGDS: Ah yes 🤦

CarlyG55 (approved):

PR #3592, opened 2023-11-21
ATO-175: Add AccountInterventionService

kerrr:
## What?

Add AccountInterventionService and associated classes.

The AccountInterventionService is a proxy wrapper to the external Account Intervention Service. Our representation makes an HTTP call to the actual service and parses the result into our internal representation. At this stage we are only interested in a subset of the returned information, namely the account status.

If the Account Intervention Service is disabled our `AccountInterventionService` will return an AccountStatus with all false values, indicating there is no intervention.

If the Account Intervention Service is enabled, but we encounter an exception at any point when connecting to the service, we catch the exception and rethrow an "AccountInterventionException"

## Why?

Required for integration with Account Intervention Service

### DIFF
--- a/oidc-api/build.gradle
+++ b/oidc-api/build.gradle
@@ -9,6 +9,7 @@ version "unspecified"
 dependencies {
 
     implementation project(path: ':ipv-api')
+    implementation project(path: ':orchestration-shared')
     testImplementation project(path: ':ipv-api')
     compileOnly             configurations.kms,
             configurations.lambda,
--- a/oidc-api/build.gradle
+++ b/oidc-api/build.gradle
@@ -9,7 +9,6 @@ version "unspecified"
 dependencies {
 
     implementation project(path: ':ipv-api')
-    implementation project(path: ':orchestration-shared')
     testImplementation project(path: ':ipv-api')
     compileOnly             configurations.kms,
             configurations.lambda,
@@ -23,6 +22,7 @@ dependencies {
             configurations.bouncycastle,
             configurations.cloudwatch,
             project(":shared"),
+            project(":orchestration-shared"),
             project(":client-registry-api"),
             project(":doc-checking-app-api"),
             project(":auth-external-api")
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
@@ -1,0 +1,3 @@
+package uk.gov.di.authentication.oidc.entity;
+
+public record AccountInterventionResponse(AccountInterventionStatus state, String auditLevel) {}
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
@@ -1,3 +1,6 @@
 package uk.gov.di.authentication.oidc.entity;
 
-public record AccountInterventionResponse(AccountInterventionStatus state, String auditLevel) {}
+import com.google.gson.annotations.Expose;
+
+public record AccountInterventionResponse(
+        @Expose AccountInterventionStatus state, @Expose String auditLevel) {}
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
@@ -1,7 +1,7 @@
 package uk.gov.di.authentication.oidc.exceptions;
 
 public class AccountInterventionException extends RuntimeException {
-    public AccountInterventionException(Exception cause) {
-        super(cause);
+    public AccountInterventionException(String message, Exception cause) {
+        super(message, cause);
     }
 }
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
@@ -1,0 +1,7 @@
+package uk.gov.di.authentication.oidc.exceptions;
+
+public class AccountInterventionException extends RuntimeException {
+    public AccountInterventionException(Exception cause) {
+        super(cause);
+    }
+}
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
@@ -6,7 +6,6 @@ import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.oidc.entity.AccountInterventionResponse;
 import uk.gov.di.authentication.oidc.entity.AccountInterventionStatus;
 import uk.gov.di.authentication.oidc.exceptions.AccountInterventionException;
-import uk.gov.di.orchestration.shared.services.AuditService;
 import uk.gov.di.orchestration.shared.services.ConfigurationService;
 
 import java.io.IOException;
@@ -19,19 +18,13 @@ import java.net.http.HttpResponse;
 public class AccountInterventionService {
 
     private static final Logger LOGGER = LogManager.getLogger(AccountInterventionService.class);
-    private final boolean accountInterventionsAuditEnabled;
     private final boolean accountInterventionsEnabled;
     private final HttpClient httpClient;
     private final URI accountInterventionServiceURI;
-    private final AuditService auditService;
 
-    public AccountInterventionService(
-            ConfigurationService configService, AuditService auditService, HttpClient httpClient) {
-        this.accountInterventionsAuditEnabled =
-                configService.isAccountInterventionServiceAuditEnabled();
+    public AccountInterventionService(ConfigurationService configService, HttpClient httpClient) {
         this.accountInterventionsEnabled = configService.isAccountInterventionServiceEnabled();
         this.accountInterventionServiceURI = configService.getAccountInterventionServiceURI();
-        this.auditService = auditService;
         this.httpClient = httpClient;
     }
 
@@ -40,10 +33,6 @@ public class AccountInterventionService {
         try {
             if (accountInterventionsEnabled) {
                 return retrieveAccountStatus(internalSubjectId);
-            }
-
-            if (accountInterventionsAuditEnabled) {
-                sendAuditEvent();
             }
 
             return new AccountInterventionStatus(false, false, false, false);
@@ -72,6 +61,4 @@ public class AccountInterventionService {
 
         return response.state();
     }
-
-    private void sendAuditEvent() {}
 }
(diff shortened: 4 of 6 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
```diff
@@ -0,0 +1,7 @@
+package uk.gov.di.authentication.oidc.exceptions;
+
+public class AccountInterventionException extends RuntimeException {
+    public AccountInterventionException(Exception cause) {
```
billrdunn, line 4: Looking at other custom exceptions in this dir, the general pattern is 
`public MyException(String message) {
        super(message);
    }`
    and a message is passed when the exception is called. Is there a reason to break that pattern here?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
```diff
@@ -0,0 +1,73 @@
...
+
+        String body = httpResponse.body();
+
+        var response = new Gson().fromJson(body, AccountInterventionResponse.class);
+
+        return response.state();
+    }
+
+    private void sendAuditEvent(){
+
```
billrdunn, line 71: Do we know what will be in here?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
```diff
@@ -0,0 +1,5 @@
+package uk.gov.di.authentication.oidc.entity;
+
+import java.util.Map;
```
billrdunn, line 3: Remove?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+        when(config.getAccountInterventionServiceURI()).thenReturn(new URI(BASE_AIS_URL));
+        when(config.isAccountInterventionServiceEnabled()).thenReturn(true);
+        when(config.isAccountInterventionServiceAuditEnabled()).thenReturn(true);
+    }
+
+    @Test
+    void shouldConstructWellFormedRequestToAccountInterventionService() throws URISyntaxException, IOException, InterruptedException {
+
+        var internalSubjectId = "some-internal-subject-id";
+        var ais = new AccountInterventionService(config, audit, httpClient);
```
billrdunn, line 75: I've not seen a service variable named like this before, maybe stick to `accountInterventionService`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,127 @@
...
+import static org.mockito.Mockito.verifyNoInteractions;
+import static org.mockito.Mockito.when;
+
+class AccountInterventionServiceTest {
+
+    private final ConfigurationService config = mock(ConfigurationService.class);
+    private final AuditService audit = mock(AuditService.class);
+    private final HttpClient httpClient = mock(HttpClient.class);
+
+    private static String ACCOUNT_INTERVENTION_SERVICE_RESPONSE_SUSPEND_REPROVE = """
```
billrdunn, line 30: As there's likely to be more of these hard-coded strings would it make sense to store them elsewhere? Or have a way of generating them from `AccountInterventionResponse`?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/AccountInterventionServiceTest.java
kerrr: In general I'm happy doing stuff like this in limited scope, but I'll change it.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/AccountInterventionException.java
kerrr: I'll can add a message, though I think the context is all the cause. Difficult to really know if it's helpful or just noise outside the context of looking at bugs and how that is handled by the calling code.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
kerrr: Not yet, but I thought in the context of the headline ticket it was reasonable to leave a placeholder here.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
kerrr: 🙄

oidc-api/build.gradle
```diff
@@ -9,6 +9,7 @@ version "unspecified"
 dependencies {
 
     implementation project(path: ':ipv-api')
+    implementation project(path: ':orchestration-shared')
```
CarlyG55, line 12: I wonder if the ipv `implementation` is only here because it has an `implementation` and `testImplementation` - seems the more common pattern is to define it in the list below (and feels a bit tidier)

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/AccountInterventionServiceTest.java
kerrr: 🤔 Do we have a pattern for this elsewhere? Part of this was to check that the extra stuff we don't need is just ignored, so our `AccountInterventionResponse` is not capable of generating this, nor should it be. I'm relatively happy for this to be strings rather than creating bunch of code to manage it. Also it's not really needed else where. The option I'd thought about was to test that our config of the JSON parsing was doing what we wanted in a test of the `AccountInterventionResponse`, in which case we would move most of this out to there. But again I think this is reasonable.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
```diff
@@ -0,0 +1,77 @@
...
+    public AccountInterventionStatus getAccountStatus(String internalSubjectId)
+            throws AccountInterventionException {
+        try {
+            if (accountInterventionsEnabled) {
+                return retrieveAccountStatus(internalSubjectId);
+            }
+
+            if (accountInterventionsAuditEnabled) {
+                sendAuditEvent();
+            }
```
CarlyG55, line 47: Surely we won't return send the audit event here if we get a valid response, because we'll have already returned the status?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
```diff
@@ -0,0 +1,3 @@
+package uk.gov.di.authentication.oidc.entity;
+
+public record AccountInterventionResponse(AccountInterventionStatus state, String auditLevel) {}
```
CarlyG55, line 3: is `state` meant to be `status`?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
kerrr: Yes, of course. Maybe I really should have ignored the audit entirely for now.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AccountInterventionResponse.java
kerrr: The name in the json is `state`.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
```diff
@@ -0,0 +1,64 @@
...
+
+            return new AccountInterventionStatus(false, false, false, false);
+
+        } catch (IOException | URISyntaxException | InterruptedException e) {
+            throw new AccountInterventionException(
+                    "Unable to connect to Account Intervention Service", e);
+        }
+    }
+
+    private AccountInterventionStatus retrieveAccountStatus(String internalSubjectId)
```
ethanmills, line 46: This should be internalPairwiseSubjectId

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
```diff
@@ -0,0 +1,64 @@
...
+    private final HttpClient httpClient;
+    private final URI accountInterventionServiceURI;
+
+    public AccountInterventionService(ConfigurationService configService, HttpClient httpClient) {
+        this.accountInterventionsEnabled = configService.isAccountInterventionServiceEnabled();
+        this.accountInterventionServiceURI = configService.getAccountInterventionServiceURI();
+        this.httpClient = httpClient;
+    }
+
+    public AccountInterventionStatus getAccountStatus(String internalSubjectId)
```
ethanmills, line 31: This should be internalPairwiseSubjectId

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,145 @@
...
+                    "appliedAt": 1696869005821,
+                    "sentAt": 1696869003456,
+                    "description": "AIS_USER_PASSWORD_RESET_AND_IDENTITY_REVERIFIED",
+                    "reprovedIdentityAt": 1696969322935
+                },
+                "state": {
+                    "blocked": false,
+                    "suspended": true,
+                    "reproveIdentity": true,
+                    "resetPassword": false\s
```
ethanmills, line 44: Why do we need the `\s`?

kerrr: That's weird. I constructed them in another text editor and then pasted them into the text block. Looking at that I can see there was a trailing space which must have been converted during the paste.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
kerrr: OK

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccountInterventionService.java
```diff
@@ -0,0 +1,64 @@
...
+                        .uri(accountInterventionServiceURI.resolve(internalPairwiseSubjectId))
+                        .GET()
+                        .build();
+
+        HttpResponse<String> httpResponse =
+                httpClient.send(request, HttpResponse.BodyHandlers.ofString());
+
+        String body = httpResponse.body();
+
+        var response = new Gson().fromJson(body, AccountInterventionResponse.class);
```
ethanmills, line 60: Can you use the SerialisationService please? We may need to annotate some of the fields on the record, but it gives us protection against extra fields etc.

kerrr: Done.

ethanmills (approved):

PR #3593, opened 2023-11-21
AUT-1844:Implement Account Interventions Lambda to call the Account Interventions API

ayoshebby:
## What?
Add the logic to the Account Interventions Lambda so it can accept a request, call out to the Account Interventions API, handle the response and then return it to the frontend. 

- Make the Account Interventions Lambda accept a request which contains email
- Create Account Interventions Stub Extension which can be used by the Integration test
- Uses existing `ACCOUNT_INTERVENTION_SERVICE_URI` in config
- Calculate the user’s internal pairwise ID required as `userID` 
- Make a GET request to the Account Interventions API using this Path `/v1/ais/{userId}`
- Return response to frontend

## Why?

Implement Account Interventions backend tasks for scenarios where where account will need to be flagged, suspended or deleted.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/AccountInterventionsService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/AccountInterventionsService.java
@@ -1,0 +1,54 @@
+package uk.gov.di.authentication.frontendapi.services;
+
+import com.google.gson.JsonParseException;
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.http.HTTPRequest;
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.frontendapi.entity.AccountInterventionsInboundResponse;
+import uk.gov.di.authentication.shared.exceptions.UnsuccessfulAccountInterventionsResponseException;
+import uk.gov.di.authentication.shared.serialization.Json;
+import uk.gov.di.authentication.shared.services.SerializationService;
+
+import java.io.IOException;
+
+import static java.lang.String.format;
+
+public class AccountInterventionsService {
+
+    private static final Logger LOG = LogManager.getLogger(AccountInterventionsService.class);
+    private final Json objectMapper = SerializationService.getInstance();
+
+    public AccountInterventionsInboundResponse sendAccountInterventionsOutboundRequest(
+            HTTPRequest request) throws UnsuccessfulAccountInterventionsResponseException {
+
+        try {
+            LOG.info("Sending account interventions outbound request");
+            var response = request.send();
+            if (!response.indicatesSuccess()) {
+                throw new UnsuccessfulAccountInterventionsResponseException(
+                        format(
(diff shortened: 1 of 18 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandlerTest.java
```diff
@@ -2,21 +2,251 @@
...
+        event.setBody(format("{ \"email\": \"%s\" }", TEST_EMAIL_ADDRESS));
+        when(authenticationService.getUserProfileByEmailMaybe(anyString()))
+                .thenReturn(Optional.of(generateUserProfile()));
+        when(accountInterventionsService.sendAccountInterventionsOutboundRequest(any()))
+                .thenReturn(generateAccountInterventionResponse());
+        var result = handler.handleRequest(event, context);
+        assertThat(result, hasStatus(200));
+
+        var accountInterventionsResponse = new AccountInterventionsResponse(true, false, false);
+        assertThat(result, hasBody(objectMapper.writeValueAsString(accountInterventionsResponse)));
```
BeckaL, line 223: This looks to me locally like it's passing for the wrong reason - both the result body and the `objectMapper.writeValueAsString(accountInterventionsResponse)` are returning empty json objects. In general, I'd prefer testing against literal json strings to make sure we can't get tests incorrectly passing like this

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/AccountInterventionsService.java
```diff
@@ -0,0 +1,70 @@
...
+            throw new UnsuccessfulAccountInterventionsResponseException(
+                    "Error parsing HTTP response", e);
+        }
+    }
+
+    private AccountInterventionsInboundResponse parseResponse(HTTPResponse response)
+            throws ParseException {
+        var contentAsString = response.getContentAsJSONObject();
+
+        var interventionJson = (JSONObject) contentAsString.get("intervention");
```
BeckaL, line 50: Might be related to why the above test is passing, or might not, but this get can return a null (as can the `getAsString`s below, and we're not handling this anywhere (assuming maybe it should throw an error?). We should probably write a test case/ some test cases for the case where the interventions json does not look as expected

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/AccountInterventionsService.java
```diff
@@ -0,0 +1,129 @@
...
+            throws ParseException {
+        var contentAsString = response.getContentAsJSONObject();
+
+        var interventionJson = (JSONObject) contentAsString.get(FIELD_INTERVENTION);
+        if (interventionJson == null) {
+            LOG.debug(ERROR_MISSING_INTERVENTION);
+            throw new ParseException(ERROR_MISSING_INTERVENTION);
+        }
+        var updatedAt =
+                checkFieldNotNull(interventionJson.getAsString(FIELD_UPDATED_AT), FIELD_UPDATED_AT);
```
BeckaL, line 73: Wondering here whether we can use some automatic deserialisation of json into our record object, as that would get rid of a lot of this code - also asked on slack in case people who know their way around java libraries better than me can immediately spot the way

BeckaL (approved):
Thanks for reworking bits of this, this looks sensible to me

PR #3597, opened 2023-11-22
AUT-1988: fixed audit event logging incorrect subjectID

alhcomer:
## What?

Updated the submitAuditEvent method calls used in the UserInfoHandler and TokenHandler to use the internal pairwise ID as a parameter instead of incorrectly passed subjectID. Also updated corresponding unit tests and integration test so that they match the audit event accepting the pairwise id and not the subject id.

## Why?

The submitAuditEvent was previously incorrectly accepting the subject id for the internal pairwise ID argument.

### DIFF
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/TokenHandler.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/TokenHandler.java
@@ -9,18 +9,22 @@ import com.nimbusds.oauth2.sdk.ErrorObject;
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.core.SdkBytes;
 import uk.gov.di.authentication.external.domain.AuthExternalApiAuditableEvent;
 import uk.gov.di.authentication.external.exceptions.AuthCodeStoreRetreivalException;
 import uk.gov.di.authentication.external.services.TokenService;
 import uk.gov.di.authentication.external.validators.TokenRequestValidator;
 import uk.gov.di.authentication.shared.entity.AuthCodeStore;
+import uk.gov.di.authentication.shared.entity.UserProfile;
 import uk.gov.di.authentication.shared.exceptions.TokenAuthInvalidException;
+import uk.gov.di.authentication.shared.helpers.ClientSubjectHelper;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
 import uk.gov.di.authentication.shared.helpers.RequestBodyHelper;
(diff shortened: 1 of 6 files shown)


auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/TokenHandler.java
```diff
@@ -150,12 +158,22 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
...
+                                    configurationService.getInternalSectorUri(),
+                                    SdkBytes.fromByteBuffer(userProfile.getSalt()).asByteArray());
+
             auditService.submitAuditEvent(
                     AuthExternalApiAuditableEvent.TOKEN_SENT_TO_ORCHESTRATION,
                     AuditService.UNKNOWN,
                     AuditService.UNKNOWN,
                     Optional.ofNullable(requestBody.get("client_id")).orElse(AuditService.UNKNOWN),
-                    Optional.ofNullable(authCodeStore.getSubjectID()).orElse(AuditService.UNKNOWN),
+                    internalPairwiseId,
```
ayoshebby, line 176: Probably best to follow the same pattern to ensure "" is returned rather than "null". So something like `Optional.ofNullable(internalPairwiseId).orElse(AuditService.UNKNOWN)`

alhcomer: Thanks mate. Will do that now

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/TokenHandler.java
```diff
@@ -150,12 +158,22 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
...
 
+            String subjectID = authCodeStore.getSubjectID();
+            UserProfile userProfile = dynamoService.getUserProfileFromSubject(subjectID);
+            String internalPairwiseId =
+                    userProfile.getSalt() == null
+                            ? AuditService.UNKNOWN
+                            : ClientSubjectHelper.calculatePairwiseIdentifier(
+                                    subjectID,
+                                    configurationService.getInternalSectorUri(),
+                                    SdkBytes.fromByteBuffer(userProfile.getSalt()).asByteArray());
```
dbes-gds, line 169: if userProfile.getSalt() is null will result in NPE

dbes-gds: this is fine, it was just the way it looked in GH, fine when looking at the actual code

dbes-gds:
Just need to add dynamo read access to the user-profile table now that the lambda reads from it.

dbes-gds (approved):

PR #3604, opened 2023-11-24
ATO-203: Read 'claims' claim from JAR

billrdunn:
## What?

Include the `claims` claim when transforming a request object to an authentication request.

## Why?

When a request object is present in a request to `/authorize`, it is transformed in order to generate a properly formed authentication request. The `claims` claim wasn't being handled in the `transform` method and so was missing from the authentication request.


oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -48,6 +49,9 @@ public static AuthenticationRequest transform(AuthenticationRequest authRequest)
             if (Objects.nonNull(jwtClaimsSet.getClaim("vtr"))) {
                 transformVtr(builder, jwtClaimsSet);
             }
+            if (Objects.nonNull(jwtClaimsSet.getClaim("claims"))) {
+                builder.customParameter("claims", jwtClaimsSet.getClaim("claims").toString());
+            }
```
CarlyG55, line 54: I don't think we want to add it as a custom parameter - I think we want to add it as the "claims" parameter, similarly how we do with state and nonce above

ethanmills (approved):

PR #3605, opened 2023-11-24
ATO-209: Prevent open redirector

ethanmills:
## What?

We should not redirect to any url not configured in the client registry.

## Why?
See https://datatracker.ietf.org/doc/html/rfc6749#section-10.15


ethanmills:
As of this change, any parse error will throw an exception. There are a couple of changes I'd like to make to improve this going forward:

- Handle 5xx response codes at the API gateway layer to return a pretty error page. This could either be done with the new orch frontend, or directly with response mapping
- Check if the redirect URI and client ID match on parse errors, and redirect to the RP if so

hannah-leong (approved):

mrwilson (approved):

PR #3609, opened 2023-11-27
BAU: Make redis connection service a singleton

ethanmills:
Previously, constructing a single handler may result in several instances of the redis connection service, each with its own connection pool. Making this a singleton should have several advantages. Firstly, it should result in fewer connections being made at runtime, which will help with scalability and performance. Secondly, it fixes the issue with integration tests failing locally due to running out of file handles. Before this change running the integration test suite opened in excess of 300 redis connections. After this change 4 connections are opened.


hannah-leong (approved):

whi-tw:
@ethanmills is this still necessary / relevant? if so, can you please fix the conflicts and request re-review. if not, can you please close it?

ethanmills:
@whi-tw not sure what to do about this one. It fixes the integration tests locally, but I'm not fully sure of the impacts of this change in production

PR #3610, opened 2023-11-27
ATO-47: Get account status from account intervention

hannah-leong:
## What?

- Move AccountInterventionService to shared so that it can be accessed from ipv-api
- Add the AIS call to ProcessingIdentityHandler so that the call is made once the SPOT call is complete

## Why?

To get the account status so that we can redirect to the appropriate page if they are blocked/suspended


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -17,26 +17,20 @@
...
         try {
-            if (accountInterventionsEnabled) {
-                return retrieveAccountStatus(internalPairwiseSubjectId);
-            }
-
-            return new AccountInterventionStatus(false, false, false, false);
-
+            return retrieveAccountStatus(internalPairwiseSubjectId);
         } catch (IOException | URISyntaxException | InterruptedException | Json.JsonException e) {
+            Thread.currentThread().interrupt();
```
ethanmills, line 33: We should only be setting this when catching an interrupted exception, not any of the others

ethanmills (dismissed):

ethanmills (approved):

PR #3612, opened 2023-11-27
ATO-211: Generate the correct identifier

ethanmills:
## What?

Use just the host of the internal sector uri to generate the internal pairwise identifier used in the authentication user info response

### DIFF
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
@@ -32,13 +32,13 @@ public class UserInfoService {
         String internalSubjectId = accessTokenInfo.getSubjectID();
         var userProfile = authenticationService.getUserProfileFromSubject(internalSubjectId);
 
-        String internalRpPairwiseId =
-                ClientSubjectHelper.calculatePairwiseIdentifier(
-                        internalSubjectId,
+        Subject internalPairwiseId =
+                ClientSubjectHelper.getSubjectWithSectorIdentifier(
+                        userProfile,
                         configurationService.getInternalSectorUri(),
-                        SdkBytes.fromByteBuffer(userProfile.getSalt()).asByteArray());
+                        authenticationService);
 
-        var userInfo = new UserInfo(new Subject(internalRpPairwiseId));
+        var userInfo = new UserInfo(internalPairwiseId);
 
         var rpPairwiseId =
                 ClientSubjectHelper.calculatePairwiseIdentifier(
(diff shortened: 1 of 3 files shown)


auth-external-api/src/test/java/uk/gov/di/authentication/external/services/UserInfoServiceTest.java
```diff
@@ -144,7 +147,7 @@ private static Stream<Arguments> provideTestData() {
     private static AccessTokenStore getMockAccessTokenStore(List<String> claims) {
         var accessTokenStore = mock(AccessTokenStore.class);
         when(accessTokenStore.getSubjectID()).thenReturn(TEST_SUBJECT.getValue());
-        when(accessTokenStore.getSectorIdentifier()).thenReturn(TEST_RP_SECTOR_URI);
+        when(accessTokenStore.getSectorIdentifier()).thenReturn(TEST_RP_SECTOR_HOST);
```
ethanmills, line 150: This reflects the real behaviour. This value comes from the access token store, which comes from the auth code store, which comes from the auth frontend request. That is set to the host here:https://github.com/govuk-one-login/authentication-frontend/blob/main/src/components/auth-code/auth-code-service.ts#L44

I've also verified this by checking the access token store dynamo in sandpit

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
```diff
@@ -200,7 +200,7 @@ void shouldReturn401ForAccessTokenThatIsPastItsTtl() {
     private UserProfile addTokenToDynamoAndCreateAssociatedUser(
             String accessToken, List<String> claims, boolean isNewAccount) {
         accessTokenStoreExtension.addAccessTokenStore(
-                accessToken, TEST_SUBJECT.getValue(), claims, isNewAccount, RP_SECTOR_ID_URI);
+                accessToken, TEST_SUBJECT.getValue(), claims, isNewAccount, RP_SECTOR_ID_HOST);
```
ethanmills, line 203: See comment on other test file

billrdunn (dismissed):

auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
```diff
@@ -32,13 +32,13 @@ public UserInfo populateUserInfo(AccessTokenStore accessTokenInfo) {
         String internalSubjectId = accessTokenInfo.getSubjectID();
         var userProfile = authenticationService.getUserProfileFromSubject(internalSubjectId);
 
-        String internalRpPairwiseId =
-                ClientSubjectHelper.calculatePairwiseIdentifier(
-                        internalSubjectId,
+        Subject internalRpPairwiseId =
```
CarlyG55, line 35: While we're here, could you rename this to remove `Rp`? I think it makes it quite unclear

CarlyG55: (otherwise happy for this to go in!

dbes-gds (approved):

CarlyG55 (approved):

PR #3617, opened 2023-11-28
ATO-NONE: Move account intervention service to shared location

hannah-leong:
## What?

Move account intervention service to shared location

## Why?

So that it can be accessed from other places


kerrr (approved):
Looks good 🚀

PR #3647, opened 2023-12-04
ATO-229: Refactor LogoutHandler 

billrdunn:
## What?

This PR is purely refactoring. There are no logical changes.

Move logic for destroying sessions and creating responses from `LogoutHandler` to new `LogoutService`. 

## Why?

Allows greater control over the logout process, which is required to implement a back-channel logout if a suspended or blocked account status is received from Account Interventions.


ethanmills (commented):
To action an intervention, would you need to call `destroySessions` and then `generateLogoutResponse` to get it audited? I think it's worth a method that just does those two things for the interventions to call

billrdunn:
> To action an intervention, would you need to call `destroySessions` and then `generateLogoutResponse` to get it audited? I think it's worth a method that just does those two things for the interventions to call

@ethanmills  I was planning to add all Account Interventions logic in a subsequent PR

ethanmills (approved):

PR #3658, opened 2023-12-06
ATO-133: Persist return codes in DynamoDB // ATO-134: Bypass SPoT

hannah-leong:
## What?

ATO-133:
- Persist return codes in DynamoDB; as it is storing the valid claims, returnCode exists there and shall be saved in the case that a return code is present

ATO-134:
- If there is an error returned from IPV, make AIS call then return ACCESS DENIED
- If does not match requested level of confidence (`userIdentityError`) then make AIS call then check if `returnCode` is present
- If `returnCode` is present, process the return code:
  - If clientRegistry does not contain `returnCode` or `returnCode` was not requested in the `authRequest`, bypass SPoT and return ACCESS DENIED
  - Else save identityClaims, generate authCode and returned RP
    - For generating the authCode, use similar process to the `AuthCodeHandler` to include all necessary logging etc (extract out to be used elsewhere)

### DIFF
--- a/ci/terraform/oidc/ipv-callback.tf
+++ b/ci/terraform/oidc/ipv-callback.tf
@@ -32,23 +32,26 @@ module "ipv-callback" {
   environment     = var.environment
 
   handler_environment_variables = {
-    DYNAMO_ENDPOINT                 = var.use_localstack ? var.lambda_dynamo_endpoint : null
-    ENVIRONMENT                     = var.environment
-    IDENTITY_ENABLED                = var.ipv_api_enabled
-    INTERNAl_SECTOR_URI             = var.internal_sector_uri
-    IPV_AUDIENCE                    = var.ipv_audience
-    IPV_AUTHORISATION_CALLBACK_URI  = var.ipv_authorisation_callback_uri
-    IPV_AUTHORISATION_CLIENT_ID     = var.ipv_authorisation_client_id
-    IPV_AUTHORISATION_URI           = var.ipv_authorisation_uri
-    IPV_BACKEND_URI                 = var.ipv_backend_uri
-    IPV_NO_SESSION_RESPONSE_ENABLED = var.ipv_no_session_response_enabled
-    IPV_TOKEN_SIGNING_KEY_ALIAS     = local.ipv_token_auth_key_alias_name
-    LOCALSTACK_ENDPOINT             = var.use_localstack ? var.localstack_endpoint : null
-    LOGIN_URI                       = "https://${local.frontend_fqdn}/"
-    OIDC_API_BASE_URL               = local.api_base_url
-    REDIS_KEY                       = local.redis_key
-    SPOT_QUEUE_URL                  = aws_sqs_queue.spot_request_queue.id
-    TXMA_AUDIT_QUEUE_URL            = module.oidc_txma_audit.queue_url
+    DYNAMO_ENDPOINT                             = var.use_localstack ? var.lambda_dynamo_endpoint : null
+    ENVIRONMENT                                 = var.environment
+    IDENTITY_ENABLED                            = var.ipv_api_enabled
+    INTERNAl_SECTOR_URI                         = var.internal_sector_uri
+    ACCOUNT_INTERVENTION_SERVICE_ACTION_ENABLED = var.account_intervention_service_action_enabled
+    ACCOUNT_INTERVENTION_SERVICE_CALL_ENABLED   = var.account_intervention_service_call_enabled
+    ACCOUNT_INTERVENTION_SERVICE_URI            = var.account_intervention_service_uri
+    IPV_AUDIENCE                                = var.ipv_audience
+    IPV_AUTHORISATION_CALLBACK_URI              = var.ipv_authorisation_callback_uri
+    IPV_AUTHORISATION_CLIENT_ID                 = var.ipv_authorisation_client_id
+    IPV_AUTHORISATION_URI                       = var.ipv_authorisation_uri
+    IPV_BACKEND_URI                             = var.ipv_backend_uri
+    IPV_NO_SESSION_RESPONSE_ENABLED             = var.ipv_no_session_response_enabled
+    IPV_TOKEN_SIGNING_KEY_ALIAS                 = local.ipv_token_auth_key_alias_name
+    LOCALSTACK_ENDPOINT                         = var.use_localstack ? var.localstack_endpoint : null
+    LOGIN_URI                                   = "https://${local.frontend_fqdn}/"
+    OIDC_API_BASE_URL                           = local.api_base_url
+    REDIS_KEY                                   = local.redis_key
+    SPOT_QUEUE_URL                              = aws_sqs_queue.spot_request_queue.id
+    TXMA_AUDIT_QUEUE_URL                        = module.oidc_txma_audit.queue_url
   }
   handler_function_name = "uk.gov.di.authentication.ipv.lambda.IPVCallbackHandler::handleRequest"
 
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
@@ -67,10 +67,6 @@ public class IPVCallbackHelper {
     private final SessionService sessionService;
     private final AwsSqsClient sqsClient;
 
-    private IPVCallbackHelper() {
-        this(ConfigurationService.getInstance());
-    }
-
     public IPVCallbackHelper(ConfigurationService configurationService) {
         this.auditService = new AuditService(configurationService);
         this.cloudwatchMetricsService = new CloudwatchMetricsService(configurationService);
@@ -174,11 +170,17 @@ public class IPVCallbackHelper {
     }
 
     public void doAccountIntervention(AccountInterventionStatus accountInterventionStatus) {
-        segmentedFunctionCall(
-                "AIS: getAccountStatus",
-                () ->
-                        this.accountInterventionService.doAccountIntervention(
-                                accountInterventionStatus));
+        if (accountInterventionStatus.blocked()) {
+            LOG.info("Account is blocked");
+            // TODO: (ATO-171) back channel logout + (ATO-170) redirect to blocked page
+        } else if (accountInterventionStatus.suspended()
+                || accountInterventionStatus.resetPassword()
+                || accountInterventionStatus.reproveIdentity()) {
+            LOG.info(
+                    "Account is suspended, requires a password reset, or requires identity to be reproved");
+            // TODO: (ATO-171) back channel logout + (ATO-170) redirect to suspended
+            // page
+        }
     }
 
     public Optional<ErrorObject> validateUserIdentityResponse(UserInfo userIdentityUserInfo)
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
@@ -1,0 +1,359 @@
+package uk.gov.di.authentication.ipv.helpers;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
+import com.nimbusds.oauth2.sdk.id.Subject;
+import com.nimbusds.openid.connect.sdk.AuthenticationErrorResponse;
+import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+import com.nimbusds.openid.connect.sdk.AuthenticationSuccessResponse;
+import com.nimbusds.openid.connect.sdk.claims.UserInfo;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.ipv.domain.IPVAuditableEvent;
+import uk.gov.di.authentication.ipv.entity.IpvCallbackException;
+import uk.gov.di.authentication.ipv.entity.LogIds;
+import uk.gov.di.authentication.ipv.entity.SPOTClaims;
+import uk.gov.di.authentication.ipv.entity.SPOTRequest;
+import uk.gov.di.orchestration.shared.entity.AccountInterventionStatus;
+import uk.gov.di.orchestration.shared.entity.ClientRegistry;
+import uk.gov.di.orchestration.shared.entity.ClientSession;
+import uk.gov.di.orchestration.shared.entity.IdentityClaims;
+import uk.gov.di.orchestration.shared.entity.LevelOfConfidence;
+import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
+import uk.gov.di.orchestration.shared.entity.Session;
+import uk.gov.di.orchestration.shared.entity.UserProfile;
+import uk.gov.di.orchestration.shared.entity.ValidClaims;
+import uk.gov.di.orchestration.shared.exceptions.UserNotFoundException;
+import uk.gov.di.orchestration.shared.serialization.Json;
+import uk.gov.di.orchestration.shared.serialization.Json.JsonException;
+import uk.gov.di.orchestration.shared.services.AccountInterventionService;
+import uk.gov.di.orchestration.shared.services.AuditService;
+import uk.gov.di.orchestration.shared.services.AuthCodeResponseGenerationService;
+import uk.gov.di.orchestration.shared.services.AuthorisationCodeService;
+import uk.gov.di.orchestration.shared.services.AwsSqsClient;
+import uk.gov.di.orchestration.shared.services.CloudwatchMetricsService;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+import uk.gov.di.orchestration.shared.services.DynamoClientService;
+import uk.gov.di.orchestration.shared.services.DynamoIdentityService;
+import uk.gov.di.orchestration.shared.services.DynamoService;
+import uk.gov.di.orchestration.shared.services.SerializationService;
+import uk.gov.di.orchestration.shared.services.SessionService;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+
+import static uk.gov.di.orchestration.shared.entity.IdentityClaims.VOT;
+import static uk.gov.di.orchestration.shared.entity.IdentityClaims.VTM;
+import static uk.gov.di.orchestration.shared.entity.ValidClaims.RETURN_CODE;
+import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
+import static uk.gov.di.orchestration.shared.helpers.ConstructUriHelper.buildURI;
+import static uk.gov.di.orchestration.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
+import static uk.gov.di.orchestration.shared.services.AuditService.MetadataPair.pair;
+
+public class IPVCallbackHelper {
+    private static final Logger LOG = LogManager.getLogger(IPVCallbackHelper.class);
+    protected final Json objectMapper;
+    private final AccountInterventionService accountInterventionService;
+    private final AuditService auditService;
+    private final AuthCodeResponseGenerationService authCodeResponseService;
+    private final AuthorisationCodeService authorisationCodeService;
+    private final CloudwatchMetricsService cloudwatchMetricsService;
+    private final ConfigurationService configurationService;
+    private final DynamoClientService dynamoClientService;
+    private final DynamoIdentityService dynamoIdentityService;
+    private final DynamoService dynamoService;
+    private final SessionService sessionService;
+    private final AwsSqsClient sqsClient;
+
+    private IPVCallbackHelper() {
+        this(ConfigurationService.getInstance());
+    }
+
+    public IPVCallbackHelper(ConfigurationService configurationService) {
+        this.auditService = new AuditService(configurationService);
+        this.cloudwatchMetricsService = new CloudwatchMetricsService(configurationService);
+        this.accountInterventionService =
+                new AccountInterventionService(
+                        configurationService, cloudwatchMetricsService, auditService);
+        this.authorisationCodeService = new AuthorisationCodeService(configurationService);
+        this.configurationService = configurationService;
+        this.dynamoClientService = new DynamoClientService(configurationService);
+        this.dynamoIdentityService = new DynamoIdentityService(configurationService);
+        this.dynamoService = new DynamoService(configurationService);
+        this.objectMapper = SerializationService.getInstance();
+        this.sessionService = new SessionService(configurationService);
+        this.sqsClient =
+                new AwsSqsClient(
+                        configurationService.getAwsRegion(),
+                        configurationService.getSpotQueueUri(),
+                        configurationService.getSqsEndpointUri());
+        this.authCodeResponseService =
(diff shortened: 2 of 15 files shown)


hannah-leong:
To do:
- Adjust tests
- Check if IPV returns a 500 error when getting `userIdentityUserInfo`

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -5,66 +5,40 @@
...
-import uk.gov.di.orchestration.shared.entity.UserProfile;
-import uk.gov.di.orchestration.shared.entity.ValidClaims;
 import uk.gov.di.orchestration.shared.exceptions.NoSessionException;
 import uk.gov.di.orchestration.shared.exceptions.UnsuccessfulCredentialResponseException;
-import uk.gov.di.orchestration.shared.helpers.ClientSubjectHelper;
-import uk.gov.di.orchestration.shared.helpers.ConstructUriHelper;
-import uk.gov.di.orchestration.shared.helpers.CookieHelper;
-import uk.gov.di.orchestration.shared.helpers.PersistentIdHelper;
+import uk.gov.di.orchestration.shared.exceptions.UserNotFoundException;
+import uk.gov.di.orchestration.shared.helpers.*;
```
billrdunn, line 26: Wildcard imports should be disabled - see 3b [here](https://govukverify.atlassian.net/wiki/spaces/Orch/pages/3763668245/New+developer+guide)

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +178,52 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                            userProfile.getSubjectID(),
+                            URI.create(configurationService.getInternalSectorUri()),
+                            dynamoService.getOrGenerateSalt(userProfile));
+
+            if (errorObject.isPresent()) {
+                var accountInterventionStatus =
+                        ipvCallbackHelper.getAccountInterventionStatus(internalPairwiseSubjectId);
+                if (configurationService.isAccountInterventionServiceEnabled()) {
+                    if (accountInterventionStatus.blocked()) {
+                        // TODO: back channel logout + redirect to blocked page
```
billrdunn, line 209: Are these TODOs needed still?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +199,44 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                            () ->
+                                    ClientSubjectHelper.getSubject(
+                                            userProfile,
+                                            clientRegistry,
+                                            dynamoService,
+                                            configurationService.getInternalSectorUri()));
+            if (errorObject.isPresent()) {
+                var accountInterventionStatus =
+                        getAccountInterventionStatus(pairwiseSubject.getValue());
+                if (configurationService.isAccountInterventionServiceEnabled()) {
```
billrdunn, line 207: `configurationService.isAccountInterventionServiceEnabled()` has been removed in this unmerged PR: https://github.com/govuk-one-login/authentication-api/pull/3698

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -309,6 +310,34 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                    if (accountInterventionStatus.blocked()) {
+                        // TODO: back channel logout + redirect to blocked page
+                        LOG.info("Account is blocked");
+                    } else if (accountInterventionStatus.suspended()
+                            || accountInterventionStatus.resetPassword()
+                            || accountInterventionStatus.reproveIdentity()) {
+                        // TODO: back channel logout + redirect to suspended page
+                        LOG.info(
```
billrdunn, line 310: There are two cases which can generate the same log messages: when `errorObject.isPresent()` and `userIdentityError.isPresent()`. Would it be better to make these messages unique to help debugging?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -309,6 +310,34 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                "Account is suspended, requires a password reset, or requires identity to be reproved");
+                    }
+                }
+
+                var returnCode = userIdentityUserInfo.getClaim(RETURN_CODE.getValue());
+                if (returnCode instanceof List<?> && !((List<?>) returnCode).isEmpty()) {
```
billrdunn, line 316: Is there any reason to cast to `List<?>` in the second condition if the first condition already ensures that `returnCode` is a `List<?>`?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -309,6 +310,34 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             var userIdentityError = validateUserIdentityResponse(userIdentityUserInfo);
             if (userIdentityError.isPresent()) {
+                var accountInterventionStatus =
```
billrdunn, line 300: Lines 219-231 and 313-326 could be extracted to a private method

ci/terraform/oidc/ipv-callback.tf
```diff
@@ -35,23 +35,23 @@ module "ipv-callback" {
...
-    IPV_BACKEND_URI                 = var.ipv_backend_uri
-    IPV_NO_SESSION_RESPONSE_ENABLED = var.ipv_no_session_response_enabled
-    IPV_TOKEN_SIGNING_KEY_ALIAS     = local.ipv_token_auth_key_alias_name
-    LOCALSTACK_ENDPOINT             = var.use_localstack ? var.localstack_endpoint : null
-    LOGIN_URI                       = "https://${local.frontend_fqdn}/"
-    OIDC_API_BASE_URL               = local.api_base_url
-    REDIS_KEY                       = local.redis_key
-    SPOT_QUEUE_URL                  = aws_sqs_queue.spot_request_queue.id
-    TXMA_AUDIT_QUEUE_URL            = module.oidc_txma_audit.queue_url
+    DYNAMO_ENDPOINT                            = var.use_localstack ? var.lambda_dynamo_endpoint : null
```
billrdunn, line 38: Would be nicer to squash the formatting changes into the previous commit, can show you how if you like

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -107,9 +107,10 @@ public IPVCallbackHandler(
 
     public IPVCallbackHandler(ConfigurationService configurationService) {
         var kmsConnectionService = new KmsConnectionService(configurationService);
+        this.cloudwatchMetricsService = new CloudwatchMetricsService(configurationService);
         this.accountInterventionService =
-                new AccountInterventionService(configurationService, newHttpClient());
-        this.cloudwatchMetricsService = new CloudwatchMetricsService();
+                new AccountInterventionService(
```
billrdunn: Looks like the changes to IPVCallbackHandler in commit 8e2077f could be squashed into the previous commit

ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandlerTest.java
```diff
@@ -3,24 +3,14 @@
...
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
-import com.nimbusds.oauth2.sdk.AccessTokenResponse;
-import com.nimbusds.oauth2.sdk.AuthorizationCode;
-import com.nimbusds.oauth2.sdk.ErrorObject;
-import com.nimbusds.oauth2.sdk.OAuth2Error;
-import com.nimbusds.oauth2.sdk.ResponseType;
-import com.nimbusds.oauth2.sdk.Scope;
-import com.nimbusds.oauth2.sdk.TokenErrorResponse;
-import com.nimbusds.oauth2.sdk.TokenRequest;
+import com.nimbusds.oauth2.sdk.*;
```
billrdunn, line 6: Remove wildcards - see other comment

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: It's not a TODO for this ticket, more of a placeholder to flag that the functionality here is incomplete and this is what needs to be done

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: Ah thanks, I thought I'd changed that in IntelliJ but that was a while ago, maybe an update changed it - will fix now

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: `returnCode` is still of type `Object` (even though we know it is a `List`) which means we cannot access operations specific to `List`, such as `isEmpty()`, hence the need to cast it

billrdunn:
Should there be any changes to or new integration tests?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: Note: there are a few other instances where wildcards are present that are unrelated to this PR but within `ipv-api`/`oidc-api`

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +194,52 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                            configurationService.getInternalSectorUri()));
+
+            var internalPairwiseSubjectId =
+                    ClientSubjectHelper.calculatePairwiseIdentifier(
+                            userProfile.getSubjectID(),
+                            URI.create(configurationService.getInternalSectorUri()),
+                            dynamoService.getOrGenerateSalt(userProfile));
+
+            if (errorObject.isPresent()) {
+                var accountInterventionStatus =
```
CarlyG55, line 221: This might be my lack on knowledge on the AIS stuff, but don't we only want to get the accountInterventionStatus if AIS is enabled?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +194,52 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        LOG.info("Account is blocked");
+                    } else if (accountInterventionStatus.suspended()
+                            || accountInterventionStatus.resetPassword()
+                            || accountInterventionStatus.reproveIdentity()) {
+                        // TODO: back channel logout + redirect to suspended page
+                        LOG.info(
+                                "Account is suspended, requires a password reset, or requires identity to be reproved");
+                    }
+
+                    return ipvCallbackHelper.generateAuthenticationErrorResponse(
```
CarlyG55, line 235: This is currently only called if AIS is enabled - is that correct?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
```diff
@@ -0,0 +1,361 @@
...
+                        .toString();
+
+        if (!trustmarkURL.equals(userIdentityUserInfo.getClaim(VTM.getValue()))) {
+            LOG.warn("VTM does not contain expected trustmark URL");
+            throw new IpvCallbackException("IPV trustmark is invalid");
+        }
+        return Optional.empty();
+    }
+
+    public APIGatewayProxyResponseEvent processReturnCode(
```
CarlyG55, line 199: Could you add a bit more detail to this name? I don't think it's super clear as to what it's doing, eg when it's called I wouldn't expect it to return an API gateway response (sorry I know naming is hard!)

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
```diff
@@ -0,0 +1,361 @@
...
+                302, "", Map.of(ResponseHeaders.LOCATION, errorResponse.toURI().toString()), null);
+    }
+
+    public AccountInterventionStatus getAccountInterventionStatus(
+            String internalPairwiseSubjectId) {
+        var accountInterventionStatus =
+                segmentedFunctionCall(
+                        "AIS: getAccountStatus",
+                        () ->
+                                this.accountInterventionService.getAccountStatus(
```
hannah-leong, line 163: This has been changed to also take in auditContext (optional) - something to include?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: Currently, in `getAccountInterventionStatus`, it checks `call_enabled`
- If true, it makes the call then audits and returns the result
- If false, it returns `noInterventionResponse` (all intervention fields equal to false)

In the next step it checks `action_enabled`
- If true, it does any required intervention steps
- If false, it moves on

The naming is slightly confusing as it seems like the AIS call is always made, but this is essentially either getting the real intervention response or a dummy intervention response. This way, when the flag is removed it just needs to be changed in the service. 

Also, as Bill pointed out, the functionality corresponding to the `action_enabled` flag should be extracted - a similar process will implemented there such that when said flag is removed, it should result in as few code changes as possible.

Does that make sense?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: Just saw that when addressing your comment above - it should be outside the conditional, shall move :)

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
hannah-leong: They are both in relation to the AIS result so when this functionality is extracted out both locations should be calling the same function -> will be using the same logging message -> should be clear that it is from the AIS result :)

hannah-leong: They also produce different logs and the `errorObject.isPresent()` situation additionally audits the error event in the `generateAuthenticationErrorResponse` method

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
CarlyG55: (I'm going to resolve this as Bill is off today and we want to get it merged 🙂)

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
CarlyG55: Yep, and I think the code reads much more nicely after this refactor you've done - thank you!

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -306,9 +303,55 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                        OAuth2Error.ACCESS_DENIED,
+                                        authRequest.getState(),
+                                        authRequest.getResponseMode());
+                        return generateApiGatewayProxyResponse(
+                                302,
+                                "",
+                                Map.of(ResponseHeaders.LOCATION, errorResponse.toURI().toString()),
+                                null);
+                    } else {
+                        LOG.info("Generating auth code response");
```
CarlyG55, line 332: This is really nit picky, but I'd argue that we should maybe mention the return code in here, as we've just gone through that step of the logic - just so it's clear for someone checking the logs that we're generating an auth code response in a slightly unhappy state

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
CarlyG55: What do you think? 🙂

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
```diff
@@ -0,0 +1,338 @@
...
+    private final ConfigurationService configurationService;
+    private final DynamoClientService dynamoClientService;
+    private final DynamoIdentityService dynamoIdentityService;
+    private final DynamoService dynamoService;
+    private final SessionService sessionService;
+    private final AwsSqsClient sqsClient;
+
+    private IPVCallbackHelper() {
+        this(ConfigurationService.getInstance());
+    }
```
ethanmills, line 71: What is this for?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
```diff
@@ -0,0 +1,338 @@
...
+                        "resetPassword",
+                        String.valueOf(accountInterventionStatus.resetPassword()),
+                        "reproveIdentity",
+                        String.valueOf(accountInterventionStatus.reproveIdentity())));
+        return accountInterventionStatus;
+    }
+
+    public void doAccountIntervention(AccountInterventionStatus accountInterventionStatus) {
+        segmentedFunctionCall(
+                "AIS: getAccountStatus",
```
ethanmills, line 177: This is the wrong segment name. I'm also not sure this needs to be a segmented function call. Are we specifically interested in the performance of this block?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
ethanmills, line 1: I'm not a particular fan of this factoring of extracting helper methods to a separate class when they're only used in one place. I understand what you're trying to achieve, but I think I would draw the boundaries in a slightly different place. 

This shouldn't block this PR being merged, because it's important that it goes in soon. When we have a spare moment, it would be good to spend a bit of time pairing on refactoring this.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +196,43 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-            }
             var userProfile =
                     dynamoService
                             .getUserProfileFromEmail(session.getEmailAddress())
                             .orElseThrow(
                                     () ->
                                             new IpvCallbackException(
                                                     "Email from session does not have a user profile"));
+            var rpPairwiseSubject =
+                    segmentedFunctionCall(
```
ethanmills, line 207: Again, no need to segment

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -125,6 +125,20 @@ private AccountInterventionStatus retrieveAccountStatus(String internalPairwiseS
...
+            // TODO: (ATO-171) back channel logout + (ATO-170) redirect to blocked page
+        } else if (accountInterventionStatus.suspended()
+                || accountInterventionStatus.resetPassword()
+                || accountInterventionStatus.reproveIdentity()) {
+            LOG.info(
+                    "Account is suspended, requires a password reset, or requires identity to be reproved");
+            // TODO: (ATO-171) back channel logout + (ATO-170) redirect to suspended
+            // page
+        }
+    }
```
ethanmills, line 140: This shouldn't be in the shared service in this way. The logic of which flags result in which interventions depends on where in the journey it is called from. For example, suspended does not result in an intervention if on an ID journey when after auth.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -212,21 +196,43 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                            userProfile,
+                                            clientRegistry,
+                                            dynamoService,
+                                            configurationService.getInternalSectorUri()));
+
+            var internalPairwiseSubjectId =
+                    ClientSubjectHelper.calculatePairwiseIdentifier(
+                            userProfile.getSubjectID(),
+                            URI.create(configurationService.getInternalSectorUri()),
+                            dynamoService.getOrGenerateSalt(userProfile));
```
ethanmills, line 220: We are calculating this here, but retrieving from the session on line 242. Why the difference?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
```diff
@@ -0,0 +1,341 @@
...
+        LOG.info("Is journey a test journey: {}", isTestJourney);
+
+        cloudwatchMetricsService.incrementCounter("SignIn", dimensions);
+        cloudwatchMetricsService.incrementSignInByClient(
+                session.isNewAccount(),
+                clientSessionId,
+                clientSession.getClientName(),
+                isTestJourney);
+
+        authCodeResponseService.saveSession(false, sessionService, session);
```
ethanmills, line 269: I wonder if we should session.setAuthenticated(true) before saving here. I think I would expect to be successfully auth'd even if my identity wasn't proven

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -123,165 +125,89 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                var client =
-                        dynamoClientService
-                                .getClient(clientID.getValue())
-                                .orElseThrow(
-                                        () -> new ClientNotFoundException(clientID.getValue()));
-                internalSubjectId =
-                        Objects.isNull(session.getEmailAddress())
-                                ? AuditService.UNKNOWN
-                                : userProfile.getSubjectID();
+                subjectId = authCodeResponseService.getSubjectId(session);
```
ethanmills, line 207: I think it's a shame that this means that we duplicate the user profile dynamo lookup. Fine for now, but let's follow this up to see if we can improve this

ipv-api/src/main/java/uk/gov/di/authentication/ipv/helpers/IPVCallbackHelper.java
ethanmills: This happens inside the method

ethanmills (approved):

PR #3673, opened 2023-12-11
AUT-2139: expose MfaMethodType in CheckUserExists API response

mark-cab:
## What?

[JIRA link](https://govukverify.atlassian.net/browse/AUT-2139)

We need to pass the users MFA type to the ui during password reset so the end user can navigate to the relevant MFA screen (auth app / sms).

## Why?

User will not be logged in so the user context at this stage will not contain the required information, we need to expose the mfa type of the user prior to login so we can complete the password reset user journey.


dbes-gds:
Mind adding some extra integration tests?

dbes-gds (approved):

PR #3682, opened 2023-12-12
ATO-177: Audit calls to the Account Intervention Services (Manual Rebase)

kerrr:
### What?

Call the AuditService when we get a response from the AccountInterventionService.

Introduce an AuditContext object to carry information required for the audit event sent to TXMA. There is a significant overlap with the data contained in this object and TxmaAuditUser, however they are used for differnet things. The TxmaAuditUser is used to construct the user section of a TxmaAuditEvent, and therefor must match that exactly. The AuditContext is intended to represent the data that we understand about the user and this request that can be shared by multiple events. To make this useful in more places, more work should be done to understand event specific information, added as metadata pairs. Further work on that is out of scope for this ticket.

Added a new AccountInterventionsAuditableEvent to carry the AuditableEvent, since this is shared.

Update the AuthenticationCallbackHandler and ProcessingIdentityHandler calls to the AccountInterventionService to provide an audit context.

### Why?
Required for ATO-43

### DIFF
--- a/ci/terraform/oidc/authentication-callback.tf
+++ b/ci/terraform/oidc/authentication-callback.tf
@@ -27,26 +27,26 @@ module "authentication_callback" {
   environment     = var.environment
 
   handler_environment_variables = {
-    ACCOUNT_INTERVENTION_SERVICE_AUDIT_ENABLED = var.account_intervention_service_audit_enabled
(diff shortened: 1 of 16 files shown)


ethanmills:
@kerrr what's your plan? Would it be useful for me to duplicate my comments on the previous PR here?

kerrr:
No need to copy comments, I'll address here.

ethanmills (approved):
Thanks!

PR #3684, opened 2023-12-12
ATO-222: Emit cloudwatch signin metrics on successful signin

ethanmills:
## What?

Emit cloudwatch signin metrics on successful signin

## Why?

These are used in several dashboards. We should keep them accurate post-split

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -26,15 +26,30 @@ import uk.gov.di.authentication.oidc.services.AuthenticationTokenService;
 import uk.gov.di.authentication.oidc.services.InitiateIPVAuthorisationService;
 import uk.gov.di.orchestration.audit.AuditContext;
 import uk.gov.di.orchestration.shared.entity.ClientRegistry;
+import uk.gov.di.orchestration.shared.entity.ClientSession;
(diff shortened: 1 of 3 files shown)


isaac-GDS (dismissed):

isaac-GDS (approved):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -404,6 +398,50 @@ public APIGatewayProxyResponseEvent handleRequest(
         }
     }
 
+    private Map<String, String> buildDimensions(
```
hannah-leong, line 401: This is similar to something done as part of the [ATO-133/ATO-134](https://github.com/govuk-one-login/authentication-api/pull/3658) - it is in the shared location.
Happy to make adjustments to better align with this so that only a shared method is needed :)

hannah-leong: https://github.com/govuk-one-login/authentication-api/pull/3658/files#:~:text=Map%3CString%2C%20String%3E%20getDimensions

ethanmills: I think this one will go in first. If you get a chance to refactor this to use your shared method that would be great, if not no big deal 🙂 Thanks!

PR #3698, opened 2023-12-18
ATO-177: Audit calls to the Account Intervention Services (Re-apply)

kerrr:
### What?
Call the AuditService when we get a response from the AccountInterventionService.

Introduce an AuditContext object to carry information required for the audit event sent to TXMA. There is a significant overlap with the data contained in this object and TxmaAuditUser, however they are used for differnet things. The TxmaAuditUser is used to construct the user section of a TxmaAuditEvent, and therefor must match that exactly. The AuditContext is intended to represent the data that we understand about the user and this request that can be shared by multiple events. To make this useful in more places, more work should be done to understand event specific information, added as metadata pairs. Further work on that is out of scope for this ticket.

Added a new AccountInterventionsAuditableEvent to carry the AuditableEvent, since this is shared.

Update the AuthenticationCallbackHandler and ProcessingIdentityHandler calls to the AccountInterventionService to provide an audit context.

### Why?
Required for ATO-43

Reapplies previous ATO-177 PR (https://github.com/govuk-one-login/authentication-api/pull/3682) with fix for problem reverted with https://github.com/govuk-one-login/authentication-api/pull/3697

### DIFF
--- a/ci/terraform/oidc/authentication-callback.tf
+++ b/ci/terraform/oidc/authentication-callback.tf
@@ -27,26 +27,26 @@ module "authentication_callback" {
   environment     = var.environment
 
   handler_environment_variables = {
-    ACCOUNT_INTERVENTION_SERVICE_AUDIT_ENABLED = var.account_intervention_service_audit_enabled
-    ACCOUNT_INTERVENTION_SERVICE_ENABLED       = var.account_intervention_service_enabled
-    ACCOUNT_INTERVENTION_SERVICE_URI           = var.account_intervention_service_uri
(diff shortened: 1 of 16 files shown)


ethanmills:
Let's wait until tomorrow to merge this one 😅 happy with the change in principle though

kerrr:
This highlights the diff between the reverted and new version
`git diff 6635280fb0a2ce90523bebbd812cb9b7a1c7c7ad 8445a95a196df1f024a6f1ef954f1947afe78f52`

andyloughran:
Do we have an interface definition for what we send to IPV that would have caught the additional claim?

ethanmills (approved):

PR #3706, opened 2023-12-19
ATO-236: Millisecond timestamp for txma audit

isaac-GDS:
## What?

Include an addition timestamp field `event_timestamp_ms` giving txma a millisecond timestamp

## Why?

HMRC requirement


hannah-leong (approved):

hannah-leong:
Before merging, pause promotion to production to test in staging just to double-check :)

PR #3708, opened 2023-12-19
AUT-1326: Remove code that sets password hint version

gtvj:
## What?

Remove code that sets password hint version but retain the `Feature` class so that we have the facility to run these kind of experiments in future. 

## Why?

The related code within `authentication-frontend` has been removed so there is now no need to set and provide this.

## Related PRs

* https://github.com/govuk-one-login/authentication-frontend/pull/1257

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/Features.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/Features.java
@@ -1,28 +1,9 @@
 package uk.gov.di.authentication.frontendapi.entity;
 
-import com.google.gson.annotations.Expose;
-import com.google.gson.annotations.SerializedName;
-
 public class Features {
-
-    @SerializedName("updatePasswordHintTextVersion")
-    @Expose
-    private String updatePasswordHintTextVersion;
-
-    public String getUpdatePasswordHintTextVersion() {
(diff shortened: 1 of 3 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/Features.java
```diff
@@ -1,28 +1,9 @@
 package uk.gov.di.authentication.frontendapi.entity;
 
-import com.google.gson.annotations.Expose;
-import com.google.gson.annotations.SerializedName;
-
 public class Features {
 
-    @SerializedName("updatePasswordHintTextVersion")
```
BeckaL, line 8: Could we remove this entire class now?

gtvj: Thanks @BeckaL - I thought the same and asked about this before raising the PR. The view was that it would be beneficial to retain the facility. What I should have done was update the PR to reflect this (I've done that now) I'll send you the Slack thread 👍🏽

BeckaL: Ah cool, makes sense

BeckaL (approved):

PR #3710, opened 2023-12-19
AUT-2113: Add reauthenticate param to buildUserStartInfo

BeckaL:
If this param is set to true, then the user is returned as not authenticated. This will form part of the reauthentication flow.

## Related PRs
https://github.com/govuk-one-login/authentication-frontend/pull/1253

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
@@ -129,6 +129,43 @@ class StartIntegrationTest extends ApiGatewayHandlerIntegrationTest {
         assertTxmaAuditEventsReceived(txmaAuditQueue, List.of(START_INFO_FOUND));
     }
 
+    @Test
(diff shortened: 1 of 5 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
```diff
@@ -383,6 +320,34 @@ private SignedJWT createSignedJWT(KeyPair keyPair, State state) throws JOSEExcep
         return signedJWT;
     }
 
+    private void verifyStandardClientInformationSetOnResponse(
+            StartResponse startResponse, Scope scope, State state) {
+        assertThat(startResponse.getClient().getClientName(), equalTo(TEST_CLIENT_NAME));
```
mrwilson, line 325: We're calling `.getClient()` in every line for this set of verifications, so a refactoring might be to make it take a `Client` rather than a `StartResponse`, and use `startResponse.getClient()` at the method's callsite instead.

Same applies for `verifyStandardUserInformationSetOnResponse` below.

ayoshebby (approved):

PR #3720, opened 2023-12-20
AUT-1998: Add new journey type to ‘VerifyMfaCodeHandler’ for auth app codes

mark-cab:
## What?

New CodeRequestType added for the 2FA password reset journey using auth app notifications. 

The relevant processor was also enhanced to process requests for the 2FA password reset journey.

## Why?

We maintain separate 2FA attempts counters for each journey type so that the user has the correct number of attempts before they are locked out.

A new counter type is required for Password Reset 2FA Auth App.


BeckaL (commented):
Code changes all look sensible to me - would you be able to add a bit more detail to the commit message and/or the PR description to explain this change a bit better / the effect of this? I think it's useful to have in the history. E.g. that this will create a new counter (I _think_ - that's what the ticket implies) so we can separately count password reset MFA attempts

dbes-gds (approved):

PR #3755, opened 2024-01-08
ATO-249: Add field to store a list of VTRs in ClientSession

billrdunn:
## What?

Add `vtrList` field to `ClientSession` class and initialise it using the existing `effectiveVectorOfTrust` field.

## Why?

The assumption that we are only returning one VTR is changing, and this is the first step to allowing multiple VTRs to be returned. This is a result of the inherit identities work.


ethanmills (dismissed):

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientSession.java
```diff
@@ -55,6 +59,7 @@ public VectorOfTrust getEffectiveVectorOfTrust() {
```
hannah-leong, line 59: Maybe add a getter method for the vtrList here? It will be needed in another ticket and I think it makes sense to have it as part of the vtrList field set-up

hannah-leong (approved):

PR #3758, opened 2024-01-08
ATO-250: Use VTR list from Client Session where required

billrdunn:
## What?

- Replace any usages of the old single VTR on `ClientSession` to the new list of VTRs.
- In the cases that our logic requires a single VTR, the VTR with the lowest credential trust value from the list is used.
- The VTR list is audited as a comma separated string.

## Why?

Currently we assume that we only receive a single VTR, but as part of the inherited identities work we need to be able to handle a list of VTRs.

### DIFF
--- a/doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandlerTest.java
+++ b/doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandlerTest.java
@@ -34,6 +34,7 @@ import uk.gov.di.orchestration.shared.entity.ClientSession;
 import uk.gov.di.orchestration.shared.entity.NoSessionEntity;
 import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
 import uk.gov.di.orchestration.shared.entity.Session;
+import uk.gov.di.orchestration.shared.entity.VectorOfTrust;
 import uk.gov.di.orchestration.shared.exceptions.NoSessionException;
 import uk.gov.di.orchestration.shared.exceptions.UnsuccessfulCredentialResponseException;
 import uk.gov.di.orchestration.shared.helpers.CookieHelper;
@@ -112,7 +113,8 @@ class DocAppCallbackHandlerTest {
     private final Session session = new Session(SESSION_ID).setEmailAddress(TEST_EMAIL_ADDRESS);
 
     private final ClientSession clientSession =
-            new ClientSession(generateAuthRequest().toParameters(), null, null, null);
+            new ClientSession(
+                    generateAuthRequest().toParameters(), null, (VectorOfTrust) null, null);
 
     @RegisterExtension
     private final CaptureLoggingExtension logging =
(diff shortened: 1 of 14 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientSession.java
```diff
@@ -79,4 +112,19 @@ public ClientSession setDocAppSubjectId(Subject docAppSubjectId) {
     public String getClientName() {
         return clientName;
     }
+
+    public String getVtrLocsAsCommaSeparatedString() {
```
CarlyG55, line 116: I don't think this sorts the vtrs - we want to have some kind of consistent ordering, so we always get the same dimensions

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AuthCodeResponseGenerationService.java
```diff
@@ -71,18 +71,11 @@ public Map<String, String> getDimensions(
...
-        var mfaNotRequired =
-                clientSession
-                        .getEffectiveVectorOfTrust()
-                        .getCredentialTrustLevel()
-                        .equals(CredentialTrustLevel.LOW_LEVEL);
-        var levelOfConfidence = LevelOfConfidence.NONE.getValue();
-        if (clientSession.getEffectiveVectorOfTrust().containsLevelOfConfidence()) {
-            levelOfConfidence =
-                    clientSession.getEffectiveVectorOfTrust().getLevelOfConfidence().getValue();
-        }
```
CarlyG55, line 83: I don't think what we have now is going to return the same thing in the case of no level of confidence - the helper returns empty string, and here we return `LevelOfConfidence.NONE.getValue()` which is `P0`

hannah-leong (dismissed):
Looks good to me :) Might want to test in staging?

doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandlerTest.java
```diff
@@ -112,7 +113,8 @@ class DocAppCallbackHandlerTest {
     private final Session session = new Session(SESSION_ID).setEmailAddress(TEST_EMAIL_ADDRESS);
 
     private final ClientSession clientSession =
-            new ClientSession(generateAuthRequest().toParameters(), null, null, null);
+            new ClientSession(
+                    generateAuthRequest().toParameters(), null, (VectorOfTrust) null, null);
```
CarlyG55, line 117: It might be a good idea to move the tests to the vector of trust list approach as well?

billrdunn: I started doing that but it got quite messy, was planning to do it in a separate PR

billrdunn:
> Looks good to me :) Might want to test in staging?

Currently testing in sandpit and will pause prod to test in staging when it's merged 👍

CarlyG55:
@ethanmills what are we doing in terms of updating code that's in both shared and orchestration-shared? should this update both?

ethanmills:
> @ethanmills what are we doing in terms of updating code that's in both shared and orchestration-shared? should this update both?

No, I think we just update ours and let Auth know about the change. It should be up to them

doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandlerTest.java
CarlyG55: okay sure 👍

CarlyG55 (approved):

PR #3760, opened 2024-01-08
BAU: Switch state and interventions JSON 'shape'

DTaylorGDS:
## What?
- Switch state and interventions JSON 'shape' 

## Why?
- State POJO reflected contents of "interventions" JSON field coming back from Account Interventions Service (or stub)
- And vice versa for Interventions POJO and "state" field
- This should help the account interventions handler to parse the JSON from AIS correctly


alhcomer (approved):

DTaylorGDS:
This PR has things the opposite way round - the state should indeed contain the blocked boolean fields as is already the case. The error is in the stub, NOT the consuming lambda

PR #3764, opened 2024-01-09
ATO-133 // ATO-134: return codes integration tests

hannah-leong:
## What?

Integration tests for the return codes piece of work

## Why?

To test that:
- (ATO-133) Return codes are saved to DynamoDB if they are present, even if the identity is not proven
- (ATO-134) SPoT is bypassed and an authcode is issued if IPV returns P0 and the RP requests return codes

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/3658


integration-tests/src/test/java/uk/gov/di/authentication/api/IPVCallbackHandlerIntegrationTest.java
```diff
@@ -301,6 +305,98 @@ void shouldRedirectToLoginWhenSuccessfullyProcessedIpvResponse() throws Json.Jso
                 startsWith(TEST_CONFIGURATION_SERVICE.getLoginURI().toString()));
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
```
billrdunn, line 310: Not suggesting any changes as I think this approach is quite neat but I think there's a case for writing two separate tests (with more descriptive names) rather than passing `validLoC` here.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -459,6 +459,10 @@ private APIGatewayProxyResponseEvent redirectToFrontendErrorPage(String errorPag
 
     private boolean rpRequestedReturnCode(
             ClientRegistry clientRegistry, AuthenticationRequest authRequest) {
+        if (authRequest.getOIDCClaims() == null
```
billrdunn, line 462: The changes to `IPVCallbackHandler` look like they should belong in a separate commit

integration-tests/src/test/java/uk/gov/di/authentication/api/IPVCallbackHandlerIntegrationTest.java
```diff
@@ -397,6 +398,66 @@ void shouldStoreReturnCodesInDynamoWhenTheyArePresent(boolean validLoC)
         assertThat(returnCode.size(), equalTo(1));
     }
 
+    @ParameterizedTest
+    @ValueSource(booleans = {true, false})
+    void shouldBypassSPoTAndReturnAuthCodeIfIPVReturnsP0ButReturnCodeIsPresentAndRequested(
```
billrdunn, line 403: I think this test name is slightly misleading as it covers a case when `returnCodeRequested` is `false`. I'd split into two tests and extract some common logic if needed.

PR #3741, opened 2024-01-04
ATO-171: Perform back-channel logout if account status is suspended or blocked (v2)

billrdunn:
## What?

If the account status received from Account Interventions is `suspended` or `blocked`, destroy all sessions and redirect to account suspended/blocked page.

This PR does not handle destroying the Authentication frontend session.

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/3647

PR #3770, opened 2024-01-10
ATO-171: Perform back-channel logout if account status is suspended or blocked

billrdunn:
_This is a duplicate of https://github.com/govuk-one-login/authentication-api/pull/3741 which was reverted (https://github.com/govuk-one-login/authentication-api/pull/3765) due to a pipeline failure._

## What?

If the account status received from Account Interventions is `suspended` or `blocked`, destroy all sessions and redirect to account suspended/blocked page.

This PR does not handle destroying the Authentication frontend session.

Note that back-channel logout has not been implemented in `ProcessingIdentityHandler` and is marked as TODO. This will be done in a separate PR.

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/3647

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -24,6 +24,7 @@ import uk.gov.di.authentication.oidc.exceptions.AuthenticationCallbackException;
 import uk.gov.di.authentication.oidc.services.AuthenticationAuthorizationService;
 import uk.gov.di.authentication.oidc.services.AuthenticationTokenService;
 import uk.gov.di.authentication.oidc.services.InitiateIPVAuthorisationService;
+import uk.gov.di.authentication.oidc.services.LogoutService;
 import uk.gov.di.orchestration.audit.AuditContext;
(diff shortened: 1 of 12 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -326,8 +332,23 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        configurationService.isAccountInterventionServiceCallEnabled()
+                                ? accountInterventionService.getAccountStatus(
+                                        userInfo.getSubject().getValue(), auditContext)
+                                : null;
+
+                if (nonNull(accountStatus)) {
+                    if (accountStatus.blocked()
+                            || accountStatus.resetPassword()
+                            || (!identityRequired && accountStatus.suspended())) {
+                        return logoutService.handleAccountInterventionLogout(
```
hannah-leong, line 344: Should there be the check for the AISActionEnabled here as logging out is an action?

billrdunn: I persisted the logic from https://github.com/govuk-one-login/authentication-api/pull/3698 but I think you're right, have changed

ethanmills (approved):

PR #3790, opened 2024-01-11
BAU: Script to calculate internal common subject id for a user

dbes-gds:
## What?

Script to calculate internal common subject id for a user.

## Why?

For testing account interventions using either the stub or in staging, we need to know the internal common subject id for a user.

### DIFF
--- a/scripts/export-ics.sh
+++ b/scripts/export-ics.sh
@@ -1,0 +1,38 @@
+#!/usr/bin/env bash
+
+set -eu
+
+if [ $# -ne 2 ]
+  then
+    echo "Usage: export-ics.sh email environment"
+    exit 1
+fi
+
+export AWS_REGION=eu-west-2
+
+sector="identity.$2.account.gov.uk"
+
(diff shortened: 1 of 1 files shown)


scripts/export-ics.sh
```diff
@@ -0,0 +1,38 @@
...
+set -eu
+
+if [ $# -eq 0 ]
+  then
+    echo "Usage: export-ics.sh email"
+    exit 1
+fi
+
+export AWS_REGION=eu-west-2
+export ENVIRONMENT_NAME=build
```
TaherKapasi, line 12: would provide ENVIRONMENT_NAME as a 2nd argument that can use to overrride the env

TaherKapasi (approved):

scripts/export-ics.sh
```diff
@@ -0,0 +1,38 @@
...
+sector="identity.build.account.gov.uk"
```
ethanmills, line 13: environment name in here?

TaherKapasi (approved):

billrdunn:
Have tested the script for my account in sandpit and staging and it works great 👍

alhcomer (approved):

PR #3796, opened 2024-01-12
Ato 23/allow custom locs per rp

gn-GDS:
## What?

Update client registry (Java class) to have optional list of LoC and change code so it’s backwards compatible.
Needs to validate level of confidence being requested, as we will be accepting new levels. Different clients can request different levels.

## Why?

Part of HMRC migration

## Related PRs

[ATO-19](https://govukverify.atlassian.net/browse/ATO-19)

[ATO-19]: https://govukverify.atlassian.net/browse/ATO-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/LevelOfConfidence.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/LevelOfConfidence.java
@@ -6,6 +6,8 @@ import java.util.stream.Collectors;
 
 public enum LevelOfConfidence {
     NONE("P0", true),
+    HMRC1("PCl200", true),
+    HMRC2("PCl250", true),
     LOW_LEVEL("P1", false),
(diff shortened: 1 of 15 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/LevelOfConfidence.java
```diff
@@ -6,6 +6,8 @@
 
 public enum LevelOfConfidence {
     NONE("P0", true),
+    HMRC1("Cl200", true),
+    HMRC2("Cl250", true),
```
hannah-leong, line 10: Think these should be `PCL200` and `PCL250`

PR #3797, opened 2024-01-12
ATO-176: Fix AIS URI

billrdunn:
## What
Fix how the AIS URI is resolved so that no part of the path is stripped.

## Why
Previously the call to AIS was stripping the environment part of the path, ie. giving
`https://some.uri.com/abc/id` instead of `https://some.uri.com/environment/abc/id`


ethanmills (commented):
Can we add a test case for this that would have failed before this change please?

alhcomer (dismissed):

ethanmills (approved):

PR #3799, opened 2024-01-12
add part phone number to user exists api

mark-cab:
## What?

Added the phoneNumberLastThree field to the CheckUserExists API response.

## Why?

Previously decided to user CheckUserExists API to present data for the new MFA PW reset journey as the user is not logged in before this journey so we require more information to proceed through the screens. This update adds the last 3 digits to API response so that the frontend can use it for the text 'We have sent a code to your phone number ending with XXX'.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/CheckUserExistsIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/CheckUserExistsIntegrationTest.java
@@ -33,7 +33,9 @@ import static java.util.Collections.singletonList;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.fail;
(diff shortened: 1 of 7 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/CheckUserExistsIntegrationTest.java
```diff
@@ -85,6 +86,9 @@ void shouldCallUserExistsEndpointAndReturnAuthenticationRequestStateWhenUserExis
         assertThat(checkUserExistsResponse.getEmail(), equalTo(emailAddress));
         assertThat(checkUserExistsResponse.getMfaMethodType(), equalTo(mfaMethodType));
         assertTrue(checkUserExistsResponse.doesUserExist());
+        if (MFAMethodType.SMS.equals(mfaMethodType)) {
+            assertThat(checkUserExistsResponse.getPhoneNumberLastThree(), equalTo("321"));
+        }
```
LazarAlexandru-Constantin, line 91: I suggest adding another if statement to check if the mfaMethodType is of type AUTH_APP. If so, then to check if the getPhoneNumberLastThree() returns an empty string.

LazarAlexandru-Constantin (changes requested):

LazarAlexandru-Constantin (approved):

BeckaL (approved):

PR #3807, opened 2024-01-15
ATO-264: Update vtr format for IPV core

hannah-leong:
## What?

- Extract only the levels of confidence into a list to send to IPV Core in the vtr claim
- Add the new format to the requests

## Why?

To provide IPV core with only the information they require in a nicer format


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/VectorOfTrust.java
```diff
@@ -150,6 +150,14 @@ private static VectorOfTrust parseVtrSet(JSONArray vtrJsonArray) {
                                                                 "Invalid VTR attribute")));
     }
 
+    public static List<String> getRequestedLevelsOfConfidence(List<VectorOfTrust> vtrList) {
+        var levelsOfConfidence = new ArrayList<String>();
```
gn-GDS, line 154: This may be cleaner as a mapping, something like:
`return vtrList.stream().map(VectorOfTrust::getLevelOfConfidence).map(LevelOfConfidence::getValue).toList();`

isaac-GDS (approved):

gn-GDS (approved):

PR #3817, opened 2024-01-18
ATO-176: Enable AIS actions flag in staging and add required env vars

billrdunn:
## What?

Enable the AIS actions flag in staging. This enables actions such as redirecting and auditing after a AIS response is received. 

Add the environment variables which are required when the flag is turned on.

## Why?

For testing in staging env.

PR #3820, opened 2024-01-19
ATO-265: Redirect to auth frontend instead of orch frontend in staging.

billrdunn:
## What?

Redirect to auth frontend instead of orch frontend in staging after blocked or suspended response from AIS.

## Why?

- This is a temporary measure until orch frontend is deployed to staging.
- This will allow testing of AIS integration in staging.

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/3817


ci/terraform/oidc/staging-overrides.tfvars
```diff
@@ -26,6 +26,8 @@ doc_app_decouple_enabled                  = true
 orch_client_id                            = "orchestrationAuth"
 auth_audience                             = "https://auth.staging.account.gov.uk"
 account_intervention_service_call_enabled = true
+account_status_suspended_uri              = "/auth-frontend/account-locked"
```
ethanmills, line 29: I don't think that's where the auth frontend lives 🙂

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/LogoutService.java
```diff
@@ -174,6 +174,32 @@ private APIGatewayProxyResponseEvent generateAccountInterventionLogoutResponse(
...
+                throw new RuntimeException("Account status must be blocked or suspended");
+            }
+            sessionId.ifPresent(
+                    t ->
+                            cloudwatchMetricsService.incrementLogout(
+                                    clientId, Optional.of(accountStatus)));
+
+            return generateLogoutResponse(
+                    ConstructUriHelper.buildURI(
+                            "https://signin.staging.account.gov.uk", redirectUri),
```
ethanmills, line 195: This needs to be environment specific. Once it's in production I would mind hardcoding the prod URL, but something like this is too easy to miss I think

PR #3824, opened 2024-01-19
ATO-190: Spike pact pipeline integration

hannah-leong:
## What?

GitHub action to publish contract tests to the pact broker

## Why?

To automate running and uploading the contract tests

## How it works

The running of the contract tests is done in the pipeline and the results are published to the pact broker. 
- Contract tests are run - these generate contracts that are stored in a `pact` directory (not committed to GitHub); files are named in the format `<consumer>-<provider>.json`. This directory is specified in the `build.gradle` as a system property `pact.rootdir`
- Contracts are then taken from this directory and published to the given pact broker using the pact package.

- Environment variables will need to be set up in GitHub secrets:
  - `PACT_USER` and `PACT_PASSWORD`: Basic auth credentials for the pact broker
  - `PACT_URL`: URL for the pact broker (make sure this is the same broker as the credentials)
  - `GIT_BRANCH` and `CONSUMER_APP_VERSION`: These are used for versioning/tagging
  - `PACT_BROKER_SOURCE_HEADER`: This is a secret value that is put into the request for further auth checks


.github/workflows/contract-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Check out repository code
+        uses: actions/checkout@v3
+      - name: Set up JDK 17
+        uses: actions/setup-java@v3
+        with:
+          java-version: '17'
+          distribution: 'temurin'
+          cache: 'gradle'
+      - name: Run contract tests
+        run: ./gradlew --parallel ipv-api:contractTest
```
isaac-GDS, line 35: Happy for this as this Spike was specific for IPV but as we add more contract tests for more coverage do we need to change this to '*:contractTest' ?

.github/workflows/contract-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Upload pacts to broker
+        if: ${{ github.actor != 'dependabot[bot]' && github.event_name != 'merge_group' }}
+        run: ./gradlew :ipv-api:pactPublish
```
isaac-GDS, line 38: Similar question to the one above.

.github/workflows/contract-tests.yml
hannah-leong: Yeah, will change when relevant

isaac-GDS (dismissed):

hannah-leong:
This GHA successfully published the consumer side contract tests for ipv-api, given the broker is set up to accept the test source in the query params.

ethanmills (approved):

PR #3876, opened 2024-01-30
ATO-297: Audit intervention response

ethanmills:
## What?

Record the response from the intervention service in the audit event. The real change here is in the third commit. The first two were just some tidying to make this change easier


billrdunn (approved):
LGTM

PR #3826, opened 2024-01-19
Aut 2148/create email check dynamo

BeckaL:
## What?

Adds a dynamo table for storing an email check result, consisting of an email address, a status and a time to live.

Also adds read and write functionality (not yet used).

## Why?

Part of the infrastructure and code that will be needed for adding the email checks - see the [epic ticket](https://govukverify.atlassian.net/jira/software/c/projects/AUT/boards/477?assignee=620a44d9a715c600691222f8&selectedIssue=AUT-2144) for the full architecture diagram.

PR #3894, opened 2024-02-02
AUT-2149: Add email check result handler

DTaylorGDS:
## What?
- Add email check result handler 
- Also add a field to the expected SQS and Dynamo store as agreed schema changed since initial work was done on that (see "Related PRs")
- This extra field will allow us to trace a request from the point that it is triggered all the way through to the result storage
- Note: the extra DB field is kept separate in the first commit. Subsequent work to implement the handler (the main focus of the Jira ticket) is in the second commit

## Why?
- This will record the results of a counter fraud check of emails (implementation details of that check are in a separate private repository)
- The results of that check can be used to block certain user journeys in case of a negative result (not currently implemented)

## Related PRs
- As noted above, the work of this PR has been augmented to add a further field reflecting a constant reference number: https://github.com/govuk-one-login/authentication-api/pull/3826

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/EmailCheckResultSqsMessage.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/EmailCheckResultSqsMessage.java
@@ -1,0 +1,54 @@
+package uk.gov.di.authentication.shared.entity;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import uk.gov.di.authentication.shared.validation.Required;
+
+public class EmailCheckResultSqsMessage {
+    @Expose
+    @SerializedName("Email")
+    @Required
+    private String email;
+
+    @Expose
+    @SerializedName("Status")
+    @Required
+    private EmailCheckResultStatus emailCheckResultStatus;
+
+    @Expose
+    @SerializedName("TimeToExist")
+    @Required
+    private long timeToExist;
+
+    @Expose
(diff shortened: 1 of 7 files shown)


BeckaL (dismissed):
LGTM

shared/src/main/java/uk/gov/di/authentication/shared/entity/EmailCheckResultSqsMessage.java
```diff
@@ -0,0 +1,54 @@
+package uk.gov.di.authentication.shared.entity;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import uk.gov.di.authentication.shared.validation.Required;
+
+public class EmailCheckResultSqsMessage {
```
BeckaL, line 7: Just wondering if we'd be able to do this as a record? I think we can't for dynamo db classes at the moment because as far as I can tell the dynamo library doesn't support them, but gson does and I'm assuming that's the only thing that matters in this case.

BeckaL (commented):
D'oh, was just looking at the first commit there. Just a question about whether we can use records.

shared/src/main/java/uk/gov/di/authentication/shared/entity/EmailCheckResultSqsMessage.java
DTaylorGDS: Yeah, good point. Thanks @BeckaL

BeckaL (approved):
Aaah records make things so much nicer :-) LGTM

PR #3896, opened 2024-02-02
AUT-2267: Update API to send message to SQS queue after sign up phone validation

LazarAlexandru-Constantin:
## What?

Update API to send message to SQS queue after sign up phone validation.

## Why?

When the Experian Phone Checker service (owned by Experian) has an outage then checks against phone numbers fail and do not happen.  There is no failure or DLQ for the transactions, so we have no way of trying to run them again.  The service has failed a few times recently which generates alerts that we have to deal with, but we are unable to recover the transactions.

## Related PRs
[AUT-2103](https://govukverify.atlassian.net/browse/AUT-2103)

[AUT-2103]: https://govukverify.atlassian.net/browse/AUT-2103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/PhoneNumberRequest.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/PhoneNumberRequest.java
@@ -1,0 +1,13 @@
+package uk.gov.di.authentication.frontendapi.entity;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import uk.gov.di.authentication.shared.entity.JourneyType;
+
+public record PhoneNumberRequest(
+        @Expose @SerializedName("phoneNumberVerified") boolean phoneNumberVerified,
+        @Expose @SerializedName("phoneNumber") String phoneNumber,
+        @Expose @SerializedName("updatedPhoneNumber") boolean updatedPhoneNumber,
+        @Expose @SerializedName("journeyType") JourneyType journeyType,
+        @Expose @SerializedName("internalCommonSubjectIdentifier")
+                String internalCommonSubjectIdentifier) {}
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/PhoneNumberCodeProcessor.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/PhoneNumberCodeProcessor.java
@@ -1,19 +1,28 @@
 package uk.gov.di.authentication.frontendapi.validation;
 
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.entity.CodeRequest;
 import uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent;
+import uk.gov.di.authentication.frontendapi.entity.PhoneNumberRequest;
 import uk.gov.di.authentication.shared.entity.CodeRequestType;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
 import uk.gov.di.authentication.shared.entity.MFAMethodType;
 import uk.gov.di.authentication.shared.entity.NotificationType;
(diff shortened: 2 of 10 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/PhoneNumberRequest.java
```diff
@@ -0,0 +1,60 @@
+package uk.gov.di.authentication.frontendapi.entity;
+
+import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
+import uk.gov.di.authentication.shared.entity.JourneyType;
+
+public class PhoneNumberRequest {
```
BeckaL, line 7: Could we make this a record to get rid of the class boilerplate?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/PhoneNumberCodeProcessor.java
```diff
@@ -99,29 +138,49 @@ public Optional<ErrorResponse> validateCode() {
...
+                        new PhoneNumberRequest(
+                                phoneNumberVerified,
+                                phoneNumber,
+                                updatedPhoneNumber,
+                                journeyType,
+                                internalCommonSubjectIdentifier);
+                try {
+                    sqsClient.send(objectMapper.writeValueAsString(phoneNumberRequest));
+                } catch (Exception e) {
+                    LOG.error("Unexpected exception: {}", e.getMessage());
```
BeckaL, line 166: I wonder about putting in a more useful error message here, like "Unexpected exception when writing phone number request to SQS" or equivalent?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/PhoneNumberCodeProcessor.java
```diff
@@ -99,29 +138,49 @@ public Optional<ErrorResponse> validateCode() {
...
+        if (codeRequest.getJourneyType() == JourneyType.REGISTRATION
+                || codeRequest.getJourneyType() == JourneyType.ACCOUNT_RECOVERY) {
+            JourneyType journeyType = codeRequest.getJourneyType();
+            UserProfile userProfile = userContext.getUserProfile().get();
+            boolean phoneNumberVerified = userProfile.isPhoneNumberVerified();
+            String phoneNumber =
+                    PhoneNumberHelper.formatPhoneNumber(codeRequest.getProfileInformation());
+            boolean updatedPhoneNumber = !phoneNumber.equals(userProfile.getPhoneNumber());
+
+            if (configurationService.isPhoneCheckerWithReplyEnabled()
```
BeckaL, line 150: I wonder particularly about pulling this check of the feature flag into the top `if` - I think it might be a bit easier to read if you could immediately see that this whole block from line 141 was only relevant with the feature flag on.

BeckaL (commented):
Generally looks sensible to me, and a nice refactoring of the existing code. A few comments / suggestions

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/validation/PhoneNumberCodeProcessor.java
mark-cab: This is the scenario I was talking about in standup. There could be a risk of regression then as currently you can update your current phone number with the same number during account recovery. Ideally we need a small story to validate this on the ui (or maybe a new errorcode on api) so that the user receives the correct feedback.
I have pulled out the sqs code into another method for visibility though.

BeckaL: Ah got you, yeah I was specifically thinking just that the feature flag check could go to the top but I think I was misunderstanding the scope of the overall if :-) The version with the method pulled out looks good, nice and readable

BeckaL (approved):
LGTM 👍

PR #3903, opened 2024-02-05
AUT-2318: audit event and user feedback on 5th failed login attempt

mark-cab:
## What?

The application will record the first instance of the AUTH_ACCOUNT_TEMPORARILY_LOCKED event on the login attempt where the user is locked out and not the subsequent attempt.

## Why?

During investigation of https://govukverify.atlassian.net/browse/AUT-2316  it was found that there was less than 15 minutes between events AUTH_ACCOUNT_TEMPORARILY_LOCKED and  AUTH_LOG_IN_SUCCESS which lead to the belief that the 15 minute timeout was not working as expected. 

The application only records the AUTH_ACCOUNT_TEMPORARILY_LOCKED event if the user makes a subsequent attempt after being locked out so the lockout actually started on the last instance of AUTH_INVALID_CREDENTIALS prior to this.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
@@ -178,7 +178,6 @@ public class LoginHandler extends BaseFrontendHandler<LoginRequest>
                 } else {
                     codeStorageService.increaseIncorrectPasswordCount(request.getEmail());
                 }
-
                 auditService.submitAuditEvent(
                         FrontendAuditableEvent.INVALID_CREDENTIALS,
                         userContext.getClientSessionId(),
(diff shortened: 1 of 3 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -178,24 +178,48 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
-                        AuditService.UNKNOWN,
-                        persistentSessionId,
-                        pair("internalSubjectId", userProfile.getSubjectID()),
-                        pair(
-                                "incorrectPasswordCount",
-                                codeStorageService.getIncorrectPasswordCount(request.getEmail())),
-                        pair("attemptNoFailedAt", configurationService.getMaxPasswordRetries()));
-
-                return generateApiGatewayProxyErrorResponse(401, ErrorResponse.ERROR_1008);
+                incorrectPasswordCount++;
```
dbes-gds, line 181: what does this do?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -178,24 +178,48 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
+                            userProfile.getPhoneNumber(),
+                            persistentSessionId,
+                            pair("internalSubjectId", userProfile.getSubjectID()),
+                            pair(
+                                    "attemptNoFailedAt",
+                                    configurationService.getMaxPasswordRetries()));
+
+                    return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1028);
+                } else {
+                    auditService.submitAuditEvent(
```
dbes-gds, line 202: should we send an INVALID_CREDENTIALS event in all cases, even in addition to ACCOUNT_TEMPORARILY_LOCKED to record the number of times that the credentials were entered incorrectly?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
mark-cab: I thought it was better to read, incrementing the count then checking against max retries after the new failure instead of comparing against configurationService.getMaxPasswordRetries() - 1 . Happy to change though

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -195,6 +194,27 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                                 codeStorageService.getIncorrectPasswordCount(request.getEmail())),
                         pair("attemptNoFailedAt", configurationService.getMaxPasswordRetries()));
 
+                if (incorrectPasswordCount + 1 >= configurationService.getMaxPasswordRetries()) {
```
dbes-gds, line 197: looking at this again, think we need to move it into the invalid credentials block, as that's the first time you know when the account is locked

dbes-gds (approved):

PR #3904, opened 2024-02-05
ATO-159: Add performance testing stub to staging

CarlyG55:
## What?

Add performance testing stub to staging db
## Why?

To allow performance testers to use the stub


hannah-leong (approved):
Looks good to me! :)

PR #3910, opened 2024-02-06
ATO-191/ipv contract tests

isaac-GDS:
## What?

Pact contract test for 

IPV-token
IPV-user-identity

 - Using the IPVcallback lambda as a way if unit testing

## Why?

Consumer contract

## Question?

Left out Additional Claims as they can be optional. Does it implies there should be more contract test for each additional claim ?

The structure and required field for claims.

The contacts have deliberately been left out of the PR as they cannot be commited to github. I would recommend viewing the contract as it showcases what is being matched. This can be done by running the pact tests locally or if I can be decided somewhere to upload the contracts.


ipv-api/build.gradle
```diff
@@ -28,14 +28,16 @@ dependencies {
             configurations.lambda,
             configurations.sqs,
             configurations.kms,
-            configurations.dynamodb
+            configurations.dynamodb,
+            "au.com.dius.pact.consumer:junit5:4.6.4"
```
ethanmills, line 32: Let's stick to the established pattern here

isaac-GDS: Do we need/want to do anything about getting it onto artifactory ?

ethanmills: Nope, artifactory is just a transparent cache at the moment

ethanmills (approved):
As far as I can tell, this looks really good. We won't really know until IPV start using it. Nice work!

PR #3917, opened 2024-02-07
ATO-248: Update vtr validation

hannah-leong:
## What?

Update vtr validation to account for the new list format. 

## Why?

- Vtrs must contain either all identity vectors (have P-values) or all non-identity vectors (do not have P-values)


ethanmills (commented):
What error will be returned to the user if this is invalid? Is it caught in a validation function somewhere?

hannah-leong:
It errors in the same way as non-compliant P2 vectors (P2 identity confidence must require at least Cl.Cm credential trust) - there is no catching of the `IllegalArgumentException` in the `IdentityHelper` or `AuthorisationHandler`

hannah-leong:
With a non-compliant list, it sends an `invalid_request` error (description: `Request vtr not valid`) to the rp redirect uri

ethanmills (approved):
Let's do the usual dance of pausing and testing in staging please

PR #3920, opened 2024-02-07
ATO-59: Removed DocAppAuthorizeHandler

BenjaminWCO:
## What?

Removing DocAppAuthorizeHandler

## Why?

No longer needed post Orch/Auth split


ethanmills:
I've spotted a couple of requests to this, need to look into them

PR #3924, opened 2024-02-08
ATO-350: Add new sandpit stubs to client registry

CarlyG55:
## What?

Add new sandpit stubs to client registry

## Why?

So we can use sandpit stubs to enter the flow

Note: I've added a docapp stub here, but it won't work currently as we don't have a cri stub set up. The resources will be generated so that in future if we do set it up it will be available to us

### DIFF
--- a/ci/terraform/shared/sandpit.tfvars
+++ b/ci/terraform/shared/sandpit.tfvars
@@ -27,6 +27,46 @@ stub_rp_clients = [
     one_login_service = false
     service_type      = "MANDATORY"
   },
+  {
+    client_name = "relying-party-stub-sandpit"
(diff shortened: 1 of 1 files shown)


hannah-leong (approved):
Looks good to me :)
Only _tiny_ thing is that in the other environments tfvars, `client_type` is generally after `identity_verification_supported` (apart from the first client in most files for some reason??). Nice that it's consistent here though, just observation that if ever we need to compare files it might be easier if they are on the equivalent line.

CarlyG55:
@hannah-leong I've left as is for now to get it in - at some point we'll want to remove the old entries so maybe something to think about then 🙂

PR #3925, opened 2024-02-08
AUT-1844: Log AIS response in test environments (configurable)

dbes-gds:
## What?

Log AIS response in test environments (configurable).

## Why?

Provide visibility on the application behaviour in test environments.


BeckaL (approved):
LGTM

PR #3936, opened 2024-02-09
Ato 270/fix duplicate key null error

gn-GDS:
## What?

- Make codeRequestTypes and Journeys the same between auth and orch. 
- Fixed Redis race condition

## Why?

Causing errors in production.


shared/src/main/java/uk/gov/di/authentication/shared/services/SessionService.java
```diff
@@ -113,17 +113,11 @@ public void deleteSessionFromRedis(String sessionId) {
...
-                        segmentedFunctionCall(
-                                "Deserialise session",
-                                () ->
-                                        OBJECT_MAPPER.readValue(
-                                                redisConnectionService.getValue(sessionId),
-                                                Session.class)));
-            } else {
-                return Optional.empty();
-            }
+            String serializedSession = redisConnectionService.getValue(sessionId);
```
gn-GDS, line 116: @ethanmills are we ok with removing the segmentedFunctionCall here?
Both `redisConnectionServic.getValue()` and `OBJECT_MAPPER.readValue()` call it anyway so we should still have good visibility?

shared/src/main/java/uk/gov/di/authentication/shared/services/SessionService.java
```diff
@@ -113,17 +113,11 @@ public void deleteSessionFromRedis(String sessionId) {
 
     public Optional<Session> readSessionFromRedis(String sessionId) {
         try {
```
ethanmills, line 115: Not sure if this is more or less nice:
```java
        String serializedSession = redisConnectionService.getValue(sessionId);
        return Optional.ofNullable(serializedSession).map(s -> {
            try {
                return OBJECT_MAPPER.readValue(s, Session.class);
            } catch (Json.JsonException e) {
                throw new RuntimeException(e);
            }
        });
```

I think I'd be tempted to define a `readValueUnchecked`, then it could just be 
```java
        String serializedSession = redisConnectionService.getValue(sessionId);
        return Optional.ofNullable(serializedSession).map(s -> OBJECT_MAPPER.readValue(s, Session.class));
```

shared/src/main/java/uk/gov/di/authentication/shared/services/SessionService.java
ethanmills: Yeah that's fine

ethanmills (dismissed):

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
ethanmills, line 1: Nice!

ethanmills (approved):

PR #3944, opened 2024-02-13
ATO-97: Create KMS signing key for storage token sent to IPV

billrdunn:
## What?

- Create a new KMS key using the ES256 signing algorithm
- Create a new policy for use of the signing key
- Attach the policy to the Lambda role for IPVAuthorisationHandler

## Why?

This will allow Orchestration to send a signed JWT in the JAR request to IPV

## Testing
Tested in sandpit: new KMS key can be retrieved and used to sign a message, only by the lambda with the permission to do so

- [✓] Impact on orch and auth mutual dependencies has been checked.

### DIFF
--- a/ci/terraform/oidc/ecc-signing-key.tf
+++ b/ci/terraform/oidc/ecc-signing-key.tf
@@ -1,0 +1,13 @@
+resource "aws_kms_key" "storage_token_signing_key_ecc" {
+  description              = "KMS signing key (ECC) for VC storage token"
+  deletion_window_in_days  = 30
+  key_usage                = "SIGN_VERIFY"
+  customer_master_key_spec = "ECC_NIST_P256"
+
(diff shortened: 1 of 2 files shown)


ci/terraform/oidc/ecc-signing-key.tf
```diff
@@ -0,0 +1,13 @@
+resource "aws_kms_key" "storage_token_signing_key_ecc" {
+  description              = "KMS signing key (ECC) for VC storage token"
+  deletion_window_in_days  = 30
+  key_usage                = "SIGN_VERIFY"
+  customer_master_key_spec = "ECC_NIST_P256"
```
ethanmills, line 5: @mrwilson if you agree I think we should make this one RSA

ethanmills (approved):

PR #3948, opened 2024-02-13
AUT-2295:  Added abortOnAccountInterventionsErrorResponse Flag and Refactored

alhcomer:
## What?

Added abortOnAccountInterventionsErrorResponse feature flag that when turned on and AIS returns an error, the lambda aborts and returns the error response. When turned off, if AIS returns an error response, a default error response is returned. A default timeout has been added for AIS to respond within.

Also in this PR, some refactoring for improved readability.

## Why?

The purpose of this switch and associated logic is to handle errors that occur when AIS is not working as intended. 

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
@@ -648,4 +648,15 @@ public class ConfigurationService implements BaseLambdaConfiguration, AuditPubli
     public URI getAccountInterventionServiceURI() {
         return URI.create(System.getenv("ACCOUNT_INTERVENTION_SERVICE_URI"));
     }
+
+    public boolean abortOnAccountInterventionsErrorResponse() {
+        return System.getenv()
+                .getOrDefault("ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR", "false")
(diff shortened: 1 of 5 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -648,4 +648,15 @@ public String getNotifyTemplateId(String templateName) {
...
+
+    public boolean abortOnAccountInterventionsErrorResponse() {
+        return System.getenv()
+                .getOrDefault("ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR", "false")
+                .equals("true");
+    }
+
+    public long getAccountInterventionServiceCallTimeout() {
+        return Long.parseLong(
+                System.getenv().getOrDefault("ACCOUNT_INTERVENTION_SERVICE_CALL_TIMEOUT", "5000"));
```
billrdunn, line 660: The Auth implementation may be different (in which case ignore me) but Orch were initially suggested to use 5000ms but that was recently dropped to 1000ms

billrdunn (commented):
I won't approve as I haven't really been following the Auth implementation of AIS work, but generally LGTM

BeckaL (approved):

PR #3960, opened 2024-02-16
AUT-2094: Create invalid email entered lock when during re-authentication

ayoshebby:
## What?

Create invalid email entered lock when during re-authentication

- Create a new lock when user has entered their sign-in details too many times
- Creates a new prefix value in Redis `multiple-incorrect-reauth-email`

## Why?

- Prevents the user from attempting to sign in multiple times

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/ErrorResponse.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/ErrorResponse.java
@@ -67,7 +67,8 @@ public enum ErrorResponse {
     ERROR_1053(1053, "Account Interventions API Bad Gateway"),
     ERROR_1054(1054, "Account Interventions API Gateway Timeout"),
     ERROR_1055(1055, "Account Interventions API Unexpected Error"),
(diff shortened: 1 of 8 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/ErrorResponse.java
```diff
@@ -67,7 +67,8 @@ public enum ErrorResponse {
     ERROR_1053(1053, "Account Interventions API Bad Gateway"),
     ERROR_1054(1054, "Account Interventions API Gateway Timeout"),
     ERROR_1055(1055, "Account Interventions API Unexpected Error"),
-    ERROR_1056(1056, "User not found or no match");
+    ERROR_1056(1056, "User not found or no match"),
+    ERROR_1057(1057, "User not found or no match");
```
dbes-gds, line 71: Are these different?

ayoshebby: They are meant to be, didn't change the message. Will update now

dbes-gds (approved):

PR #3966, opened 2024-02-16
[AUT-2437] Remove references to localstack from terraform

whi-tw:
## What?

Remove all references to localstack from our terraform

Fix a resource that always reports a change unnecessarily (`aws_api_gateway_rest_api.di_auth_ext_api`)

## Why?

As localstack isn't used for terraform, we're creating 'lists' of resources unnecessarily. This is ugly and can be unexpected.

## How to review

1. Code review
2. Ensure that every time a `count` check for `var.use_localstack` has been removed, a `moved` block has been added.
3. Ensure that all references to localstack have actually been removed
4. deploy `main` to a dev env, then deploy this on top - check there are no changes other than `x has been moved` and the standard 'the lambda zip was rebuilt' ones.
5. repeat 4 on another dev environment

## How to deploy

Pause production, observe in Integration first.

## Who can review this?

- Not me
- Anyone with a good understanding of terraform (likely another SRE)


ethanmills:
Thanks for picking this up! Did you spot this thread? I think we were hoping to break this down into some smaller changes to reduce the risk https://gds.slack.com/archives/C02KUCT0ZAS/p1707748312185769

PR #3969, opened 2024-02-16
ATO-336: Configure JAR request requirements in the client registry

hannah-leong:
## What?

Configure JAR request requirements in the client registry

## Why?

To validate JAR requests for various clients (not just doc app)


MartinH-DI:
I have reviewed this from Adoption Pod viewpoint (self-service admin portal) and can see no issues with the code from our end. This appears to be transparent from our usage.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/OrchestrationAuthorizationService.java
```diff
@@ -222,6 +222,8 @@ public void storeState(String sessionId, State state) {
     }
 
     public boolean isJarValidationRequired(ClientRegistry client) {
-        return client.getScopes().contains(CustomScopeValue.DOC_CHECKING_APP.getValue());
+        return client.getScopes().contains(CustomScopeValue.DOC_CHECKING_APP.getValue())
+                ? client.getScopes().contains(CustomScopeValue.DOC_CHECKING_APP.getValue())
+                : client.isJarValidationRequired();
```
ethanmills, line 227: Is this different from
```suggestion
        return client.getScopes().contains(CustomScopeValue.DOC_CHECKING_APP.getValue()) || client.isJarValidationRequired();
```
?

hannah-leong: Good point, shall change that :)

ethanmills (approved):

PR #3965, opened 2024-02-16
Aut 2399/fix logging on timeouts

BeckaL:
## What?

* Adds a specific log message for when the request to the Account Interventions Service times out
* Refactors the Account Interventions Service to use the standard java http library, rather than the OAuth library (since we're not making an OAuth request to the Account Interventions service). This also makes it easier to catch the specific timeout exception and log.
* Moves all the logic to call the account interventions api to the Account Interventions service, so it's all in one place.

## Why?

We're seeing generic IOExceptions being thrown in build and staging, we think because the call to the API is timing out. It is useful for debugging purposes to specifically log when a request has timed out, since it's more difficult to know what happened with the more generic error.

PR #3977, opened 2024-02-19
AUT-2397: Add feature flags for auth account interventions

BeckaL:
Adds two feature flags:

1. ACCOUNT_INTERVENTIONS_SERVICE_ACTION_ENABLED - when false,  we still make a call to the account interventions service, but we discard the result (whether a 4xx
 or 5xx, or a successful response that may indicate interventions) and return the no interventions response.
2. ACCOUNT_INTERVENTIONS_CALL_ENABLED - when false, the account intervention handler takes no action (does not call the account interventions api) and just returns a no interventions response, so this has the effect of being a feature flag for all the behaviour.

This will allow us to validate the implement account interventions slowly, by adding a step where we can analyse the results of the interventions without having an impact on the user journey.

## 🔴  Depends on 🔴 

I've based this off the changes in [this PR](https://github.com/govuk-one-login/authentication-api/pull/3965) since they involve some refactoring of the code here, so that PR should be merged first, and this PR should then be changed to be against a base of `main` (will happen automatically when that PR merged)

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandlerTest.java
@@ -97,6 +97,7 @@ public class AccountInterventionsHandlerTest {
         UserProfile userProfile = generateUserProfile();
         when(authenticationService.getUserProfileByEmailMaybe(TEST_EMAIL_ADDRESS))
                 .thenReturn(Optional.of(userProfile));
+        when(configurationService.accountInterventionsServiceActionEnabled()).thenReturn(true);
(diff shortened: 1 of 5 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandlerTest.java
```diff
@@ -223,6 +225,21 @@ void shouldReturnErrorResponseWithGivenHttpStatusCode(
                         false, false, false));
     }
 
+    @Test
+    void shouldReturn200NotCallAccountInterventionsServiceWhenCallIsDiabled()
```
billrdunn, line 229: Typo

billrdunn (commented):
LGTM. I won't approve as I'll leave that to an Auth dev

PR #3981, opened 2024-02-20
ATO-95: Add storage token signing key alias env variable

billrdunn:
## What?

- Rename TOKEN_SIGNING_KEY_ALIAS variables to prevent ambiguity when new storage token variable is introduced. Note that the variable name in Auth shared has not been changed, as Auth are not introducing a storage token.
- Add the alias name of the storage token signing key to AuthenticationCallbackHandler as an environment variable.

## Why?

This alias is required to access the key which will be used to sign the storage token (implementation in [this PR](https://github.com/govuk-one-login/authentication-api/pull/3983))

## Testing

Tested in sandpit on auth-only and identity journeys 


orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/HandlerIntegrationTest.java
```diff
@@ -270,7 +270,7 @@ public IntegrationTestConfigurationService(
         }
 
         @Override
-        public String getTokenSigningKeyAlias() {
+        public String getIdTokenSigningKeyAlias() {
             return tokenSigningKey.getKeyAlias();
```
hannah-leong, line 274: If this is the ID token, should `tokenSigningKey` (and all the other bits that leads to...) also be named idToken?

orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/HandlerIntegrationTest.java
```diff
@@ -89,7 +89,7 @@ public abstract class HandlerIntegrationTest<Q, S> {
             new KmsKeyExtension("ipv-encryption-key", KeyUsageType.ENCRYPT_DECRYPT);
 
     @RegisterExtension
-    protected static final TokenSigningExtension tokenSigner = new TokenSigningExtension();
+    protected static final TokenSigningExtension idTokenSigner = new TokenSigningExtension();
```
hannah-leong, line 92: Is this extension for generic tokens?

hannah-leong: Sure, I was just asking in case it should be renamed too :))

hannah-leong (approved):

ethanmills:
Do we also sign the access token with this key?

hannah-leong (approved):

ethanmills (approved):

PR #3996, opened 2024-02-22
AUT-2434: decrease default retry limit for auth app codes during sign up

mark-cab:
## What?

Change retry limit default for auth app codes during sign up to 5 from 999999.

## Why?

This was noticed during 2hr lockout unit testing and as it's a production issue should be addressed separately and quickly.

original decision to effectively remove retry limit in this [ticket](https://govukverify.atlassian.net/browse/AUT-1322).

Latest decision from security is to restore the retry limit.


dbes-gds:
No changes will be required here.

PR #3997, opened 2024-02-22
AUT-2398: Raise alarm when receiving invalid AIS responses/auth AIS lambda fails

alhcomer:
## What?

- Added alarm that goes off when receiving a specified amount of 4xx or 5xx responses from AIS
- Added an alarm that goes off when the auth AIS lambda fails specified number of times. This then notifies whoever is on pager duty.

## Why?

Alarms needed to be implemented since it indicates issues with the interventions api, or our integration with it. 


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandlerTest.java
```diff
@@ -198,6 +199,8 @@ void shouldReturnErrorResponseWithGivenHttpStatusCode(
                 .thenThrow(
                         new UnsuccessfulAccountInterventionsResponseException(
                                 "Any 4xx/5xx error valid here", 404));
+        System.out.println(apiRequestEventWithEmail(TEST_EMAIL_ADDRESS));
+        System.out.println(context);
```
ayoshebby, line 203: Is this still required? same with the line above this

ayoshebby (approved):

PR #4002, opened 2024-02-23
Aut 2433/adding metadata pair journey type in audit events

VladGavrilet:
## What?

Adding metada pairs for journey type to audit events as per AUT 2433

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/MfaHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/MfaHandler.java
@@ -45,6 +45,7 @@ import static uk.gov.di.authentication.shared.entity.NotificationType.VERIFY_PHO
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateEmptySuccessApiGatewayResponse;
 import static uk.gov.di.authentication.shared.helpers.LogLineHelper.attachSessionIdToLogs;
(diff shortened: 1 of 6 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/MfaHandler.java
```diff
@@ -151,7 +152,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                                 .getClient()
                                 .map(ClientRegistry::getClientID)
                                 .orElse(AuditService.UNKNOWN),
-                        AuditService.UNKNOWN,
+                        String.valueOf(pair("journey-type", journeyType)),
```
dbes-gds, line 155: What's the behaviour if journeyType is null?  Is there a test to show that this doesn't generate a null pointer exception?

VladGavrilet: No, it will just pass the null value, it will not throw out an exception :)

dbes-gds, line 155: is this pair in the right place? pairs at the end?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/MfaHandler.java
```diff
@@ -171,7 +172,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
```
dbes-gds, line 175: in the right place?

dbes-gds (approved):

PR #4003, opened 2024-02-23
ATO-361: Add wallet subject ID

hannah-leong:
## What?

Add wallet subject ID as a scope and calculate the value for it when requested

## Why?

Wallet ID is a scope rather than a `ValidClaim` as `ValidClaim`s are those that we pass on to IPV and so, as they are not concerned by the walletSubjectID, it has been added as a scope instead.
The reasoning for this decision can be found in RFC 0065.

## Related PRs


oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AuthUserInfoClaims.java
```diff
@@ -12,6 +12,7 @@ public enum AuthUserInfoClaims {
     EMAIL_VERIFIED("email_verified"),
     PHONE_NUMBER("phone_number"),
     PHONE_VERIFIED("phone_number_verified"),
+    WALLET_SUBJECT_ID("https://vocab.account.gov.uk/v1/walletSubjectID"),
```
CarlyG55, line 15: Really picky but it might be nice to keep all the id's together

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/UserInfoService.java
```diff
@@ -224,4 +236,13 @@ private void incrementClaimIssuedCounter(String claimName, String clientID) {
                         "Claim",
                         claimName));
     }
+
+    private String calculateWalletSubjectID(AccessTokenInfo accessTokenInfo) {
+        var sectorID =
+                ClientSubjectHelper.getSectorIdentifierForClient(
+                        dynamoClientService.getClient(accessTokenInfo.getClientID()).orElseThrow(),
```
CarlyG55, line 243: Could you add a more helpful error here?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/ClientSubjectHelper.java
```diff
@@ -101,7 +102,7 @@ public static String calculatePairwiseIdentifier(
             md.update(sectorHost.getBytes(StandardCharsets.UTF_8));
             md.update(subjectID.getBytes(StandardCharsets.UTF_8));
 
-            byte[] bytes = md.digest(salt);
+            byte[] bytes = Objects.nonNull(salt) ? md.digest(salt) : md.digest();
```
CarlyG55, line 105: I don't love the idea of us changing this - I think it's a bit unclear in what cases this is null. I assume it also will change the failure state if the salt is null for some reason where it shouldn't be. I think it would be fine to basically copy most of the logic into the function below, or even commonise the logic above this point - but this feels like commonising for the sake of commonising.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/AuthUserInfoClaims.java
ethanmills: This is not an auth user info claim

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/UserInfoService.java
```diff
@@ -129,6 +129,10 @@ private void populateInfoOrchSplitEnabled(UserInfo userInfo, AccessTokenInfo acc
                     AuthUserInfoClaims.PUBLIC_SUBJECT_ID.getValue(),
                     tmpUserInfo.getClaim(AuthUserInfoClaims.PUBLIC_SUBJECT_ID.getValue()));
         }
+        if (accessTokenInfo.getScopes().contains(CustomScopeValue.WALLET_SUBJECT_ID.getValue())) {
+            var walletSubjectID = calculateWalletSubjectID(accessTokenInfo);
+            userInfo.setClaim("wallet_subject_id", walletSubjectID);
```
hannah-leong, line 138: Still a bit confusing having it set as a claim when returning it

hannah-leong:
Note: Wellknown endpoint does not include `wallet-subject-id` in the supported scopes - currently only displays the OIDC scopes (as these are the only required ones). In the future would be good to include this, however there are some custom scopes that we don't want to include so not addressing this at the moment

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/UserInfoService.java
ethanmills: Discussed the difference between claim and requested claim

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/UserInfoHandler.java
```diff
@@ -66,6 +66,7 @@ public UserInfoHandler(ConfigurationService configurationService) {
                 new UserInfoService(
                         new DynamoService(configurationService),
                         new DynamoIdentityService(configurationService),
+                        new DynamoClientService(configurationService),
```
ethanmills, line 69: Are there any terraform changes needed to give this lambda permission to use this table?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/UserInfoService.java
```diff
@@ -224,4 +236,19 @@ private void incrementClaimIssuedCounter(String claimName, String clientID) {
...
                         claimName));
     }
+
+    private String calculateWalletSubjectID(AccessTokenInfo accessTokenInfo) {
+        var client =
+                dynamoClientService
+                        .getClient(accessTokenInfo.getClientID())
+                        .orElseThrow(
+                                () ->
+                                        new UserInfoException(
```
ethanmills, line 246: Not sure this is the right exception here?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/ClientSubjectHelper.java
```diff
@@ -95,17 +95,27 @@ public static String calculatePairwiseIdentifier(String subjectID, URI uri, byte
...
+        byte[] bytes = md.digest(salt);
+        var sb = Base64Url.encode(bytes);
+        return PAIRWISE_PREFIX + sb;
+    }
+
+    public static String calculateWalletSubjectIdentifier(String sectorID, String commonSubjectID) {
+        var md = getMessageDigest(sectorID, commonSubjectID);
+        byte[] bytes = md.digest();
+        var sb = Base64Url.encode(bytes);
+        return PAIRWISE_PREFIX + sb;
```
ethanmills, line 108: Asked for confirmation we're using the same pairwise prefix: https://gds.slack.com/archives/C066NQKE95G/p1709043871964709

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/UserInfoHandler.java
hannah-leong: It has the `dynamo_client_registry_read_access_policy` attached to it which should be adequate

ethanmills (approved):

PR #4018, opened 2024-02-27
AUT-2453: Increase retry attempt count for email OPT codes during registration journey

mark-cab:
## What?

Email OTP code retry limit has been removed for registration journey. VerifyCodeHandler required updating and some minor refactoring elsewhere.

## Why?

From a review of existing OPT/MFA code attempts and retry functionality it was highlighted that there is no security need for blocking users due to incorrect email OTP code entries for registration journey only.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
@@ -145,9 +145,15 @@ public class ConfigurationService implements BaseLambdaConfiguration, AuditPubli
         return Integer.parseInt(System.getenv().getOrDefault("CODE_MAX_RETRIES", "5"));
     }
 
-    public int getCodeMaxRetriesRegistration() {
+    public int getIncreasedCodeMaxRetries() {
         return Integer.parseInt(
-                System.getenv().getOrDefault("CODE_MAX_RETRIES_REGISTRATION", "999999"));
+                System.getenv().getOrDefault("CODE_MAX_RETRIES_INCREASED", "999999"));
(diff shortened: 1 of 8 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -145,9 +145,9 @@ public int getCodeMaxRetries() {
         return Integer.parseInt(System.getenv().getOrDefault("CODE_MAX_RETRIES", "5"));
     }
 
-    public int getCodeMaxRetriesRegistration() {
+    public int getIncreasedCodeMaxRetries() {
         return Integer.parseInt(
-                System.getenv().getOrDefault("CODE_MAX_RETRIES_REGISTRATION", "999999"));
+                System.getenv().getOrDefault("CODE_MAX_RETRIES_INCREASED", "999999"));
```
dbes-gds, line 150: was this env var added in another PR?

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -145,9 +145,15 @@ public int getCodeMaxRetries() {
...
+    }
+
+    public boolean removeRetryLimitForRegistrationEmailCodeEntry() {
+        return System.getenv()
+                .getOrDefault("REMOVE_RETRY_LIMIT_REGISTRATION_EMAIL_CODE", "false")
```
dbes-gds, line 155: will need to inject the new env var into the lambda config for the lambdas that use it

PR #4037, opened 2024-02-28
BAU: Clean logging context on lambda invocation

ethanmills:
## What?

We need to call ThreadContext.clearMap(); on every invocation

## Why?

 otherwise we log identifiers from the previous request

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/2891


ethanmills:
Will put in a ticket for a base handler which does this to make this easier

PR #4039, opened 2024-02-29
ATO-98: Added C1/C2 Support To VTR

BenjaminWCO:
## What?

- Added support for C1/C2 values To VTR to supplement Cl/Cl.Cm respectively.
- Improvements to VectorOfTrust so that it can express both legacy and new values.
- Unit tests reflect that we accept arbitrary order or values in VectorsOfTrust. (This appears to be a deviation from our original spec, but we have to continue to support it going forward and the upshot is it is compliant with rfc8485)

## Why?

Cl.Cm caused confusion around whether Cm could be used a standalone value.


orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/entity/.DS_Store
ethanmills, line 1: We should not be committing these. Please feel free to add to gitignore

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/IdentId.java
ethanmills, line 1: What's the conceptual difference between IdentId and LevelOfConfidence?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/VectorOfTrust.java
```diff
@@ -0,0 +1,220 @@
...
+import static uk.gov.di.orchestration.shared.entity.vectoroftrust.VotConstants.IDENT_NONE;
+
+/** A Vector of Trust represents a combined authentication and identity trust level. */
+public class VectorOfTrust implements Comparable<VectorOfTrust> {
+
+    /** AKA "Credential Trust Level" */
+    @NotNull @Expose private final VotComponent<AuthId> authComponent;
+
+    /** AKA "Level Of Confidence" */
+    @NotNull @Expose private final VotComponent<IdentId> identComponent;
```
ethanmills, line 30: So why not call them this?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/VotComponent.java
ethanmills, line 1: It feels a little odd that this takes a type parameter rather than being abstract. Is there a reason we've chosen this form of composition over inheritance?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/IdentId.java
BenjaminWCO: My thinking was that like we have multiple VoT auth codes that have the same meaning P0 and Empty have the same meaning, so it made sense to be symmetric with AuthId.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/VectorOfTrust.java
BenjaminWCO: I thought you had mentioned they were not particularly clear names (which I also think is was true -  "LoC" is vague and I believe I've seen CTLs assigned to a variable called "confidence". Hadn't initially appreciated how widely the terms were used.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/VotComponent.java
BenjaminWCO: I suppose you could argue it would read clearer with inheritance. In this case both implementations would do effectively the same thing though and now there'd two concrete classes to unit test.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/IdentId.java
BenjaminWCO: I've changed the VoT back to using CTL and LoC enums, but with the addition of two new fields CTLCode and LoCCode which replace Ident/Auth Component. Although it's a little more code, it's conceptually a smaller change just having Ident/Auth Component to represent the raw VoT and normalising them to get something that can be used internally. Having both types make the intension of their usage clearer also. I've tried to shift all of the string conversions as late as possible to help following the logic as it's a little uncle when see something like List<String> vtr in the code, what it actually means.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/vectoroftrust/VectorOfTrust.java
BenjaminWCO: As above, I've changed the names back. Even if they are a little unclear probably best to be consistent given the widespread usage.

PR #4040, opened 2024-02-29
ATO-419: Implement Snapstart

gn-GDS:
## What?

Implement Snapstart for new lambda

## Why?

To reduce cold start time

## Related PRs

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -29,6 +29,11 @@ Globals:
   Function:
     MemorySize: 1536
     Timeout: 30
+    SnapStart:
+      ApplyOn: PublishedVersions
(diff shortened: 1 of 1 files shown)


ethanmills:
If you haven't already it would be good to read over https://docs.aws.amazon.com/lambda/latest/dg/snapstart-uniqueness.html and the linked https://github.com/aws/aws-lambda-snapstart-java-rules to make sure this is safe

ethanmills:
I think there are a couple of warmup functions that might not make sense with snapstart

ethanmills (approved):

PR #4060, opened 2024-03-04
ATO-361: Add wallet subject id to auth user info

hannah-leong:
## What?

Adding it to the AuthUserInfoClaims so that it can be part of the authenticationRequest

## Why?

Request for wallet subject id isn't going through; not showing in the stubs

## Related PRs


billrdunn (approved):

ethanmills (changes requested):
This is the wrong place

PR #4071, opened 2024-03-06
[DRAFT] Don't test so much all the time

whi-tw:
## What

Ensure that we only run tests that are relevant to the PR.

## How to review

- Code review
- Check over the yaml files, make sure that the `paths` specifiers make sense

### DIFF
--- a/.github/workflows/checkov.yml
+++ b/.github/workflows/checkov.yml
@@ -4,9 +4,15 @@ on:
   push:
     branches:
       - main
+    paths:
+      - "ci/terraform/**"
+      - "**/template.yaml"
   pull_request:
     branches:
       - main
+    paths:
+      - "ci/terraform/**"
+      - "**/template.yaml"
 
   workflow_dispatch:
 
--- a/.github/workflows/checkov.yml
+++ b/.github/workflows/checkov.yml
@@ -3,10 +3,10 @@ name: checkov
 on:
   push:
     branches:
-    - main
+      - main
   pull_request:
     branches:
-    - main
+      - main
 
   workflow_dispatch:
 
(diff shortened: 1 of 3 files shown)


ethanmills:
Do you have a plan for when the required checks don't run? I think this would make it impossible to merge a README change as we'd be waiting forever for the tests to pass?

whi-tw:
@ethanmills just saw your message as I'm looking at this again.

No, not as-yet. I'm trying to work out if there's another way to do the required checks, one that would allow for more targetted tests, while simultaneously enforcing specific checks.

Is there a reason that we've got specific tests as part of branch protection rule instead of using the standard 'PR is blocked if any check fails' behaviour?

AFAIK, the only *good* reason to have these tests is if you're using external services which only submit a 'check passed/failed' status (and not a 'check running' status) - it's possible to quickly merge the PR before that status comes in, thus skipping the check.

These checks are all replicated by the merge queue anyway, so any merged PR is forced to pass those tests anyway.

ethanmills:
> Is there a reason that we've got specific tests as part of branch protection rule instead of using the standard 'PR is blocked if any check fails' behaviour?

Do you have a docs link for this? This was not my understanding of how branch protection rules work.

> These checks are all replicated by the merge queue anyway, so any merged PR is forced to pass those tests anyway.

ethanmills:
The way I've seen this done is with `if`s on the individual steps to skip them when you don't want them run. That way the required checks still pass

whi-tw:
> The way I've seen this done is with `if`s on the individual steps to skip them when you don't want them run. That way the required checks still pass

Yeah, i've seen that too. problem is, it marks the test as passed, which seems disingenuous.

I'll have a think.

PR #4073, opened 2024-03-06
ATO-436: Add a lambda to manually delete accounts

ethanmills:
## What

This lambda is intended to me manually invoked from the AWS console, and provides a method to emit the correct audit events when actioning manual account deletions.

## How to review
Pay particular attention to the error handling I think. We always want to know the IDs of the account being deleted.
The intended use of this is that it would be invoked manually from the console using the "test" functionality. The return value can then be copied into the account deletion ticket.


ethanmills:
converting to draft while looking at auditing of which user invoked the lambda

ethanmills:
We've got a data trail in cloudtrail which logs all the information we need

isaac-GDS (approved):

PR #4076, opened 2024-03-07
ATO-96: Create new well-known endpoint serving storage token signing public key

billrdunn:
## What

Create new well-known endpoint for storage token signing public key

## Testing

Tested on sandpit and key is available at https://oidc.sandpit.account.gov.uk/.well-known/storage-token-jwk.json

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/3944


ci/terraform/oidc/storage-token-jwk.tf
```diff
@@ -0,0 +1,60 @@
...
+
+  policies_to_attach = [
+    aws_iam_policy.storage_token_kms_signing_policy.arn
+  ]
+}
+
+module "storage_token_jwk" {
+  source = "../modules/endpoint-module"
+
+  endpoint_name   = "storage-token-jwk.json"
```
ethanmills, line 15: @TaherKapasi are you good with this? Your key will be available on `oidc.(environment.)account.gov.uk/storage-token-jwk.json`

TaherKapasi (dismissed):

ci/terraform/oidc/authentication-callback.tf
ethanmills, line 1: I think this is fine because it's not in production, but it's worth bearing in mind we would otherwise need to be careful about moving resources like this. This will cause some downtime and errors in staging

ethanmills: Does the production code definitely not access this key in any way?

ethanmills (commented):
Could we have some tests for the new lambda please? Ideally unit and integration

ethanmills:
See https://sonarcloud.io/summary/new_code?id=govuk-one-login_authentication-api&pullRequest=4076

ci/terraform/oidc/authentication-callback.tf
billrdunn: The prod code doesn't access the key as the use of the key is feature flagged using `SEND_STORAGE_TOKEN_TO_IPV_ENABLED`. However this isn't a necessary change so I can revert if it's not worth the risk.

ethanmills: Nope, happy to leave this in, no need to revert. I think if this were production we would add a `moved` block so the resource wasn't deleted and recreated

ethanmills (approved):

PR #4096, opened 2024-03-11
AUT-2551-Adding redis daily Automated backup

pskushwaha1:
## What

AUT-2551 Adding redis daily Automated backup

## How to review

This can be review using  this branch and try to deploy to Authdev1 ( Shared & Account mgmt api)  , it must show below high lighted changes

  **# aws_elasticache_replication_group.sessions_store[0] will be updated in-place**
  ~ resource "aws_elasticache_replication_group" "sessions_store" {
        id                          = "authdev1-sessions-store"
```
      ~ snapshot_retention_limit    = 0 -> 5
      ~ snapshot_window             = "00:00-01:00" -> "00:00-04:00"
```

**# aws_elasticache_replication_group.account_management_sessions_store[0] will be updated in-place**
  ~ resource "aws_elasticache_replication_group" "account_management_sessions_store" {
        id                          = "authdev1-acct-mgmt-session-store"
```
      ~ snapshot_retention_limit    = 0 -> 5
      ~ snapshot_window             = "04:00-05:00" -> "00:00-04:00"
```

### DIFF
--- a/ci/terraform/account-management/redis.tf
+++ b/ci/terraform/account-management/redis.tf
@@ -35,6 +35,8 @@ resource "aws_elasticache_replication_group" "account_management_sessions_store"
   port                        = local.redis_port_number
   maintenance_window          = "tue:22:00-tue:23:00"
   notification_topic_arn      = data.aws_sns_topic.slack_events.arn
+  snapshot_retention_limit    = 5
+  snapshot_window             = "00:00-04:00"
 
   multi_az_enabled = true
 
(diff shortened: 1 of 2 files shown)


ethanmills:
Why do we want to back these up? Is there anything stored in redis with a lifetime of longer than one hour? Under what circumstances would we restore a backup? 

We would need to involve security and IA in this decision, as it increases our PII retention times

pskushwaha1:
> Member

The main reason to enable backup was save guard us  if we encountered any issue whilst we update the node from t2 small to M4 large , this activity will happen in maintenance window  Midnight on Wednesday , we will  have something to restore from , I am not aware of PII data retention time but if you point me to contact i can get some guidance on PII.

 I have set retention to 5 days on day 6 the first backup will be auto  delated , if the is is against PII retention  we can bring down to 1 or 2 days

ethanmills:
How was that decision made? I think restoring from a backup would carry more risk than just accepting the loss of data, especially at that time of day. If the backup were more than an hour old it would be functionally equivalent to no backup anyway, as all data would immediately expire

PR #4102, opened 2024-03-12
ATO-323: Integrate dynatrace

gn-GDS:
## What

- Update Sam Template
- Order parameters.json alphabetically

## How to review

See [here](https://github.com/govuk-one-login/observability-infrastructure/blob/main/lambdalayer/README.md#prerequisites) for additional info.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -25,8 +25,40 @@ Conditions:
   UseCodeSigning:
     !Not [ !Equals [ none, !Ref CodeSigningConfigArn ] ]
 
+Mappings:
+  EnvironmentConfiguration:
+    dev:
+      dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+    build:
+      dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
+    staging:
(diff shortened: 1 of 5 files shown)


template.yaml
```diff
@@ -25,9 +25,41 @@ Conditions:
...
+          - '{{resolve:secretsmanager:${SecretArn}:SecretString:DT_CLUSTER_ID}}'
+          - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
+        DT_LOG_COLLECTION_AUTH_TOKEN: !Sub
+          - '{{resolve:secretsmanager:${SecretArn}:SecretString:DT_LOG_COLLECTION_AUTH_TOKEN}}'
+          - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
+        DT_TENANT: !Sub
+          - '{{resolve:secretsmanager:${SecretArn}:SecretString:DT_TENANT}}'
+          - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
+        DT_OPEN_TELEMETRY_ENABLE_INTEGRATION: "true"
+    MemorySize: 2048
```
ethanmills, line 62: I think we should be ok with 1.5GB, not 2? That's what we're running in production I think

ethanmills (approved):

PR #4104, opened 2024-03-12
Aut 2553/pass timestamp to orch

BeckaL:
## What

Adds the ability to send a "password reset time" through in the request to the auth code handler. 

If this is present in the request to auth code, we:

* Add it to the auth code table in the Authentication Auth Code Handler
* Add it to the access token table in the Token Handler (which reads from the auth code table)
* Send this back as a claim in the user info response.

This will enable orchestration to implement logic to ignore a password reset intervention when a password has been reset more recently than the intervention was applied. This makes our system more robust, meaning we're not reliant on the async process which updates the account interventions data when a password is reset.

This is very similar to what we did for `isNewAccount`.

## How this has been tested

We've deployed this to sandpit on its own, and checked that a normal sign in journey works (we haven't broken anything existing). Then, we deployed the change to the frontend that sends this password reset timestamp through, and verified it appears in all tables and is sent back in the claim.

## How to review

1. [Optional] you can repeat our testing in sandpit, by:

- deploying this to sandpit
- doing a normal sign in journey and seeing no existing breakages
- deploy the frontend to sandpit
- give yourself a password reset intervention
- go through a sign-in journey and reset your password
- see that the relevant tables are populated 
- we also added logging to log the specific claim that we're adding in user info, this is optional

## Related PRs

* [PR in frontend ](https://github.com/govuk-one-login/authentication-frontend/pull/1449)

(The plan is to merge this PR first, and then the above frontend PR)

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
@@ -34,7 +34,8 @@ public class AccessTokenService extends BaseDynamoService<AccessTokenStore> {
             String subjectID,
             List<String> claims,
             boolean isNewAccount,
-            String sectorIdentifier) {
+            String sectorIdentifier,
+            Long passwordResetTime) {
         if (isAccessTokenStoreEnabled) {
             var tokenStore =
                     get(accessToken)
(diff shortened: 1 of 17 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
```diff
@@ -34,7 +34,8 @@ public void addAccessTokenStore(
             String subjectID,
             List<String> claims,
             boolean isNewAccount,
-            String sectorIdentifier) {
+            String sectorIdentifier,
+            Long resetPasswordTime) {
```
billrdunn, line 38: Extremely pedantic comment but stick to `passwordResetTime`?

billrdunn (approved):

dbes-gds (approved):

PR #4108, opened 2024-03-13
ATO-440: Account interventions reset password race condition

billrdunn:
## What

- Redefined account interventions class structure. Two immutable classes are defined which hold the the raw response from AIS (`AccountInterventionState` and `AccountInterventionDetails`). A further class, `AccountIntervention`, handles account interventions logic applicable to Orchestration.  The idea is that in future, a developer should only need to use `AccountIntervention` and none of the other classes in production code.
- Updated codebase to use new structure.
- Serialised more of the response from account intervention service into `AccountInterventionDetails` which includes `appliedAt` value (the time that the intervention was applied).
- Read `password_reset_time` claim from Authentication
- Implement fix for race condition: if `password_reset_time` is after `appliedAt` then ignore `passwordReset` on the intervention.

## How to review
1. Code Review - commit by commit will be easiest

## Related PRs
PR to send `password_reset_time` claim from Authentication: https://github.com/govuk-one-login/authentication-api/pull/4104

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
@@ -1,0 +1,97 @@
+package uk.gov.di.orchestration.shared.entity;
+
+public class AccountIntervention {
+
+    private final AccountInterventionDetails details;
+    private final AccountInterventionState state;
+    private final AccountInterventionStatus status;
+    private final boolean ignoreResetPassword;
+
+    public AccountIntervention(AccountInterventionState state) {
+        this.details =
+                new AccountInterventionDetails(
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        "",
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE);
+        this.state = state;
+        this.ignoreResetPassword = false;
+        this.status = deduceStatus();
+    }
+
+    public AccountIntervention(
+            AccountInterventionDetails details,
+            AccountInterventionState state,
+            Long passwordResetTime) {
+        this.details = details;
+        this.state = state;
+        this.ignoreResetPassword = details.appliedAt() < passwordResetTime;
+        this.status = deduceStatus();
+    }
+
+    public AccountIntervention(AccountInterventionState state, Long passwordResetTime) {
+        this.details =
+                new AccountInterventionDetails(
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        "",
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE);
+        this.state = state;
+        this.ignoreResetPassword = details.appliedAt() < passwordResetTime;
+        this.status = deduceStatus();
+    }
+
+    public AccountInterventionStatus getStatus() {
+        if (status == null) {
+            return deduceStatus();
+        }
+        return status;
+    }
+
+    public boolean getSuspended() {
+        return state.suspended();
+    }
+
+    public boolean getBlocked() {
+        return state.blocked();
+    }
+
+    public boolean getReproveIdentity() {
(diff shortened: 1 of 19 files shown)


BeckaL (commented):
Approving on the basis that this looks like the logic we agreed. Will let someone from your team have the final sign off :-)

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
```diff
@@ -0,0 +1,97 @@
...
+
+    private final AccountInterventionDetails details;
+    private final AccountInterventionState state;
+    private final AccountInterventionStatus status;
+    private final boolean ignoreResetPassword;
+
+    public AccountIntervention(AccountInterventionState state) {
+        this.details =
+                new AccountInterventionDetails(
+                        Long.MIN_VALUE,
```
ethanmills, line 13: Would `0L` make more sense here? Feels weird to have negative time

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
```diff
@@ -0,0 +1,97 @@
...
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        "",
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE);
+        this.state = state;
+        this.ignoreResetPassword = details.appliedAt() < passwordResetTime;
```
ethanmills, line 44: This is always true, right?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
```diff
@@ -0,0 +1,97 @@
...
+                new AccountInterventionDetails(
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        "",
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE);
+        this.state = state;
+        this.ignoreResetPassword = false;
+        this.status = deduceStatus();
```
ethanmills, line 21: Nitpick,but `calculate` feels more conventional than `deduce`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
billrdunn: I thought about that and then decided on `MIN_VALUE` as it's a bit safer.  Using `MIN_VALUE` means `ignorePasswordReset` will never be `true`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
```diff
@@ -0,0 +1,97 @@
...
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE,
+                        "",
+                        Long.MIN_VALUE,
+                        Long.MIN_VALUE);
+        this.state = state;
+        this.ignoreResetPassword = details.appliedAt() < passwordResetTime;
+        this.status = deduceStatus();
+    }
```
ethanmills, line 46: This constructor feels a little odd. Given it's only used in one test I think I'd construct the details object in the test and remove this constructor.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
billrdunn: Will remove this constructor

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccountIntervention.java
ethanmills: Nah, fine to leave

ethanmills (approved):

PR #4109, opened 2024-03-13
ATO 418 - ipv feedback contract tests

isaac-GDS:
## What

Updated changes requested from IPV and regenerated the signed signature and a method to regenerate.


ipv-api/src/test/java/uk/gov/di/authentication/ipv/contract/IpvTokenTest.java
```diff
@@ -275,9 +214,70 @@ void getIPVTokenErrorResponse(MockServer mockServer) {
...
+        assertThat(tokenResponse.indicatesSuccess(), equalTo(true));
+    }
+
     private SignResponse mockKmsReturn() {
+        byte[] signature_bytes;
+
+        try {
+            signature_bytes =
+                    ECDSA.transcodeSignatureToDER(
+                            Base64.getDecoder().decode(CLIENT_ASSERTION_SIGNATURE));
```
ethanmills, line 274: Nice, this make much more sense. At this point we probably don't need the complexity of the disabled methods, we can just use jwt.io

ethanmills (approved):

PR #4113, opened 2024-03-14
AUT-2583: Verify account locked for user not found during reauthentication

ayoshebby:
## What

- When a user attempts to reauthenticate with a different email that does not exist in the user profiles store, the user is allowed to make multiple attempts
- This gives the user a large number of tries and is also a backdoor way to find out which emails have accounts
- Verify if the account has been locked and return the appropriate error

## How to review

1. Code Review
2. Test in build only as reauthentication journey unable to test in sandpit environments


BeckaL (approved):
Looks good to me!

PR #4118, opened 2024-03-15
ATO-404: Add new client registry entries for stub using for_each

CarlyG55:
## What

We need to remove the old stub client registry entries. However we currently can't do this without causing stub downtime, as removing one entry will update them all, as they are generated using 'count'. I've a second set of entries indexed by the name of the stub, which means if we remove one from the list, it will only affect that entry (checked with a terraform plan).

### DIFF
--- a/ci/terraform/shared/outputs.tf
+++ b/ci/terraform/shared/outputs.tf
@@ -157,6 +157,18 @@ output "stub_rp_client_credentials" {
   sensitive = true
 }
 
+output "stub_relying_party_client_credentials" {
+  value = [
+    for i, rp in var.stub_rp_clients : {
+      client_name = rp.client_name
+      client_id   = random_string.stub_relying_party_client_id[rp.client_name].result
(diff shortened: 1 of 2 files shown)


ci/terraform/shared/outputs.tf
```diff
@@ -157,6 +157,18 @@ output "stub_rp_client_credentials" {
   sensitive = true
 }
 
+output "stub_relying_party_client_credentials" {
+  value = [
+    for i, rp in var.stub_rp_clients : {
+      client_name = rp.client_name
+      client_id   = random_string.stub_relying_party_client_id[rp.client_name].result
+      private_key = tls_private_key.stub_relying_party_client_private_key[rp.client_name].private_key_pem_pkcs8
```
CarlyG55, line 165: this is a different format to the other output, but it's the format we actually use

ethanmills (approved):

PR #4121, opened 2024-03-15
ATO-322: Add alerting for lambdas

hannah-leong:
## What

Add alerting for lambdas to the cloudformation template; done for OpenIdConfigurationFunction - will be copied and pasted for others as they are moved across

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -172,7 +172,7 @@ Resources:
       ActionsEnabled: true
       AlarmActions:
         - !Ref SlackEvents
-      AlarmDescription: !Sub "10 or more errors have occurred in the ${Environment} openid-configuration lambda.ACCOUNT: ${AWS::AccountId}"
+      AlarmDescription: !Sub "10 or more errors have occurred in the ${Environment} openid-configuration lambda.ACCOUNT: di-orchestration-${Environment}"
       AlarmName: !Sub ${Environment}-openid-configuration-alarm
       ComparisonOperator: GreaterThanOrEqualToThreshold
       EvaluationPeriods: 1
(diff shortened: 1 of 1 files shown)


hannah-leong:
Note: currently using `AWS::AccountId` rather than the alias as there seems to be no easy way of doing it. Will look into this on Monday

template.yaml
```diff
@@ -80,10 +80,39 @@ Globals:
         - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
 
 Resources:
+  MainKmsKey:
+    Type: AWS::KMS::Key
+    Properties:
+      Description: Key used to encrypt the Orchestration
```
ethanmills, line 86: this doesn't sound quite right

hannah-leong: :(

template.yaml
```diff
@@ -110,4 +139,85 @@ Resources:
...
+              MetricName: Errors
+              Dimensions:
+                - Name: FunctionName
+                  Value: !Ref OpenIdConfigurationFunction.Alias
+            Period: 60
+            Stat: Sum
+            Unit: Count
+
+
+  OpenIdConfigurationFunctionSlackEvents:
```
ethanmills, line 219: I think this resource name is wrong. We're only going to have one of these, not one per lambda right?

hannah-leong: Ah yes, good point, shall change that now

ethanmills:
> Note: currently using `AWS::AccountId` rather than the alias as there seems to be no easy way of doing it. Will look into this on Monday

I'd be tempted to define a mapping between account number and name if there's no easier way. I think it's quite important that we have something human readable, as that's what shows up in the text of the alert and that's where TSD start from

ethanmills (approved):
Discussed we still need to figure out how to get these into slack

PR #4124, opened 2024-03-18
BAU: Explicitly specify the API Key Source for the API Gateway

ethanmills:
## What

Manually specify the API Key Source for API Gateway. It defaults to "HEADER" when no value is specified. SSE tool depends on this. This was manually changed in one environment, and lead to some hard to find issues


hannah-leong (approved):

billrdunn (approved):

billrdunn:
Why does`HEADER` need to be set if it's there by default?
https://registry.terraform.io/providers/hashicorp/aws/5.38.0/docs/resources/api_gateway_rest_api#api_key_source

PR #4126, opened 2024-03-18
ATO-440: Replace instances of `Long.MIN_VALUE` with `0L`

billrdunn:
## What

When an account intervention should not ignore the password reset flag, this is enforced by setting `passwordResetTime` to a small number. This works because `ignorePasswordReset` is `false` when the time that the intervention was applied is greater than or equal to `passwordResetTime` (comparison of unix timestamps).

Replace instances of `Long.MIN_VALUE` with `0L`, because it doesn't make sense to set `passwordResetTime` to a negative value, and it doesn't affect functionality.

## How to review

Code review only

## Related PRs

PR where `Long.MIN_VALUE` was introduced: https://github.com/govuk-one-login/authentication-api/pull/4108


ethanmills (approved):
thanks!

PR #4131, opened 2024-03-19
ATO-359: Account Interventions contract tests

billrdunn:
## What
- Add contract tests for communication with Account Intervention Service.
- Tests cover all the possible intervention outcomes based on the data in the response from AIS.
- Tests also cover error response statuses from AIS (from the spec [here](https://github.com/govuk-one-login/account-interventions-service/blob/main/src/specs/api.yaml)), and whether we abort or assume no intervention, depending on the value of the abort flag.

## How to review

1. Code Review
2. Run the tests
3. Check the [AIS API spec](https://github.com/govuk-one-login/account-interventions-service/blob/main/src/specs/api.yaml) - are the tests missing anything?

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
@@ -1,0 +1,423 @@
+package uk.gov.di.authentication.oidc.contract;
+
+import au.com.dius.pact.consumer.MockServer;
+import au.com.dius.pact.consumer.dsl.DslPart;
+import au.com.dius.pact.consumer.dsl.PactDslWithProvider;
+import au.com.dius.pact.consumer.junit.MockServerConfig;
+import au.com.dius.pact.consumer.junit5.PactConsumerTest;
+import au.com.dius.pact.consumer.junit5.PactTestFor;
+import au.com.dius.pact.core.model.PactSpecVersion;
+import au.com.dius.pact.core.model.RequestResponsePact;
+import au.com.dius.pact.core.model.annotations.Pact;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import uk.gov.di.orchestration.audit.AuditContext;
+import uk.gov.di.orchestration.shared.entity.AccountIntervention;
+import uk.gov.di.orchestration.shared.entity.AccountInterventionStatus;
+import uk.gov.di.orchestration.shared.exceptions.AccountInterventionException;
+import uk.gov.di.orchestration.shared.services.*;
+
+import java.net.URI;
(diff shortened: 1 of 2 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,606 @@
...
+            return builder.uponReceiving(
+                            "A request that results in an error response with status " + status)
+                    .path("/v1/ais/" + INTERNAL_PAIRWISE_SUBJECT_ID)
+                    .method("GET")
+                    .willRespondWith()
+                    .status(status)
+                    .body("")
+                    .toPact();
+        }
+    }
```
ethanmills, line 605: We should definitely be unit testing these error cases, but I'm not sure how much value there is in Pact testing all of them. The contract we're forming is that we both do the correct things based on the scenario we're setting up. It's going to be really hard to write a provider test for a 502, and not very valuable. 

On the other hand, a contact that says something like "given 1234 does not appear in your database, upon receiving a request for 1234 we receive a response of false,false,false,false"

oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,606 @@
...
+            boolean blocked, boolean suspended, boolean reproveIdentity, boolean resetPassword) {
+        return newJsonBody(
+                        (body) -> {
+                            body.object(
+                                    "intervention",
+                                    (obj) -> {
+                                        obj.numberType("updatedAt", TIME_NOW);
+                                        obj.numberType("appliedAt", TIME_NOW);
+                                        obj.numberType("sentAt", TIME_NOW);
+                                        obj.stringValue("description", "TEST_DESCRIPTION");
```
ethanmills, line 558: Should this be `stringType`?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
```diff
@@ -0,0 +1,606 @@
...
+                                        obj.numberType("reprovedIdentityAt", TIME_NOW);
+                                        obj.numberType("resetPasswordAt", TIME_NOW);
+                                    });
+                            body.object(
+                                    "state",
+                                    (obj) -> {
+                                        obj.booleanType("blocked", blocked);
+                                        obj.booleanType("suspended", suspended);
+                                        obj.booleanType("reproveIdentity", reproveIdentity);
+                                        obj.booleanType("resetPassword", resetPassword);
```
ethanmills, line 568: Do these need to be `booleanValue`?

ethanmills (approved):

PR #4143, opened 2024-03-21
ATO-98: Added JsonUpdateHelper class to avoid overwriting ClientSession fields in Redis

BenjaminWCO:
## What
- ClientSessionService now updates new / altered ClientSession fields without deleting existing fields if they are not maintained in both Orchestration and Authentication versions of ClientSession

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/4039

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSessionService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSessionService.java
@@ -1,10 +1,12 @@
 package uk.gov.di.orchestration.shared.services;
 
+import com.google.gson.JsonSyntaxException;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.orchestration.shared.entity.ClientSession;
 import uk.gov.di.orchestration.shared.entity.VectorOfTrust;
 import uk.gov.di.orchestration.shared.helpers.IdGenerator;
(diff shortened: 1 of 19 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSessionService.java
```diff
@@ -94,22 +97,33 @@ public Optional<ClientSession> getClientSession(String clientSessionId) {
...
+
         try {
-            redisConnectionService.saveWithExpiry(
-                    CLIENT_SESSION_PREFIX.concat(clientSessionId),
-                    objectMapper.writeValueAsString(clientSession),
-                    configurationService.getSessionExpiry());
-        } catch (JsonException e) {
+            var oldClientSession = redisConnectionService.getValue(clientSessionKey);
+            var newClientSession = objectMapper.writeValueAsString(clientSession);
+            var updatedClientSession =
```
ethanmills, line 116: This is going to have a big performance hit, and lead to some nasty race conditions, right?

BenjaminWCO: I don't know how bing the performance hit would be (we could add a local cache for the raw json string). Strictly speaking best practice would be to use a distributed lock, but I thought it was unnecessary. If we ever were somehow updating a client session in two places on a journey, the way we were updating is already problematic.

dbes-gds (approved):

PR #4145, opened 2024-03-21
ATO-452: Instrument AIS calls

ethanmills:
## What
Add instrumentation of response time for AIS calls. For use in monitoring dashboard during go-live

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
@@ -180,12 +180,17 @@ public class AccountInterventionService {
 
     private HttpResponse<String> sendRequestToAis(HttpRequest request) {
         HttpResponse<String> httpResponse = null;
+        var start = System.nanoTime();
         try {
             httpResponse = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
 
         } catch (IOException | InterruptedException e) {
             logAndThrowAccountInterventionException(
                     "Failed to send request to Account Intervention Service.");
+        }
+        var durationMs = (System.nanoTime() - start) / 1_000_000L;
+        if (httpResponse != null) {
+            instrumentResponse(durationMs, Integer.toString(httpResponse.statusCode()));
         }
         validateResponse(httpResponse);
         return httpResponse;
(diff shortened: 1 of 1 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -218,6 +223,11 @@ private AccountInterventionResponse serializeResponse(HttpResponse<String> httpR
         return accountInterventionResponse;
     }
 
+    private void instrumentResponse(Double duration, String status) {
+        cloudwatchMetricsService.putEmbeddedValue(
+                "AISTiming", duration, Map.of("statusCode", status));
```
billrdunn, line 228: maybe "AISResponseTime"?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -218,6 +223,11 @@ private AccountInterventionResponse serializeResponse(HttpResponse<String> httpR
         return accountInterventionResponse;
     }
 
+    private void instrumentResponse(Double duration, String status) {
```
billrdunn, line 226: Should this be `double duration`?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -180,13 +180,18 @@ private AccountIntervention retrieveAccountIntervention(
...
         HttpResponse<String> httpResponse = null;
+        var start = System.nanoTime();
         try {
             httpResponse = httpClient.send(request, HttpResponse.BodyHandlers.ofString());
 
         } catch (IOException | InterruptedException e) {
             logAndThrowAccountInterventionException(
                     "Failed to send request to Account Intervention Service.");
         }
+        var durationMs = (System.nanoTime() - start) / 1000L;
```
billrdunn, line 191: Use `currentTimeMillis()` instead of converting?

billrdunn: Ah I think this will end up in microseconds not milliseconds

billrdunn: was that the intention?

ethanmills: See https://stackoverflow.com/questions/1770010/how-do-i-measure-time-elapsed-in-java/1776053#1776053

billrdunn (approved):

PR #4161, opened 2024-03-26
BAU: fix orchestration CF stack reference

isaac-GDS:
## What

orch frontend stack name has changed so the reference in terraform needs to be updated.


ethanmills (approved):

isaac-GDS:
Same change raised - https://github.com/govuk-one-login/authentication-api/pull/4160#pullrequestreview-1960907062

PR #4174, opened 2024-03-28
ATO-438: Publishing and verifying contracts with GHA

hannah-leong:

### DIFF
--- a/.github/workflows/contract-tests.yml
+++ b/.github/workflows/contract-tests.yml
@@ -9,7 +9,15 @@ env:
   PACT_BROKER_SOURCE_HEADER: ${{ secrets.PACT_BROKER_SOURCE_HEADER }}
 
 on:
-  workflow_dispatch:
+  pull_request:
+    types:
+      - opened
+      - reopened
+      - ready_for_review
+      - synchronize
+  merge_group:
+    # merge main
+    branches: [ main ]
 
 jobs:
   contract-testing:
@@ -17,14 +25,20 @@ jobs:
     steps:
       - name: Check out repository code
         uses: actions/checkout@v4
+
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
           cache: 'gradle'
+
       - name: Run contract tests
-        run: ./gradlew --parallel ipv-api:contractTest
+        run: ./gradlew contractTest
+
       - name: Upload pacts to broker
         if: ${{ github.actor != 'dependabot[bot]' && github.event_name != 'merge_group' }}
-        run: ./gradlew :ipv-api:pactPublish
+        run: ./gradlew pactPublish
+
+      - name: Verify pacts
+        run: ./gradlew pactProviderTests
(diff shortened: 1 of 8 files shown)


.github/workflows/contract-tests.yml
```diff
@@ -17,14 +18,20 @@ jobs:
...
       - name: Set up JDK 17
         uses: actions/setup-java@v4
         with:
           java-version: '17'
           distribution: 'temurin'
           cache: 'gradle'
+
       - name: Run contract tests
-        run: ./gradlew --parallel ipv-api:contractTest
+        run: ./gradlew contractTest
```
ethanmills, line 30: Does this just do the consumer tests?

hannah-leong:
Testing:
- Have run the pipeline to check the consumer and provider test jobs run
- Checked the contracts are uploaded (consumer)
- Manually uploaded an example contract to dev broker and verified against that (provider)

integration-tests/build.gradle
```diff
@@ -39,6 +40,61 @@ dependencies {
...
+
+pactPublish.enabled = false
+
+task pactProviderTests (type: Test, group: "verification") {
+    useJUnitPlatform()
+    include "uk/gov/di/authentication/contract/**"
+    environment "PROVIDER_UNDER_TEST", "ClientRegistryProvider"
 
     jacoco {
     }
```
ethanmills, line 100: Do we need any of this jacoco stuff in the pact task? Are we expecting to get coverage data on this? If so I think it won't be reported to sonar because of the way the github actions are structured

oidc-api/build.gradle
```diff
@@ -42,10 +44,25 @@ test {
...
     useJUnitPlatform()
     include '/uk/gov/di/authentication/oidc/contract/**'
-    systemProperties['pact.rootDir'] = "$rootDir/oidc-api/src/test/pact"
+    systemProperties['pact.rootDir'] = pactDir
+}
+
+pact {
+    publish {
+        pactDirectory = pactDir
+        pactBrokerUrl = "$System.env.PACT_URL".toString() + "?testSource=" +"$System.env.PACT_BROKER_SOURCE_HEADER".toString()
```
DanCorderIPV, line 58: This can be simplified a little: https://github.com/govuk-one-login/ipv-core-back/pull/1802/files#diff-49a96e7eea8a94af862798a45174e6ac43eb4f8b4bd40759b5da63ba31ec3ef7

oidc-api/build.gradle
```diff
@@ -42,10 +44,25 @@ test {
...
+        pactBrokerUsername = "$System.env.PACT_USER".toString()
+        pactBrokerPassword = "$System.env.PACT_PASSWORD".toString()
+        consumerVersion = "$System.env.CONSUMER_APP_VERSION".toString()
+        tags = [
```
DanCorderIPV, line 62: For IPV core we're putting the git branch into the `consumerBranch` property instead. We don't have confirmation yet from Lime that they can use that property yet when pulling down pacts though (they are going to try to get that to work).

ethanmills (approved):

PR #4184, opened 2024-04-02
AUTH-2643 log Subscription for API execution logs

pskushwaha1:
## What

Log subscription for Auth ext API & test service API  (api Gateway )  not configured


pskushwaha1:
Auth ext API Loggroup Already exist in staging & prod  ( created outside of Terraform seems) 

 so creating with same name will fail hence creating a new log group with Name API-Gateway-Execution-Logs**_New_**  , once this log group is created will delete the old one

dbes-gds (approved):

PR #4192, opened 2024-04-03
ATO-410: Remove Usage SUPPORT_AUTH_ORCH_SPLIT - Authorize

BenjaminWCO:
Remove usage of SUPPORT_AUTH_ORCH_SPLIT in AuthorisationHandler
- Additionally fixed ID Token hint extract VoT bug - orchestrationAuthorizationService.getVtrList(authRequest) never returns null


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -373,7 +375,7 @@ private static String getRpSid(AuthenticationRequest authRequest) {
     private List<VectorOfTrust> getVtrList(
             boolean reauthRequested, AuthenticationRequest authRequest)
             throws java.text.ParseException {
-        if (reauthRequested && orchestrationAuthorizationService.getVtrList(authRequest) == null) {
+        if (reauthRequested && isNull(authRequest.getCustomParameter(VTR_PARAM))) {
```
ethanmills, line 378: Why do you prefer `isNull`, out of interest?

ethanmills (commented):
We should set the default value on the configuration service to be true before removing the env var I think, in case we've missed a service way down in the depths somewhere that checks the value

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
BenjaminWCO: I don't have a strong preference. I've seen both and try to be consistent. I guess maybe I gravitated towards it because C# which I'm more familiar with has the syntax `some_var is null`. I'm going to assume you have a different preference give you asked so will use == from now on.

PR #4202, opened 2024-04-04
AUT-1709: cleanup isNotifyTemplatePerLanguage feature flag

mark-cab:
## What

cleanup of unused feature flag **isNotifyTemplatePerLanguage**. There are 2 commits. The first removes the dependency on the feature flag. The second is subsequent cleanup of language dependent template ids that we may or may not want to keep for additional language support in future.

## How to review

Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
@@ -45,14 +45,7 @@ public enum NotificationType implements TemplateAware {
     @Override
     public String getTemplateId(
             SupportedLanguage language, ConfigurationService configurationService) {
-        String templateId = configurationService.getNotifyTemplateId(getTemplateName(language));
-        if (!configurationService.isNotifyTemplatePerLanguage()
-                || templateId == null
-                || templateId.length() == 0) {
-            return configurationService.getNotifyTemplateId(templateName);
-        } else {
-            return templateId;
-        }
+        return configurationService.getNotifyTemplateId(templateName);
     }
 
     String getTemplateName(SupportedLanguage language) {
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
@@ -43,8 +43,7 @@ public enum NotificationType implements TemplateAware {
     }
 
     @Override
-    public String getTemplateId(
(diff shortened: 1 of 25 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
```diff
@@ -45,14 +45,7 @@ public enum NotificationType implements TemplateAware {
...
             SupportedLanguage language, ConfigurationService configurationService) {
-        String templateId = configurationService.getNotifyTemplateId(getTemplateName(language));
-        if (!configurationService.isNotifyTemplatePerLanguage()
-                || templateId == null
-                || templateId.length() == 0) {
-            return configurationService.getNotifyTemplateId(templateName);
-        } else {
-            return templateId;
-        }
+        return configurationService.getNotifyTemplateId(templateName);
```
BeckaL, line 47: Should this be `.getNotifyTemplateId(language)`?

account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
```diff
@@ -45,14 +45,7 @@ public enum NotificationType implements TemplateAware {
     @Override
     public String getTemplateId(
             SupportedLanguage language, ConfigurationService configurationService) {
-        String templateId = configurationService.getNotifyTemplateId(getTemplateName(language));
-        if (!configurationService.isNotifyTemplatePerLanguage()
-                || templateId == null
```
BeckaL, line 50: We're losing these two checks (for templateId == null and templateId.length() == 0), are we sure we don't need them?

mark-cab: configurationService.isNotifyTemplatePerLanguage() was always false so it never hit this code anyway. Also this is part of the second commit which separates templateId and language so dependency is only on templateName now.

account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/NotificationType.java
BeckaL: Ah sorry, I'd missed that the feature flag was off in all environments, not on

BeckaL (approved):
Sorry, comments were based on me missing that the feature flag was off, not on. LGTM

LazarAlexandru-Constantin:
@mark-cab branch has conflicts which need to be solved

BeckaL (approved):

PR #4224, opened 2024-04-08
ATO-508: Log at ERROR level on error response from SPOT

ethanmills:
This will enable SPOT failures to be caught by our general alerting


billrdunn:
Do lines 120 - 131 mean that we never reach `processingStatus = ProcessingIdentityStatus.ERROR` ?

ethanmills:
Yep @billrdunn the case we're interested in here is PROCESSING -> gone (ERROR)

PR #4225, opened 2024-04-08
Ato 472/dynamodb cross account access

hannah-leong:
## What

Added cross-account access to the client registry dynamodb table

## How to review

Code review
I shall be testing it against sandpit


hannah-leong:
Note: Upgraded terraform to `5.43.0`. [There is another PR to update it to the latest `5.44.0`](https://github.com/govuk-one-login/authentication-api/pull/4227), which will have any comments about changes that may have an effect

ethanmills (approved):

PR #4266, opened 2024-04-12
ATO-514: Set up orch back channel request lambda

CarlyG55:
## What

Set up the back channel logout lambda in orch account

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -84,6 +84,8 @@ Globals:
         DT_OPEN_TELEMETRY_ENABLE_INTEGRATION: "true"
     MemorySize: 1536
     Timeout: 30
+    SnapStart:
+      ApplyOn: PublishedVersions
     Runtime: java17
     Architectures:
       - x86_64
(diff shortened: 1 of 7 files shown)


template.yaml
```diff
@@ -31,22 +31,37 @@ Mappings:
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
       authApiId: "aqyf5d32s2"
       authAccountId: "761723964695"
+      serviceDomain: dev.account.gov.uk
```
ethanmills, line 34:
```suggestion
      serviceDomain: sandpit.account.gov.uk
```

template.yaml
```diff
@@ -71,8 +86,6 @@ Globals:
         DT_OPEN_TELEMETRY_ENABLE_INTEGRATION: "true"
     MemorySize: 1536
     Timeout: 30
-    SnapStart:
-      ApplyOn: PublishedVersions
```
ethanmills, line 75: I think we still want this

template.yaml
```diff
@@ -88,6 +101,11 @@ Globals:
       - !Sub
         - '{{resolve:secretsmanager:${SecretArn}:SecretString:JAVA_LAYER}}'
         - SecretArn: !FindInMap [ EnvironmentConfiguration, !Ref Environment, dynatraceSecretArn ]
+    VpcConfig:
+      SubnetIds:
+        - Fn::ImportValue: !Sub ${VpcStackName}-PrivateSubnetIdA
+        - Fn::ImportValue: !Sub ${VpcStackName}-PrivateSubnetIdB
+        - Fn::ImportValue: !Sub ${VpcStackName}-PrivateSubnetIdC
```
ethanmills, line 108: Don't think this works without the security group

ethanmills (approved):

PR #4288, opened 2024-04-15
ATO-414: Remove SUPPORT_AUTH_ORCH_SPLIT Config

BenjaminWCO:
## What

Removed SUPPORT_AUTH_ORCH_SPLIT Flag config.

## How to review
Code Review
## Checklist


shared-test/src/main/java/uk/gov/di/authentication/sharedtest/basetest/HandlerIntegrationTest.java
```diff
@@ -187,13 +187,7 @@ public String getTxmaAuditQueueUrl() {
...
                     spotQueue,
                     docAppPrivateKeyJwtSigner,
-                    configurationParameters) {
-
-                @Override
-                public boolean isAuthOrchSplitEnabled() {
-                    return true;
-                }
-            };
+                    configurationParameters) {};
```
isaac-GDS, line 190: Can this all be removed if we are just using `TEST_CONFIGURATION_SERVICE` for the the integration test ?

isaac-GDS (dismissed):

isaac-GDS:
Is it possible to rebase onto main instead of merging it. Revert the latest `merge into main commit`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccessTokenService.java
```diff
@@ -1,85 +0,0 @@
-package uk.gov.di.orchestration.shared.services;
```
isaac-GDS, line 1: Wanted to double check that this is correct - should orch and auth both have this service. Or does only auth components use this service ?

BenjaminWCO: Yeah, it's unused. If you take a look at it in InteliJ, the only reference to any of the contractors of that class is from  within another one of it's contractors. I had to modify Auth's copy, but thought it made more sense to just delete ours.

isaac-GDS (approved):

PR #4290, opened 2024-04-15
ATO-430: Add IP restriction for orchestration frontend non-production environments 

billrdunn:
### Summary
Add an authorizer to the orchestration frontend route of the di-authentication-api API Gateway. The authorizer permits all IP addresses in production and integration, but only permits specific IPs in lower environments (sandpit, build, and staging).

The authorizer is configured to cache its authorization result differently in production/integration vs lower environments. In the former, caching is based on `apiId` with a TTL of 1 hour, which means that the cached (permitted) result will be used for nearly all requests, as `apiId` will be the same for all requests in a particular environment. This configuration means that there isn't a significant performance drop by using an authorizer in these environments. 

In lower environments, where performance is not as much of an issue and authorization depends on IP address, caching is based on `sourceIp` with a TTL of 1 minute. 

### API Gateway Lambda authorization flow
1. API Gateway`/orch-frontend` resource is called 
2. API Gateway invokes Lambda authorizer
3. Lambda authenticates based on the caller's IP address
4. Lambda returns an "Allow" or "Deny" IAM policy 
5. If "Allow", `/orch-frontend` is invoked by API Gateway
    If "Deny", API Gateway responds with a 4xx status code
6. The result (policy) of the lambda authorizer is cached (see above)

### Testing
The authorizer lambda has been tested in sandpit independently of API Gateway by sending and test events containing 1) an allowed IP and 2) a non-allowed IP. The lambda successfully returns an "Allow" or "Deny" policy respectively (only "Deny" shown):
image

The integration of the authorizer with API Gateway has also been tested:
image

## How to review / questions for reviewer

1. Code review 
(easier to review commits separately as the first commit is just housekeeping)
2. How can the full authorization flow be tested?
3. Is `orch_frontend_authorizer_invoke_permission.source_arn` too broad? Or OK as the authorizer is only attached to `orch_frontend_proxy_method`?
4. Are the checkov errors relevant?

## Checklist


billrdunn:
Closed as superseded by https://github.com/govuk-one-login/authentication-api/pull/4336

PR #4303, opened 2024-04-18
ATO-545: Allow cross-account access for auth callback userinfo table

hannah-leong:
## What

CMK is already set up for this
- Add the necessary policies to allow for cross-account access
- Slight reordering of flags in shared variables
- 
## Checklist

### DIFF
--- a/ci/terraform/shared/kms.tf
+++ b/ci/terraform/shared/kms.tf
@@ -532,27 +532,63 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
   key_usage                = "ENCRYPT_DECRYPT"
   customer_master_key_spec = "SYMMETRIC_DEFAULT"
   enable_key_rotation      = true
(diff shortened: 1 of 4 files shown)


ci/terraform/shared/kms.tf
```diff
@@ -532,29 +532,65 @@ resource "aws_kms_key" "authentication_callback_userinfo_encryption_key" {
...
+    ]
+    resources = ["*"]
+
+    principals {
+      type        = "AWS"
+      identifiers = [var.orchestration_account_id]
+    }
+  }
+}
+
```
CarlyG55, line 593: In my PR I generalised the ones you made of these for client registry, because the only specific thing is the sid. If you've got thoughts on that lmk 🙂

hannah-leong: Looks good to me, I copy them over :)

CarlyG55 (approved):

PR #4307, opened 2024-04-18
ATO-546: Allow cross-account access for identity credentials table

hannah-leong:
## What

Added policies etc to allow for cross-account access to the identity credentials table

## Checklist

### DIFF
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -1,5 +1,6 @@
 locals {
-  client_registry_kms_key_arn = data.terraform_remote_state.shared.outputs.client_registry_kms_key_arn
+  client_registry_kms_key_arn      = data.terraform_remote_state.shared.outputs.client_registry_kms_key_arn
+  identity_credentials_kms_key_arn = data.terraform_remote_state.shared.outputs.identity_credentials_kms_key_arn
 }
 
 data "aws_dynamodb_table" "user_credentials_table" {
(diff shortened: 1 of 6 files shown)


ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -1,5 +1,6 @@
 locals {
-  client_registry_kms_key_arn = data.terraform_remote_state.shared.outputs.client_registry_kms_key_arn
+  client_registry_kms_key_arn      = data.terraform_remote_state.shared.outputs.client_registry_kms_key_arn
+  identity_credentials_kms_key_arn = data.terraform_remote_state.shared.outputs.identity_credentials_kms_key_arn
```
CarlyG55, line 3: would be nice to pull these into the shared.tf file with other locals (I've moved in my PR)

ci/terraform/shared/dynamodb.tf
```diff
@@ -542,6 +542,12 @@ resource "aws_dynamodb_resource_policy" "client_registry_table_policy" {
   policy       = data.aws_iam_policy_document.cross_account_table_resource_policy_document.json
 }
 
+resource "aws_dynamodb_resource_policy" "identity_credentials_table_policy" {
+  count        = var.identity_credentials_cross_account_access_enabled ? 1 : 0
+  resource_arn = aws_dynamodb_table.identity_credentials_table.arn
+  policy       = data.aws_iam_policy_document.cross_account_table_resource_policy_document.json
```
CarlyG55, line 548: this policy doc doesn't allow deletion, and one of the policies in the file enough is a deletion policy so a lambda will probably need that

hannah-leong: If adding `dynamodb:DeleteItem` to the cross-account policy, then will apply to all tables. As we won't give all lambdas permission to delete, this should be okay but shall have a look into it next week to see if it is possible to just apply it to the identity credentials

CarlyG55: Yeah agreed that maybe in this case we want a different policy

CarlyG55 (approved):

CarlyG55 (approved):

PR #4319, opened 2024-04-19
RFF-2: Refactor submitting audit messages

mrwilson:
This PR refactors the `submitAuditMessage` method to take fewer parameters, so we avoid large callsites in the application code.

PR #4320, opened 2024-04-19
ATO-none: use correct persistent session id

isaac-GDS:
Used a persistent session id that passed the regex checks in `PersistentIdHelper`


ethanmills (approved):
Thanks!

mrwilson:
Thanks for fixing this! Two things pop into my mind:
1. Is this pattern affecting any other tests in the repo?
2. Should we have some helpers to create well-formed identifiers?

We have `IdGenerator.generate()` in `orchestration-shared` which might be good enough. If not we can create a deliberate test-only helper

isaac-GDS:
@mrwilson Some good points. Will have a look at this outside of the PR to get https://github.com/govuk-one-login/authentication-api/pull/4319 merged in.

PR #4321, opened 2024-04-19
ATO-513: Template and script to deploy Cloudfront WAF

ethanmills:
## What

This copies the configuration of the WAF applied to the OIDC API Gateway into a cloudfront WAF in us-east-1


gn-GDS:
Might make more sense to put this under the manual stacks directory? i.e
`ci/stack-orchestration/manual-stacks/waf`

ci/cloudfront-orchestration/waf/template.yaml
```diff
@@ -0,0 +1,184 @@
...
+          Priority: 1
+          Statement:
+            RateBasedStatement:
+              Limit: !If [IsStaging, 600000, 3600]
+              AggregateKeyType: "IP"
+          VisibilityConfig:
+            CloudWatchMetricsEnabled: true
+            MetricName: !Sub "${Environment}OidcWafMaxRequestRate"
+            SampledRequestsEnabled: true
+        - Name: !Sub "${Environment}-oic-common-rule-set"
```
CarlyG55, line 40: typo 🙂

ci/cloudfront-orchestration/waf/template.yaml
```diff
@@ -0,0 +1,184 @@
...
+                    Count: {}
+                - Name: "SizeRestrictions_QUERYSTRING"
+                  ActionToUse:
+                    Count: {}
+                - !If
+                  - IsProduction
+                  - Name: "EC2MetaDataSSRF_BODY"
+                    ActionToUse:
+                      Count: {}
+                  - !Ref AWS::NoValue
```
CarlyG55, line 69: Is this not the opposite logic to the terraform? Also would be good to do testing around this as we've had issues around it in the past

ci/cloudfront-orchestration/waf/template.yaml
```diff
@@ -0,0 +1,184 @@
...
+                          SearchString: !Ref SmokeTestClientId
+                          TextTransformations:
+                            - Priority: 0
+                              Type: "NONE"
+            VisibilityConfig:
+              CloudWatchMetricsEnabled: true
+              MetricName: "ProductionSmokeTestClientExceptionRule"
+              SampledRequestsEnabled: true
+          - !Ref AWS::NoValue
+        - Name: !Sub "${Environment}-default-query-param-limit"
```
CarlyG55, line 114: I don't know if you're bothered, but in the terraform this has underscores instead of dashes & doesn't have an environment (but yours looks better!)

ethanmills: Yeah, I went for consistency with the others rather than keeping it the same. These are going to end up in a different region, so metrics will change anyway

ci/cloudfront-orchestration/waf/template.yaml
```diff
@@ -0,0 +1,184 @@
...
+      VisibilityConfig:
+        CloudWatchMetricsEnabled: true
+        MetricName: !Sub "${Environment}OidcWafRules"
+        SampledRequestsEnabled: true
+
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub "aws-waf-logs-cf-oidc-${Environment}"
+      RetentionInDays: 30
```
CarlyG55, line 169: Should we provide a key for encryption?

ci/cloudfront-orchestration/waf/template.yaml
ethanmills: I need to take a look at this. The logic in terraform makes no sense because `(NOT prod) OR (NOT sandpit)` is always true...

ethanmills:
> I assume we will create the association in the distribution config?
Yep: https://github.com/govuk-one-login/devplatform-deploy/tree/main/cloudfront-distribution#:~:text=of%20the%20Origin-,CloudfrontWafAcl,-String

ci/cloudfront-orchestration/waf/template.yaml
CarlyG55: That's a very good point and maybe why we've had issues in the past changing it...

ethanmills:
> Might make more sense to put this under the manual stacks directory? i.e `ci/stack-orchestration/manual-stacks/waf`

I'm going to keep it separate for now because those stacks are typically deployed into a different set of accounts (the orchestration accounts, these are going in to where the domains point now)

ci/cloudfront-orchestration/waf/template.yaml
```diff
@@ -0,0 +1,212 @@
...
+            Condition:
+              ArnEquals:
+                "kms:EncryptionContext:aws:logs:arn": !Sub "arn:aws:logs:${AWS::Region}:${AWS::AccountId}:log-group:aws-waf-logs-cf-oidc-${Environment}"
+
+
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Sub "aws-waf-logs-cf-oidc-${Environment}"
+      RetentionInDays: 30
```
hannah-leong, line 196: Is there a reason why `${Environment}` is at the end of the name here but for other names it is at the beginning?

hannah-leong (approved):

ci/cloudfront-orchestration/waf/template.yaml
ethanmills: WAF log groups must start with `aws-waf-logs` because of some silly amazon rules

ethanmills: https://github.com/govuk-one-login/authentication-api/pull/4321#discussion_r1576018223

PR #4324, opened 2024-04-22
ATO-554: Create read policy for SPoT queue

CarlyG55:
## What

Create read policy to add to lambda for spot queue

PR #4328, opened 2024-04-23
ATO-554 Remove read permission from spot queue from resource policy 

CarlyG55:
## What

We should move to identity policies for access to our queues. I have set up identity based access to the spot queue in the previous PR https://github.com/govuk-one-login/authentication-api/pull/4324 and this PR will start the rollout of removing the resource policies.

## How to review

Check it's feature flagged, and the logic is the correct way around


ethanmills (approved):
Can we have followup ticket to clean up the flag and old policy please? Or are you planning on doing it all as part of this ticket?

PR #4331, opened 2024-04-23
ATO-516: Bypass oidc origin rate limiting for requests containing cloudfront header

ethanmills:
Once traffic is routed through cloudfront, all requests will be coming from a small pool of cloudfront IPs. It does not make sense to rate limit based on them. We are ratelimiting with a WAF at edge. This covers the origin for the transition to cloudfront.

I've set up deployment secrets in all environments for this


ethanmills:
Draft until devplatform update the header name

ethanmills:
Will mark as ready once https://github.com/govuk-one-login/devplatform-deploy/pull/1733 is in and I've updated the sandpit stack

PR #4333, opened 2024-04-23
Bau/refactor start handler

BeckaL:
## What

Converts response types used in the start handler from classes to records to take advantage of the reduced boilerplate as introduced in more recent java versions. This may also allow for future test refactorings, as it's easier to check equality across two equivalent records.

## How to review


alhcomer (approved):
LGTM

PR #4336, opened 2024-04-24
ATO-430: Add orch frontend authorizer for IP restriction

billrdunn:
### Summary
Add an authorizer to the orchestration frontend route of the di-authentication-api API Gateway. The authorizer permits all IP addresses in production and integration, but only permits specific IPs in lower environments (sandpit, build, and staging).

The authorizer is configured to cache its authorization result differently in production/integration vs lower environments. In the former, caching is based on `apiId` with a TTL of 1 hour, which means that the cached (permitted) result will be used for nearly all requests, as `apiId` will be the same for all requests in a particular environment. This configuration means that there isn't a significant performance drop by using an authorizer in these environments. 

In lower environments, where performance is not as much of an issue and authorization depends on IP address, caching is based on `sourceIp` with a TTL of 1 minute. 

### API Gateway Lambda authorization flow
1. API Gateway`/orch-frontend` resource is called 
2. API Gateway invokes Lambda authorizer
3. Lambda authenticates based on the caller's IP address
4. Lambda returns an "Allow" IAM policy or throws an exception
5. If "Allow", `/orch-frontend` is invoked by API Gateway
    If exception, API Gateway responds with unauthorized response
6. The result (policy) of the lambda authorizer is cached (see above)

### Testing in sandpit
- The authorizer lambda has been tested in sandpit independently of API Gateway by sending test events containing an allowed IP and a non-allowed IP. The lambda successfully returns an Allow policy or throws an exception respectively.
- The integration of the authorizer with API Gateway has also been tested via the authorizers test feature in the AWS console and functions correctly.
- Once deployed to sandpit, [oidc.sandpit.account.gov.uk/orch-frontend/error](https://oidc.sandpit.account.gov.uk/orch-frontend/error) is available from allows IPs (eg. while on the VPN), but not others.
- Once changes are in production it should be confirmed that caching is set up correctly and /orch-frontend is accessible from all IPs.

## How to review

1. Code review (easier to review commit-by-commit)

## Questions
- How can the full range of allowed IPs be tested? Is this even necessary since CloudFront will soon be in front of API Gateway?

## Checklist


ci/terraform/oidc/authorizer-orch-frontend.tf
```diff
@@ -0,0 +1,199 @@
...
+  runtime       = "java17"
+
+  s3_bucket         = aws_s3_bucket.source_bucket.bucket
+  s3_key            = aws_s3_object.oidc_api_release_zip.key
+  s3_object_version = aws_s3_object.oidc_api_release_zip.version_id
+
+  code_signing_config_arn = local.lambda_code_signing_configuration_arn
+
+  publish     = true
+  timeout     = 900
```
isaac-GDS, line 31: Where was this value taken from ? The previous authorizer was set to 30.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/OrchFrontendAuthorizerHandlerTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @BeforeEach
+    public void setUp() {
+        handler = new OrchFrontendAuthorizerHandler(configurationService);
+        when(configurationService.getEnvironment()).thenReturn("test-environment");
+        when(context.getAwsRequestId()).thenReturn(REQUEST_ID);
+    }
+
+    @Test
+    public void shouldReturnAuthPolicyForAllowedIpAddress() {
```
isaac-GDS, line 34: Do we need equivalent tests for production and integration

ci/terraform/oidc/authorizer-orch-frontend.tf
billrdunn: Good spot have set to 30

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/OrchFrontendAuthorizerHandlerTest.java
```diff
@@ -0,0 +1,87 @@
...
+        when(configurationService.getEnvironment()).thenReturn(NON_PROD_ENV);
+        APIGatewayProxyRequestEvent event = new APIGatewayProxyRequestEvent();
+        event.setRequestContext(contextWithSourceIp(TRUSTED_IP));
+        event.setResource(RESOURCE);
+
+        AuthPolicy authPolicy = handler.handleRequest(event, context);
```
isaac-GDS, line 43: Picky comment - but a space between the `act` and the `assert` - same comment for the ones below

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/OrchFrontendAuthorizerHandler.java
```diff
@@ -0,0 +1,87 @@
...
+import static uk.gov.di.orchestration.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
+
+public class OrchFrontendAuthorizerHandler
+        implements RequestHandler<APIGatewayProxyRequestEvent, AuthPolicy> {
+
+    private static final Logger LOG = LogManager.getLogger(OrchFrontendAuthorizerHandler.class);
+
+    private final ConfigurationService configurationService;
+
+    private final String[] validIps = {
```
ethanmills, line 22: can we add a link to `https://sites.google.com/a/digital.cabinet-office.gov.uk/gds/working-at-gds/gds-internal-it/gds-internal-it-network-public-ip-addresses` in a comment please?

ethanmills: I think there are some we can get rid of from this list too

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AuthPolicy.java
ethanmills, line 1: Is there not something in the SDK we can use? e.g. https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/policybuilder/iam/package-summary.html

ethanmills (approved):

PR #4250, opened 2024-04-10
ATO-472: Add dynamo cross account policy

hannah-leong:
## What

Add resource policy to allow access from corresponding orchestration account and update terraform version to allow for this

## How to review

For example:

1. Code Review
1. Check policy 
1. Check feature flagged for environments
1. Check account IDs are correct

## 🚨⚠️ Orchestration and Authentication mutual dependencies ⚠️ 🚨

Be careful when making changes to code in 'shared' components where each team has a copy.
Check with counterparts to see if changes need to be made in the other team's code.

In particular pay attention to classes representing Session data where changes need to be applied on both sides to avoid deserialization errors.

- [ ] Impact on orch and auth mutual dependencies has been checked.

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/4225

PR #4362, opened 2024-04-26
ATO-554: Add spot request queue cross account access

CarlyG55:
## What

Allow orchestration access to the spot request queue behind a feature flag


ci/terraform/oidc/spot-sqs.tf
```diff
@@ -56,7 +98,7 @@ resource "aws_sqs_queue_policy" "spot_request_queue_policy" {
   ]
```
CarlyG55, line 98: was tempted to get rid of this depends on as it seems a bit wrong... but obviously don't want to break anything spot related!

isaac-GDS: This does seem wrong - if we do keep should we add `cross_account_spot_request_queue_policy_document` . If there was a need for this feels like `cross_account_spot_request_queue_policy_document`  would be important.

isaac-GDS (approved):

ci/terraform/oidc/spot-sqs.tf
CarlyG55: on chatting with ethan, going to remove the depends on and pause promotion beyond staging to test

ci/terraform/oidc/spot-sqs.tf
```diff
@@ -27,6 +27,48 @@ resource "aws_sqs_queue" "spot_request_dead_letter_queue" {
...
 
+data "aws_iam_policy_document" "cross_account_spot_request_queue_policy_document" {
+  statement {
+    sid    = "AllowSpotAccountToReceive"
+    effect = "Allow"
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${aws_ssm_parameter.spot_account_number.value}:root"]
+    }
```
hannah-leong, line 38: I think the identifier can be shortened to just `[${aws_ssm_parameter.spot_account_number.value}]` ([example here](https://github.com/govuk-one-login/authentication-api/pull/4250/files#diff-331c55b2d84d1db997c377c4169580d4c13c837f9193e2217a8814089ed948e9))

CarlyG55: Oh yeah, thanks!

hannah-leong (approved):

CarlyG55:
Need to pause int & prod to merge

hannah-leong (approved):

ethanmills (approved):
Have you checked the plan for this with the flag off?

PR #4364, opened 2024-04-26
RFF-2: Migrate to new audit service style

mrwilson:
This moves the orchestration code to use the new style of audit service method calling, as set up in #4319

### DIFF
--- a/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ClientRegistrationHandler.java
+++ b/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ClientRegistrationHandler.java
@@ -12,6 +12,7 @@ import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.clientregistry.entity.ClientRegistrationRequest;
 import uk.gov.di.authentication.clientregistry.entity.ClientRegistrationResponse;
 import uk.gov.di.authentication.clientregistry.services.ClientConfigValidationService;
+import uk.gov.di.orchestration.audit.TxmaAuditUser;
 import uk.gov.di.orchestration.shared.helpers.IpAddressHelper;
 import uk.gov.di.orchestration.shared.serialization.Json;
 import uk.gov.di.orchestration.shared.serialization.Json.JsonException;
@@ -27,6 +28,7 @@ import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.ge
 import static uk.gov.di.orchestration.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.LogFieldName.CLIENT_ID;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachLogFieldToLogs;
+import static uk.gov.di.orchestration.shared.services.AuditService.UNKNOWN;
 
 public class ClientRegistrationHandler
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
@@ -67,16 +69,12 @@ public class ClientRegistrationHandler
     public APIGatewayProxyResponseEvent clientRegistrationRequestHandler(
             APIGatewayProxyRequestEvent input, Context context) {
         String ipAddress = IpAddressHelper.extractIpAddress(input);
-        auditService.submitAuditEvent(
-                REGISTER_CLIENT_REQUEST_RECEIVED,
-                AuditService.UNKNOWN,
-                context.getAwsRequestId(),
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN,
-                ipAddress,
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN);
+
+        var user =
+                TxmaAuditUser.user()
+                        .withTransactionId(context.getAwsRequestId())
+                        .withIpAddress(ipAddress);
+        auditService.submitAuditEvent(REGISTER_CLIENT_REQUEST_RECEIVED, UNKNOWN, user);
 
         try {
             LOG.info("Client registration request received");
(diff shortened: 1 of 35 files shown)


client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ClientRegistrationHandler.java
```diff
@@ -67,16 +69,12 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN,
-                ipAddress,
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN);
+
+        var user =
+                TxmaAuditUser.user()
+                        .withSessionId(context.getAwsRequestId())
```
ethanmills, line 75: This doesn't seem right to me?

ethanmills, line 75: Is this correct? It looks like this was originally added explicitly as the request ID: https://github.com/govuk-one-login/authentication-api/commit/7ef78e33315866863e07f203d2bf97e4e48f589c#diff-a7dd2094ffa7958a40dfc6b333bf0d588539d0eccc5bf89d37efc6cdc13fbec7

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -306,13 +287,8 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
...
                         AUTH_CODE_ISSUED,
                         clientId,
-                        clientSessionId,
-                        session.getSessionId(),
-                        clientSession.getDocAppSubjectId().getValue(),
-                        session.getEmailAddress(),
-                        IpAddressHelper.extractIpAddress(input),
-                        AuditService.UNKNOWN,
-                        AuditService.UNKNOWN,
+                        user.withEmail(session.getEmailAddress())
```
ethanmills, line 290: I know this isn't changing this, but I think the email address will never be known because they're docapp?

mrwilson: Agree - this shouldn't percolate into the txma stream because that value will be null. What's your preference here, make it explicit and not add email at this callsite?

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ClientRegistrationHandler.java
mrwilson: Discussed on Slack, while this is not a regression, it doesn't really make sense as a value - would probably be a better fit for `transactionId`

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -78,6 +80,7 @@
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.LogFieldName.CLIENT_SESSION_ID;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.LogFieldName.GOVUK_SIGNIN_JOURNEY_ID;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.LogFieldName.PERSISTENT_SESSION_ID;
+import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.UNKNOWN;
```
ethanmills, line 83: I think this the wrong UNKNOWN?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -510,10 +496,8 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
                     authenticationRequest.getResponseMode(),
                     OIDCError.UNMET_AUTHENTICATION_REQUIREMENTS,
-                    ipAddress,
-                    persistentSessionId,
                     authenticationRequest.getClientID().getValue(),
-                    clientSessionId);
+                    user);
         }
         var session = existingSession.orElseGet(sessionService::createSession);
         attachSessionIdToLogs(session);
```
ethanmills, line 503: Would it be worth doing something like `user = user.withSessionId....` here instead of setting in the individual audit event? I'd be worried that we miss this off if we make changes in future

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
ethanmills: I think so, yes. I think it's misleading to be looking for something we know will never be there

orchestration-shared/src/main/java/uk/gov/di/orchestration/audit/TxmaAuditUser.java
```diff
@@ -1,114 +1,118 @@
...
+                phone,
+                ipAddress,
+                sessionId,
+                persistentSessionId,
+                govukSigninJourneyId);
     }
 
     public String getPhone() {
         return phone;
     }
```
ethanmills, line 117: Why do we need this?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AuditService.java
ethanmills, line 1: I think I'd be inclined to follow this up with a change to remove the audit context altogether - this seems to achieve most of what it was setting out to do. What do you think?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -146,16 +147,15 @@ public APIGatewayProxyResponseEvent generateLogoutResponse(
...
-                LogoutAuditableEvent.LOG_OUT_SUCCESS,
-                clientId.orElse(AuditService.UNKNOWN),
-                AuditService.UNKNOWN,
-                sessionId.orElse(AuditService.UNKNOWN),
-                AuditService.UNKNOWN,
-                AuditService.UNKNOWN,
-                IpAddressHelper.extractIpAddress(input),
-                AuditService.UNKNOWN,
-                PersistentIdHelper.extractPersistentIdFromCookieHeader(input.getHeaders()));
+
```
ethanmills, line 150: @billrdunn relevant to your work

orchestration-shared/src/main/java/uk/gov/di/orchestration/audit/TxmaAuditUser.java
mrwilson: Now that it's a record, we don't, good shout. We need it because in `AuditService` we add an extension based on the country code of the phone number. We used to get phone number as a parameter, but now it's on the object.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AuditService.java
mrwilson: I agree

ethanmills (dismissed):

ethanmills (approved):

PR #4371, opened 2024-04-29
BAU: Refactored VerifyCodeHandler

alhcomer:
## What

Refactored the VerifyCodeHandler to remove noise and make it more readable. Also refactored the VerifyCodeRequest class to be a record.

## How to review

1. Code Review, going through commits one at a time


BeckaL (approved):
Looks good to me, nice one ⭐ 

In terms of the audit stuff, I think we'll end up refactoring in the same way as auth so in general we probably don't want to do too much more custom stuff in each class for this before we do that bigger refactoring, but generally a nice improvement!

PR #4372, opened 2024-04-29
ATO-526: Pass txma-audit-encoded header in ProcessingIdentityHandler

gn-GDS:
## What
Pass txma-audit-encoded header in ProcessingIdentityHandler, using attachTxmaAuditFieldFromHeaders()

## Checklist


ethanmills:
I think we need to be careful with this one. This is called by the auth frontend, not the user.

hannah-leong (approved):

PR #4381, opened 2024-04-30
BAU: fix interventions alarms & AIS detailed logging

alhcomer:
## What

Auth AIS alarms in account-interventions-alerts.tf have been deleted. This is because alarms are already configured for every endpoint automatically in endpoint-module/alerts.tf

Also fixed isAisDetailedLoggingEnabled always returning false due to mismatched types being compared.

## How to review

1. Code Review


ci/terraform/oidc/account-interventions-alerts.tf
```diff
@@ -32,12 +32,12 @@ resource "aws_cloudwatch_metric_alarm" "account_interventions_p1_cloudwatch_alar
 
 data "aws_cloudwatch_log_group" "auth_account_interventions_lambda_log_group" {
   depends_on = [module.account_interventions]
-  count      = var.environment == "production" || var.environment == "integration" ? 0 : 1
+  count      = var.environment == "production" || var.environment == "integration" ? 1 : 0
```
BeckaL, line 35: I think for the log groups and metric filters, these should just be fully on in all environments (as per orch's code). I think worth cross referencing each of the things here against orch's ones above and aligning them

alhcomer: Going to delete our alarms in this file as @dbes-gds pointed out that there are already alarms set up for every endpoint automatically for auth in endpoint-module/alerts.tf

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -166,7 +166,8 @@ public int getAuthAppCodeAllowedWindows() {
     }
 
     public boolean isAisDetailedLoggingEnabled() {
-        return !List.of(Environment.PRODUCTION, Environment.INTEGRATION).contains(getEnvironment());
+        return !List.of(Environment.PRODUCTION.getValue(), Environment.INTEGRATION.getValue())
```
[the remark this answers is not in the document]
alhcomer: That makes complete sense

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AccountInterventionsHandler.java
```diff
@@ -251,14 +251,12 @@ private void submitAuditEvents(
...
-                    accountInterventionsInboundResponse.state().suspended(),
-                    accountInterventionsInboundResponse.state().resetPassword(),
-                    accountInterventionsInboundResponse.state().reproveIdentity());
-        }
+        LOG.info(
+                "AIS Response: blocked: {} suspended: {} resetPassword: {} reproveIdentity: {}",
+                accountInterventionsInboundResponse.state().blocked(),
+                accountInterventionsInboundResponse.state().suspended(),
+                accountInterventionsInboundResponse.state().resetPassword(),
+                accountInterventionsInboundResponse.state().reproveIdentity());
```
mrwilson, line 259: Records generate `toString` by default, so you can likely have something like:

```
LOG.info("AIS Response: {}", accountInterventionsInboundResponse.state());
```

which would give
```
AIS Response: State[blocked=false, suspended=true, reproveIdentity=true, resetPassword=false]
```

dbes-gds (approved):

PR #4386, opened 2024-05-01
ATO-585: Add information to the AUTH_LOG_OUT_SUCCESS audit event

billrdunn:
## What

- Add user ID (`subjectId`), when it is available, to the `AUTH_LOG_OUT_SUCCESS` event
- Add RP pairwise ID, when it is available, to the `AUTH_LOG_OUT_SUCCESS` event

I'd like to spend some time in the near future refactoring the logout flows as they are quite convoluted.

## How to review

1. Code review only, easiest commit by commit

## Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
@@ -100,14 +100,8 @@ public class LogoutHandler
                         () -> sessionService.getSessionFromSessionCookie(input.getHeaders()));
         attachSessionToLogsIfExists(sessionFromSessionCookie, input.getHeaders());
 
-        Optional<String> subjectId = Optional.empty();
-        Optional<String> sessionId = Optional.empty();
-        if (sessionFromSessionCookie.isPresent()) {
-            subjectId =
-                    Optional.ofNullable(
-                            sessionFromSessionCookie.get().getInternalCommonSubjectIdentifier());
-            sessionId = Optional.ofNullable(sessionFromSessionCookie.get().getSessionId());
-        }
+        var subjectId = sessionFromSessionCookie.map(Session::getInternalCommonSubjectIdentifier);
+        var sessionId = sessionFromSessionCookie.map(Session::getSessionId);
 
         var auditUser =
                 TxmaAuditUser.user()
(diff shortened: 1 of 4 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -153,7 +162,8 @@ public APIGatewayProxyResponseEvent generateLogoutResponse(
                         .withIpAddress(extractIpAddress(input))
                         .withPersistentSessionId(
                                 extractPersistentIdFromCookieHeader(input.getHeaders()))
-                        .withSessionId(sessionId.orElse(null));
+                        .withSessionId(sessionId.orElse(null))
```
ethanmills, line 165: Would it make sense to build the audit user in the handler and pass that through, rather than the individual fields?

billrdunn: It's possible, and it would make things neater, but also means that the audit user is quite separated from where it's actually used. I'm not sure if it's that valuable right now. I'd like to spend some time refactoring LogoutHandler / LogoutService after this ticket

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
```diff
@@ -97,25 +100,37 @@ public APIGatewayProxyResponseEvent logoutRequestHandler(APIGatewayProxyRequestE
...
         attachSessionToLogsIfExists(sessionFromSessionCookie, input.getHeaders());
 
+        Optional<String> subjectId = Optional.empty();
+        Optional<String> sessionId = Optional.empty();
+        if (sessionFromSessionCookie.isPresent()) {
+            subjectId =
+                    Optional.ofNullable(
+                            sessionFromSessionCookie.get().getInternalCommonSubjectIdentifier());
+            sessionId = Optional.ofNullable(sessionFromSessionCookie.get().getSessionId());
+        }
```
mrwilson, line 110: We can collapse these using `.map()` rather than testing if `sessionFromSessionCookie` exists directly.

```
var subjectId = sessionFromSessionCookie.map(Session::getInternalCommonSubjectIdentifier);
var sessionId = sessionFromSessionCookie.map(Session::getSessionId);
```

ethanmills (approved):

ethanmills:
Can you update the spreadsheet before merging this one please?

isaac-GDS (approved):

PR #4390, opened 2024-05-01
BAU: Add check for if clientID is null in the access token

hannah-leong:
## What

Add a check for the `clientID` in the AccessToken, to make sure that if it is `null` it throws an exception instead of querying the database with a null clientID and erroring.

https://gds.slack.com/archives/C02HCLREVV5/p1714570226681109

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccessTokenService.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccessTokenService.java
@@ -81,14 +81,18 @@ public class AccessTokenService {
                         "Unable to validate AccessToken signature", BearerTokenError.INVALID_TOKEN);
             }
             var clientID = signedJWT.getJWTClaimsSet().getStringClaim("client_id");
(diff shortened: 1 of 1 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccessTokenService.java
```diff
@@ -81,14 +81,18 @@ public AccessTokenInfo parse(String authorizationHeader, boolean identityEnabled
                         "Unable to validate AccessToken signature", BearerTokenError.INVALID_TOKEN);
             }
             var clientID = signedJWT.getJWTClaimsSet().getStringClaim("client_id");
-            var client = clientService.getClient(clientID).orElse(null);
-
             attachLogFieldToLogs(CLIENT_ID, clientID);
+            if (Objects.isNull(clientID)) {
+                LOG.warn("Client not found");
```
billrdunn, line 86: "ClientId not found" ?

billrdunn (approved):

jkunle (approved):

PR #4392, opened 2024-05-01
AUT-2640: Update audit event component id value to 'AUTH'

alhcomer:
## What
Changed componentId audit field assignment to have the value "AUTH".

Security require the component id field to be populated with a signifier that allows them to distinguish audit events from our service (or component) from audit events from other services. At least for the time being, "AUTH" has been chosen as an appropriate component id value.

Previously we were populating this field getting the env var OIDC_API_BASE_URL. This env var was not declared in our terraform and so was returning null anyway, but even if the correct value for OIDC_API_BASE_URL were returned, that would not be an appropriate componentId for auth as it is not our service uri.

## How to review

1. Code Review

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
@@ -21,6 +21,7 @@ public class AuditService {
     private final Clock clock;
     private final ConfigurationService configurationService;
     private final AwsSqsClient txmaQueueClient;
+    private final String COMPONENT_ID = "AUTH";
 
     public AuditService(
             Clock clock, ConfigurationService configurationService, AwsSqsClient txmaQueueClient) {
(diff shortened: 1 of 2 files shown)


ethanmills:
I think this should be `https://signin.<env.>account.gov.uk/`. Have you seen https://govukverify.atlassian.net/wiki/spaces/~712020ca5ac93cd97f4e89b368ea8355f2af55/pages/3901685867/Component+ID+Analysis or talked to Dom S about this?

shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
```diff
@@ -64,7 +64,7 @@ public void submitAuditEvent(
         var txmaAuditEvent =
                 auditEventWithTime(event, () -> Date.from(clock.instant()))
                         .withClientId(clientId)
-                        .withComponentId(configurationService.getOidcApiBaseURL().orElse("UNKNOWN"))
+                        .withComponentId("AUTH")
```
BeckaL, line 67: Shall we pull this string out into a constant? (I appreciate we haven't always been super consistent with this kind of thing so far :-D)

BeckaL (commented):
Looks good to me! Minor comment about pulling a string out into a constant

BeckaL (approved):

dbes-gds (approved):

PR #4398, opened 2024-05-02
ATO-573: Create orch token lambda

CarlyG55:
## What
Set up token lambda in orchestration

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -645,6 +645,7 @@ Resources:
     # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
     # checkov:skip=CKV_AWS_117: VPC settings are global
     Properties:
+      FunctionName: !Sub ${Environment}-TokenFunction
       AutoPublishAlias: latest
       CodeUri: ./oidc-api
       Handler: uk.gov.di.authentication.oidc.lambda.TokenHandler::handleRequest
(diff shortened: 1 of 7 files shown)


isaac-GDS (dismissed):

template.yaml
```diff
@@ -620,6 +635,199 @@ Resources:
...
+              - kms:ReEncrypt*
+              - kms:GenerateDataKey*
+              - kms:CreateGrant
+              - kms:DescribeKey
+            Resource: !FindInMap [ EnvironmentConfiguration, !Ref Environment, clientRegistryTableKeyArn ]
+          - Sid: AllowUserProfileTableReadAndWriteAccess
+            Effect: "Allow"
+            Action:
+              - dynamodb:BatchGetItem
+              - dynamodb:DescribeStream
```
ethanmills, line 720: Don't think we need all of these?

ethanmills (approved):

template.yaml
CarlyG55: I just took these permissions from the terraform permissions - I can have a bit more of a think about them

template.yaml
```diff
@@ -625,6 +640,201 @@ Resources:
...
+            Action:
+              - dynamodb:BatchGetItem
+              - dynamodb:DescribeTable
+              - dynamodb:Get*
+              - dynamodb:Query
+              - dynamodb:Scan
+            Resource: !Sub
+              - 'arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-client-registry'
+              - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+                AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
```
hannah-leong, line 715: Very nit-picky but do the quotation marks need to be here? Just wondering for consistency as most of the other statements (or at least most bits of them) don't have them

hannah-leong (approved):

ethanmills (approved):

template.yaml
CarlyG55: I just copied them over but I can change in both places

CarlyG55: Although I might do this later on, as I probably want to check it deploys happily and sandpit is busy today

PR #4402, opened 2024-05-03
Bau/refactor audit mob

BeckaL:
## What

Introduces a new method to submit an audit service, which takes an audit user as an argument as opposed to individual fields. This will allow us to reduce a lot of noise and duplication in our handlers.

This converts a single handler, the Login Handler, to use this new method.

## How to review

1. Code Review

## Related PRs

An equivalent refactor has [already been done for orchestration ](https://github.com/govuk-one-login/authentication-api/pull/4319)

PR #4417, opened 2024-05-08
AUT-2663: account management api consumes encoded txma audit header

mark-cab:
## What

Read the new CF TXMA encoded header in the account management api lambda handlers.  If present send to txma in each audit event submitted.
Value is being stored in ThreadContext (in line with Orch) similar to how we store session id. This is then read and the data is added to audit events.

## How to review

1. Code Review

## Checklist

<!-- 🚨⚠️ Orchestration and Authentication mutual dependencies ⚠️ 🚨

### DIFF
--- a/ci/terraform/account-management/variables.tf
+++ b/ci/terraform/account-management/variables.tf
@@ -224,6 +224,12 @@ variable "support_email_check_enabled" {
   description = "Feature flag which toggles the Experian email check on and off"
 }
 
+variable "txma_audit_encoded_enabled" {
+  default     = false
+  type        = bool
+  description = "Feature flag which toggles submitting additional Txma headers for audit"
+}
+
 locals {
   default_performance_parameters = {
     memory          = var.endpoint_memory_size
--- a/shared/src/main/java/uk/gov/di/audit/AuditContext.java
+++ b/shared/src/main/java/uk/gov/di/audit/AuditContext.java
@@ -1,0 +1,160 @@
+package uk.gov.di.audit;
+
+import uk.gov.di.authentication.shared.services.AuditService.MetadataPair;
+
+import java.util.Arrays;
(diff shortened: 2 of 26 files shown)


shared/src/main/java/uk/gov/di/audit/AuditContext.java
```diff
@@ -0,0 +1,160 @@
+package uk.gov.di.audit;
+
+import uk.gov.di.authentication.shared.services.AuditService.MetadataPair;
+
+import java.util.Arrays;
+import java.util.Objects;
+
+public class AuditContext {
```
BeckaL, line 8: This can be a record (in general the only things that I'd still make classes are things that encapsulate e.g. services and lambdas and dynamo db entities which aren't compatible with records yet)

BeckaL: Ah sorry I see that you're updating fields on this as you go, and the syntax for doing this in records isn't particularly nice yet. For consistency in terms of refactoring this, I'd probably want to go for the same approach we've done in the main api that instead constructs an audit user since this is the thing that is eventually used in the audit service anyway - see [here](https://github.com/govuk-one-login/authentication-api/pull/4402). 

If this ends up being a bigger conversation and holds this work back too much, maybe we can just introduce the changes we need for this work in this PR, and refactor separately.

ci/terraform/account-management/variables.tf
```diff
@@ -224,6 +224,12 @@ variable "support_email_check_enabled" {
   description = "Feature flag which toggles the Experian email check on and off"
 }
 
+variable "txma_audit_encoded_enabled" {
```
BeckaL, line 227: Is there a requirement to feature flag this? (I didn't think so but might have missed it)

PR #4429, opened 2024-05-09
Add DocAppBackendAPI Class

BenjaminWCO:
## Checklist


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/api/DocAppBackendAPI.java
ethanmills, line 1: I think the name here is a little confusing. We have a docapp api, but this is referring to the API of the docapp CRI

PR #4438, opened 2024-05-09
Add FrontEndPages Class

BenjaminWCO:
## Checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -112,7 +112,8 @@ subprojects {
         apache "commons-codec:commons-codec:1.16.1",
                 "org.apache.httpcomponents:httpclient:4.5.14",
                 "org.apache.commons:commons-collections4:4.1",
-                "commons-net:commons-net:3.10.0"
+                "commons-net:commons-net:3.10.0",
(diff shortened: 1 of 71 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/api/FrontEndAPI.java
```diff
@@ -0,0 +1,86 @@
...
+import com.nimbusds.openid.connect.sdk.Prompt;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Optional;
+
+import static uk.gov.di.orchestration.shared.helpers.ConstructUriHelper.buildURI;
+
+public class FrontEndAPI {
```
ethanmills, line 13: I think this is the wrong name. The frontend API exists, but this refers to the frontend

BenjaminWCO: Yeah that did occur to me. Not really sure what a ogod name fore it is.

PR #4451, opened 2024-05-10
ATO-518: Enforce cloudfront for OIDC in sandpit and staging

ethanmills:
## What

Switch the WAF on the OIDC API Gateway to the one provided by dev platform that just blocks everything without the secret header that cloudfront sets

### DIFF
--- a/ci/terraform/oidc/api-gateway.tf
+++ b/ci/terraform/oidc/api-gateway.tf
@@ -703,7 +703,7 @@ resource "aws_wafv2_web_acl" "wafregional_web_acl_oidc_api" {
 }
 
 resource "aws_wafv2_web_acl_association" "oidc_waf_association" {
-  count        = var.use_localstack ? 0 : 1
+  count        = var.enforce_cloudfront ? 0 : 1
   resource_arn = aws_api_gateway_stage.endpoint_stage.arn
   web_acl_arn  = aws_wafv2_web_acl.wafregional_web_acl_oidc_api[count.index].arn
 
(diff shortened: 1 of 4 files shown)


ci/terraform/oidc/api-gateway.tf
```diff
@@ -853,6 +863,8 @@ data "aws_cloudformation_stack" "orch_frontend_stack" {
 locals {
   nlb_dns_name = length(data.aws_cloudformation_stack.orch_frontend_stack) > 0 ? data.aws_cloudformation_stack.orch_frontend_stack[0].outputs["OrchFrontendNlbDnsName"] : null
   nlb_arn      = length(data.aws_cloudformation_stack.orch_frontend_stack) > 0 ? data.aws_cloudformation_stack.orch_frontend_stack[0].outputs["OrchFrontendNlbArn"] : null
+
+  oidc_origin_cloaking_waf_export_name = var.environment == "sandpit" ? "dev-oidc-cloudfront-CloakingOriginWebACLArn" : "${var.environment}-oidc-cloudfront-CloakingOriginWebACLArn"
```
CarlyG55, line 867: I was planning on pulling out a local for the 'secure pipelines' environment (eg dev for sandpit, environment otherwise) which might be useful here (although I can always update when I get to that)
Also if someone tries to deploy oidc to authdev is it going to complain?

ethanmills: I think so long as they don't turn on this feature it won't complain in authdev. Would be good to tidy with that local, but I'd like to get this in now

CarlyG55: Yeah totally fine, I can do that later

PR #4455, opened 2024-05-13
ATO-595: Store client session ID (Journey ID) in the access token store

billrdunn:
## What

Store client session ID (Journey ID) in the access token store

This is required so that the journey ID is available in UserInfoHandler, and it can be included in the AUTH_USER_INFO_RETURNED audit event.


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/AccessTokenStore.java
```diff
@@ -10,13 +10,19 @@ public class AccessTokenStore {
 
     @Expose private String internalPairwiseSubjectId = "missing";
 
+    @Expose private String journeyId;
```
ethanmills, line 13: Do we need to set a default on this like the one above?

ethanmills (approved):

PR #4457, opened 2024-05-13
ATO-627: Update AIS tests to cover case that fields are missing in AIS response

billrdunn:
## What

AIS may not send us all the fields under "intervention" in the response. This PR modifies the contract test to let AIS know we do not need these fields, and adds unit and integration tests to cover this situation.

## Checklist

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
@@ -39,8 +39,8 @@ class AccountInterventionServiceTest {
     private final HttpClient httpClient = HttpClient.newHttpClient();
     private AccountInterventionService accountInterventionService;
 
-    private static final String PROVIDER_NAME = "ExternalAccountInterventionService";
-    private static final String CONSUMER_NAME = "OrchestrationAccountInterventionService";
+    private static final String PROVIDER_NAME = "AccountInterventionServiceProvider";
+    private static final String CONSUMER_NAME = "OrchAccountInterventionServiceConsumer";
     private static final String INTERNAL_PAIRWISE_SUBJECT_ID = "internal-pairwise-subject-id";
     private static final String INVALID_INTERNAL_PAIRWISE_SUBJECT_ID =
             "invalid-internal-pairwise-subject-id";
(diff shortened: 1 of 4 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/contract/AccountInterventionServiceTest.java
```diff
@@ -358,12 +358,7 @@ private static DslPart createAisResponseBody(
                             body.object(
                                     "intervention",
                                     (obj) -> {
-                                        obj.numberType("updatedAt", TIME_NOW);
                                         obj.numberType("appliedAt", TIME_NOW);
-                                        obj.numberType("sentAt", TIME_NOW);
-                                        obj.stringType("description", "TEST_DESCRIPTION");
-                                        obj.numberType("reprovedIdentityAt", TIME_NOW);
-                                        obj.numberType("resetPasswordAt", TIME_NOW);
                                     });
```
isaac-GDS, line 362: Is the default to return only "appliedAt". Do we care about the other fields at any point in the `intervention` object ?

billrdunn: No the default is to return either all the fields or all the fields excluding `reprovedIdentityAt` and `resetPasswordAt`. The only field we use is `appliedAt` and we don't care about any other fields

isaac-GDS: Cool - this is fine

shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/AccountInterventionsStubExtension.java
```diff
@@ -91,6 +91,39 @@ public void initWithAccountStatus(
...
+        register(
+                "/v1/ais/" + userId,
+                200,
+                "application/json",
+                "{"
+                        + "  \"intervention\": {"
+                        + "    \"updatedAt\": 1696969322935,"
+                        + "    \"appliedAt\": 1696869005821,"
+                        + "    \"sentAt\": 1696869003456,"
+                        + "    \"description\": \"EXAMPLE_DESCRIPTION\""
```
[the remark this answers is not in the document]
isaac-GDS: The answer above has resolved this. Happy for the acceptance test to capture the two states of response objects that can be returned

isaac-GDS (dismissed):

isaac-GDS (approved):

PR #4458, opened 2024-05-13
AUT-2264: Add txma audit header to the UserContext to allow retrieval…

andrew-moores:
… in the lambdas when building a txma audit event.  This commit makes use of this in the LoginHandler to establish how it will work.

## Checklist


shared-test/src/main/java/uk/gov/di/authentication/sharedtest/helper/AuditAssertionsHelper.java
```diff
@@ -27,6 +27,45 @@ public static void assertNoTxmaAuditEventsReceived(SqsQueueExtension txmaAuditQu
                         () -> assertThat(txmaAuditQueue.getApproximateMessageCount(), equalTo(0)));
     }
 
+    public static void assertUnvalidatedTxmaAuditEventsReceived(
```
BeckaL, line 30: how about something like `assertTxmaAuditEventsSubmittedWithMatchingNames`?
I guess reading the `unvalidated` bit of this in a test would be a bit surprising to me, I think I'd spend a bit of time working out what was meant by unvalidated

BeckaL (commented):
Looks good to me :-) Just a few minor comments / suggestions

BeckaL (dismissed):
All looks good to me

BeckaL (approved):

PR #4459, opened 2024-05-13
ATO-56: Decouple IPV Callback From Auth

ghost:
Note: Requires a rebase once https://github.com/govuk-one-login/authentication-api/pull/4479 is merged.

## What

- Added ORCH_FRONTEND_ENABLED Flag to toggle IPV Callback Handler usage of Orch Frontend instead of Auth Frontend - currently false.
- Renamed FRONTEND_BASE_URL to AUTH_FRONTEND_BASE_URL.
- Added ORCH_FRONTEND_BASE_URL - currently no value is set.

## How to review

Code Review

## Checklist

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -23,6 +23,8 @@ import uk.gov.di.authentication.ipv.services.IPVTokenService;
 import uk.gov.di.orchestration.audit.AuditContext;
 import uk.gov.di.orchestration.audit.TxmaAuditUser;
 import uk.gov.di.orchestration.shared.api.AuthFrontend;
+import uk.gov.di.orchestration.shared.api.CommonFrontend;
+import uk.gov.di.orchestration.shared.api.OrchFrontend;
 import uk.gov.di.orchestration.shared.entity.AccountIntervention;
 import uk.gov.di.orchestration.shared.entity.ClientRegistry;
 import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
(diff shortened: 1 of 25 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -148,7 +150,10 @@ public IPVCallbackHandler(ConfigurationService configurationService) {
         this.noSessionOrchestrationService =
                 new NoSessionOrchestrationService(configurationService);
         this.ipvCallbackHelper = new IPVCallbackHelper(configurationService);
-        this.authFrontend = new AuthFrontend(configurationService);
+        this.frontend =
+                configurationService.getOrchFrontendEnabled()
+                        ? new OrchFrontend(configurationService)
+                        : new AuthFrontend(configurationService);
     }
```
ethanmills, line 157: Is it possible to write a test covering this?

ethanmills (commented):
needs a rebase

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
BenjaminWCO: Not really possible to unit test the constructor given how it creates new instances of all the services it uses, but I can add a public method getFrontend(ConfigurationService) and call that from the constructor.

BenjaminWCO: Done. Might make sense to factor that out into a "FrontendProviderService" class if we're going to duplicate this logic elsewhere, but I think this makes sense for now.

ethanmills (approved):

ethanmills (approved):

PR #4481, opened 2024-05-15
AUT-2165: log the status of experian email check during update email process

mark-cab:
## What

In UpdateEmail lambda in the Account Management API, added a step check the status of the users fraudulent email check for the new email address that the user is proposing to update their profile to use and log the result.

Interim stage of email phone check process where we will log the status of the asynchronous request sent to Experian to get an idea of how long it takes

## How to review

1. Code Review

## Checklist

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
@@ -15,6 +15,7 @@ import uk.gov.di.accountmanagement.exceptions.InvalidPrincipalException;
 import uk.gov.di.accountmanagement.helpers.PrincipalValidationHelper;
 import uk.gov.di.accountmanagement.services.AwsSqsClient;
 import uk.gov.di.accountmanagement.services.CodeStorageService;
+import uk.gov.di.authentication.shared.entity.EmailCheckResultStatus;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.exceptions.UserNotFoundException;
 import uk.gov.di.authentication.shared.helpers.AuditHelper;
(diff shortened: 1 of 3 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
```diff
@@ -107,6 +113,13 @@ public APIGatewayProxyResponseEvent updateEmailRequestHandler(
         try {
             UpdateEmailRequest updateInfoRequest =
                     objectMapper.readValue(input.getBody(), UpdateEmailRequest.class);
+            dynamoEmailCheckResultService
+                    .getEmailCheckStore(updateInfoRequest.getReplacementEmailAddress())
+                    .ifPresent(
+                            result ->
+                                    LOG.info(
+                                            "UpdateEmailHandler: Experian email verification status: {}",
+                                            result.getStatus()));
```
mattvot, line 122: Is it possible to add a log if it's not present to say the status is PENDING?

mattvot (approved):

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/UpdateEmailHandler.java
mark-cab: From what you're saying the assumption is that this lambda is called only after an email check has been requested and can't be called without. 
I don't like adding something like this to a separate lambda unless you know it's been requested but I'm not as close to this as you are so will go with your suggestion and add it.

mattvot: This lambda is called by Home team to update the users email at any point and it is possible that we don't have a result in the database by the time this lambda is called. It is an expected that sometimes we won't have a result for the check at this point. 

Happy to leave it as is though.

mark-cab: as-is is with the additional logging now so if you can give it a once over and then I will merge it.

mattvot (approved):
Thanks, approved

PR #4488, opened 2024-05-16
AUT-2267: async Experian phone check is triggered after change of phone number

mark-cab:
## What?

Update API to send message to SQS queue after sign up phone validation.

## Why?

When the Experian Phone Checker service (owned by Experian) has an outage then checks against phone numbers fail and do not happen.  There is no failure or DLQ for the transactions, so we have no way of trying to run them again.  The service has failed a few times recently which generates alerts that we have to deal with, but we are unable to recover the transactions.

## Related PRs
[contra indicator PR](https://github.com/govuk-one-login/authentication-contra-indicators/pull/144)

[AUT-2103]: https://govukverify.atlassian.net/browse/AUT-2103?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ci/terraform/oidc/authdev1.tfvars
+++ b/ci/terraform/oidc/authdev1.tfvars
@@ -49,6 +49,9 @@ email_acct_creation_otp_code_ttl_duration = 60
 
 orch_client_id = "orchestrationAuth"
 
+contra_state_bucket      = "di-auth-development-tfstate"
+phone_checker_with_retry = false
+
 orch_frontend_api_gateway_integration_enabled = false
 
 orch_redirect_uri = "https://oidc.authdev1.sandpit.account.gov.uk/orchestration-redirect"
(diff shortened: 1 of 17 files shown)


ci/terraform/oidc/authdev1.tfvars
```diff
@@ -49,6 +49,12 @@ email_acct_creation_otp_code_ttl_duration = 60
 
 orch_client_id = "orchestrationAuth"
 
+contra_state_bucket      = "di-auth-development-tfstate"
+phone_checker_with_retry = true
+
+support_auth_orch_split           = true
```
dbes-gds, line 55: were these missing? is this what we have in the other authdevs?

dbes-gds (approved):

PR #4492, opened 2024-05-16
ATO-572: Create Auth Code Lambda In Orch Account

BenjaminWCO:
## What

Set up Auth Code lambda in orchestration account.

### DIFF
--- a/ci/terraform/oidc/api-gateway.tf
+++ b/ci/terraform/oidc/api-gateway.tf
@@ -160,7 +160,8 @@ resource "aws_api_gateway_deployment" "deployment" {
       var.orch_doc_app_callback_enabled,
       var.orch_token_enabled,
       var.orch_jwks_enabled,
-      var.orch_authorisation_enabled
+      var.orch_authorisation_enabled,
+      var.orch_auth_code_enabled,
     ]))
   }
 
@@ -187,7 +188,8 @@ resource "aws_api_gateway_deployment" "deployment" {
     aws_api_gateway_integration.orch_doc_app_callback_integration,
     aws_api_gateway_integration.orch_token_integration,
     aws_api_gateway_integration.orch_jwks_integration,
-    aws_api_gateway_integration.orch_authorisation_integration
+    aws_api_gateway_integration.orch_authorisation_integration,
+    aws_api_gateway_integration.orch_auth_code_integration,
   ]
 }
 
@@ -1009,6 +1011,28 @@ resource "aws_api_gateway_method" "orch_authorisation_method" {
   authorization = "NONE"
 }
 
+resource "aws_api_gateway_resource" "orch_auth_code_resource" {
+  count       = var.orch_auth_code_enabled ? 1 : 0
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
+  parent_id   = aws_api_gateway_rest_api.di_authentication_api.root_resource_id
+  path_part   = "auth-code"
+  depends_on = [
+    module.auth-code
+  ]
+}
+
+resource "aws_api_gateway_method" "orch_auth_code_method" {
+  count       = var.orch_auth_code_enabled ? 1 : 0
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
+  resource_id = aws_api_gateway_resource.orch_auth_code_resource[0].id
+  http_method = "GET"
+
+  depends_on = [
+    aws_api_gateway_resource.orch_auth_code_resource
+  ]
+  authorization = "NONE"
+}
+
 resource "aws_api_gateway_integration" "orch_authorisation_integration" {
   for_each    = var.orch_authorisation_enabled ? toset(["GET", "POST"]) : []
   rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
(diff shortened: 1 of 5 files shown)


ci/terraform/oidc/api-gateway.tf
```diff
@@ -946,3 +948,38 @@ resource "aws_api_gateway_integration" "orch_token_integration" {
...
+resource "aws_api_gateway_integration" "orch_auth_code_integration" {
+  count       = var.orch_auth_code_enabled ? 1 : 0
+  rest_api_id = aws_api_gateway_rest_api.di_authentication_api.id
+  resource_id = aws_api_gateway_resource.orch_auth_code_resource[0].id
+  http_method = aws_api_gateway_method.orch_auth_code_method[0].http_method
+  depends_on = [
+    aws_api_gateway_resource.orch_auth_code_resource
+  ]
+  type                    = "AWS_PROXY"
+  integration_http_method = "GET"
```
CarlyG55, line 983: Good spot but I think this actually needs to be POST for all of them - previous version of this is [here](https://github.com/govuk-one-login/authentication-api/blob/5b820c9144773f6de6790749ea56b0f11427baf1/ci/terraform/modules/endpoint-module/api-gateway.tf#L31)

template.yaml
```diff
@@ -835,6 +835,196 @@ Resources:
...
+      Handler: uk.gov.di.authentication.oidc.lambda.AuthCodeHandler::handleRequest
+      ReservedConcurrentExecutions: 1
+      LoggingConfig:
+        LogGroup: !Ref AuthCodeLogGroup
+      Environment:
+        Variables:
+          # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
+          ENVIRONMENT: !Sub ${Environment}
+          EXTERNAL_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, idTokenKeyArn ]
+          EXTERNAL_TOKEN_SIGNING_KEY_RSA_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, idTokenKeyRsaArn ]
```
CarlyG55, line 855: Does this lambda need access to these?

ci/terraform/oidc/api-gateway.tf
BenjaminWCO: Ah, okay, thanks.

template.yaml
BenjaminWCO: I don't think it does. Will remove.

template.yaml
```diff
@@ -1012,6 +1012,194 @@ Resources:
...
+              - kms:GetPublicKey
+            Resource:
+              - !FindInMap [ EnvironmentConfiguration, !Ref Environment, idTokenKeyArn ]
+          - Sid: AllowIdTokenSigningKmsRsa
+            Effect: Allow
+            Action:
+              - kms:Sign
+              - kms:GetPublicKey
+            Resource:
+              - !FindInMap [ EnvironmentConfiguration, !Ref Environment, idTokenKeyRsaArn ]
```
CarlyG55, line 1056: We also don't need the permissions to get this key

BenjaminWCO: fixed

template.yaml
```diff
@@ -1315,6 +1315,180 @@ Resources:
...
+            Effect: Allow
+            Action:
+              - dynamodb:DescribeTable
+              - dynamodb:Get*
+              - dynamodb:Query
+              - dynamodb:Scan
+            Resource: !Sub
+              - 'arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-user-profile'
+              - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+                AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
```
CarlyG55, line 1392:
```suggestion
          - Sid: AllowUserProfileTableReadAndWriteAccess
            Effect: "Allow"
            Action:
              - dynamodb:DescribeTable
              - dynamodb:Get*
              - dynamodb:Query
              - dynamodb:Scan
              - dynamodb:UpdateItem
              - dynamodb:PutItem
            Resource:
              - !Sub
                - arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-user-profile
                - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
                  AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
              - !Sub
                - arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-user-profile/index/*
                - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
                  AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
```
There's a small edge case where we might need write profile to the user profile table from this lambda, and access to the indices

CarlyG55: It's in `getOrGenerateSalt`, which will basically never get used as a write permission in this scenario, but we should cover it just in case (even though it's not currently covered!)

PR #4503, opened 2024-05-17
RFF-4: Refactoring test data

BeckaL:
## What

We introduce `CommonTestVariables` in the `frontend-api` module to encapsulate frequently used test data like email, password, phone number.

This introduces an abstraction so that we can make future improvements such as randomly generated input data to prevent dependency on specific inputs, or specific test cases / personas for users.

## Checklist

This code is contained within the `frontend-api` module only.


mrwilson (dismissed):

alhcomer (approved):
Awesome

PR #4517, opened 2024-05-21
ATO-644: Expose IdTokenSigningAlgorithm In Client Registry API

BenjaminWCO:
## What

Exposing IdTokenSigningAlgorithm Client Registry API

## How to review

Code Review

## Checklist

### DIFF
--- a/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationRequest.java
+++ b/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationRequest.java
@@ -10,6 +10,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 
+import static com.nimbusds.jose.JWSAlgorithm.ES256;
+
 public class ClientRegistrationRequest {
 
     @SerializedName("client_name")
@@ -79,6 +81,10 @@ public class ClientRegistrationRequest {
     @Expose
     private boolean jarValidationRequired;
 
+    @SerializedName("id_token_signing_algorithm")
+    @Expose
+    private String idTokenSigningAlgorithm = ES256.getName();
+
     public ClientRegistrationRequest() {}
 
     public ClientRegistrationRequest(
@@ -94,7 +100,8 @@ public class ClientRegistrationRequest {
             String subjectType,
             boolean identityVerificationRequired,
             List<String> claims,
-            String clientType) {
+            String clientType,
+            String idTokenSigningAlgorithm) {
         this(
                 clientName,
                 redirectUris,
@@ -109,6 +116,7 @@ public class ClientRegistrationRequest {
                 identityVerificationRequired,
                 claims,
                 clientType,
+                idTokenSigningAlgorithm,
                 null);
     }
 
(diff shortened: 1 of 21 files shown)


client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/services/ClientConfigValidationService.java
```diff
@@ -196,4 +217,8 @@ private boolean areClientLoCsValid(List<String> clientLoCs) {
         }
         return true;
     }
+
+    private boolean isValidIdTokenSigningAlgorithm(String idTokenSigningAlgorithm) {
+        return validIdTokenSigningAlgorithms.contains(idTokenSigningAlgorithm);
```
mrwilson, line 222: You could probably retool this method to do all the Optional-based logic that's currently in the if-statements, and then the two callsites would become
something like
```java
if (isNotValidIdTokenSigningAlgorithm(updateRequest.getIdTokenSigningAlgorithm())) {
...
}
```

BenjaminWCO: Yeah, it's a bit of an odd pattern. Would help with readability. I'm not really sure why we don't throw an exception rather than returning an an error object. Both validation calls are already in a try catch.

BenjaminWCO: I'll open a small BAU PR to refactor this when I have a moment, but I think it's makes sense to leave it as is for now as it's consistent with what's already there.

BenjaminWCO: I've opened a separate RP for the above change: https://github.com/govuk-one-login/authentication-api/pull/4629. It'll require a rebase once this RP is merged.

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationRequest.java
```diff
@@ -150,6 +157,7 @@ public ClientRegistrationRequest(
             clientType = ClientType.WEB.getValue();
         }
         this.clientType = clientType;
+        this.idTokenSigningAlgorithm = idTokenSigningAlgorithm;
```
ethanmills, line 160: We need to make this default to es256 so that this is not a breaking API change. As-is I think this would stop the self service tool from working until they updated to send the new field?

ethanmills (changes requested):

ethanmills (commented):
We expect (for some reason) `RSA256` in the client reg, not RS256: https://github.com/govuk-one-login/authentication-api/blob/eec178f2dfa70f7c1f6562b8ab088effea99580c/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java#L317

Does this account for that?

BenjaminWCO:
> We expect (for some reason) `RSA256` in the client reg, not RS256:
> 
> https://github.com/govuk-one-login/authentication-api/blob/eec178f2dfa70f7c1f6562b8ab088effea99580c/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java#L317
> 
> Does this account for that?

No it doesn't. I've amended. Given the logic in token handler it doesn't make sense to ever accept "RS256" as that'd actually end up defaulting to "ES256".

ethanmills (approved):

ethanmills:
Would be good to see some tests around RSA256 if possible

ethanmills (approved):

PR #4519, opened 2024-05-21
AUT-2155: add cloudwatch metrics

LazarAlexandru-Constantin:
## What
Add a Cloudwatch metric with a dimension of timeFromInitial request, which is the delta between TimeOfInitialRequest (as captured in the SQS message) and now().

## Why

The intent here is to be able to assess the length of time it has taken to get a result as a metric.

## How to review

1. Code Review

### DIFF
--- a/utils/src/main/java/uk/gov/di/authentication/utils/lambda/EmailCheckResultWriterHandler.java
+++ b/utils/src/main/java/uk/gov/di/authentication/utils/lambda/EmailCheckResultWriterHandler.java
@@ -9,24 +9,31 @@ import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.shared.entity.EmailCheckResultSqsMessage;
 import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.serialization.Json.JsonException;
+import uk.gov.di.authentication.shared.services.CloudwatchMetricsService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 import uk.gov.di.authentication.shared.services.DynamoEmailCheckResultService;
 import uk.gov.di.authentication.shared.services.SerializationService;
 
 import static uk.gov.di.authentication.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
+import static uk.gov.di.authentication.shared.helpers.NowHelper.now;
 
 public class EmailCheckResultWriterHandler implements RequestHandler<SQSEvent, Void> {
 
     private static final Logger LOG = LogManager.getLogger(EmailCheckResultWriterHandler.class);
     private final Json objectMapper = SerializationService.getInstance();
     private final DynamoEmailCheckResultService db;
+    private final CloudwatchMetricsService cloudwatchMetricsService;
 
-    public EmailCheckResultWriterHandler(DynamoEmailCheckResultService databaseService) {
+    public EmailCheckResultWriterHandler(
+            DynamoEmailCheckResultService databaseService,
+            CloudwatchMetricsService cloudwatchMetricsService) {
         this.db = databaseService;
+        this.cloudwatchMetricsService = cloudwatchMetricsService;
     }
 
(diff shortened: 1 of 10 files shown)


utils/src/main/java/uk/gov/di/authentication/utils/lambda/EmailCheckResultWriterHandler.java
```diff
@@ -59,7 +68,11 @@ public Void emailCheckResultWriterHandler(SQSEvent event) {
                 LOG.info(
                         "Message for email check reference {} written to database",
                         emailCheckResult.referenceNumber());
+                var timeFromInitialRequest =
+                        now().getTime() - emailCheckResult.timeOfInitialRequest();
 
+                cloudwatchMetricsService.putEmbeddedValue(
+                        "NumberOfAccounts", timeFromInitialRequest, Map.of());
```
mrwilson, line 75: Should this be something like `EmailCheckDuration` rather than `NumberOfAccounts`?

LazarAlexandru-Constantin: Yes, thanks!

utils/src/main/java/uk/gov/di/authentication/utils/lambda/EmailCheckResultWriterHandler.java
```diff
@@ -59,7 +68,11 @@ public Void emailCheckResultWriterHandler(SQSEvent event) {
                 LOG.info(
                         "Message for email check reference {} written to database",
                         emailCheckResult.referenceNumber());
+                var timeFromInitialRequest =
+                        now().getTime() - emailCheckResult.timeOfInitialRequest();
 
+                cloudwatchMetricsService.putEmbeddedValue(
+                        "timeFromInitialRequest", timeFromInitialRequest, Map.of());
```
mattvot, line 75: At the top of `java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java` I notice we have a set of imports to some defined metrics.

Can we do the same here, rather than create the string `"timeFromInitialRequest"`?

Looking at the pattern already set up here, I wonder if we'd be better off naming this metric something like `EmailCheckDuration`

mattvot (changes requested):
Looking good @LazarAlexandru-Constantin, just a couple of small things.

utils/src/main/java/uk/gov/di/authentication/utils/lambda/EmailCheckResultWriterHandler.java
mattvot: Good points to raise. Agreed. I would probably suggest adding a couple of dimensions. Looking at the other implementations in `CloudwatchMetricsService.java` I would suggest the following could be helpful dimensions to add
```
ACCOUNT.getValue(),
accountState.name(),
ENVIRONMENT.getValue(),
configurationService.getEnvironment(),
IS_TEST.getValue(),
Boolean.toString(isTestJourney),
```

The ones about the client are kind of irrelevant for the knowing how long the email checker took.

LazarAlexandru-Constantin: Alright, agreed
Adding them now

LazarAlexandru-Constantin: Matt, may I suggest just adding the environment, email and email check status to the cloudwatch Logs?
The other information can be retrieved only from a `userSession`, which we do not have in this handler and I am quite reluctant to add it to the `EmailCheckResultWriterHandler` as it is in the shared `utils` project

mattvot: Hey, makes sense. 
- Environment 👍, go for it
- Email, probably not, as it's PII.

mattvot (approved):

PR #4524, opened 2024-05-22
BAU/remove-case-insensitive-headers

ethanmills:
## What
This was hardcoded to false, and commit history shows it was only required for localstack


ethanmills:
> Just to double-check, this is to remove it from the Orch side only? There seem to still be some instances but only in `frontend-api`, `auth-external-api` and `shared` (which I assume is left in for the auth instances)

Yep, not touched anything that uses auth's configuration service

hannah-leong (approved):

ethanmills:
I think we should do all headers truly case insensitive. Closing this

PR #4527, opened 2024-05-22
ATO-571: create authentication callback lambda

isaac-GDS:
## Checklist


template.yaml
```diff
@@ -52,14 +57,20 @@ Mappings:
       userProfileTableKeyArn: arn:aws:kms:eu-west-2:761723964695:key/79b55133-ed01-45d8-94cd-99be01711e2c
       txmaEventEncryptionKeyArn: arn:aws:kms:eu-west-2:816047645251:key/d74787b0-8d11-4dd9-8691-7c0e26856225
       orchToAuthTokenSigningKeyAlias: arn:aws:kms:eu-west-2:761723964695:key/f2b0090b-56aa-4ff1-80fd-e8f8334199a7
+      ipvAud: undefined
```
CarlyG55, line 60: Rather than adding another mapping value, you can calculate this and give `!Ref AWS::NoValue` for the relevant environments - you can take a look at how I've done it in my draft ipv callback pr 🙂

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+      ReservedConcurrentExecutions: 1
+      LoggingConfig:
+        LogGroup: !Ref AuthenticationCallbackFunctionLogGroup
+      Environment:
+        Variables:
+          # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
+          ENVIRONMENT: !Sub ${Environment}
+          ACCOUNT_INTERVENTION_SERVICE_ACTION_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_CALL_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_URI: true
```
CarlyG55, line 950: I think this is wrong haha - we need a secret, I've set one up in my pr so you can copy from that if you like

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+          AUTHENTICATION_BACKEND_URI: !Sub
+            - https://${AuthExternalApiId}-${VpcId}.execute-api.eu-west-2.amazonaws.com/${AuthEnvironment}/
+            - AuthExternalApiId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authExternalApiId ]
+              VpcId:
```
CarlyG55, line 954: This should be the vpc endpoint id not the vpc id I think. It's also outputted from the vpc stack

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+                Fn::ImportValue: !Sub ${VpcStackName}-VpcId
+              AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
+          ACCOUNT_INTERVENTIONS_ERROR_METRIC_NAME: "AISException"
+          DYNAMO_ARN_PREFIX: !Sub
+            - 'arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-'
+            - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+              AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
+          IDENTITY_ENABLED: true
```
CarlyG55, line 962: This is false for build

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+          REDIS_KEY: session
+          TXMA_AUDIT_QUEUE_URL:
+            Fn::ImportValue: !Sub orchestration-${Environment}-txma-QueueURL
+          OIDC_API_BASE_URL: !Sub
+            - https://oidc.${ServiceDomain}/
+            - ServiceDomain: !FindInMap [ EnvironmentConfiguration, !Ref Environment, serviceDomain ]
+          FRONTEND_BASE_URL: !Sub
+            - https://signin.${ServiceDomain}/
+            - ServiceDomain: !FindInMap [ EnvironmentConfiguration, !Ref Environment, serviceDomain ]
+          ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR: true
```
CarlyG55, line 989: We don't want this true for prod or staging

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+          ACCOUNT_INTERVENTION_SERVICE_CALL_TIMEOUT: "3000"
```
CarlyG55, line 990: We don't actually need this as it falls back to 3000

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+          STORAGE_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, storageTokenKeyArn ]
+          SEND_STORAGE_TOKEN_TO_IPV_ENABLED: true
```
CarlyG55, line 992: also looks like this is false for prod/int

template.yaml
```diff
@@ -897,6 +927,287 @@ Resources:
...
+              - dynamodb:DescribeStream
+              - dynamodb:DescribeTable
+              - dynamodb:Get*
+              - dynamodb:Query
+              - dynamodb:Scan
+            Resource:
+              - !Sub
+                - 'arn:aws:dynamodb:eu-west-2:${AccountId}:table/${AuthEnvironment}-user-credentials'
+                - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+                  AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
```
CarlyG55, line 1092: This shouldn't actually need access to user credentials table I believe - the old permissions were a bit too lax

template.yaml
isaac-GDS: Just to confirm prod or integration

template.yaml
```diff
@@ -934,6 +962,299 @@ Resources:
...
+      ReservedConcurrentExecutions: 1
+      LoggingConfig:
+        LogGroup: !Ref AuthenticationCallbackFunctionLogGroup
+      Environment:
+        Variables:
+          # checkov:skip=CKV_AWS_173: These environment variables do not require encryption.
+          ENVIRONMENT: !Sub ${Environment}
+          ACCOUNT_INTERVENTION_SERVICE_ACTION_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_CALL_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_URI: !Ref AccountInterventionsServiceUriSecret
```
CarlyG55, line 985:
```suggestion
          ACCOUNT_INTERVENTION_SERVICE_URI:  !Sub
            - '{{resolve:secretsmanager:${SecretArn}:SecretString}}'
            - SecretArn: !Ref AccountInterventionsServiceUriSecret
```

I've realised the old one doesn't work, my bad 😁

template.yaml
```diff
@@ -934,6 +962,299 @@ Resources:
...
+          AUTHENTICATION_BACKEND_URI: !Sub
+            - https://${AuthExternalApiId}-${VpcId}.execute-api.eu-west-2.amazonaws.com/${AuthEnvironment}/
+            - AuthExternalApiId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authExternalApiId ]
+              VpcId: !ImportValue vpc-ExecuteApiGatewayEndpointId
+              AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
+          ACCOUNT_INTERVENTIONS_ERROR_METRIC_NAME: "AISException"
```
CarlyG55, line 991: Where did you get this from? I can't seem to find it being set in the old lambda

template.yaml
```diff
@@ -1133,7 +1454,7 @@ Resources:
           ORCH_REDIRECT_URI: !Sub
             - https://oidc.${ServiceDomain}/orchestration-redirect/
             - ServiceDomain: !FindInMap [ EnvironmentConfiguration, !Ref Environment, serviceDomain ]
-          ORCH_TO_AUTH_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, orchToAuthTokenSigningKeyAlias ]
+          ORCH_TO_AUTH_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, orchToKeyArn ]
```
CarlyG55, line 1457:
```suggestion
          ORCH_TO_AUTH_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, orchToAuthKeyArn ]
```

template.yaml
isaac-GDS: Its set as the default in the variables.tf file

isaac-GDS: Or are you just talking about the ENV variable itself ? https://github.com/govuk-one-login/authentication-api/blob/main/ci/terraform/oidc/authentication-callback.tf#L37

CarlyG55: Yes completely missed that 🤦 sorry!

template.yaml
CarlyG55: yep int sorry

CarlyG55 (dismissed):

gn-GDS (approved):

gn-GDS (approved):

gn-GDS (dismissed):

hannah-leong (dismissed):

PR #4529, opened 2024-05-22
ATO-577: Create ipv callback lambda

CarlyG55:
## What

create ipv callback in orch
## Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -33,7 +33,7 @@ Conditions:
   AisAbortOnError:
     !Not [ !Or [ !Equals [ integration, !Ref Environment ], !Equals [ production, !Ref Environment ] ] ]
   IdentityEnabled:
-    !Or [ !Equals [ staging, !Ref Environment ], !Equals [ integration, !Ref Environment ], !Equals [ production, !Ref Environment ]]
+    !Or [ !Equals [ staging, !Ref Environment ], !Equals [ integration, !Ref Environment ], !Equals [ production, !Ref Environment ] ]
 
 Mappings:
   EnvironmentConfiguration:
@@ -1634,13 +1634,14 @@ Resources:
 
   IpvCallbackFunction:
     Type: AWS::Serverless::Function
+    IgnoreGlobals: [ VpcConfig ]
     # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
     # checkov:skip=CKV_AWS_117: VPC settings are global
     Properties:
       FunctionName: !Sub ${Environment}-IpvCallbackFunction
       AutoPublishAlias: latest
       CodeUri: ./ipv-api
-      Handler: uk.gov.di.authentication.oidc.lambda.IPVCallbackHandler::handleRequest
+      Handler: uk.gov.di.authentication.ipv.lambda.IPVCallbackHandler::handleRequest
       ReservedConcurrentExecutions: 1
       LoggingConfig:
         LogGroup: !Ref IpvCallbackFunctionLogGroup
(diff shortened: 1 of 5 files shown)


CarlyG55:
deployment succeeded, will test in staging

template.yaml
```diff
@@ -857,6 +878,333 @@ Resources:
...
+            - ServiceDomain: !FindInMap [ EnvironmentConfiguration, !Ref Environment, serviceDomain ]
+          ACCOUNT_INTERVENTION_SERVICE_ACTION_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_CALL_ENABLED: true
+          ACCOUNT_INTERVENTION_SERVICE_URI: !Sub
+            - '{{resolve:secretsmanager:${SecretArn}:SecretString}}'
+            - SecretArn: !Ref AccountInterventionsServiceUriSecret
+          ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR: !If
+            - AisAbortOnError
+            - true
+            - false
```
hannah-leong, line 940: Unsure if this is possible but can this just reference `AisAbortOnError` as it is a bool of the same value?
e.g., `ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR: AisAbortOnError`

/ might this be more readable but not consistent? 
```suggestion
          ACCOUNT_INTERVENTION_SERVICE_ABORT_ON_ERROR: !If [ AisAbortOnError,  true,  false ]
```

template.yaml
```diff
@@ -857,6 +878,333 @@ Resources:
...
+            - !Ref AWS::NoValue
+          IPV_NO_SESSION_RESPONSE_ENABLED: true
+          IPV_TOKEN_SIGNING_KEY_ALIAS: !FindInMap [ EnvironmentConfiguration, !Ref Environment, ipvTokenAuthKeyArn ]
+          SPOT_QUEUE_URL: !If
+            - IdentityEnabled
+            - !Sub
+              - https://sqs.eu-west-2.amazonaws.com/${AccountId}/${AuthEnvironment}-spot-request-queue
+              - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+                AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
+            - !Ref AWS::NoValue
```
hannah-leong, line 977: Just to check, are these `IdentityEnabled` cases correct in being based off of the condition (stage, integration or production)? just because the `IDENTITY_ENABLED` variable getting passed into the function is based off of whether it is build so don't quite align but I'm aware they're meant to not be quite the same

template.yaml
```diff
@@ -857,6 +878,333 @@ Resources:
...
+              - arn:aws:sqs:eu-west-2:${AccountId}:${AuthEnvironment}-spot-request-queue
+              - AccountId: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authAccountId ]
+                AuthEnvironment: !FindInMap [ EnvironmentConfiguration, !Ref Environment, authEnvironment ]
+          - Sid: AllowEncryptionAccessToSpotQueueKey
+            Effect: Allow
+            Action:
+              - kms:GenerateDataKey
+              - kms:Decrypt
+            Resource: !FindInMap [ EnvironmentConfiguration, !Ref Environment, spotRequestQueueKmsArn ]
+        - !Ref AccountInterventionsServiceUriSecretAccessPolicy
```
hannah-leong, line 1084: What is the reasoning behind this policy being separated out but the others being statements all within the same policy? Is it worth looking into splitting them (or at least the re-used statements) up into different policies so that there is less repeated code?

template.yaml
CarlyG55: It's a bit confusing, I'm going to add a comment. It's the difference between having the ipv api turned on (IDENTITY_ENABLED), and environments where identity actually works (IdentityEnabled). Sandpit has ipv api turned on for testing even though identity doesn't work

template.yaml
CarlyG55: This one is because I copied code from elsewhere 😅 when we have a bit more time it will be nice to separate out the reused statements into policies like this, but for now I think we should leave as is because of time pressure

hannah-leong (approved):

hannah-leong (approved):

PR #4543, opened 2024-05-23
AUT-2264: Update StartHandler to correctly retrieve the TxMA header.

andrew-moores:
## What

The StartHandler was attempting to retrieve the header from the user context.  For other lambdas the user context is created in the base class, BaseFrontendHandler, and the TxMA header added at that point.  The StartHandler does not extend BaseFrontendHandler and so does not benefit from its creation and population of the user context, instead it needs to retrieve the header itself.

## How to review

Code review.

## Checklist


BeckaL (approved):
Looks sensible to me :-) Thanks for picking this up

PR #4544, opened 2024-05-23
Rff 5 signup handler test

BeckaL:
## What

Refactors to the signup handler test to clear up some noise and make the intention of tests clearer.


BeckaL:
Ah this is old now, likelihood we'll get round to fixing merge conflicts and merging is low, I'll think about doing this again

PR #4574, opened 2024-05-28
BAU: Fix ipv callback user profile call

CarlyG55:
## What

Currently ipv callback needs the user credentials table to get a user profile, but this should not be necessary


ethanmills (approved):

isaac-GDS:
Does this mean UserProfile could be null and if so do we need a test case for it ?

PR #4577, opened 2024-05-29
ATO-656: Add all missing txma permissions to orch lambdas

CarlyG55:
## What

Some lambdas were missing permissions to push to the audit queue


isaac-GDS (approved):

ethanmills (approved):

ethanmills:
Might make sense to define a single policy with these permissions then attach it to all the lambdas, instead of repeating inline?

PR #4579, opened 2024-05-29
AUT-2165: log the status of Experian email check during update email process

mark-cab:
## What

In UpdateEmail lambda in the Account Management API, added logging to check the status of the users fraudulent email check for the new email address that the user is proposing to update their profile to use and log the result.

Interim stage of email phone check process where we will log the status of the asynchronous request sent to Experian to get an idea of how long it takes.

Terraform was added to enable account management API access to fraud check dynamoDB table.

## How to review

1. Code Review

## Checklist


mattvot (approved):
Approved, but not to be merged until we pause the pipeline between build deploy and build promotion stages for manual testing with the home build environment to trigger the call to the UpdateEmail handler.

PR #4590, opened 2024-05-30
ATO-655: Add journeyid mapping

hannah-leong:
## What

Add journeyId to the auditUser if present

## How to review

Check that:
- If there is an IdToken present, the journeyId is taken from there
- If no IdToken, take the journeyId from the cookies
- If there is no journeyId in the cookies, return a null journeyId

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
@@ -26,6 +26,7 @@ import uk.gov.di.orchestration.shared.services.TokenValidationService;
 import java.net.URI;
 import java.text.ParseException;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 
 import static uk.gov.di.orchestration.shared.helpers.AuditHelper.attachTxmaAuditFieldFromHeaders;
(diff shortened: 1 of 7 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
```diff
@@ -198,6 +207,7 @@ public APIGatewayProxyResponseEvent logoutRequestHandler(APIGatewayProxyRequestE
                     Optional.of(clientID));
         }
 
+        var finalAuditUser = auditUser;
```
ethanmills, line 210: Probably don't need this?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
```diff
@@ -228,12 +238,20 @@ private void attachSessionToLogsIfExists(
...
 
             attachSessionIdToLogs(session);
-            attachLogFieldToLogs(CLIENT_SESSION_ID, sessionCookieIds.getClientSessionId());
-            attachLogFieldToLogs(GOVUK_SIGNIN_JOURNEY_ID, sessionCookieIds.getClientSessionId());
+            attachLogFieldToLogs(CLIENT_SESSION_ID, clientSessionId);
+            attachLogFieldToLogs(GOVUK_SIGNIN_JOURNEY_ID, clientSessionId);
+        }
+    }
+
+    private String extractClientSessionIdFromCookieHeaders(Map<String, String> headers) {
```
ethanmills, line 249: Might be nicer if this returned an Optional?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/LogoutHandler.java
hannah-leong: image
I added it in as IntelliJ was getting upset about it not being final or effectively final

PR #4630, opened 2024-06-04
ATO-682: Remove orch function name env variables

CarlyG55:
## What

Refactor to remove the need for function name env variables. Created a new local for secure pipelines equivalent environment
## How to review

Make sure the old names match up with new names


isaac-GDS:
Nice ! Much better then repeating it in the tf.vars file. Good stuff.

isaac-GDS:
Saw that for the TrustmarkFunction we never had specified function name will that cause problems ?

CarlyG55:
@isaac-GDS I also noticed that 😆 I'm assuming we never actually tested it - I don't think adding the name will cause issues

isaac-GDS:
Think Cloudformation is smart enough to delete the old lambda and recreate with the updated name. So happy enough :)

isaac-GDS (approved):

CarlyG55:
@isaac-GDS this won't change cloudformation, these changes are in terraform, and when I did a plan I couldn't see any changes as the string used remains the same

isaac-GDS:
Ah yes of course - silly me. All fine then :)

PR #4637, opened 2024-06-05
ATO-686: Disable spot response event mapping in int and prod

CarlyG55:
## What

If the event mapping is not disabled, the lambda will pick up messages from the queue as soon as it's deployed


template.yaml
```diff
@@ -3213,6 +3215,7 @@ Resources:
     Type: AWS::Lambda::EventSourceMapping
     Condition: IpvExists
     Properties:
+      Enabled: !If [ SpotResponseEventMappingDisabled, false, true ]
```
CarlyG55, line 3218: I know this is a bit ugly but it will only take a boolean, not a condition 🙃

hannah-leong: Is it possible to add it as another condition on the resource (where `IpvExists` is) or do we still want it to be created, just not enabled?

ethanmills (approved):

hannah-leong (approved):

PR #4639, opened 2024-06-06
ATO-661: Add method to generate storage access token for Auth to send to IPV

billrdunn:
## What

Add method to generate a signed storage access token. 

Authentication will send this storage token to IPV as a claim (named https://vocab.account.gov.uk/v1/storageAccessToken) in the JAR when identity re-verification is required. More details [here](https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/4188799112/Storage+access+token).

This is required as part of the MFA reset initiative.

## How to review

1. Code review only
2. Confirm that sonarcloud analysis can be dismissed because:
  - ConfigurationService tests are not required
  - Code duplication is between authentication shared and orchestration shared which is expected

## Checklist

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
@@ -514,4 +514,26 @@ public class ConfigurationService implements BaseLambdaConfiguration, AuditPubli
     public String getAccountInterventionsErrorMetricName() {
         return System.getenv().getOrDefault("ACCOUNT_INTERVENTIONS_ERROR_METRIC_NAME", "");
     }
+
+    public String getIPVAudience() {
+        return System.getenv().getOrDefault("IPV_AUDIENCE", "");
+    }
+
+    public String getStorageTokenSigningKeyAlias() {
+        return System.getenv("STORAGE_TOKEN_SIGNING_KEY_ALIAS");
+    }
+
+    public URI getCredentialStoreURI() {
+        return getURIOrDefault("CREDENTIAL_STORE_URI", "https://credential-store.account.gov.uk");
+    }
+
+    private URI getURIOrDefault(String envVar, String defaultUri) {
+        return getOptionalURI(envVar).orElseGet(() -> URI.create(defaultUri));
+    }
+
+    private Optional<URI> getOptionalURI(String envVar) {
+        return System.getenv().containsKey(envVar)
+                ? Optional.of(URI.create(System.getenv(envVar)))
+                : Optional.empty();
+    }
 }
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
@@ -244,7 +244,7 @@ public class TokenService {
         idTokenClaims.setClaim("vtm", trustMarkUri.toString());
 
         try {
-            return generateSignedJWT(
+            return generateSignedJwtUsingExternalKey(
                     idTokenClaims.toJWTClaimsSet(), Optional.empty(), signingAlgorithm);
         } catch (com.nimbusds.oauth2.sdk.ParseException e) {
             LOG.error("Error when trying to parse IDTokenClaims to JWTClaimSet", e);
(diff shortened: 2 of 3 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -514,4 +514,26 @@ public long getAccountInterventionServiceCallTimeout() {
     public String getAccountInterventionsErrorMetricName() {
         return System.getenv().getOrDefault("ACCOUNT_INTERVENTIONS_ERROR_METRIC_NAME", "");
     }
+
+    public String getIPVAudience() {
```
ethanmills, line 518: Could you leave a comment on the ticket that uses this to pass this variable please?

billrdunn: [Done](https://govukverify.atlassian.net/browse/ATO-660?focusedCommentId=152305)

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
```diff
@@ -278,14 +278,43 @@ private SignedJWT generateIDToken(
...
+        LOG.info("Storage token being created with JWTID: {}", jwtID);
+
+        List<String> aud =
+                List.of(
+                        configService.getCredentialStoreURI().toString(),
+                        configService.getIPVAudience());
+
+        JWTClaimsSet.Builder claimSetBuilder =
+                new JWTClaimsSet.Builder()
+                        .issuer(configService.getFrontendBaseUrl())
```
ethanmills, line 304: Let's go with this for now, but I wonder if this should be the auth frontend API domain instead? Let's start a converstation about this with Auth and VC Storage

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
```diff
@@ -278,14 +278,43 @@ private SignedJWT generateIDToken(
...
+                        .audience(aud)
+                        .expirationTime(expiryDate)
+                        .issueTime(NowHelper.now())
+                        .subject(internalPairwiseSubject.getValue())
+                        .jwtID(jwtID);
```
ethanmills, line 309: Think we're missing the `reverification` scope here?

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
ethanmills: This one doesn't block merge

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
ethanmills: This one does

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
```diff
@@ -352,13 +384,29 @@ private RefreshToken generateAndStoreRefreshToken(
         return refreshToken;
     }
 
-    public SignedJWT generateSignedJWT(
+    public SignedJWT generateSignedJwtUsingExternalKey(
```
ethanmills, line 387: Can this be private?

shared/src/main/java/uk/gov/di/authentication/shared/services/TokenService.java
```diff
@@ -352,13 +384,29 @@ private RefreshToken generateAndStoreRefreshToken(
...
+    private SignedJWT generateSignedJwtUsingStorageKey(
+            JWTClaimsSet claimsSet, Optional<String> type) {
+        return generateSignedJWT(
+                claimsSet,
+                type,
+                JWSAlgorithm.ES256,
+                configService.getStorageTokenSigningKeyAlias());
+    }
+
+    public SignedJWT generateSignedJWT(
```
ethanmills, line 405: Same for this, can it be private?

ethanmills (dismissed):

ethanmills:
I've noticed that this make two calls to KMS, one to fetch the key ID from the alias, then another to sign. How much do we gain by using the alias instead of the key ID directly? Is it worth that extra call? This applies every time we sign something, so it's potentially quite a lot of overhead

ethanmills (approved):

PR #4661, opened 2024-06-11
ATO-702: Implement provisioned concurrency

CarlyG55:
## What

Implement provisioned concurrency, and allow it to be configured per lambda if necessary. Note that cloudformation does not allow you to set it to zero, so had to do a bit of a hacky way around it! No settings for client config lambdas.


ethanmills:
I assume we can't do anything nice like set the default value to NoValue instead of 0?

CarlyG55:
@ethanmills nope, tried that :/

ethanmills (approved):

PR #4669, opened 2024-06-12
ATO-none: update the log retention for orch lambdas to 30 days

isaac-GDS:
## What

Set the default for orch lambda log retention to 30 days


template.yaml
ethanmills, line 1: Does the Globals section work for this?

ethanmills (approved):

PR #4672, opened 2024-06-12
ATO-687: Add Optional Type Adapter

BenjaminWCO:
## What

Added json type adapter for optionals.
Optionals of some value will be serialised to the inner value and vice versa.
Empty optionals will be serialised to null and vice versa.

## Why

The existing pubic key and new JWKS endpoint fields in client registry need to be nullable via the client registry APIs update handler. The update handler currently interprets both null and undefined fields in the update request as to be ignored. With this type adapter and by making those fields optionals, we will have a means of differentiating between explicit nulls which will deserialise to Optional.Empty and undefined fields which will deserialise to null.

## How to review

Code Review

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/serialization/OptionalAdapter.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/serialization/OptionalAdapter.java
@@ -1,0 +1,57 @@
+package uk.gov.di.orchestration.shared.serialization;
+
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.TypeAdapterFactory;
+import com.google.gson.reflect.TypeToken;
+import com.google.gson.stream.JsonReader;
+import com.google.gson.stream.JsonToken;
+import com.google.gson.stream.JsonWriter;
+
+import java.io.IOException;
+import java.lang.reflect.ParameterizedType;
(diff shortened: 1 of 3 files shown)


ethanmills:
[A variable whose type is Optional should never itself be null; it should always point to an Optional instance.](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html#:~:text=A%20variable%20whose%20type%20is%20Optional%20should%20never%20itself%20be%20null%3B%20it%20should%20always%20point%20to%20an%20Optional%20instance.)

BenjaminWCO:
> [A variable whose type is Optional should never itself be null; it should always point to an Optional instance.](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Optional.html#:~:text=A%20variable%20whose%20type%20is%20Optional%20should%20never%20itself%20be%20null%3B%20it%20should%20always%20point%20to%20an%20Optional%20instance.)

I appreciate it might seem like an anti-pattern, but given gson will always set a field to null if it is undefined in the JSON object (even if there would normally be no way of doing this in code) by not making an exception here we're effectively saying you can't use optionals in such a class. I suppose one could add a required attribute but then I suspect as it is written it would always fail the required field validation whether null or undefined in the JSON. I suppose a simple fix in this case is to use an optional like class with a name that makes it a little less unclear.

PR #4676, opened 2024-06-13
BAU: use batch-get-secret-value in read_secrets.sh

whi-tw:
## What

This change updates the `read_secrets.sh` script to use the `batch-get-secret-value` command from the AWS CLI to retrieve secrets from the Secrets Manager. This command allows us to retrieve multiple secrets in a single call, which should speed up the process of retrieving secrets.

Instead of reading the secrets making as many calls as there are secrets to retrieve, it pulls them in batches of 10 and exports them at once.

## How to review

Test deploying with `deploy-authdevs.sh` - if the deploy works ok, the secrets were retrieved correctly.

### DIFF
--- a/ci/terraform/account-management/read_secrets.sh
+++ b/ci/terraform/account-management/read_secrets.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail
 
 [[ "${BASH_SOURCE[0]}" != "${0}" ]] || {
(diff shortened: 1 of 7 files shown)


mattvot:
Hey @whi-tw, just tried this but got the following error:

`An error occurred (AccessDeniedException) when calling the BatchGetSecretValue operation: User: arn:aws:sts::REDACTED:assumed-role/AWSReservedSSO_REDACTED is not authorized to perform: secretsmanager:BatchGetSecretValue with an explicit deny in a service control policy`

This is after running `./scripts/set-up-sso.sh` and locally rebasing to `main` 38c3a194f1163325e2432711e73038bed6fcafef

mattvot (approved):

PR #4682, opened 2024-06-13
ATO-708: Fix identityVerificationSupported for client registry

gn-GDS:
## What

- Allow updating identityVerificationSupported in client registry
- Change identity_verification_required to identity_verification_supported

## Checklist

## Related PRs
https://govukverify.atlassian.net/jira/software/c/projects/SSE/boards/222?selectedIssue=SSE-3447

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/UpdateClientConfigRequest.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/UpdateClientConfigRequest.java
@@ -67,6 +67,10 @@ public class UpdateClientConfigRequest {
     @Expose
     private String idTokenSigningAlgorithm;
 
+    @SerializedName("identity_verification_supported")
+    @Expose
+    private Boolean identityVerificationSupported;
+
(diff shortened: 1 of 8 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/UpdateClientConfigRequest.java
```diff
@@ -67,6 +67,10 @@ public class UpdateClientConfigRequest {
     @Expose
     private String idTokenSigningAlgorithm;
 
+    @SerializedName("identity_verification_supported")
+    @Expose
+    private Boolean identityVerificationSupported;
+
```
ethanmills, line 73: Do we need to make this default to false? If we ship this as-is what impact does it have on the live SSE?

ethanmills: Discussed and agreed to add some unit tests

ethanmills (approved):
Thanks!

PR #4685, opened 2024-06-13
ATO-666: Re-verification result from IPV

isaac-GDS:
Add Base lambda
Add lambda infrastructure
Add api gateway endpoint on the fronted infrastructure

Policies and environment variables will need to be added as the lambda is developed.

### DIFF
--- a/ci/terraform/oidc/api-gateway-frontend.tf
+++ b/ci/terraform/oidc/api-gateway-frontend.tf
@@ -87,6 +87,8 @@ resource "aws_api_gateway_deployment" "frontend_deployment" {
       module.identity_progress.method_trigger_value,
       module.mfa_reset_storage_token_jwk.integration_trigger_value,
       module.mfa_reset_storage_token_jwk.method_trigger_value,
+      module.reverification_result.integration_trigger_value,
+      module.reverification_result.method_trigger_value,
       local.deploy_account_interventions_count == 1 ? module.account_interventions[0].integration_trigger_value : null,
       local.deploy_account_interventions_count == 1 ? module.account_interventions[0].method_trigger_value : null,
       local.deploy_reauth_user_count == 1 ? module.check_reauth_user[0].integration_trigger_value : null,
@@ -119,6 +121,7 @@ resource "aws_api_gateway_deployment" "frontend_deployment" {
     module.orch_auth_code,
     module.check_reauth_user,
     module.mfa_reset_storage_token_jwk,
+    module.reverification_result,
   ]
 }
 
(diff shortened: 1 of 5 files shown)


ci/terraform/oidc/re-verification-result.tf
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  handler_function_name = "uk.gov.di.authentication.frontendapi.lambda.ReverificationResultHandler::handleRequest"
+
+  rest_api_id      = aws_api_gateway_rest_api.di_authentication_frontend_api.id
+  root_resource_id = aws_api_gateway_rest_api.di_authentication_frontend_api.root_resource_id
+  execution_arn    = aws_api_gateway_rest_api.di_authentication_frontend_api.execution_arn
+
+  memory_size                 = lookup(var.performance_tuning, "re-verification-result", local.default_performance_parameters).memory
+  provisioned_concurrency     = lookup(var.performance_tuning, "re-verification-result", local.default_performance_parameters).concurrency
```
ethanmills, line 30: We can set the provisioned concurrence to 1 on this where it defaults to 3. It's going to be very low volume I expect

ci/terraform/oidc/re-verification-result.tf
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  handler_function_name = "uk.gov.di.authentication.frontendapi.lambda.ReverificationResultHandler::handleRequest"
+
+  rest_api_id      = aws_api_gateway_rest_api.di_authentication_frontend_api.id
+  root_resource_id = aws_api_gateway_rest_api.di_authentication_frontend_api.root_resource_id
+  execution_arn    = aws_api_gateway_rest_api.di_authentication_frontend_api.execution_arn
+
+  memory_size                 = lookup(var.performance_tuning, "re-verification-result", local.default_performance_parameters).memory
+  provisioned_concurrency     = 1
```
ethanmills, line 30: Ah sorry, this is not what I meant. We need to set the values in the environment files. There should still be no provisioned concurrency in dev, like the default

ethanmills (commented):
Sorry I've only just spotted this, but I think it should be reverification, not re-verification, to match the IPV endpoint and the scope name

isaac-GDS:
> Sorry I've only just spotted this, but I think it should be reverification, not re-verification, to match the IPV endpoint and the scope name

Ah nice - yeah shall do. Good spot.

ethanmills (dismissed):

ethanmills (approved):

PR #4703, opened 2024-06-14
BAU: Refactor the audit service

BeckaL:
## What

Introduces a new way of calling the audit service, intended to limit the number of parameters we pass in and repeat in each handler, making handlers easier to read and making it easier to see, for each audit event, what the important information is.

The idea is to introduce an AuditContext object (see prior art/closed PRs [here](https://github.com/govuk-one-login/authentication-api/commit/c2b3169e191f997be84b7f67c0504e970493a20d), [here](https://github.com/govuk-one-login/authentication-api/pull/4292) and [here](https://github.com/govuk-one-login/authentication-api/pull/4434), reflecting multiple independent people thinking along the same lines) to capture the common things that are submitted on all audit events within a handler, and generally either don't change between audit events, or are added to as we go in a handler as we learn more about a user.

This makes calls to the audit service much less noisy.

In this PR, we've converted two of our handlers to use this new pattern.

The idea is to incrementally introduce this across all of our handlers, so this is just a starter PR that gets the basics in and demonstrates how to migrate a handler.

## How to review

## Related PRs

The approach here is similar to a [previous refactoring](https://github.com/govuk-one-login/authentication-api/pull/4402) we did, although since we've changed our mind about this approach, hence reversing it in the login handler here.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
@@ -77,6 +77,20 @@ public class CheckReAuthUserHandler extends BaseFrontendHandler<CheckReauthUserR
             UserContext userContext) {
         LOG.info("Processing CheckReAuthUser request");
 
+        AuditContext auditContext =
+                new AuditContext(
+                        userContext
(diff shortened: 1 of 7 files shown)


gtvj (approved):
Hi @BeckaL - reviewing this was a good reminder of what we did last week. It looks good to me.

PR #4713, opened 2024-06-17
ATO-707: Fix AIS secret reference

CarlyG55:
## Checklist


billrdunn (approved):

ethanmills (approved):

billrdunn:
is three colons the correct syntax?
`SecretString:::${SecretId}`

CarlyG55:
@billrdunn it's to do with what you're referencing - you can drop colons at the end, but because I want to add a version id (not the same as other examples) you need to add the colons so that it knows what you're trying to reference: https://docs.aws.amazon.com/secretsmanager/latest/userguide/cfn-example_reference-secret.html

PR #4714, opened 2024-06-17
ATO-688: Add DynamoDB table to cache RP public keys

billrdunn:
## What

When an RP starts to share their signing key using a JWKS endpoint, Orchestration will need to retrieve and cache the key. This PRs adds a new DynamoDB table to store cached RP public keys. A key will be stored as a string in JWK format.

The table has a primary key schema with `clientId` as the partition key (HASH) and `keyId` as the sort key (RANGE).

A new KMS key is defined and used to encrypt the table.

TTL is enabled as keys will be fetched once per day.

## Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -277,6 +277,65 @@ Resources:
               ArnLike:
                 kms:EncryptionContext:aws:logs:arn: !Sub arn:aws:logs:${AWS::Region}:${AWS::AccountId}:*
 
+  #region RP Public Key DynamoDB Table
+
+  RpPublicKeyTableEncryptionKey:
+    Type: AWS::KMS::Key
+    Properties:
+      Description: KMS encryption key for RP public key DynamoDB table
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: 2012-10-17
+        Statement:
+          - Sid: AllowIamManagement
+            Effect: Allow
+            Principal:
+              AWS: !Sub arn:aws:iam::${AWS::AccountId}:root
+            Action: kms:*
+            Resource: "*"
+          - Sid: AllowDynamodbAccessToEncryptionKey
+            Effect: Allow
+            Principal:
+              Service: dynamodb.amazonaws.com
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -277,6 +277,30 @@ Resources:
...
+          AttributeType: S
+      KeySchema:
+        - AttributeName: clientId
+          KeyType: HASH
+        - AttributeName: keyId
+          KeyType: RANGE
+      BillingMode: PAY_PER_REQUEST
+      SSESpecification:
+        SSEEnabled: true
+        KMSMasterKeyId: !GetAtt MainKmsKey.Arn
```
billrdunn, line 297: I've assumed it's OK to use `MainKmsKey` for encrypting the table. Is that OK or should a separate key be defined?

billrdunn: In fact, I added encryption just to appease checkov, but is it even necessary as the data in this table isn't sensitive?

isaac-GDS: Looks like we have specific keys for tables but I think the question is more if we need encryption on this table.

isaac-GDS:
Looks good - curious on what others think on the encryption on the table.

template.yaml
ethanmills: We should have it. We don't really need the extra control or visibility it gives us in this case, but we don't really lose anything and it's not worth justifying an exception

template.yaml
```diff
@@ -277,6 +277,30 @@ Resources:
...
+      PointInTimeRecoverySpecification:
+        PointInTimeRecoveryEnabled: true
```
ethanmills, line 299: I know you've done this to appease checkov, but I think we should remove this and make an exception. It would be actively harmful to restore from a historical backup of this data, and it is intended to be fetched if not present anyway.

ethanmills (commented):
We need to enable TTL on this table

ethanmills (approved):

PR #4728, opened 2024-06-18
AUT-2701: Improve management of redis connection resources in lambdas.

andrew-moores:
The lambdas create a lot of new redis connections each time they are invoked.  These conections are not closed explicitly.  We can improve the management of these connections by moving their creation to the lambda runtime init phase.

## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
@@ -30,11 +30,28 @@ import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.g
 public class CheckReAuthUserHandler extends BaseFrontendHandler<CheckReauthUserRequest>
         implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
 
+    // INIT PHASE
     private static final Logger LOG = LogManager.getLogger(CheckReAuthUserHandler.class);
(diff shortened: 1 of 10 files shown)


BeckaL (approved):
Test approval - not merging

BeckaL (dismissed):
Another test approval - not merging

andrew-moores:
LGTM

pskushwaha1 (approved):
Approved as already reviewed by Andrew and Becka

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
```diff
@@ -30,11 +30,29 @@
...
-    private final CodeStorageService codeStorageService;
+    static {
+        if (System.getProperty("TEST") == null) {
+            redis = new RedisConnectionService(ConfigurationService.getInstance());
+        }
+    }
+
+    // FUNCTION INIT
+    public CheckReAuthUserHandler() {
+        this(ConfigurationService.getInstance(), redis);
```
andyloughran, line 45: If the connection breaks; is there anything here that will re-connect - or will the lambda simply have a broken connection?

andrew-moores: At the moment the first call that uses the broken connection will fail and cause an RTE to be thrown which, in turn, will cause the lambda to restart.  If we see improvement from this change then the next step will be to add in better handling of errors on the redis client including broken connections.

PR #4741, opened 2024-06-19
ATO-7: Turn on provisioned concurrency in production

CarlyG55:
## Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -39,7 +39,7 @@ Conditions:
   SpotResponseEventMappingDisabled:
     !Or [ !Equals [ integration, !Ref Environment ], !Equals [ production, !Ref Environment ] ]
   EnableProvisionedConcurrency:
-    !Or [ !Equals [!Ref Environment, staging], !Equals [!Ref Environment, build], !Equals [!Ref Environment, integration]]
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -39,7 +39,7 @@ Conditions:
   SpotResponseEventMappingDisabled:
     !Or [ !Equals [ integration, !Ref Environment ], !Equals [ production, !Ref Environment ] ]
   EnableProvisionedConcurrency:
-    !Or [ !Equals [!Ref Environment, staging], !Equals [!Ref Environment, build], !Equals [!Ref Environment, integration]]
+    !Or [ !Equals [!Ref Environment, staging], !Equals [!Ref Environment, build], !Equals [!Ref Environment, integration], !Equals [!Ref Environment, production]]
```
isaac-GDS, line 42: Can we switch to`!Not [ !Equals [!Ref Environment, dev] ]` eventually ?

isaac-GDS (approved):

template.yaml
CarlyG55: Yep, I was thinking to do this when cleaning up after. but just wanted to definitely convince myself we're doing the right thing in production this time 😅

isaac-GDS: Cool - was thinking the same thing. Good stuff!

PR #4745, opened 2024-06-19
Bau/update mfa handler audit calls

BeckaL:
## What

Converts the MFA Handler to use the new submitAuditEvent signature with an audit context.

This removes a lot of noise from the handler as well as its tests, and makes it easier to read the information that is most important in both.

Also a small refactor to introduce a variable for the metadata pairs which are the same in all cases in this handler.

## How to review


alhcomer (approved):
LGTM

PR #4747, opened 2024-06-19
ATO-689: Create lambda to fetch RP JWKS

billrdunn:
### Overview
We want to be able to fetch RP public keys from a JWKS endpoint that RPs own and publish, but lambdas in our existing VPC cannot call a public URL. The solution is to create a new lambda, outside our VPC, which can call public URLs. 

### New lambda
The new lambda (FetchJwksFunction) accepts two parameters, `keyId` and `url`, and returns a response object which will contain either a `JWK` or `ErrorObject` object. It will be directly invoked by AuthorisationHandler (implementation in a subsequent PR).

### VPC endpoint
To allow communication between AuthorisationHandler (in our VPC) and FetchJwksHandler, a lambda VPC endpoint has been configured. As we use secure pipelines we don’t provision the VPC in our own template file - instead a parameter (LambdaApiEnabled) has been set in stack-orchestration VPC parameters. Note that this has only been enabled in dev so far, to allow for careful testing before deploying to higher environments.

### Permissions boundary
Dev platform publish information about the [programmatic permissions boundary](https://govukverify.atlassian.net/wiki/spaces/PLAT/pages/3394142260/Programmatic+Permissions+Boundary) which contains an allowed list of services. This PR adds Lambda as an allowed service, again only to the dev environment for now.

### Next up
A subsequent PR will include:
- The code to actually invoke FetchJwksFunction, including IAM permissions
- Alerting configuration in case fetching the JWKS fails

## How to review

1. Code Review, commit by commit
2. Confirm that the changes to DocAppAuthorisationService don't affect any functionality

## Checklist

### DIFF
--- a/ci/stack-orchestration/configuration/dev/vpc/parameters.json
+++ b/ci/stack-orchestration/configuration/dev/vpc/parameters.json
@@ -30,5 +30,9 @@
   {
     "ParameterKey": "DynamoDBApiEnabled",
     "ParameterValue": "Yes"
+  },
+  {
+    "ParameterKey": "LambdaApiEnabled",
+    "ParameterValue": "Yes"
   }
 ]
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/FetchJwksResponse.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/FetchJwksResponse.java
@@ -3,31 +3,4 @@ package uk.gov.di.authentication.oidc.entity;
 import com.nimbusds.jose.jwk.JWK;
 import com.nimbusds.oauth2.sdk.ErrorObject;
 
-public class FetchJwksResponse {
-
-    private JWK jwk = null;
-    private ErrorObject error = null;
-
-    public FetchJwksResponse() {}
-
-    public FetchJwksResponse(JWK jwk, ErrorObject error) {
-        this.jwk = jwk;
-        this.error = error;
-    }
-
-    public JWK getJwk() {
-        return jwk;
-    }
-
-    public void setJwk(JWK jwk) {
-        this.jwk = jwk;
-    }
-
-    public ErrorObject getError() {
-        return error;
-    }
-
-    public void setError(ErrorObject error) {
-        this.error = error;
-    }
-}
+public record FetchJwksResponse(JWK jwk, ErrorObject error) {}
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/FetchJwksResponse.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/FetchJwksResponse.java
@@ -1,0 +1,33 @@
+package uk.gov.di.authentication.oidc.entity;
+
+import com.nimbusds.jose.jwk.JWK;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+
+public class FetchJwksResponse {
+
+    private JWK jwk = null;
+    private ErrorObject error = null;
+
+    public FetchJwksResponse() {}
+
+    public FetchJwksResponse(JWK jwk, ErrorObject error) {
+        this.jwk = jwk;
+        this.error = error;
+    }
+
+    public JWK getJwk() {
+        return jwk;
+    }
+
+    public void setJwk(JWK jwk) {
+        this.jwk = jwk;
+    }
+
+    public ErrorObject getError() {
+        return error;
+    }
+
+    public void setError(ErrorObject error) {
+        this.error = error;
+    }
+}
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandler.java
@@ -1,0 +1,63 @@
+package uk.gov.di.authentication.oidc.lambda;
+
+import com.amazonaws.services.lambda.runtime.Context;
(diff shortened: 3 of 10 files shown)


template.yaml
```diff
@@ -335,6 +335,41 @@ Resources:
 
   #endregion
 
+  #region Fetch JWKS Lambda
+
+  FetchJwksFunction:
+    Type: AWS::Serverless::Function
+    # checkov:skip=CKV_AWS_115: Reserved concurrency will be looked into in ATO-657
+    # checkov:skip=CKV_AWS_116: DLQ is not required
+    # checkov:skip=CKV_AWS_117: Lambda needs to exist outside our VPC so default AWS Lambda service VPC is OK
```
billrdunn, line 344: I've assumed it's OK not to explicitly define a VPC for the lambda and just let exist in the default location which has public internet access
https://docs.aws.amazon.com/lambda/latest/operatorguide/networking-vpc

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandler.java
```diff
@@ -0,0 +1,60 @@
...
+            response.setJwk(jwk);
+        } catch (KeySourceException e) {
+            LOG.error("Failed to fetch JWKS: could not find key with provided key ID", e);
+            response.setError(OAuth2Error.SERVER_ERROR);
+        } catch (MalformedURLException e) {
+            LOG.error("Failed to fetch JWKS: URL is malformed", e);
+            response.setError(OAuth2Error.SERVER_ERROR);
+        } catch (IllegalArgumentException e) {
+            LOG.error("Failed to fetch JWKS: url and/or keyId parameters not present", e);
+            response.setError(OAuth2Error.SERVER_ERROR);
```
billrdunn, line 56: Is an OAuth2Error appropriate here? If so, is SERVER_ERROR reasonable?

ci/stack-orchestration/configuration/dev/vpc/parameters.json
```diff
@@ -30,5 +30,9 @@
   {
     "ParameterKey": "DynamoDBApiEnabled",
     "ParameterValue": "Yes"
+  },
+  {
+    "ParameterKey": "LambdaApiEnabled",
+    "ParameterValue": "Yes"
```
billrdunn, line 36: I've only edited the parameters for dev, not other envs for now, as making changes to the VPC will require careful testing in dev. 
ATO-694 (writing the code to invoke FetchJwksHandler) should be completed first in order to do proper testing.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandlerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.util.Map;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
+
+public class FetchJwksHandlerTest {
```
billrdunn, line 18: I spent a while but couldn't figure out a way of testing the happy path by mocking the response from `jwksService.retrieveJwkFromURLWithKeyId`. I thought it would be OK to cover this case in an integration test once the code to invoke FetchJwksHandler is merged (in a subsequent PR).

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/FetchJwksResponse.java
ethanmills, line 1: Can this be a record?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandler.java
ethanmills: It feels a bit odd to be using this here. I agree this is what we should be returning to the RP, but it feels like we're leaking that context into a function that could be pretty separate from that

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/JwksService.java
```diff
@@ -75,15 +75,11 @@ public JWK retrieveJwkFromURLWithKeyId(URL url, String keyId) {
...
-                    .orElseThrow(() -> new KeySourceException("No key found with given keyID"));
-        } catch (KeySourceException e) {
-            LOG.error("Unable to load JWKSet", e);
-            throw new RuntimeException(e);
-        }
+
+        LOG.info("Retrieving JWKSet with URL: {}", url);
+        return jwkSource.get(selector, null).stream()
+                .findFirst()
+                .orElseThrow(() -> new KeySourceException("No key found with given keyID"));
```
ethanmills, line 82: Can we add the key ID to this or log it somewhere please?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandlerTest.java
billrdunn: Update: figured it out

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/FetchJwksHandler.java
billrdunn: Have updated it, see what you think

ethanmills (approved):

PR #4751, opened 2024-06-19
ATO-687: Add Public Key Source and JWKS URL fields to Client Registry…

BenjaminWCO:
… API

## What

- Added JWKS URL Field
- Added Public Key Source Field to determine usage of Public Key / JWKS URL fields
- Added Validation logic to make sure only on of Public Key or JWKS URL is set
- Added logic to handle existing RP config that doesn't confirm to new expectations (missing Public Key Source field - will still allow Public Key only if JWKS URL field is null and Public Key Source is null to allow release before SSE start sending Public Key Source)

## How to review


BenjaminWCO:
Looks like SonarCloud doesn't like the duplication between the create request, update request and response. It does make changing the API a bit of a pain. Perhaps we cold do something like use a single class and move the non null field requirements into the validator class? I think there might be a field or two already missing from the response which I assume was unintentional.

BenjaminWCO:
> Looks like SonarCloud doesn't like the duplication between the create request, update request and response. It does make changing the API a bit of a pain. Perhaps we could do something like use a single class and move the non null field requirements into the validator class? I think there might be a field or two already missing from the response which I assume was unintentional.

Yeah, it looks like we don't reply with "identity_verification_supported" or "accepted_levels_of_confidence"

ethanmills:
> Added logic to handle existing RP config that doesn't confirm to new expectations (missing Public Key Source field)

Could you expand on this point a little please? From a first look it seems to me that this change would stop the Self Service Tool from working until it's updated to send the new field?

BenjaminWCO:
> > Added logic to handle existing RP config that doesn't confirm to new expectations (missing Public Key Source field)
> 
> Could you expand on this point a little please? From a first look it seems to me that this change would stop the Self Service Tool from working until it's updated to send the new field? As such, I'm surprised the contract tests pass.

I've made it backwards compatible so this can released without SSE sending the source field.

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationRequest.java
```diff
@@ -29,11 +30,18 @@ public class ClientRegistrationRequest {
...
+    @SerializedName("public_key_source")
+    @Expose
+    private String publicKeySource = PublicKeySource.STATIC.getValue();
+
     @SerializedName("public_key")
     @Expose
-    @Required
     private String publicKey;
 
+    @SerializedName("jwks_url")
```
ethanmills, line 41:
```suggestion
    @SerializedName("jwks_uri")
```
See https://openid.net/specs/openid-connect-registration-1_0.html#:~:text=Section%C2%A02.1.-,jwks_uri,-OPTIONAL.%20URL%20for

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
```diff
@@ -75,9 +78,23 @@ public ClientRegistry withClientName(String clientName) {
         return this;
     }
 
+    @DynamoDbAttribute("PublicKeySource")
+    public String getPublicKeySource() {
+        return Optional.ofNullable(publicKeySource).orElse(PublicKeySource.STATIC.getValue());
```
ethanmills, line 83: Doesn't matter much here because it's a simple method with no side effects, but calling a method in an `orElse` is typically not what we want to do because it's evaluated even if the value is not needed. Instead we should do something like 
```suggestion
        return Optional.ofNullable(publicKeySource).orElseGet(PublicKeySource.STATIC::getValue);
```
which will only execute if it's needed

BenjaminWCO: Done

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/entity/ClientRegistrationRequest.java
BenjaminWCO: Fixed in request and response classes

ethanmills (approved):

PR #4767, opened 2024-06-21
ATO-726: fix storage token claim and update test

isaac-GDS:
## What
 
- returns storage token in claims request
 - update test to assert claims
 - add test to look at storage claim being passed through with test on flag

## How to review

if storage token claim is being tested correctly

## Testing

To thoroughly test will need to see if claim appears in IPV request in staging. Check with IPV for decrypted payload to see if storage token claim is there. Shouldn't have a need to pause to production as the storage token code exists there already.

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationServiceTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationServiceTest.java
@@ -262,19 +262,20 @@ public class InitiateIPVAuthorisationServiceTest {
                         LEVELS_OF_CONFIDENCE);
 
         assertThat(response, hasStatus(302));
-        verify(tokenService).generateStorageToken(any());
+        verify(tokenService).generateStorageToken(any(Subject.class));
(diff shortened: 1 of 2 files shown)


Ryan-Andrews99 (approved):

oidc-api/src/test/java/uk/gov/di/authentication/oidc/services/InitiateIPVAuthorisationServiceTest.java
```diff
@@ -312,7 +360,7 @@ public static Map<String, String> splitQuery(String stringUrl) {
         return query_pairs;
     }
 
-    private AuthenticationRequest createAuthenticationRequest() {
+    private AuthenticationRequest createAuthenticationRequest(ClaimsSetRequest claimsSetRequest) {
```
CarlyG55, line 363: I might be misreading this, but I think it would be nice to change this variable name, as it seems to be the same as the thing we're passing in?

Ryan-Andrews99: Have updated!

CarlyG55 (approved):

PR #4769, opened 2024-06-21
ATO-7: Turn on logout and spot response event mapping in production, disable auth spot response integration

CarlyG55:
## Checklist


template.yaml
```diff
@@ -2513,7 +2511,7 @@ Resources:
     Type: AWS::Lambda::EventSourceMapping
     Condition: IpvExists
     Properties:
-      Enabled: !If [ SpotResponseEventMappingDisabled, false, true ]
+      Enabled: true
```
CarlyG55, line 2514: I know this is unnecessary but it makes me less nervous 😅 will remove later

PR #4778, opened 2024-06-24
ATO-739: Remove kms cross account feature flag

CarlyG55:
## What

Removing the feature flag for the auth orch split to all kms keys. I've done a tf plan against sandpit and can't see any changes
## Checklist

### DIFF
--- a/ci/terraform/oidc/ecc-signing-key.tf
+++ b/ci/terraform/oidc/ecc-signing-key.tf
@@ -4,7 +4,7 @@ resource "aws_kms_key" "storage_token_signing_key_ecc" {
   key_usage                = "SIGN_VERIFY"
   customer_master_key_spec = "ECC_NIST_P256"
 
-  policy = var.kms_cross_account_access_enabled ? data.aws_iam_policy_document.storage_token_signing_key_access_policy_with_orch_access.json : data.aws_iam_policy_document.storage_token_signing_key_access_policy.json
+  policy = data.aws_iam_policy_document.storage_token_signing_key_access_policy.json
 
   tags = local.default_tags
 }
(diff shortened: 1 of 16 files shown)


ci/terraform/oidc/ecc-signing-key.tf
```diff
@@ -14,7 +14,7 @@ resource "aws_kms_alias" "storage_token_signing_key_alias" {
   target_key_id = aws_kms_key.storage_token_signing_key_ecc.key_id
 }
 
-data "aws_iam_policy_document" "storage_token_signing_key_access_policy_with_orch_access" {
+data "aws_iam_policy_document" "storage_token_signing_key_access_policy" {
```
CarlyG55, line 17: After I made this rename I wasn't sure I liked it - I guess it depends where these KMS keys are going to end up - if they stay in auth, at some point we're going to want to remove cross account access

ethanmills: This one can move to Orch accounts. We can check with Bravo, but we could probably just do it now given it's not in use yet

CarlyG55: Sorry I meant in general changing these names to remove `with_orch_access`

ethanmills: I think , given auth's migration plans, it's unlikely any of these keys stay in this aws account long term. They'll either be moved to ours or theirs. I'm happy to just drop the `with_orch_access` and these will be cleaned up in time

ethanmills (approved):

ci/terraform/oidc/ecc-signing-key.tf
CarlyG55: Nice, thanks

PR #4782, opened 2024-06-24
ATO-674: Enforce cloudfront in production

ethanmills:
## What

Same as https://github.com/govuk-one-login/authentication-api/pull/4677 but for production. This changes the firewall on the API gateway to one which requires all traffic to have the secret header set by CloudFront.

I have:
 - Checked all traffic is routed via CF: 
image

 - Checked the WAF expects the correct values for the secret header


billrdunn (approved):

billrdunn:
Is there a ticket for removing the `enforce_cloudfront` flag?

PR #4794, opened 2024-06-25
Bau: Use new audit service method in mfa handlers

BeckaL:
## What

Uses our updated audit service method, using an audit context to share some common variables, in two more handlers.

Also refactors some of the audit service code in these handlers (construction of metadata pairs) to further tidy calls to the audit service up.

## How to review

## Related PRs

An [example PR](https://github.com/govuk-one-login/authentication-api/pull/4766) where we previously converted another handler

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -32,6 +32,7 @@ import uk.gov.di.authentication.shared.services.DynamoAccountModifiersService;
 import uk.gov.di.authentication.shared.services.SessionService;
 import uk.gov.di.authentication.shared.state.UserContext;
 
+import java.util.ArrayList;
 import java.util.Map;
 import java.util.Optional;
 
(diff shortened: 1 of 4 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -254,8 +239,33 @@ private void processSuccessfulCodeRequest(
...
+            JourneyType journeyType,
+            VerifyCodeRequest codeRequest,
+            Integer loginFailureCount,
+            Boolean successfulRequest) {
+        AuditService.MetadataPair[] basicMetadataPairs = {
+            pair("notification-type", notificationType.name()),
+            pair("account-recovery", journeyType.equals(JourneyType.ACCOUNT_RECOVERY)),
+            pair("journey-type", String.valueOf(journeyType))
+        };
+        List<AuditService.MetadataPair> elementsList = new ArrayList<>(List.of(basicMetadataPairs));
```
mrwilson, line 259: We're starting with an array, then creating a list, then converting back to an array in the final line.

Could we start with just a list instead? e.g. `var elementsList = Arrays.asList(pair("notification-type", ...)...)`

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -254,8 +239,33 @@ private void processSuccessfulCodeRequest(
...
+        if (notificationType.equals(MFA_SMS)) {
+            elementsList.add(pair("mfa-type", MFAMethodType.SMS.getValue()));
+            elementsList.add(pair("loginFailureCount", loginFailureCount));
+            elementsList.add(pair("MFACodeEntered", codeRequest.code()));
+            if (Boolean.FALSE.equals(successfulRequest)) {
```
mrwilson, line 264: Is there a reason we're using boxed booleans in this method rather than primitive booleans? We could write `if (!successfulRequest) {...}`

BeckaL: Yeah because sonarcloud complains about that first syntax 😢

mrwilson (approved):

BeckaL:
Just kicked this off the queue because I want to shepherd another change in first, will re-merge when that's all done

PR #4798, opened 2024-06-25
ATO-738: Add orch account ids to dev environments

CarlyG55:
## What

This env variable is currently not set in authdev environments, so it may cause issues when the cross account flags are removed
## Checklist


whi-tw (approved):
LGTM

PR #4811, opened 2024-06-27
BAU: Minor tidyings to AuditService (auth)

mrwilson:
I was pairing with Becka and noticed a few things that I could just make a small PR for.


BeckaL (approved):
Lovely stuff, LGTM

PR #4816, opened 2024-06-28
AUT-2871-Codedeploy auth lambdas

pskushwaha1:
## What

Code Deploy to shift lambda traffic  Post new version of Lambda Deployment 

## Why 

Observation during Performance testing have shown that when lambda is running in concurrency to process request & we have a new version of lambda getting deployed same time  the Traffic switch to new lambda is done instantly which cause lambda Throttling as old version is processing existing request before the concurrency is released back & new version scale up to Process new request .

To mitigate this situation we are having a Blue green deployment  with Linear shifting the Traffic by 10% every 1 minute hopefully this will help processing the new request on new Version  & existing request on old version Gracefully without throttling 

What will this solution do :
1) It will Shift Alias traffic to new version in linear way Blue/Green
2) It will Auto rollback in case of Lambda Deployment Failure  in init phase or other Failure to previous version
3) In case of Manual roll back we can always re deploy the Previous Version on Lambda using click ops Single click operation.

**This PR also include the trust initiative related to have a Blue/green deployment in pipeline** 

**_`**Below Both version throttling  when instance traffic switch is done**`_** 
image

**_`**Provisioned Currency getting reseted on new version**`_** 

image

**_`**Code Deploy traffic Shift**`_**

image

**_`**Weight in Lambda Version**`_**

image

**_`**Code Deploy App View**`_** 
image

**_`**Deployment**`_** 

image


pskushwaha1:
Closing this PR Created new PR ( 5292 ) with Commit msg fix

PR #4817, opened 2024-06-28
Bau: Convert more handlers to use the new audit service method

BeckaL:
## What

Another PR to convert handlers to use the new way of calling the audit service. I had ambitions for this to be the final PR for frontend handlers but not quite there and didn't want to make this PR too big.

This simplifies call sites and tests by introducing a shared audit context object with common variables used across all events in a handler.

## How to review

## Related PRs

A previous PR where we've done this for some other handlers: https://github.com/govuk-one-login/authentication-api/pull/4794

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
@@ -7,9 +7,9 @@ import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent
 import com.nimbusds.oauth2.sdk.id.Subject;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent;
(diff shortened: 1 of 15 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
```diff
@@ -105,28 +106,23 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
                         AuditService.UNKNOWN,
                         request.getEmail(),
                         IpAddressHelper.extractIpAddress(input),
                         AuditService.UNKNOWN,
                         PersistentIdHelper.extractPersistentIdFromHeaders(input.getHeaders()),
-                        restrictedSection);
+                        Optional.ofNullable(userContext.getTxmaAuditEncoded()));
+
+        if (passwordValidationError.isEmpty()) {
+            LOG.info("No password validation errors found");
```
andrew-moores, line 122: Is this going to be useful in prod?

BeckaL: You're probably right that it's not, but I didn't add this line here - github diff here not being particularly smart. So I'd propose not to touch it, but I think maybe we as a team should discuss logging strategy.

andrew-moores (approved):

PR #4836, opened 2024-07-03
BAU: Introduce constructor method for audit context

BeckaL:
In most handlers, half of the fields in the audit context come from the user context. Therefore, creating a constructor which handles the retrieval of values from the user context makes the handler code less verbose.

I used my editor to optimise the imports in some handlers, which also reversed some wildcard imports, which we should be avoiding.

## How to review

1. Code Review


LazarAlexandru-Constantin (approved):
Nice refactoring job!

gtvj (approved):

PR #4842, opened 2024-07-03
BAU: Final conversion of all handlers to the new audit service method

BeckaL:
## What

Conversion of all remaining authentication code to use the new method of calling the audit service, which gets an audit context passed through to it in order to clean up some of the calling code and tests.

Removes the old method, as our conversion is now complete!

## How to review

## Related PRs

An example PR where we've done this conversion most recently: https://github.com/govuk-one-login/authentication-api/pull/4848

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
@@ -73,10 +73,6 @@ public class AuditService {
                                 txmaAuditEvent.addExtension("phone_number_country_code", country));
     }
 
-    public record RestrictedSection(Optional<String> encoded) {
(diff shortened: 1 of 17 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
```diff
@@ -103,34 +99,6 @@ public void submitAuditEvent(
...
-                        subjectId,
-                        email,
-                        ipAddress,
-                        phoneNumber,
-                        persistentSessionId,
-                        restrictedSection.encoded);
-
-        submitAuditEvent(event, auditContext, metadataPairs);
-    }
-
```
andrew-moores, line 133: 👍

andrew-moores (approved):

PR #4848, opened 2024-07-03
Bau: Convert account management api to new audit service method

BeckaL:
## What

Converts all the handlers in account management to use the new audit service method.

In the majority of cases this cleans up the call sites (as it introduces audit context that can be shared between all calls in the handler). It's also a step towards retiring the old method.

## How to review

1. Code Review

## Related PRs

A previous example PR where we did similar conversions in the frontend api: https://github.com/govuk-one-login/authentication-api/pull/4824


VladGavrilet (approved):
:rocket:

dbes-gds (approved):

PR #4866, opened 2024-07-08
ATO-887: Return bad request when redirect uri is invalid

CarlyG55:
## What

We're getting a lot of alarms go off in integration due to invalid redirect uri's being provided - we should return 400 rather than 500. Also renamed exception to make it clearer what case we are returning 400 in.

## Checklist


hannah-leong (dismissed):

ethanmills (approved):

ethanmills:
Would be good to get a test of the validator, that it throws the corrct exception

ethanmills (approved):

PR #4877, opened 2024-07-10
ATO-732: Pass Reauthentication Failures to RP

BenjaminWCO:
## What

- Refactored Logout Service / Logout Handler to destroy session in Logout Service rather than Logout Handler.
- Pass Reauthentication Failures To RP
- Add "logout reason" extension to LOG_OUT_SUCCESS audit event.

## How to review

Code Review

## Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -4,8 +4,6 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
-import com.nimbusds.oauth2.sdk.ErrorObject;
-import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.ResponseMode;
 import com.nimbusds.oauth2.sdk.TokenResponse;
@@ -23,6 +21,7 @@ import uk.gov.di.authentication.ipv.services.IPVAuthorisationService;
 import uk.gov.di.authentication.oidc.domain.OidcAuditableEvent;
 import uk.gov.di.authentication.oidc.domain.OrchestrationAuditableEvent;
 import uk.gov.di.authentication.oidc.exceptions.AuthenticationCallbackException;
+import uk.gov.di.authentication.oidc.exceptions.AuthenticationCallbackValidationException;
 import uk.gov.di.authentication.oidc.services.AuthenticationAuthorizationService;
 import uk.gov.di.authentication.oidc.services.AuthenticationTokenService;
 import uk.gov.di.authentication.oidc.services.InitiateIPVAuthorisationService;
@@ -228,6 +227,7 @@ public class AuthenticationCallbackHandler
         ThreadContext.clearMap();
         LOG.info("Request received to AuthenticationCallbackHandler");
         attachTxmaAuditFieldFromHeaders(input.getHeaders());
+
         try {
             CookieHelper.SessionCookieIds sessionCookiesIds =
                     cookieHelper.parseSessionCookie(input.getHeaders()).orElse(null);
@@ -273,13 +273,11 @@ public class AuthenticationCallbackHandler
             String clientId = authenticationRequest.getClientID().getValue();
             attachLogFieldToLogs(CLIENT_ID, clientId);
 
-            boolean requestValid =
-                    authorisationService.validateRequest(
-                            input.getQueryStringParameters(), userSession.getSessionId());
-
-            if (!requestValid) {
-                return generateAuthenticationErrorResponse(
-                        authenticationRequest, OAuth2Error.SERVER_ERROR, user);
+            var validationFailureResponse =
+                    generateAuthenticationErrorResponseIfRequestInvalid(
+                            authenticationRequest, input, user, userSession);
+            if (validationFailureResponse.isPresent()) {
+                return validationFailureResponse.get();
             }
 
             auditService.submitAuditEvent(
@@ -542,26 +540,57 @@ public class AuthenticationCallbackHandler
         return dimensions;
     }
 
+    private Optional<APIGatewayProxyResponseEvent>
+            generateAuthenticationErrorResponseIfRequestInvalid(
+                    AuthenticationRequest authenticationRequest,
+                    APIGatewayProxyRequestEvent input,
+                    TxmaAuditUser user,
+                    Session session) {
+        try {
+            authorisationService.validateRequest(
+                    input.getQueryStringParameters(), session.getSessionId());
+        } catch (AuthenticationCallbackValidationException e) {
+            return Optional.of(
+                    generateAuthenticationErrorResponse(
+                            authenticationRequest, input, e, user, session));
+        }
+        return Optional.empty();
+    }
+
     private APIGatewayProxyResponseEvent generateAuthenticationErrorResponse(
             AuthenticationRequest authenticationRequest,
-            ErrorObject errorObject,
-            TxmaAuditUser user) {
+            APIGatewayProxyRequestEvent input,
+            AuthenticationCallbackValidationException exception,
+            TxmaAuditUser user,
+            Session session) {
+        var error = exception.getError();
         LOG.warn(
-                "Error in Authentication Authorisation Response. ErrorCode: {}. ErrorDescription: {}",
-                errorObject.getCode(),
-                errorObject.getDescription());
+                "Error in Authentication Authorisation Response. ErrorCode: {}. ErrorDescription: {}.{}",
+                error.getCode(),
+                error.getDescription(),
+                exception.getLogoutRequired() ? " Logging out." : "");
         auditService.submitAuditEvent(
                 OrchestrationAuditableEvent.AUTH_UNSUCCESSFUL_CALLBACK_RESPONSE_RECEIVED,
                 authenticationRequest.getClientID().getValue(),
                 user);
-        var errorResponse =
+        var errorResponseUri =
                 new AuthenticationErrorResponse(
-                        authenticationRequest.getRedirectionURI(),
-                        errorObject,
-                        authenticationRequest.getState(),
-                        authenticationRequest.getResponseMode());
-        return generateApiGatewayProxyResponse(
-                302, "", Map.of(ResponseHeaders.LOCATION, errorResponse.toURI().toString()), null);
+                                authenticationRequest.getRedirectionURI(),
+                                error,
+                                authenticationRequest.getState(),
+                                authenticationRequest.getResponseMode())
+                        .toURI();
+
+        if (exception.getLogoutRequired()) {
+            return logoutService.handleReauthenticationFailureLogout(
+                    session,
+                    input,
+                    authenticationRequest.getClientID().getValue(),
+                    errorResponseUri);
+        } else {
+            return generateApiGatewayProxyResponse(
+                    302, "", Map.of(ResponseHeaders.LOCATION, errorResponseUri.toString()), null);
+        }
     }
 
     private Long getPasswordResetTimeClaim(UserInfo userInfo) {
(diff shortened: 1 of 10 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -149,23 +148,25 @@ public APIGatewayProxyResponseEvent handleLogout(
         return generateLogoutResponse(uri, auditUser, clientId, rpPairwiseId);
     }
 
+    public APIGatewayProxyResponseEvent handleReauthenticationFailureLogout(
+            Session session,
+            APIGatewayProxyRequestEvent input,
+            String clientId,
+            URI errorRedirectUri) {
+        var auditUser = createAuditUser(input, session);
+        destroySessions(session);
```
CarlyG55, line 162: is there a reason we don't check for the existence of the sessions here?

CarlyG55 (commented):
Nothing major here, just a few little bits. The splitting out of the code into separate commits here was super helpful, made it a lot easier to review - thank you!

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
```diff
@@ -253,20 +249,21 @@ void successfullyReturnsErrorLogoutResponse() {
...
 
         verify(auditService)
                 .submitAuditEvent(
                         LOG_OUT_SUCCESS,
-                        AuditService.UNKNOWN,
+                        CLIENT_ID,
                         auditUser,
                         AuditService.MetadataPair.pair("rpPairwiseId", rpPairwiseId.get()));
-        verifyNoInteractions(cloudwatchMetricsService);
+        verify(cloudwatchMetricsService).incrementLogout(Optional.of(CLIENT_ID));
```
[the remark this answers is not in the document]
BenjaminWCO: Ah, so previously destroying the session and incrementing the logout metric was handled in the logout handler class in the case of a logout request but in the relevant handle method in the logout service in the case of an intervention. Given I was adding a third type of logout handle method to the logout service I thought it was best to refactor the service so all three followed the same pattern and did this logic in their respective logout service handle methods. This `verifyNoIntercations` check was presumably here to make sure the metric didn't get incremented twice (once in the logout handler and once in the logout service). It doesn't get called in the logout handler anymore and it actually should get called in the logout service now.

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
```diff
@@ -328,46 +325,24 @@ void shouldDeleteSessionFromRedisWhenNoCookieExists() {
...
-        verify(backChannelLogoutService)
-                .sendLogoutMessage(
-                        argThat(withClientId("client-id-3")), eq(EMAIL), eq(INTERNAL_SECTOR_URI));
-
-        verify(clientSessionService).deleteStoredClientSession("client-session-id-1");
-        verify(clientSessionService).deleteStoredClientSession("client-session-id-2");
-        verify(clientSessionService).deleteStoredClientSession("client-session-id-3");
-        verify(sessionService, times(1)).deleteSessionFromRedis(SESSION_ID);
-    }
-
```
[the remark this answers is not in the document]
BenjaminWCO: related to the above, `destroySessions` was public because it was called in the case of a logout handler in the case of a logout request but this logic has been moved into the handle logout method. I should have made destroySessions private. Will fix.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -542,26 +539,56 @@ private Map<String, String> buildDimensions(
         return dimensions;
     }
 
+    private Optional<APIGatewayProxyResponseEvent> validateRequest(
```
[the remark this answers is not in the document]
BenjaminWCO: To be honest I only moved this to a method because sonar didn't like the introduction of a nested try-catch (despite this already being present). A bit of a mouthful but I've renamed to `generateAuthenticationErrorResponseIfRequestInvalid`

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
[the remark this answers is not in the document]
BenjaminWCO: Done

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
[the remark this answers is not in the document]
CarlyG55: I see, makes sense

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
CarlyG55: Ah if we're already doing a previous session exists check then I'm not too worried

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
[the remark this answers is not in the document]
CarlyG55: I still don't think we should be deleting this test - this is an important bit of logic that we want to make sure works. We should still be doing this testing somewhere

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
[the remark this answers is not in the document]
CarlyG55: Makes sense - thanks for that

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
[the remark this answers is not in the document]
BenjaminWCO: I think most of this logic is covered in other tests but you're right, it does look like none of the other tests check that `backChannelLogoutService.sendLogoutMessage` is called or test for multiple sessions so I think it makes sure these are still covered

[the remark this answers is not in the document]
BenjaminWCO: Okay, I've added verifications for calls to `backChannelLogoutService.sendLogoutMessage` in the various other tests and a `sessionsAreAllDeletedOnLogout` tests which does much the same as the above but via a `handleLogout` call rather than a call to `logoutService.destroySessions`.

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
```diff
@@ -328,15 +375,20 @@ void shouldDeleteSessionFromRedisWhenNoCookieExists() {
...
         logoutService.handleLogout(
+                Optional.of(session),
                 Optional.empty(),
                 Optional.of(CLIENT_LOGOUT_URI),
                 Optional.of(STATE.getValue()),
                 auditUserWhenNoCookie,
                 Optional.empty(),
                 rpPairwiseId);
 
-        verify(sessionService, times(0)).deleteSessionFromRedis(SESSION_ID);
```
BenjaminWCO, line 338: Although due to the refactor this test is now testing for a session being deleted as it's name suggests, the fact that it wasn't before makes me think this was for some old logic that has now changed. The auditUserWhenNoCookie doesn't seem to trigger any specific logic?

BenjaminWCO: After looking into this it appears this test always asserted deleteSessionFromRedis was not called despite it's name. I'm removing as it's confusingly named and for the current version f the logout service doesn't test any unique logic.

isaac-GDS: Hesitant to delete the test even though its confusingly not properly tested. Since this has gone through a refactor is it that the test hasn't been setup correctly. Seems like the test was purposefully named and should be kept and updated unless someone can confirm this logic change is fine.

isaac-GDS: Should this test be moved into the logoutHandler as thats where the cookie logic is - testing that the `logoutService.destroySession` is still called when no cookie exists.

BenjaminWCO: @billrdunn do you have any thoughts? It looks like this was here since you created this class: https://github.com/govuk-one-login/authentication-api/pull/3647/commits/f374da8b0bf5193ae982c005e38ba94a69323adc

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -80,21 +82,15 @@ public LogoutService(
 
     private APIGatewayProxyResponseEvent generateLogoutResponse(
             URI logoutUri,
-            Optional<String> state,
             TxmaAuditUser auditUser,
             Optional<String> clientId,
             Optional<String> rpPairwiseId) {
         LOG.info("Generating logout response using URI: {}", logoutUri);
```
ethanmills, line 90: Not sure we should be logging the state here as it's arbitrary RP controlled input and we aren't sure it's not PII

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
ethanmills: Looks like there is a change in behaviour around emitting metrics though?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -177,21 +192,41 @@ public APIGatewayProxyResponseEvent handleAccountInterventionLogout(
...
-                        .orElse(new AuditService.MetadataPair[] {});
-        auditService.submitAuditEvent(LOG_OUT_SUCCESS, auditClientId, auditUser, metadata);
+        var metadata = new LinkedList<AuditService.MetadataPair>();
+        metadata.add(pair(LOGOUT_REASON, logoutReason.getValue()));
+        rpPairwiseId.ifPresent(i -> metadata.add(pair("rpPairwiseId", i)));
+        auditService.submitAuditEvent(
+                LOG_OUT_SUCCESS,
+                auditClientId,
+                auditUser,
+                metadata.toArray(new AuditService.MetadataPair[0]));
```
ethanmills, line 215: What's the difference between this and 
```suggestion
                metadata.toArray(AuditService.MetadataPair[]::new));
```
?

ethanmills:
I might be wrong, but it looks like we were only previously incrementing the cloudwatch metric if there was a session, but now we always do?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
BenjaminWCO: Sure, only logging the path now.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
BenjaminWCO: Agreed, the second one is better

BenjaminWCO:
> I might be wrong, but it looks like we were only previously incrementing the cloudwatch metric if there was a session, but now we always do?

That's true. I've fixed to remain consistent and added a test.

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
BenjaminWCO: As discussed yesterday, will maintain the current behaviour for now, which dosen't delete the session if the cookie is missing. I've opened a new ticket to investigate the desired behaviour.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -80,21 +83,17 @@ public LogoutService(
...
             TxmaAuditUser auditUser,
             Optional<String> clientId,
             Optional<String> rpPairwiseId) {
-        LOG.info("Generating logout response using URI: {}", logoutUri);
-        var uri =
-                state.map(s -> buildURI(logoutUri, Map.of(STATE_PARAMETER_KEY, s)))
-                        .orElse(logoutUri);
-
-        sendAuditEvent(auditUser, clientId, rpPairwiseId);
+        LOG.info("Generating logout response using URI: {}", logoutUri.getPath());
```
ethanmills, line 90: I think this needs host and path to be meaningful?

BenjaminWCO: Yes, fixed

ethanmills (approved):

PR #4879, opened 2024-07-10
BAU: Fix Error On Update Client Without PK

BenjaminWCO:
## What

- Fix updates to Client Reg without PK or JWKS_URL failing. Only need to enforce the presence of a public key source on creation.

## How to review

Code Review

## Checklist

### DIFF
--- a/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/services/ClientConfigValidationService.java
+++ b/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/services/ClientConfigValidationService.java
@@ -81,7 +81,9 @@ public class ClientConfigValidationService {
                 .orElse(true)) {
             return Optional.of(RegistrationError.INVALID_REDIRECT_URI);
         }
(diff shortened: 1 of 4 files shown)


client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/services/ClientConfigValidationService.java
```diff
@@ -213,8 +217,8 @@ private boolean isPublicKeySourceValid(String publicKeySource) {
     }
 
     private boolean isPublicKeySourceConsistent(
-            String publicKeySource, String publicKey, String jwksUri) {
-        return (publicKeySource == null && publicKey == null && jwksUri == null)
+            String publicKeySource, String publicKey, String jwksUri, boolean isUpdate) {
+        return (publicKeySource == null && (isUpdate || publicKey != null) && jwksUri == null)
```
BenjaminWCO, line 221: It's okay for the public key to be null on updates.

PR #4884, opened 2024-07-11
BAU: Better GHA gradle caching

whi-tw:
## What

Improve dependency caching across our GHA workflows.

This will hopefully ensure that we retain a slightly more stable set of transitive dependencies and will reduce the number of unexpected rebuilds. Additionally, it will ensure that all subprojects receive the same set of cached dependencies.

Additionally, i've introduced dependency submission / review workflows - details in the commit.

Finally, use corretto instead of temurin, to ensure we are replicating the runtime environment as closely as possibly.

## How to review

- Ensure workflows have all successfully run on this branch
- Ensure that `deploy-api-modules-dev` workflow runs successfully against this branch. [I've already run this](https://github.com/govuk-one-login/authentication-api/actions/runs/9892363244), but it would likely be worth running again yourself.

### DIFF
--- a/.github/workflows/dependency-review-on-pr.yml
+++ b/.github/workflows/dependency-review-on-pr.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
       - uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 #v4.2.1
         with:
-          distribution: temurin
+          distribution: corretto
           java-version: 17
 
       - name: Generate and submit dependency graph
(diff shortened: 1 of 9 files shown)


.github/workflows/dependency-review-on-pr.yml
```diff
@@ -0,0 +1,23 @@
...
+  dependency-submission:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 17
+
+      - name: Generate and submit dependency graph
```
andyloughran, line 19: @whi-tw this will be rejected from the PR branch as it doesn't collect metadata so this action can be removed except on push to main.

andrew-moores (approved):

PR #4899, opened 2024-07-12
AUT-2790: Update logic to not lock the user in the reauth journey whe…

VladGavrilet:
## What
AUT-2790: Reset the emailCountLock before we send the response so the user can be logged out instead of locked out.

## How to review

1. Code Review

## Related PRs

https://github.com/govuk-one-login/authentication-frontend/pull/1805

### DIFF
--- a/ci/terraform/oidc/authdev1.tfvars
+++ b/ci/terraform/oidc/authdev1.tfvars
@@ -15,6 +15,7 @@ account_intervention_service_action_enabled = true
 account_intervention_service_abort_on_error = true
 send_storage_token_to_ipv_enabled           = true
 call_ticf_cri                               = true
+support_reauth_signout_enabled              = true
 
 auth_frontend_public_encryption_key = <<-EOT
 -----BEGIN PUBLIC KEY-----
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
@@ -93,11 +93,13 @@ public class CheckReAuthUserHandler extends BaseFrontendHandler<CheckReauthUserR
             UserContext userContext) {
         LOG.info("Processing CheckReAuthUser request");
 
+        var emailUserIsSignedInWith = userContext.getSession().getEmailAddress();
+
         var auditContext =
                 auditContextFromUserContext(
                         userContext,
                         AuditService.UNKNOWN,
-                        request.email(),
+                        emailUserIsSignedInWith,
                         IpAddressHelper.extractIpAddress(input),
                         AuditService.UNKNOWN,
                         PersistentIdHelper.extractPersistentIdFromHeaders(input.getHeaders()));
@@ -130,7 +132,7 @@ public class CheckReAuthUserHandler extends BaseFrontendHandler<CheckReauthUserR
                                         auditContext);
                             })
                     .map(rpPairwiseId -> generateSuccessResponse())
-                    .orElseGet(() -> generateErrorResponse(request.email(), auditContext));
+                    .orElseGet(() -> generateErrorResponse(emailUserIsSignedInWith, auditContext));
         } catch (AccountLockedException e) {
 
             auditService.submitAuditEvent(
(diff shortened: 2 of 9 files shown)


ci/terraform/oidc/authdev1.tfvars
```diff
@@ -15,6 +15,7 @@ account_intervention_service_action_enabled = true
 account_intervention_service_abort_on_error = true
 send_storage_token_to_ipv_enabled           = true
 call_ticf_cri                               = true
+support_reauth_signout_enabled              = true
```
di-fabs, line 18: Do we need to add this also to authdev2.tfvars if we want to keep our 2 dev envs synchronise?

VladGavrilet: Yes. By default the value for this feature switch is false.

di-fabs (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
```diff
@@ -108,6 +108,8 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                     .flatMap(
                             userProfile -> {
                                 if (hasEnteredIncorrectEmailTooManyTimes(userProfile.getEmail())) {
+                                    if (configurationService.supportReauthSignoutEnabled())
+                                        removeEmailCountLock(userProfile.getEmail());
```
andrew-moores, line 112:
```suggestion
                                    {
                                        removeEmailCountLock(userProfile.getEmail());
                                    }
```

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
```diff
@@ -181,6 +183,7 @@ private APIGatewayProxyResponseEvent generateSuccessResponse() {
     private APIGatewayProxyResponseEvent generateErrorResponse(
             String email, AuditContext auditContext) {
         if (hasEnteredIncorrectEmailTooManyTimes(email)) {
+            if (configurationService.supportReauthSignoutEnabled()) removeEmailCountLock(email);
```
andrew-moores, line 186:
```suggestion
            if (configurationService.supportReauthSignoutEnabled()) {
                removeEmailCountLock(email);
            }
```

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -292,6 +298,10 @@ public long getIDTokenExpiry() {
         return Long.parseLong(System.getenv().getOrDefault("ID_TOKEN_EXPIRY", "120"));
     }
 
+    public Optional<String> getNotifyApiUrl() {
+        return Optional.ofNullable(System.getenv("NOTIFY_URL"));
+    }
+
```
andrew-moores, line 304: Why did this change?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
andrew-moores: Do we have tests for enabled and disabled?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
andrew-moores: Do we have tests for enabled and disabled?

andrew-moores (changes requested):

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
VladGavrilet: it was deleted by accident in a previous commit it was just added back

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
VladGavrilet: No, we have not, considering that in the current test we only check for the response that the lambda gives back I don't think it's necessary. I believe this will change when we have the full scope of this feature, with the audit events and error messages that need to be added in the short future.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckReAuthUserHandler.java
VladGavrilet: I don't think it is needed for the following reasons:

1 - The feature switch comes by default false. So in the case the reauth is switch on in an env, the user will just be locked out if the feature switch isn't modified.
2 - We do not still have the full scope for reauth. So until we have more details about what errors and what audit events we should be sending, I propose we wait and write them once :)

alhcomer (approved):
LGTM

BeckaL (dismissed):

BeckaL (approved):

PR #4901, opened 2024-07-12
BAU: Indicate whether account deleted manually in audit message

dbes-gds:
## What

Indicate whether account deleted manually in audit message.

HMRC have indicated that certain fields are missing from the AUTH_DELETE_ACCOUNT audit event.  Account deletion can either be triggered by users, or manually as required when we have to delete accounts.  For manual deletion many of the usual fields such as session ids will be missing as there is not a user session.  To make this explicit a new field has been added to the event stating whether the account was deleted manually or not.

## How to review

1. Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
@@ -87,12 +87,12 @@ public class AccountDeletionService {
 
         String persistentSessionID = AuditService.UNKNOWN;
         String ipAddress = AuditService.UNKNOWN;
-        Boolean manualDeletion = Boolean.TRUE;
+        boolean manualDeletion = true;
         if (input.isPresent()) {
             ipAddress = PersistentIdHelper.extractPersistentIdFromHeaders(input.get().getHeaders());
             attachLogFieldToLogs(PERSISTENT_SESSION_ID, ipAddress);
             ipAddress = IpAddressHelper.extractIpAddress(input.get());
-            manualDeletion = Boolean.FALSE;
+            manualDeletion = false;
         }
 
         try {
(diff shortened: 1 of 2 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
```diff
@@ -86,10 +87,12 @@ public DeletedAccountIdentifiers removeAccount(
 
         String persistentSessionID = AuditService.UNKNOWN;
         String ipAddress = AuditService.UNKNOWN;
+        Boolean manualDeletion = Boolean.TRUE;
         if (input.isPresent()) {
             ipAddress = PersistentIdHelper.extractPersistentIdFromHeaders(input.get().getHeaders());
             attachLogFieldToLogs(PERSISTENT_SESSION_ID, ipAddress);
             ipAddress = IpAddressHelper.extractIpAddress(input.get());
+            manualDeletion = Boolean.FALSE;
```
mrwilson, line 95: Is there a reason we are using boxed booleans rather than primitive e.g. `true` and `false`?

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
```diff
@@ -113,6 +116,10 @@ public DeletedAccountIdentifiers removeAccount(
                                             RequestHeaderHelper.getHeaderValueOrElse(
                                                     n.getHeaders(), SESSION_ID_HEADER, ""))
                             .orElse(AuditService.UNKNOWN);
+            var pairs =
+                    new AuditService.MetadataPair[] {
+                        pair("manualDeletion", manualDeletion.toString())
+                    };
```
mrwilson, line 122: This variable can probably be inlined as line 134 takes a varargs e.g.

`auditService.submitAuditEvent(DELETE_ACCOUNT, auditContext, pair("manualDeletion", manualDeletion"))`

ethanmills:
I think this may make sense as a `deletion_reason` field, as we're going to have more in the future such as deletion of inactive accounts

PR #4917, opened 2024-07-18
AUT-3373: Fixed incorrect DELETE_ACCOUNT audit context

alhcomer:
## What

We were not correctly sending the persistent session ID in the audit context. This is because persistentSessionId was never being reassigned after initially being assigned to null, instead incorrectly calling the extractPersistentIdFromHeaders and assigning the result to ipAddress var.

This fixes that and makes sure that they are tested in the audit context

## How to review

1. Code Review


VladGavrilet (approved):
🚀

mattvot (approved):

PR #4925, opened 2024-07-19
AUT-3412: Ensure null values in Notify Delivery Receipts are handled correctly

mattvot:
## What

Change the `NotifyDeliveryReceipt` `completedAt` and `sentAt` fields to not required, allowing for `null` values that are can be sent by Notify.

https://docs.notifications.service.gov.uk/rest-api.html#delivery-receipts

We do not use the values anywhere, but errors were being thrown by the object mapper when values were expected but not found.

## How to review

1. Code Review

## Checklist


VladGavrilet (approved):
🚀

alhcomer (approved):

andrew-moores (approved):

PR #4931, opened 2024-07-22
Ato 668/reverification endpoint contract tests

isaac-GDS:
Contract consumer test for Auth and IPV interaction

 - test for IPV token
 - test for IPV reverification result
 

ClientId is `authOrchestrator` at the moment. This should be changed to what auth want in the future.

Removed the resource field in the token request. This is not needed from IPV  when doing it from the orch side so excluding unless they ask. There is a ticket to update it for Orch.


isaac-GDS:
Might need approval first - https://govukverify.atlassian.net/browse/ATO-489

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/ReverificationResultService.java
```diff
@@ -79,7 +76,7 @@ public TokenRequest constructTokenRequest(String authCode) {
                 generatePrivateKeyJwt(claimsSet),
                 codeGrant,
                 null,
-                singletonList(ipvTokenURI),
+                null,
```
ethanmills, line 79: Is this the resource change you mentioned?

ethanmills (approved):

PR #4942, opened 2024-07-24
ATO-694: Throw ClientSignatureValidationException when PublicKeySource is static but PublicKey is null

billrdunn:
## What

Previously, when PublicKeySource = static and PublicKey is not provided, a NullPointerException was thrown. Instead throw more useful ClientSignatureValidationException with error message.

## How to review

1. Code Review only

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
@@ -132,12 +132,20 @@ public class ClientSignatureValidationService {
     }
 
     private PublicKey retrievePublicKey(ClientRegistry client, String kid)
-            throws NoSuchAlgorithmException, InvalidKeySpecException, JwksException {
+            throws NoSuchAlgorithmException,
+                    InvalidKeySpecException,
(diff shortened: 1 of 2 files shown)


ethanmills:
Would be good to see a test for this if possible

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
```diff
@@ -132,11 +132,19 @@ public void validateTokenClientAssertion(PrivateKeyJWT privateKeyJWT, ClientRegi
...
         LOG.info("kid: " + kid);
         try {
             if (client.getPublicKeySource().equals(PublicKeySource.STATIC.getValue())) {
                 LOG.info("returning static");
+                String publicKey = client.getPublicKey();
+                if (publicKey == null) {
+                    throw new ClientSignatureValidationException(
+                            "PublicKeySource is static but PublicKey is null");
+                }
                 return convertPemToPublicKey(client.getPublicKey());
```
GHSwallow, line 148: can just use `publicKey` here

ethanmills (approved):

GHSwallow (approved):

PR #4944, opened 2024-07-24
ATO-935: Remove test logging and update logging when validating signature using RP public key

billrdunn:
## What

Remove test logging and update logging when validating signature using RP public key

## How to review

1. Code Review - double check that change to `getRpPublicKeyCacheData` doesn't affect logic

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
@@ -80,9 +80,7 @@ public class ClientSignatureValidationService {
         try {
             PublicKey publicKey;
             if (configurationService.fetchRpPublicKeyFromJwksEnabled()) {
-                LOG.info("flag enabled in signedJWT");
                 publicKey = retrievePublicKey(client, signedJWT.getHeader().getKeyID());
-                LOG.info("returning key in signedJWT: " + publicKey.toString());
             } else {
                 publicKey = convertPemToPublicKey(client.getPublicKey());
             }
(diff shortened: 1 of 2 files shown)


isaac-GDS (approved):

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
```diff
@@ -136,15 +132,18 @@ private PublicKey retrievePublicKey(ClientRegistry client, String kid)
...
+            String clientId = client.getClientID();
             if (client.getPublicKeySource().equals(PublicKeySource.STATIC.getValue())) {
-                LOG.info("returning static");
                 String publicKey = client.getPublicKey();
                 if (publicKey == null) {
                     throw new ClientSignatureValidationException(
                             "PublicKeySource is static but PublicKey is null");
                 }
+                LOG.info(
+                        "Returning static RP public signing key with key ID {} for client ID {}.",
```
ethanmills, line 144: We typically expect the key ID to be null in this case, right?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
```diff
@@ -182,6 +182,10 @@ private PublicKey retrievePublicKey(ClientRegistry client, String kid)
             }
 
             JWK jwk = JWK.parse(unescapedPayload);
+            LOG.info(
+                    "Caching RP public signing key with key ID {} for client ID {}.",
```
ethanmills, line 186: In general I prefer not having `.` on these as it looks like part of the value. The client ID should be attached to every log so it may be unnecessary anyway

PR #4951, opened 2024-07-25
BAU: Add SonarCloud break glass mechanism 

mattvot:
## What

Now that a successful SonarCloud analysis run is required to merge a PR, we have discovered a need to have a "break glass" mechanism to bypass the analysis in cases where we must push code quickly without an analysis on code smells / code coverage.

This means we can merge PRs that revert previously merged work that would now by hit by the failed analysis.

This is implemented by allowing the GitHub `run-sonar-analysis` job to be skipped if the PR has the label `bypass-sonarcloud`.

Various ways to do this were explored and discounted:

- Using the `if` key within a job would mean the job is skipped if true, which would fail the branch protection rule that requires a pass
- Trying to exit a job early didn't work either:
  - Having a step with a `run` key that used `exit 0` didn't work, as the step completed and the next step was started.
    - We could make it exit with an error code, but that would fail the job, thus fail the branch protection rule.
  - Using the `gh` CLI tool in `run` block to cancel the job failed, as it didn't have the right permission, also the job needs to succeed for the branch protection rule.
- Instead each step needs to check if it can be run with the new job environment variable.
  - Here we elevate the `github.actor != 'dependabot[bot]` condition that was in the last step into all steps, as the intent behind the condition is the same as the new label (skip the check but mark as soccess)

Also this will now re-trigger pre-merge-checks when the pull request labels change. Instead of requiring the PR author to jump through some hoops to get jobs to re-run after they assess the need to bypass, this will now trigger the checks when the label is added/removed.

## How to review

1. Code Review
1. Look at the logs of this PR's `run-sonar-analysis` job run to see it was bypassed due to having the label.

## Checklist

### DIFF
--- a/.github/workflows/pre-merge-checks.yml
+++ b/.github/workflows/pre-merge-checks.yml
@@ -300,23 +300,29 @@ jobs:
       - run-unit-tests
       - run-integration-tests
       - run-account-management-and-delivery-receipts-tests
+    env:
+      RUN: ${{ !contains(github.event.pull_request.labels.*.name , 'bypass-sonarcloud') && github.actor != 'dependabot[bot]' }}
     if: github.event_name != 'merge_group'
     steps:
       - name: Check out repository code
+        if: ${{ env.RUN == 'true' }}
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # v4.1.7
         with:
           fetch-depth: 0
       - name: Set up JDK 17
+        if: ${{ env.RUN == 'true' }}
         uses: actions/setup-java@99b8673ff64fbf99d8d325f52d9a5bdedb8483e9 #v4.2.1
         with:
           java-version: ${{ env.JAVA_VERSION }}
           distribution: ${{ env.JAVA_DISTRIBUTION }}
       - name: Set up Gradle
+        if: ${{ env.RUN == 'true' }}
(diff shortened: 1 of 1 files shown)


whi-tw (commented):
traditionally this would be done with a known string in a commit message (typically `[skip-ci]`) - is there a reason a tag was chosen instead?

With the commit message, it'd be obvious down the line from the git history that something went in without the checks.

mattvot:
> traditionally this would be done with a known string in a commit message (typically [skip-ci]) - is there a reason a tag was chosen instead?

No reason other than not knowing the prior art. Would every commit in the PR would have required this string in the message?

I suppose an weak argument against is the case of having to revert a merged PR. With using a commit message string you'd need to modify commit (as said, weak argument against)

whi-tw:
> No reason other than not knowing the prior art. Would every commit in the PR be required to have this string in the message?

No, it would just be the one at the tip of the branch. So you'd just add `[skip-sonar]` to the latest commit message and it'd do the thing.

> I suppose an weak argument against is the case of having to revert a merged PR. With using a commit message string you'd need to modify commit (as said, weak argument against)

Yeah, I see your point. Is it likely that a reversion would trigger sonar? I guess if we reverted a PR that had fixed some sonar issues, it'd kick off. I guess maybe a thought is that we should never gate with sonarqube on a revert?

PR #4955, opened 2024-07-26
BAU: Turn off noisy back channel logout alarms in int

CarlyG55:
## What

It's quite common for RPs to try integrating with back channel logout in integration and have issues, but leave the config in place so there are continual failures/alarms
## Checklist


CarlyG55:
@isaac-GDS we're getting a lot of noise in int from RPs that have unsuccessfully integrated with back channel logout but seem to have left the config

PR #4956, opened 2024-07-26
BAU: Move pr-data to reusable workflow

whi-tw:
## What

Rather than duplicating the same step in both prod and dev workflows,
move it to a reusable workflow instead.

Also, use the short commit sha in the summary, to reduce the length of
the string in AWS - it is annoying when it's truncated. For the same
reason, swap the order of PR number and repository name - so we
front-load as much 'useful' stuff as possible.

## How to review

- Code Review
- Run the dev deploy pipeline from this branch and ensure that the metadata is present. [done once here](https://github.com/govuk-one-login/authentication-api/actions/runs/10110652880).

### DIFF
--- a/.github/workflows/call_get_pr_data.yml
+++ b/.github/workflows/call_get_pr_data.yml
@@ -1,0 +1,107 @@
+name: Get PR Data
+on:
+  workflow_call:
+    outputs:
+      data:
+        description: "Object containing PR data"
+        value: ${{ jobs.get-pr-data.outputs.data }}
+
+jobs:
+  get-pr-data:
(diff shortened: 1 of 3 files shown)


BeckaL (approved):
Changes look sensible to me and visible metadata that ends up in the pipeline much more useful now! Ta

BeckaL:
(removed from merge queue so it doesn't trigger a late Friday deploy, but happy for this to go in on Monday!)

whi-tw:
TODO: update frontend / smoke tests / etc with this same change

PR #4959, opened 2024-07-26
Bau: add metric for notify delivery durations

BeckaL:
## What

It's useful for us to be able to see the length of time it's taken notify to successfully send out emails and texts as it gives us visibility as to whether there are any widespread issues when reports of slow sms / emails come to us, and notify already give us this information via their delivery receipts. This will also aid us in identifying, in the case of a widespread issue, whether the problem is on our side (e.g. delayed async processing) or on Notify's.

Notify send us ([docs link](https://docs.notifications.service.gov.uk/java.html#delivery-receipts)):
* created at timestamp
* completed at timestamp
* sent at timestamp

I think the relevant time difference here is the duration between created at (when notify sends the request to its deliverer partners) and updated at (which for a delivered notification should be the time it got delivered). My understanding is that sent at happens between these two things, but isn't as relevant to us looking at the end to end response time.

Also contains a few small refactors / code removals while I was in the area.

## How to review

### DIFF
--- a/delivery-receipts-api/src/main/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandler.java
+++ b/delivery-receipts-api/src/main/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandler.java
@@ -154,14 +154,12 @@ public class NotifyCallbackHandler
             var completedAt = Instant.parse(deliveryReceipt.completedAt());
             var createdAt = Instant.parse(deliveryReceipt.createdAt());
             double duration = Duration.between(createdAt, completedAt).toMillis();
+            var metricsContext =
+                    Map.ofEntries(
+                            Map.entry("Environment", configurationService.getEnvironment()),
+                            Map.entry("NotificationType", deliveryReceipt.notificationType()));
             cloudwatchMetricsService.putEmbeddedValue(
-                    "NotifyDeliveryDuration",
-                    duration,
-                    Map.of(
-                            "Environment",
-                            configurationService.getEnvironment(),
-                            "NotificationType",
-                            deliveryReceipt.notificationType()));
+                    "NotifyDeliveryDuration", duration, metricsContext);
         } catch (DateTimeParseException e) {
             LOG.warn(
                     format(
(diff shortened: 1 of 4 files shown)


delivery-receipts-api/src/main/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandler.java
```diff
@@ -140,6 +129,47 @@ public APIGatewayProxyResponseEvent notifyCallbackRequestHandler(
...
+        if (deliveryReceipt.status().equals("delivered")) {
+            sendDurationMetrics(deliveryReceipt);
+        }
+    }
+
+    private void sendDurationMetrics(NotifyDeliveryReceipt deliveryReceipt) {
+        if (Objects.isNull(deliveryReceipt.completedAt())
+                || Objects.isNull(deliveryReceipt.createdAt())) {
+            return;
+        }
```
mrwilson, line 152: We could put a `hasBeenDelivered` (or appropriate name) on the record that returns `completedAt == null || createdAt == null` and use that here instead?

BeckaL: Hmm, I'm going to leave this for now because I think it's more useful to see we've done null checks before doing the below lines, and I couldn't think of a good way of expressing the check here, but one for maybe some future refactoring

delivery-receipts-api/src/main/java/uk/gov/di/authentication/deliveryreceiptsapi/lambda/NotifyCallbackHandler.java
```diff
@@ -81,51 +86,35 @@ public APIGatewayProxyResponseEvent notifyCallbackRequestHandler(
         NotifyDeliveryReceipt deliveryReceipt;
         try {
             deliveryReceipt = objectMapper.readValue(input.getBody(), NotifyDeliveryReceipt.class);
-            if (deliveryReceipt.getNotificationType().equals("sms")) {
+            if (deliveryReceipt.notificationType().equals("sms")) {
                 LOG.info("Sms delivery receipt received");
```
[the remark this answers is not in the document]
andrew-moores: Will do later.

andrew-moores (approved):

PR #4960, opened 2024-07-29
Aut 3368/auth attempts store

VladGavrilet:
## What

AUT-3368 and AUT-3369.

Transitioned the storage medium for TTL variables from Redis to DynamoDB. This required introducing the DynamoAuthenticationAttemptsService class and creating new DynamoDB tables to handle authentication codes and counts with TTL constraints.

Added logic for CRUD Operations: 

- Create: Add new codes and counts.
- Read: Retrieve codes and counts.
- Update: Modify existing codes and counts.
- Delete: Remove codes and counts.

Added a new integration test class to verify the functionality of CRUD operations and TTL management with DynamoDB.

## How to review

## Checklist

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
@@ -1,0 +1,121 @@
+package uk.gov.di.authentication.shared.entity;
+
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbAttribute;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+
+@DynamoDbBean
+public class AuthenticationAttempts {
+
+    private String attemptIdentifier;
+    private String email;
+    private String journeyType;
+    private String authenticationType;
+    private String code;
+    private Integer count;
+    private long timeToExist;
+    private String created;
+    private String updated;
+
+    @DynamoDbPartitionKey
+    @DynamoDbAttribute("AttemptIdentifier")
+    public String getAttemptIdentifier() {
+        return attemptIdentifier;
+    }
+
+    public void setAttemptIdentifier(String attemptIdentifier) {
+        this.attemptIdentifier = attemptIdentifier;
+    }
+
+    public AuthenticationAttempts withAttemptIdentifier(String attemptIdentifier) {
+        this.attemptIdentifier = attemptIdentifier;
+        return this;
+    }
+
+    @DynamoDbAttribute("Email")
+    public String getEmail() {
+        return email;
+    }
+
+    public void setEmail(String email) {
+        this.email = email;
+    }
+
+    @DynamoDbAttribute("JourneyType")
+    public String getJourneyType() {
+        return journeyType;
+    }
+
+    public void setJourneyType(String journeyType) {
+        this.journeyType = journeyType;
+    }
+
+    @DynamoDbAttribute("AuthenticationType")
+    public String getAuthenticationType() {
+        return authenticationType;
+    }
+
+    public void setAuthenticationType(String authenticationType) {
+        this.authenticationType = authenticationType;
+    }
+
+    @DynamoDbAttribute("Code")
+    public String getCode() {
+        return code;
+    }
+
+    public void setCode(String code) {
+        this.code = code;
+    }
+
+    @DynamoDbAttribute("Count")
+    public Integer getCount() {
+        return count;
+    }
+
+    public void setCount(Integer count) {
+        this.count = count;
+    }
+
+    @DynamoDbAttribute("TimeToExist")
+    public long getTimeToExist() {
+        return timeToExist;
+    }
+
+    public void setTimeToExist(long timeToExist) {
+        this.timeToExist = timeToExist;
+    }
+
+    public AuthenticationAttempts withTimeToExist(long timeToExist) {
+        this.timeToExist = timeToExist;
+        return this;
+    }
+
+    @DynamoDbAttribute("Created")
+    public String getCreated() {
+        return created;
+    }
+
+    public void setCreated(String created) {
+        this.created = created;
+    }
+
+    public AuthenticationAttempts withCreated(String created) {
+        this.created = created;
+        return this;
+    }
+
+    @DynamoDbAttribute("Updated")
+    public String getUpdated() {
+        return updated;
+    }
+
+    public void setUpdated(String updated) {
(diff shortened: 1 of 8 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
```diff
@@ -0,0 +1,131 @@
...
+    public void setJourneyType(String journeyType) {
+        this.journeyType = journeyType;
+    }
+
+    @DynamoDbAttribute("AuthenticationType")
+    public String getAuthenticationType() {
+        return authenticationType;
+    }
+
+    public void setAuthenticationType(String authenticationType) {
```
mattvot, line 58: Some fields have builder pattern `withX()` methods, and some don't (like `journeyType` and `authenticationType`.

Can we add them to all missing fields? Makes me wonder if we can remove `set*()` methods too.

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationType.java
```diff
@@ -0,0 +1,18 @@
+package uk.gov.di.authentication.shared.entity;
+
+public enum AuthenticationType {
```
mattvot, line 3: Can we rename `AuthenticationType` to `AuthenticationMethod` which IMO seems more accurate to what we are representing with this enum (+ is more inline with the terminology we use for "managing your auth methods".

This will require changes to the AuthenticationAttempts entity field name too

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
```diff
@@ -0,0 +1,73 @@
...
+
+import java.time.Clock;
+import java.time.temporal.ChronoUnit;
+import java.util.Optional;
+
+public class DynamoAuthenticationAttemptsService extends BaseDynamoService<AuthenticationAttempts> {
+
+    public DynamoAuthenticationAttemptsService(ConfigurationService configurationService) {
+        super(AuthenticationAttempts.class, "authentication-attempts", configurationService);
+        new NowHelper.NowClock(Clock.systemUTC());
```
mattvot, line 14: From my eyes this seems to throw away the `NowClock` instance and could probably be removed.

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public void setAttemptIdentifier(String attemptIdentifier) {
+        this.attemptIdentifier = attemptIdentifier;
+    }
+
+    public AuthenticationAttempts withAttemptIdentifier(String attemptIdentifier) {
+        this.attemptIdentifier = attemptIdentifier;
+        return this;
+    }
```
mattvot, line 33: Will `attemptIdentifier` contain `internalSubjectId`?

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    public void addCode(String attemptIdentifier, long ttlSeconds, String code) {
```
mattvot, line 17: Does `addCode` also need to accept and store the `authenticationType/Method`?

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
```diff
@@ -0,0 +1,73 @@
...
+                get(attemptIdentifier)
+                        .orElse(new AuthenticationAttempts())
+                        .withAttemptIdentifier(attemptIdentifier)
+                        .withCode(code)
+                        .withTimeToLive(ttlEpochSeconds);
+
+        update(authenticationAttempt);
+    }
+
+    public void createOrIncrementCount(String attemptIdentifier, long ttl) {
```
mattvot, line 31: Does `createOrIncrementCount ` also need to accept and store the `authenticationType/Method`?

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
```diff
@@ -0,0 +1,131 @@
...
+
+    private String attemptIdentifier;
+    private String email;
+    private String journeyType;
+    private String authenticationType;
+    private String code;
+    private Integer count;
+    private long timeToLive;
+    private String created;
+    private String updated;
```
mattvot, line 18: Is there a way we can mark some of these fields as required?

I would have expected later tests to fail as `authenticationType` is missing (and very much needed).

mattvot (changes requested):

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
VladGavrilet: @dbes-gds can you please have a look over this?

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
```diff
@@ -0,0 +1,131 @@
...
+
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbAttribute;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+
+@DynamoDbBean
+public class AuthenticationAttempts {
+
+    private String attemptIdentifier;
+    private String email;
```
mattvot, line 11: @dbes-gds, shall we remove email from this? I see it's proposed as a nice addition but shall we delay adding it until we see we need it?

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
dbes-gds: I think we should merge this but there may be an enhancement around this that we can pickup next week to make it more explicit what the identifier contains.

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthenticationAttempts.java
dbes-gds: yes can remove it, the idea was to make testing a bit easier as you can see which emails have counts, but can just use the internal subject id.

mattvot (approved):

PR #4267, opened 2024-04-12
Adding lifecycle ignore for provisioned_concurrent_executions

pskushwaha1:
## What

Adding lifecycle ignore for provisioned_concurrent_executions

## Why
We are experiencing lambda throttling during lambda running in scaled   provisioned concurrent. while deployment in happing which is try to re provision  the provisioned concurrent from scaled 10 to 3 
Evidence below
image

PR #4978, opened 2024-07-31
Revert "Adding lifecycle ignore for provisioned_concurrent_executions"

pskushwaha1:
Reverts govuk-one-login/authentication-api#4267

This does the work but we have a small concern when lambda Deployment fails the concurrency is not shifted to next good version we have to manually delete the  Provisioned concurrency configurations to fix

we have better solution for Lambda throttling to be tested


pskushwaha1:
Not Required

PR #4979, opened 2024-07-31
ATO-522: add ability to disable RPs

GHSwallow:
## What

Added `isActive` property to `ClientRegistry`, which defaults to true if it is null.
If an RP is not active, redirect back to RP with unauthorised user error.

## How to review

1. Code Review
2. Deploy and check that the authentication journey is unchanged by default and when `isActive` is true, but fails with a redirect and descriptive message when `isActive` is false.

## Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -334,7 +334,7 @@ public class AuthorisationHandler
                             authRequest.getState(),
                             authRequest.getResponseMode());
             return generateApiGatewayProxyResponse(
-                    400,
+                    401,
                     "",
                     Map.of(ResponseHeaders.LOCATION, errorResponse.toURI().toString()),
                     null);
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -326,17 +326,14 @@ public class AuthorisationHandler
         if (!client.getIsActive()) {
             String errorMsg = "Client configured as not active in Client Registry";
             LOG.error(errorMsg);
-            var errorResponse =
-                    new AuthenticationErrorResponse(
-                            authRequest.getRedirectionURI(),
-                            new ErrorObject(UNAUTHORIZED_CLIENT_CODE, errorMsg),
-                            authRequest.getState(),
-                            authRequest.getResponseMode());
-            return generateApiGatewayProxyResponse(
-                    401,
-                    "",
-                    Map.of(ResponseHeaders.LOCATION, errorResponse.toURI().toString()),
-                    null);
+            LOG.warn("Redirecting");
+            return generateErrorResponse(
+                    authRequest.getRedirectionURI(),
+                    authRequest.getState(),
+                    authRequest.getResponseMode(),
+                    new ErrorObject(UNAUTHORIZED_CLIENT_CODE, errorMsg),
+                    authRequest.getClientID().getValue(),
+                    user);
         }
 
         try {
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -323,7 +323,7 @@ public class AuthorisationHandler
             }
         }
 
-        if (!client.getIsActive()) {
+        if (!client.isActive()) {
             String errorMsg = "Client configured as not active in Client Registry";
             LOG.error(errorMsg);
             LOG.warn("Redirecting");
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -324,14 +324,12 @@ public class AuthorisationHandler
         }
 
         if (!client.isActive()) {
-            String errorMsg = "Client configured as not active in Client Registry";
-            LOG.error(errorMsg);
-            LOG.warn("Redirecting");
+            LOG.error("Client configured as not active in Client Registry");
             return generateErrorResponse(
                     authRequest.getRedirectionURI(),
                     authRequest.getState(),
                     authRequest.getResponseMode(),
-                    new ErrorObject(UNAUTHORIZED_CLIENT_CODE, errorMsg),
+                    new ErrorObject(UNAUTHORIZED_CLIENT_CODE, "client deactivated"),
                     authRequest.getClientID().getValue(),
                     user);
         }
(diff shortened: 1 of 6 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
```diff
@@ -232,6 +233,20 @@ public ClientRegistry withSubjectType(String subjectType) {
         return this;
     }
 
+    @DynamoDbAttribute("IsActive")
+    public Boolean getIsActive() {
```
billrdunn, line 237: Generally we tend to use primitive `boolean` instead of `Boolean`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
```diff
@@ -232,6 +233,20 @@ public ClientRegistry withSubjectType(String subjectType) {
...
+        return isActive;
```
billrdunn, line 238: Will this default to `true` if `isActive` is `null`? Or do we need to use `Optional.ofNullable` like for `PublicKeySource`?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -317,6 +318,23 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
+                    String.format("Client not active for ClientId: %s", client.getClientID());
+            LOG.error(errorMsg);
+            var errorResponse =
+                    new AuthenticationErrorResponse(
+                            authRequest.getRedirectionURI(),
+                            new ErrorObject(UNAUTHORIZED_CLIENT_CODE, errorMsg),
+                            authRequest.getState(),
+                            authRequest.getResponseMode());
+            return generateApiGatewayProxyResponse(
+                    400,
```
billrdunn, line 332: Would 401 or 403 be more appropriate?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -317,6 +318,23 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
             }
         }
 
+        if (!client.getIsActive()) {
+            String errorMsg =
+                    String.format("Client not active for ClientId: %s", client.getClientID());
```
billrdunn, line 323: I don't think that `clientId` needs to be explicitly logged as it's included in all log messages by default

billrdunn:
I think you'll also need to make changes to ClientRegistrationHandler to make sure `isActive` is set for new clients

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
GHSwallow: Agree, I think 401

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
GHSwallow: Because null should mean `true` here, I think perhaps 
```
if (isActive == null) {
  return true
}
return isActive;
```

 Does the default assignment `private boolean isActive = true;` not cover this though? Or does the `DynamoDbAttribute` decorator overwrite it even if the entry is null?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -317,6 +318,22 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
             }
         }
 
+        if (!client.getIsActive()) {
+            String errorMsg = String.format("Client not active", client.getClientID());
+            LOG.error(errorMsg);
```
billrdunn, line 323:
```suggestion
    if (!client.getIsActive()) {
            LOG.error('Client configured as not active in Client Registry");
```

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
billrdunn: Actually 302 :)

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
billrdunn: Tested this together in dev and happy the current implementation works as intended

andrew-moores (dismissed):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -322,6 +323,19 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
             }
         }
 
+        if (!client.getIsActive()) {
+            String errorMsg = "Client configured as not active in Client Registry";
+            LOG.error(errorMsg);
+            LOG.warn("Redirecting");
```
ethanmills, line 329: Not sure we need these extra lines. They don't give any additional information. I know they were copied from elsewhere, but still...

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -322,6 +323,19 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
             }
         }
 
+        if (!client.getIsActive()) {
+            String errorMsg = "Client configured as not active in Client Registry";
```
ethanmills, line 327: this makes sense as a log message, but I think less so as an RP facing error message. "Client Registry" won't mean anything to them. For the error response we could go with something like `client disabled`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
```diff
@@ -232,6 +233,20 @@ public ClientRegistry withSubjectType(String subjectType) {
         return this;
     }
 
+    @DynamoDbAttribute("IsActive")
+    public boolean getIsActive() {
+        return isActive;
```
mrwilson, line 238: Conventionally, methods that return booleans are named `isXXX` rather than `getIsXXX` because it reads more fluently by consumers.

Is there a reason why we're using this naming approach?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
billrdunn: Gabe originally had something like this and I suggested he changed it :)
https://github.com/govuk-one-login/authentication-api/pull/4979#discussion_r1698342819

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
GHSwallow: No reason, just new to OOP, thanks for the tip:)

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -976,6 +976,39 @@ void shouldValidateRequestObjectWhenJARValidationIsNotRequired()
                                     "Redirect URI invalid-redirect-uri is invalid for client")));
         }
 
+        @Test
+        void shouldThrowErrorWhenClientIsNotActive() {
```
ethanmills, line 980: nitpick but this probably wants to be renamed?

ethanmills (dismissed):

ethanmills (approved):

ethanmills (approved):

ethanmills (approved):

PR #4989, opened 2024-08-02
BAU: Refactor NotificationHandler and test class

VladGavrilet:
## What

This PR refactors the NotificationHandler class to improve exception handling by using the notifyException functions

## How to review

1. Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
@@ -62,24 +62,21 @@ public class NotificationHandler implements RequestHandler<SQSEvent, Void> {
                 () -> notificationRequestHandler(event));
     }
 
-    public Void notificationRequestHandler(SQSEvent event) throws NotificationClientException {
-
+    public Void notificationRequestHandler(SQSEvent event) {
         for (SQSMessage msg : event.getRecords()) {
             try {
                 LOG.info("Message received from SQS queue");
                 NotifyRequest notifyRequest =
                         objectMapper.readValue(msg.getBody(), NotifyRequest.class);
-                processNotificationRequest(notifyRequest);
-            } catch (JsonException e) {
+                processNotificationRequest(notifyRequest, 0);
+            } catch (JsonException | NotificationClientException e) {
                 LOG.error("Error when mapping message from queue to a NotifyRequest", e);
-                throw new NotificationClientException(
-                        "Error when mapping message from queue to a NotifyRequest", e);
             }
         }
         return null;
     }
 
-    private void processNotificationRequest(NotifyRequest notifyRequest)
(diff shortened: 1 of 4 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
```diff
@@ -58,104 +59,104 @@ public NotificationHandler(ConfigurationService configurationService) {
     public Void handleRequest(SQSEvent event, Context context) {
         return segmentedFunctionCall(
                 "account-management-api::" + getClass().getSimpleName(),
-                () -> notificationRequestHandler(event, context));
+                () -> notificationRequestHandler(event));
     }
 
-    public Void notificationRequestHandler(SQSEvent event, Context context) {
+    public Void notificationRequestHandler(SQSEvent event) throws NotificationClientException {
```
BeckaL, line 65: I'm not sure, but I remember @andrew-moores found that removing a parameter from a handler that looked like it was unused caused dynatrace problems, can we check?

VladGavrilet: Had a discussion with Andrew, while it can cause problems, this is not one of the handlers that is affected by this type of change. Deployment was also successful in a dev environment which means dynatrace is not affected by this change 😄

andrew-moores: The issue with DynaTrace was removing the over-ridden handleRequest method where the BaseHandler class is being extended.  There is a DynaTrace test to check this method exists in the handler classes to ensure DynaTrace will be able to instrument the lambda.  As this class doesn't extend BaseHandler and there is still a handleRequest method with the existing signature this shouldn't impact DynaTrace but we should have a post deployment check in an environment where DynaTrace is deplopyed.

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
```diff
@@ -58,104 +59,104 @@ public NotificationHandler(ConfigurationService configurationService) {
...
-                                    notifyRequest.getNotificationType()),
-                            e);
-                }
+                processNotificationRequest(notifyRequest);
             } catch (JsonException e) {
-                LOG.error("Error when mapping message from queue to a NotifyRequest");
-                throw new RuntimeException(
-                        "Error when mapping message from queue to a NotifyRequest");
+                LOG.error("Error when mapping message from queue to a NotifyRequest", e);
+                throw new NotificationClientException(
```
andrew-moores, line 75: Throwing an exception from the lambda will cause it to be restarted.  Rather than refactoring the exception to a new more specific type it would be better to remove the thrown exception if the error is temporary and the next request will likely be successful.

PR #4998, opened 2024-08-05
Aut 3481/generic mfa

BeckaL:
## What

We will be storing both AuthAppMfa and SmsMfa as mfa methods in the user credentials table.

Currently, we only store AuthAppMfa methods. This commit is an initial step, which should have no behavioural impact on anything, to make mfa methods generic.

This will allow us to store and work with types that are specific to the data being stored - for example, auth app mfa data has a credential value, whereas sms doesn't. Being able to work with different types means less annoying null handling.

We had to define a custom converter for the serialisation and deserialisation from the dynamo db objects, since we're now deserialising an interface which may take multiple forms

## How to review

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
@@ -19,11 +19,11 @@ import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent;
 import uk.gov.di.authentication.frontendapi.entity.CheckUserExistsResponse;
 import uk.gov.di.authentication.frontendapi.helpers.CommonTestVariables;
+import uk.gov.di.authentication.shared.entity.AuthAppMFAMethod;
 import uk.gov.di.authentication.shared.entity.ClientRegistry;
 import uk.gov.di.authentication.shared.entity.ClientSession;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
-import uk.gov.di.authentication.shared.entity.MFAMethod;
 import uk.gov.di.authentication.shared.entity.MFAMethodType;
 import uk.gov.di.authentication.shared.entity.Session;
 import uk.gov.di.authentication.shared.entity.TermsAndConditions;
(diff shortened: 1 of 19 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/conditions/MfaHelper.java
```diff
@@ -48,7 +56,7 @@ public static UserMfaDetail getUserMFADetail(
         var mfaMethodType = isPhoneNumberVerified ? MFAMethodType.SMS : MFAMethodType.NONE;
 
         var mfaMethod = getPrimaryMFAMethod(userCredentials);
-        if (mfaMethod.filter(MFAMethod::isMethodVerified).isPresent()) {
+        if (mfaMethod.filter(AuthAppMFAMethod::isMethodVerified).isPresent()) {
```
dbes-gds, line 59: Does this throw a ClassCastException if it's not an AuthAppMFAMethod?  Though I guess that's the only subclass we have at the moment.  Did MFAMethod::isMethodVerified still work as that method is on the interface?

shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/MfaMethodListConverter.java
```diff
@@ -0,0 +1,70 @@
...
+    public List<MFAMethod> transformTo(AttributeValue input) {
+        if (input == null || input.s() == null || input.s().isEmpty()) {
+            return new ArrayList<>(); // Return an empty list if the input is null or empty
+        }
+
+        String json = input.s();
+        List<Map<String, String>> deserializedList =
+                gson.fromJson(json, new TypeToken<List<Map<String, String>>>() {}.getType());
+        List<MFAMethod> result = new ArrayList<>();
+        for (Map<String, String> item : deserializedList) {
```
dbes-gds, line 47: Does this assume that all attribute in MFAMethod are strings?

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
```diff
@@ -155,11 +155,12 @@ void setup() {
 
         @Test
         void shouldReturn200WithRelevantMfaMethod() {
-            MFAMethod mfaMethod1 = verifiedMfaMethod(MFAMethodType.AUTH_APP, false);
-            MFAMethod mfaMethod2 = verifiedMfaMethod(MFAMethodType.SMS, true);
+            AuthAppMFAMethod authAppMfaMethod1 = verifiedMfaMethod(MFAMethodType.AUTH_APP, false);
+            AuthAppMFAMethod authAppMfaMethod2 = verifiedMfaMethod(MFAMethodType.SMS, true);
```
dbes-gds, line 159: these tests look a bit weird now as currently mfa-methods doesn't support SMS

shared/src/main/java/uk/gov/di/authentication/shared/dynamodb/MfaMethodListConverter.java
```diff
@@ -0,0 +1,70 @@
...
+import software.amazon.awssdk.enhanced.dynamodb.EnhancedType;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+import uk.gov.di.authentication.shared.entity.MFAMethod;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class MfaMethodListConverter implements AttributeConverter<List<MFAMethod>> {
```
dbes-gds, line 16: Does this need some tests?  Sonar can't be happy :)

BeckaL:
Closing in favour of less sketchy approach here: https://github.com/govuk-one-login/authentication-api/pull/5000

PR #5000, opened 2024-08-05
Aut 3481/update mfa schema

BeckaL:
## Checklist


shared/src/main/java/uk/gov/di/authentication/shared/entity/MFAMethod.java
```diff
@@ -111,6 +179,20 @@ public MFAMethod withUpdated(String updated) {
...
+    public void setMfaIdentifier(int mfaIdentifier) {
+        this.mfaIdentifier = mfaIdentifier;
+    }
+
+    public MFAMethod withMfaIdentifier(int mfaIdentifier) {
+        this.mfaIdentifier = mfaIdentifier;
+        return this;
+    }
+
     AttributeValue toAttributeValue() {
```
dbes-gds, line 196: please add the new fields in here as this is used by update email to copy the user

BeckaL: Ah good spot, I've added it to equals in the relevant commits. This toAttributeValue method is unused (or at least my editor is telling me it is) - can I remove it instead or is it used in some sneak way that my editor is not picking up?

BeckaL: I've updated the update email integration tests to check explicitly (see last three commits here), and they pass without toAttributeValue, so I still think it's unused?

andrew-moores: Suggest we remove this method until we find where its needed.  If we ever have to re-instate then a unit test illustrating what its used for might be a good way to record this piece of knowledge as its not clear from the code or documentation what it could be used by.

integration-tests/src/test/java/uk/gov/di/authentication/services/DynamoServiceIntegrationTest.java
```diff
@@ -1,11 +1,15 @@
 package uk.gov.di.authentication.services;
 
 import com.nimbusds.oauth2.sdk.id.Subject;
+import org.junit.jupiter.api.BeforeEach;
```
andrew-moores, line 4: Minor typo in the commit message for this change.  (oof vs of)

BeckaL: Good spot, fixed!

andrew-moores (approved):

PR #5011, opened 2024-08-07
AUT-3497: primaryKey refactoring to use partition and sort keys

LazarAlexandru-Constantin:
## What

Add dynamoDb range key to the AuthenticationAttempt table. This way, the primary key of the table will be composed of 
a partition key: internalSubjectID
and a composite range key: mfaMethd + journeyType

## How to review

1. Code Review

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
@@ -21,16 +21,22 @@ public class DynamoAuthenticationAttemptsService extends BaseDynamoService<Authe
         long ttlEpochSeconds =
                 NowHelper.nowPlus(ttlSeconds, ChronoUnit.SECONDS).toInstant().getEpochSecond();
 
-        var authenticationAttempt =
-                get(internalSubjectId, buildSortKey(authenticationMethod, journeyType))
-                        .orElse(new AuthenticationAttempts())
-                        .withInternalSubjectId(internalSubjectId)
-                        .withCode(code)
-                        .withAuthenticationMethod(authenticationMethod)
-                        .withTimeToLive(ttlEpochSeconds)
-                        .withJourneyType(journeyType);
-
-        update(authenticationAttempt);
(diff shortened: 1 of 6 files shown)


shared/src/test/java/uk/gov/di/authentication/shared/entity/AuthenticationAttemptsTest.java
```diff
@@ -14,8 +14,8 @@ class AuthenticationAttemptsTest {
     void testFluentSetters() {
         AuthenticationAttempts attempts = new AuthenticationAttempts();
 
-        AuthenticationAttempts result = attempts.withAttemptIdentifier("attempt-1234");
-        assertEquals("attempt-1234", result.getAttemptIdentifier());
+        AuthenticationAttempts result = attempts.withInternalSubjectId("attempt-1234");
```
BeckaL, line 17: Minor, should we update this value to look more like a subject id (or just sub out attempt for subject)

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoAuthenticationAttemptsService.java
```diff
@@ -13,59 +13,74 @@ public DynamoAuthenticationAttemptsService(ConfigurationService configurationSer
...
 
         update(authenticationAttempt);
     }
 
     public void createOrIncrementCount(
-            String attemptIdentifier, long ttl, String authenticationMethod) {
-        Optional<AuthenticationAttempts> authenticationAttempt = get(attemptIdentifier);
+            String internalSubjectId, long ttl, String authenticationMethod, String journeyType) {
+        Optional<AuthenticationAttempts> authenticationAttempt =
+                get(internalSubjectId, authenticationMethod + "_" + journeyType);
```
BeckaL, line 39: I think it's worth introducing a helper function somewhere to replace all the cases we manually do this concatenation (here and in the authentication attempts method) so that it's only defined in one place

BeckaL (commented):
Nice one! In general looks good to me, a couple of small code suggestions

dbes-gds (approved):

PR #5020, opened 2024-08-08
Ato 958/dcmaw contract test feedback

isaac-GDS:
Updated error code to 401
Removed 404 test

### DIFF
--- a/doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/contract/DcmawUserInfoTest.java
+++ b/doc-checking-app-api/src/test/java/uk/gov/di/authentication/app/contract/DcmawUserInfoTest.java
@@ -164,7 +164,7 @@ public class DcmawUserInfoTest {
                         "^(?i)Bearer (.*)(?-i)",
                         tokens.getAccessToken().toAuthorizationHeader())
                 .willRespondWith()
-                .status(403)
+                .status(401)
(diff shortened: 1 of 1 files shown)


ethanmills (approved):

sandy-may (approved):
Thanks for this. I think there are two remaining things which would be helpful to have but it would be good to merge this as-is so we can work on getting the two 401 interactions passing on our side. 

1) Adding the 404 scenario: https://govukverify.atlassian.net/wiki/spaces/DCMAW/pages/3618078730/Userinfo+errors#Option-3---Use-alternate-HTTP-response-code-to-allow-differentiated-consumer-behaviour

2) The signed credential you are asserting against in the happy path scenario isn't one that follows our schema - therefore we can't easily produce it in our pact test. I think @jamesamooney has (or was going to) send over an example for you to use: https://gds.slack.com/archives/C05EW21TW3A/p1722509606259299?thread_ts=1722432056.112959&cid=C05EW21TW3A

isaac-GDS:
We don't look at 404 scenario specifically as we just handle errors the same. For the signed credential @jamesamooney has said its fine as is at the moment.

sandy-may:
> @sandy-may We don't look at 404 scenario specifically as we just handle errors the same. For the signed credential @jamesamooney has said its fine as is at the moment.

Thanks Isaac - all good with me! Thank you

PR #5041, opened 2024-08-12
ATO-489: Remove resource field in token request to IPV

Ryan-Andrews99:
## What

- Removes the `resource` field in the token request to IPV.
- Removes this field in the contract tests too

## How to review

- Code review commit by commit
- This field is unused but we should double check before merging that this does not break any of IPV's contract tests

## Checklist


isaac-GDS:
This will need to be tested in staging. We should also notify IPV when we do so.

PR #5048, opened 2024-08-13
ATO-936: Make nonce optional for auth only rps

gn-GDS:
## What

Make nonce optional for auth only rps. This is because several RPs identity platforms don't support Nonce out of the box.

## Checklist

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
@@ -42,6 +42,7 @@ import uk.gov.di.orchestration.shared.services.RedisConnectionService;
 import uk.gov.di.orchestration.shared.services.SerializationService;
 import uk.gov.di.orchestration.shared.services.SessionService;
 
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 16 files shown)


ethanmills:
Have we checked all the usages of nonce to ensure they're null-safe?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
```diff
@@ -451,4 +453,18 @@ public ClientRegistry withClientLoCs(List<String> clientLoCs) {
         this.clientLoCs = clientLoCs;
         return this;
     }
+
+    @DynamoDbAttribute("PermitMissingNonce")
+    public boolean getPermitMissingNonce() {
+        return !identityVerificationSupported && permitMissingNonce;
+    }
+
```
mrwilson, line 461: Can this be called `permitMissingNonce` as it returns a boolean?

This would read clearer at call sites e.g. `if (client.permitMissingNonce()) { ... }`

mrwilson: Also, is this method not just for persistence in the record? We would return the field value rather than joining it with whether it's an identity RP, and have a different method for that.

Maybe the `permitMissingNonce` method should be the joined form, and the `getPermitMissingNonce` method should just be the raw value getter.

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -315,15 +316,20 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
                                 null,
                                 authenticationRequest.getResponseMode());
 
+                var metadataPairs = new ArrayList<AuditService.MetadataPair>();
+                metadataPairs.add(pair("internalSubjectId", AuditService.UNKNOWN));
+                metadataPairs.add(pair("isNewAccount", session.isNewAccount()));
+                metadataPairs.add(pair("rpPairwiseId", AuditService.UNKNOWN));
+                metadataPairs.add(pair("authCode", authCode));
+                if (authenticationRequest.getNonce() != null) {
+                    metadataPairs.add(pair("nonce", authenticationRequest.getNonce()));
```
ethanmills, line 325: I've noticed this doesn't currently work. It's missing a `.getValue()` on the nonce

ethanmills (approved):

PR #5053, opened 2024-08-14
ATO-964: create new browser session cookie

GHSwallow:
## What
Adds a browser session id to cookie, and stores it in Redis. This can be used to check that a user hasn't closed their browser, and use this to have them only signed in with that browser.

Currently nothing is done with the browser session id, so there is no functional change. It is also feature flagged to true only on dev.

## How to review
1. Code Review

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
@@ -103,6 +103,18 @@ public class CookieHelper {
                 });
     }
 
+    public static Optional<String> parseBrowserSessionCookie(Map<String, String> headers) {
+        Optional<HttpCookie> httpCookie =
+                getHttpCookieFromRequestHeaders(headers, BROWSER_SESSION_COOKIE_NAME);
+        if (httpCookie.isEmpty()) {
+            return Optional.empty();
+        }
+
+        String cookieValues = httpCookie.get().getValue();
+
+        return Optional.of(cookieValues);
+    }
+
     public static Optional<String> parsePersistentCookie(Map<String, String> headers) {
         Optional<HttpCookie> httpCookie =
                 getHttpCookieFromRequestHeaders(headers, PERSISTENT_COOKIE_NAME);
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
@@ -17,6 +17,7 @@ public class CookieHelper {
     public static final String RESPONSE_COOKIE_HEADER = "Set-Cookie";
     public static final String PERSISTENT_COOKIE_NAME = "di-persistent-session-id";
     public static final String SESSION_COOKIE_NAME = "gs";
+    public static final String BROWSER_SESSION_COOKIE_NAME = "browser-session-id";
 
     public static final String LANGUAGE_COOKIE_NAME = "lng";
 
@@ -169,6 +170,11 @@ public class CookieHelper {
         return format(
                 "%s=%s; Max-Age=%d; Domain=%s; %s",
                 cookieName, cookieValue, maxAge, domain, attributes);
+    }
+
+    public static String buildCookieString(
+            String cookieName, String cookieValue, String attributes, String domain) {
+        return format("%s=%s; Domain=%s; %s", cookieName, cookieValue, domain, attributes);
     }
 
     public static String appendTimestampToCookieValue(String cookieValue) {
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/SessionService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/SessionService.java
@@ -44,7 +44,12 @@ public class SessionService {
     }
 
     public Session createSession() {
-        return new Session(IdGenerator.generate()).withBrowserSessionId(IdGenerator.generate());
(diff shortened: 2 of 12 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/OrchestrationAuthorizationService.java
```diff
@@ -232,6 +233,14 @@ public void storeState(String sessionId, State state) {
                 configurationService.getSessionExpiry());
     }
 
+    public void storeBrowserSessionState(String browserSessionId, State state) {
+        LOG.info("Storing browser session state");
+        redisConnectionService.saveWithExpiry(
+                BROWSER_AUTHENTICATION_STATE_STORAGE_PREFIX + browserSessionId,
+                state.getValue(),
+                configurationService.getSessionExpiry());
+    }
```
ethanmills, line 242: what's this for?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/SessionService.java
```diff
@@ -44,7 +44,7 @@ public SessionService(ConfigurationService configurationService) {
     }
 
     public Session createSession() {
-        return new Session(IdGenerator.generate());
+        return new Session(IdGenerator.generate()).withBrowserSessionId(IdGenerator.generate());
```
GHSwallow, line 47: Decided that rather than rewriting the Session interface to have to include a browser session id, it is instead added with this builder pattern where necessary. Makes this PR a smaller change and means it is more easily kept separate and optional, but if browser session ids become used everywhere, perhaps would be better to include in the interface.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
```diff
@@ -102,6 +103,20 @@ public String getClientSessionId() {
...
     }
 
+    public static Optional<String> parseBrowserSessionCookie(Map<String, String> headers) {
+        Optional<HttpCookie> httpCookie =
+                getHttpCookieFromRequestHeaders(headers, BROWSER_SESSION_COOKIE_NAME);
+        if (httpCookie.isEmpty()) {
+            return Optional.empty();
+        }
+
+        String[] cookieValues = httpCookie.get().getValue().split("\\.");
```
CarlyG55, line 113: What's this for?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
CarlyG55, line 1: Do any of these other files in orchestration-shared exist in shared still? If so, we try and keep them up to date as well just in case

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
GHSwallow: Ah yea, that's not needed. It would split "code1.code2" into ["code1", "code2"], but no need for that here.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
GHSwallow: Yep, I'll add them over there too 👍

GHSwallow: The places to add would be in the CookieHelper (add a parser for this browser session id cookie) and in SessionService (add browser session id when creating a session), but the cookie parsers currently in shared authentication are only used in tests, and the `createSession` method has no uses anywhere.

The ConfigurationService contained a feature flag, but if neither CookieHelper nor SessionService have any use for it, it would also be not used anywhere. 

Is there still worth including these unused methods?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
CarlyG55: I don't think you've pushed this change if you've made it 🙂

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
CarlyG55: It makes me a bit nervous not updating the createSession method - I don't think they would, but if someone from auth decided that they want to use that for some reason we'd get some weird behaviour

ethanmills: I think I'm fairly comfortable with that. It would be such a fundamental change that it would either break everything or we would spend a long time thinking about it.

ethanmills: I think what we should do is follow this PR up with one to remove some of the unused auth methods that scare us

CarlyG55: That makes sense

GHSwallow: sounds good:)

GHSwallow: `createSession` deleted here: https://github.com/govuk-one-login/authentication-api/pull/5073

CarlyG55 (approved):

CarlyG55 (approved):

PR #5060, opened 2024-08-15
BAU: Use provisioner from submodule

whi-tw:
## What

Instead of using a relative path to run provisioner.sh, instead include devplatform-deploy as a submodule, and copy the provisioner alongside `provision_all.sh` before running.

This means that the working directory is consistent with the expectation of provisioner.sh.

## How to review

1. Code review
1. Run `provision_all.sh dev` - ensure it works as expected!


billrdunn (approved):
Thanks for this Tom, works as intended for me. I agree with the removal of `&` - I haven't seen a noticeable slowdown and I don't think we'd typically update multiple stacks at once anyway

ethanmills:
the `&` was added because provisioning a new account takes an age. We shouldn't be doing too much of that though, so I'm not too precious about it.

Submodules seems to be a high complexity solution to this, but if you're sure it's the right solution could you update the attached readme please?

whi-tw:
> the `&` was added because provisioning a new account takes an age. We shouldn't be doing too much of that though, so I'm not too precious about it.

More than happy to add it back - I'm just a bit wary with having it in production scripts, as unsupervised processes can be a bit nasty.

> Submodules seems to be a high complexity solution to this, but if you're sure it's the right solution could you update the attached readme please?

It's just more reliable. This way we don't have to worry about people having local tooling up to date / checked out in the right directory. It's a bit overkill, but it feels better than `curl | bash`

PR #5071, opened 2024-08-19
BAU: Add script to prune lambda versions.

andrew-moores:
## What

Re-instating the script that purges old versions of lambdas.  Doing local deploys to the authdev environments created lots of versions of the lambdas that needed pruning, this script is a safe way to do that.

## How to review

1. Code Review
2. Use the di-auth-development account and run with and without specifying an environment to prune. 
3. Check script can be terminated if no environment is specified.
4. Check the help options print a usage message that makes sense.

## Checklist

### DIFF
--- a/scripts/prune-lambdas.sh
+++ b/scripts/prune-lambdas.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# ----------------------------------------------------------------------------------------------------------------------
+# Description:
+# A script to remove old versions of lambdas.  The script requires AWS CLI credentials which should be set in the
+# environment variables.  These credentials can be retrieved from the AWS access portal using 'Access keys' link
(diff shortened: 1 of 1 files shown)


scripts/prune-lambdas.sh
```diff
@@ -0,0 +1,83 @@
...
+    *)
+      echo "Operation aborted by user."
+      exit 1
+      ;;
+  esac
+else
+  echo "pruning lambdas with ${FUNCTION_PREFIX} prefix"
+fi
+
+#VERSIONS_TO_KEEP=5
```
dbes-gds, line 70: lots commented out here

andrew-moores: Restored.

andrew-moores:
No longer needed as the max number of old version of lambdas is not limited.

PR #5072, opened 2024-08-19
ATO-988: Hash state before logging

ethanmills:
State is set by the RP and may contain PII.
We should avoid logging it raw, but it is still useful for correlating requests with an RP when diagnosing issues.


billrdunn (approved):

billrdunn:
Is this the only place where we're logging raw data from an RP which may contain PII?

PR #5077, opened 2024-08-19
BAU: Ignore pact json files with pre-commit cfn-python-lint

whi-tw:
## What

Exclude pact files from cfn-python-lint pre-commit check.

These are not cfn templates and shouldn't be in scope!

## How to review

`pre-commit run cfn-python-lint --all-files` in this branch should return `Passed`


whi-tw:
Already done here: https://github.com/govuk-one-login/authentication-api/pull/4988

PR #5090, opened 2024-08-21
AUT-3583: Added consistent read retry to access token store query

alhcomer:
## What

Added a consistent read retry when querying the access token store. This is to remedy a suspected race condition where the access token store was not being populated before the query was happening. 
This retry strategy aims to fix this by performing a second query with strongly consistent reading (as opposed to eventually consistent reading). This second query happens with a higher latency and cost but ensures up to date DB data.

Have also added cloudwatch metrics incrementing a counter for attempted and successful access token queries for both standard initial reads (eventually consistent reads) and the consistent reads, allowing us to see how often queries are failing. 

See docs here: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html

## How to review

1. Code Review

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
@@ -5,6 +5,8 @@ import com.nimbusds.oauth2.sdk.token.AccessTokenType;
 import com.nimbusds.oauth2.sdk.token.BearerTokenError;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.enhanced.dynamodb.Key;
+import software.amazon.awssdk.enhanced.dynamodb.model.GetItemEnhancedRequest;
 import uk.gov.di.authentication.shared.entity.token.AccessTokenStore;
 import uk.gov.di.authentication.shared.exceptions.AccessTokenException;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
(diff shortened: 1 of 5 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
```diff
@@ -81,4 +83,19 @@ public Optional<AccessTokenStore> setAccessTokenTtlTestOnly(String accessToken,
...
+        Key partitionKey = Key.builder().partitionValue(partition).build();
+        Optional<AccessTokenStore> accessTokenStore =
+                Optional.ofNullable(dynamoTable.getItem(partitionKey));
+
+        if (accessTokenStore.isPresent()) {
+            return accessTokenStore;
+        } else {
+            GetItemEnhancedRequest getItemEnhancedRequest =
+                    GetItemEnhancedRequest.builder().key(partitionKey).consistentRead(true).build();
+            return Optional.ofNullable(dynamoTable.getItem(getItemEnhancedRequest));
```
ethanmills, line 98: I don't think there's any way from this to work out how often we're retrying, and how often that is making a difference

shared/src/main/java/uk/gov/di/authentication/shared/services/AccessTokenService.java
```diff
@@ -81,4 +113,39 @@ public Optional<AccessTokenStore> setAccessTokenTtlTestOnly(String accessToken,
...
+    @Override
+    public Optional<AccessTokenStore> get(String partition) {
+        Key partitionKey = Key.builder().partitionValue(partition).build();
+        Optional<AccessTokenStore> accessTokenStore =
+                Optional.ofNullable(dynamoTable.getItem(partitionKey));
+        incrementCloudwatchCounter(ACCESS_TOKEN_SERVICE_INITIAL_QUERY_ATTEMPT.getValue());
+
+        if (accessTokenStore.isPresent()) {
+            incrementCloudwatchCounter(ACCESS_TOKEN_SERVICE_INITIAL_QUERY_SUCCESS.getValue());
+        } else {
```
ethanmills, line 126: No need to change, but just as a general comment I personally find an early return pattern more readable in situations like this

ethanmills (approved):

andrew-moores (approved):

andrew-moores (approved):

PR #5098, opened 2024-08-22
BAU: Noop segmentation code and remove X-Ray dependency

mrwilson:
These can then be automatically refactored out.

### DIFF
--- a/orchestration-shared/build.gradle
+++ b/orchestration-shared/build.gradle
@@ -22,7 +22,6 @@ dependencies {
             configurations.s3,
             configurations.sqs,
             configurations.ssm,
-            configurations.xray,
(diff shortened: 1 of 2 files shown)


ethanmills:
Would it be worth keeping the call sites and replacing the implementation with something OTEL? I think Dynatrace can read the OTEL spans

ethanmills (approved):
Happy with this change though

ethanmills:
https://docs.dynatrace.com/docs/setup-and-configuration/setup-on-cloud-platforms/amazon-web-services/amazon-web-services-integrations/aws-lambda-integration/opentelemetry-interoperability/lambda-otel-bridge-java

PR #5101, opened 2024-08-22
ATO-964: add tests for different journeys

GHSwallow:
## What
- Added a null check for the session's `browser-session-id`, so that if it's null, nothing is sent back in the cookies. This means if a user is already in an existing session when this change is deployed, their journey will continue unaffected - only new journeys will have a browser session id be created.
- Added tests for expected journeys through `/authorize` for specifically browser session ids. Currently, browser session id has essentially no effect on the journey, so these will need updating in ATO-967

## How to review

1. Code Review

## Checklist

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
@@ -1564,7 +1564,7 @@ class AuthorisationHandlerTest {
         void shouldSetTheRelevantCookiesInTheHeader() {
             when(configService.isBrowserSessionCookieEnabled()).thenReturn(true);
             Session sessionWithBrowserSessionId =
-                    new Session("a-session-id").withBrowserSessionId("a-browser-session-id");
+                    new Session(SESSION_ID).withBrowserSessionId(BROWSER_SESSION_ID);
             when(sessionService.createSession()).thenReturn(sessionWithBrowserSessionId);
 
             Map<String, String> requestParams = buildRequestParams(null);
@@ -1589,7 +1589,9 @@ class AuthorisationHandlerTest {
                             .get(ResponseHeaders.SET_COOKIE)
                             .get(2)
                             .contains(
-                                    "browser-session-id=a-browser-session-id; Domain=auth.ida.digital.cabinet-office.gov.uk; Secure; HttpOnly;"));
+                                    format(
+                                            "browser-session-id=%s; Domain=auth.ida.digital.cabinet-office.gov.uk; Secure; HttpOnly;",
+                                            BROWSER_SESSION_ID)));
         }
 
         @Test
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
@@ -130,6 +130,7 @@ import static org.mockito.Mockito.doThrow;
 import static org.mockito.Mockito.inOrder;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.mockStatic;
+import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.spy;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoInteractions;
@@ -191,6 +192,10 @@ class AuthorisationHandlerTest {
     private static final String AWS_REQUEST_ID = "aws-request-id";
     private static final ClientID CLIENT_ID = new ClientID("test-id");
     private static final String SESSION_ID = "a-session-id";
+    private static final String NEW_SESSION_ID = "a-new-session-id";
+    private static final String BROWSER_SESSION_ID = "a-browser-session-id";
+    private static final String DIFFERENT_BROWSER_SESSION_ID = "a--different-browser-session-id";
+    private static final String NEW_BROWSER_SESSION_ID = "a-new-browser-session-id";
     private static final String REDIRECT_URI = "https://localhost:8080";
     private static final String DOC_APP_REDIRECT_URI = "/doc-app-authorisation";
     private static final String SCOPE = "email openid profile";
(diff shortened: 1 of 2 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
...
+                                        new RequestIdentity().withSourceIp("123.123.123.123")));
+                if (browserSessionIdFromCookie != null) {
+                    event.setHeaders(Map.of("browser-session-id", browserSessionIdFromCookie));
+                }
+
+                return makeHandlerRequest(event);
+            }
+
+            String browserSessionIdCookieFromResponse(APIGatewayProxyResponseEvent response) {
+                return response.getMultiValueHeaders().get(ResponseHeaders.SET_COOKIE).get(2);
```
ethanmills, line 1725: This feels a little fragile. Would this break if we re-ordered the set-cookie code?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
...
+            }
+
+            @Test
+            void noExistingSessionAndNoBrowserSessionIdInHeaders() {
```
ethanmills, line 1729: ideally test names describe a behaviour we're testing for, rather than a scenario

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
                             BASE_AUDIT_USER.withSessionId(SESSION_ID),
                             pair("description", expectedErrorObject.getDescription()));
         }
+
+        @Nested
+        class BrowserSessionIdJourney {
+
+            APIGatewayProxyResponseEvent setupWithExistingSession(
```
isaac-GDS, line 1701: Preference on private functions being defined at the bottom, after the test definitions

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
                             BASE_AUDIT_USER.withSessionId(SESSION_ID),
                             pair("description", expectedErrorObject.getDescription()));
         }
+
+        @Nested
+        class BrowserSessionIdJourney {
```
ethanmills, line 1699: This is not a distinct journey

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
...
+            void existingSessionAndNoBrowserSessionIdInHeaders() {
+                ArgumentCaptor<Session> sessionCaptor = ArgumentCaptor.forClass(Session.class);
+                APIGatewayProxyResponseEvent response =
+                        setupWithExistingSession(
+                                new Session("existing-session-id").withBrowserSessionId("same-id"),
+                                null);
+
+                verify(sessionService, never()).createSession();
+                verify(sessionService).save(sessionCaptor.capture());
+                assertEquals("existing-session-id", sessionCaptor.getValue().getSessionId());
```
ethanmills, line 1769: This feels like an odd thing to test, because we also cycle the session ID, right? This test only passes because we've mocked `sessionService.updateSessionId` to be a noop - it doesn't actually describe the desired system behaviour

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
...
+                        setupWithExistingSession(
+                                new Session("existing-session-id").withBrowserSessionId(null),
+                                "id-from-cookie");
+
+                verify(sessionService, never()).createSession();
+                verify(sessionService).save(sessionCaptor.capture());
+                assertEquals("existing-session-id", sessionCaptor.getValue().getSessionId());
+                assertNull(sessionCaptor.getValue().getBrowserSessionId());
+                assertEquals(
+                        2, response.getMultiValueHeaders().get(ResponseHeaders.SET_COOKIE).size());
```
ethanmills, line 1789:
```suggestion
                assertTrue(
                        response.getMultiValueHeaders().get(ResponseHeaders.SET_COOKIE).stream()
                                .noneMatch(it -> it.startsWith("browser-session-id=")));
```

ethanmills: We want a much stronger assertion here, and one which won't break the next time we touch the cookie code

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -1693,6 +1694,136 @@ void shouldErrorIfIdTokenHasIncorrectClient() throws JOSEException {
...
+            @Test
+            void noExistingSessionAndNoBrowserSessionIdInHeaders() {
+                ArgumentCaptor<Session> sessionCaptor = ArgumentCaptor.forClass(Session.class);
+                APIGatewayProxyResponseEvent response = setupWithExistingSession(null, null);
+
+                verify(sessionService).createSession();
+                verify(sessionService).save(sessionCaptor.capture());
+                assertEquals("a-new-session-id", sessionCaptor.getValue().getSessionId());
+                assertEquals(
+                        "a-new-browser-session-id", sessionCaptor.getValue().getBrowserSessionId());
```
isaac-GDS, line 1737: Can reference constants

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
GHSwallow: Yea I thought about this too. If a user has a session and they have the correct ids in their request to use it, we always cycle the `sessionId` to something new (in this test, we don't, so it's remaining as "existing-session-id"); if not, we generate a new one ("a-new-session-id"). So whether the cycler is mocked or does nothing to the existing session, we still have a 1:1 map of outcomes (I think?), and the session cycler isn't what's being tested. But that's not clear here. 

I think it makes it readable for when we trigger a new session to be created, but that is already tested by the session's browser session id. What do you reckon?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
GHSwallow: replaced with a `.filter().findAny()`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
ethanmills: This makes sense. Happy to keep it, just want to make sure it's been considered properly

ethanmills (approved):

GHSwallow:
Squashed markup commits and rebased onto main

PR #5105, opened 2024-08-22
BAU: stubs Use new api-gateway terraform module

whi-tw:
## What

Previously, these were deployed using 'bare' terraform. As we now have a module that was based from the terraform I've removed in this commit, we can now use that module

Annoyingly, as we are now 'taking ownership' of the execution logging, this requires two `terraform import` statements per-environment:

terraform import -var-file ${environment}.tfvars module.interventions-api-stub_gateway.module.api-gateway.aws_cloudwatch_log_group.execution_logs API-Gateway-Execution-Logs_${gatewayid}/${environment} terraform import -var-file ${environment}.tfvars module.ticf-cri-stub_gateway.module.api-gateway.aws_cloudwatch_log_group.execution_logs API-Gateway-Execution-Logs_${gatewayid}/${environment}

As these imports are dynamic, we can't 'just' use `import` statements within the terraform - it has to be done manually :(

## Checklist


whi-tw:
Thought to get around manually running the imports:

as they are predictable, we could add something to read_secrets.sh to do them there. maybe. but that seems like a super gross hack.

PR #5108, opened 2024-08-23
ATO-964: replace 'browser-session-id' with 'bsid' in cookies

GHSwallow:
## What
Replaces the header cookie 'browser-session-id' with 'bsid', to match other gov.uk cookie names and to make it less obvious what its role is.

## How to review
1. Code Review


ethanmills:
we'll need to update https://github.com/govuk-one-login/authentication-frontend/pull/1970 too

ethanmills (approved):

PR #5125, opened 2024-08-28
ATO-1005: Create dynamo table for auth session

gn-GDS:
## What

Create a new DynamoDB table which will be used to store the Auth session.

## How to review

- Ensure table is setup correctly
- Ensure kms key is setup correctly
- Ensure policies allow correct access (and only from auth)

## Checklist

### DIFF
--- a/ci/terraform/shared/dynamodb.tf
+++ b/ci/terraform/shared/dynamodb.tf
@@ -660,3 +660,35 @@ data "aws_iam_policy_document" "cross_account_identity_credentials_table_resourc
     resources = ["*"]
   }
 }
+
+resource "aws_dynamodb_table" "auth_session_table" {
+  name         = "${var.environment}-auth-session"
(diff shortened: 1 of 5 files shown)


ci/terraform/shared/dynamodb.tf
```diff
@@ -660,3 +660,35 @@ data "aws_iam_policy_document" "cross_account_identity_credentials_table_resourc
...
+  }
+
+  ttl {
+    attribute_name = "TimeToLive"
+    enabled        = true
+  }
+
+  point_in_time_recovery {
+    enabled = true
+  }
```
ethanmills, line 682: We should probably think about if we want pitr on for this or not. It's session data, so I'm not sure if we would ever want to restore it or not

ethanmills (approved):

BeckaL (approved):
Looks sensible to me

ci/terraform/shared/dynamodb.tf
billrdunn: Will make a ticket for it

PR #5133, opened 2024-08-28
ATO-1006: Make Auth and Orch Sessions Independent (1/2)

BenjaminWCO:
## What

- Modified SessionService to allow Auch and Orch Sessions to contain fields that the other does not.
- Renamed SessionService methods to remove ambiguity and be consistent with method names in ClientSessionService.
    - save -> storeOrUpdate
    - updateSessionId -> updateWithNewSessionId
    - createSession -> generateSession

Note: I've made a second PR that removes some unused session fields and adds an integration test for the above change. The reason for splitting this into two PRs is that the above change needs to be in production before it's safe to remove any fields as there could be a race condition between Auth and Orch releases.

## How to review

Code Review

## Checklist

## Related PRs

Part 2: https://github.com/govuk-one-login/authentication-api/pull/5145

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
@@ -203,7 +203,7 @@ public class DocAppCallbackHandler
             }
             var session =
                     sessionService
-                            .readSessionFromRedis(sessionCookiesIds.getSessionId())
+                            .getSession(sessionCookiesIds.getSessionId())
                             .orElseThrow(
                                     () -> {
                                         throw new DocAppCallbackException("Session not found");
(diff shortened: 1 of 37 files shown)


billrdunn:
Is the first commit a prerequisite for the second commit? In some cases I think I prefer the original method names rather than the ones from ClientSession

BenjaminWCO:
> Is the first commit a prerequisite for the second commit? In some cases I think I prefer the original method names rather than the ones from ClientSession

No, it's not a prerequisite. I was a bit hesitant to make the name changes for the same reason. However when trying to modify the logic to support independent session classes I was initially a bit confused because the names weren't quite as verbose as the client session service. _save_ is used for both creation and updating, but the client session service it has separate methods for those. _updateSessionId_ will also update any other fields changed from what is in Redis. I figured given others might be making changes to how and when the session service is called, slightly verbose names might actually be useful for a stop gap until we've fully removed it as a dependancy.

billrdunn:
> > Is the first commit a prerequisite for the second commit? In some cases I think I prefer the original method names rather than the ones from ClientSession
> 
> No, it's not a prerequisite. I was a bit hesitant to make the name changes for the same reason. However when trying to modify the logic to support independent session classes, I was initially a bit confused because the names weren't quite as explicit as the client session service. _save_ is used for both creation and updating, but the client session service it has separate methods for those. _updateSessionId_ will also update any other fields changed from what is in Redis. I figured given others might be making changes to how and when the session service is called, slightly verbose names might actually be useful for a stop gap until we've fully removed it as a dependancy.

Fair enough, I'm convinced 👍

dbes-gds (dismissed):

BenjaminWCO:
Fixed some merge conflicts.

PR #5145, opened 2024-08-29
ATO-1006: Remove Unused Fields From Session Classes (2/2)

BenjaminWCO:
## What

- Removed unused field _browserSessionId_ from Auth Session Class
- Removed unused field _processingIdentityAttempts_ from Auth Session Class
- Removed unused field _passwordResetCount_ from Orch Session Class
- Removed unused method _updateWithNewSessionId_ (formerly named _updateSessionId_) from Auth SessionService
- Replaced AuthOrchSerializationServicesIntegrationTest with SessionServiceIntegrationTest - The former tested only serialisation / deserialisation between Auth and Orch without simulating Redis.

Note: This is the second of two PRs and will require a rebase once the first goes in.

## How to review

Code Review

## Checklist

## Related PRs

Part 1: https://github.com/govuk-one-login/authentication-api/pull/5133

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
@@ -1,37 +1,148 @@
 package uk.gov.di.authentication.shared.services;
 
-import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.mockito.internal.matchers.apachecommons.ReflectionEquals;
 import uk.gov.di.authentication.shared.entity.Session;
-import uk.gov.di.authentication.shared.serialization.Json;
 
-import static org.mockito.Mockito.mock;
+import java.util.Optional;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.empty;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
 
 class AuthOrchSerializationServicesIntegrationTest {
-    private final ConfigurationService configuration = mock(ConfigurationService.class);
-    private final Json authObjectMapper = SerializationService.getInstance();
-    private final uk.gov.di.orchestration.shared.serialization.Json OrchobjectMapper =
-            uk.gov.di.orchestration.shared.services.SerializationService.getInstance();
+
+    private static final String REDIS_HOST =
+            System.getenv().getOrDefault("REDIS_HOST", "localhost");
+    private static final Optional<String> REDIS_PASSWORD =
+            Optional.ofNullable(System.getenv("REDIS_PASSWORD"));
+    private static final String BROWSER_SESSION_ID = "browser-session-id";
+    private static final String CLIENT_SESSION_ID = "client-session-id";
+
+    private uk.gov.di.orchestration.shared.services.SessionService orchSessionService;
+    private uk.gov.di.authentication.shared.services.SessionService authSessionService;
+
+    @BeforeEach
+    void setup() {
+        var orchConfigurationService =
+                uk.gov.di.orchestration.shared.services.ConfigurationService.getInstance();
+        var authConfigurationService =
+                uk.gov.di.authentication.shared.services.ConfigurationService.getInstance();
+        var orchRedisConnectionService =
+                new uk.gov.di.orchestration.shared.services.RedisConnectionService(
+                        REDIS_HOST, 6379, false, REDIS_PASSWORD, false);
+        var authRedisConnectionService =
+                new uk.gov.di.authentication.shared.services.RedisConnectionService(
+                        REDIS_HOST, 6379, false, REDIS_PASSWORD, false);
+
+        orchSessionService =
+                new uk.gov.di.orchestration.shared.services.SessionService(
+                        orchConfigurationService, orchRedisConnectionService);
+        authSessionService =
+                new uk.gov.di.authentication.shared.services.SessionService(
+                        authConfigurationService, authRedisConnectionService);
+    }
 
     @Test
-    void sessionsSerializedByAuthCanBeDeserializedByOrch()
-            throws Json.JsonException,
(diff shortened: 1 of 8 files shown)


billrdunn (commented):
Logic looks sound to me, few thoughts:

- I prefer something like `AuthOrchSerializationServicesIntegrationTest` to `SessionServiceIntegrationTest` as there's nothing in the latter that says it involves Auth and Orch (particularly as it lives is in Auth code).
- I don't think the comments in the new integration test are necessary (and not what we generally do style-wise).
- Is there definitely no value in keeping the old integration test?
- [This field ](https://github.com/govuk-one-login/authentication-api/pull/5256/commits/204daf577bd8cca25baf2c5aaf6ad0eca26ac5c1) was recently added to both Sessions, can it be removed from Orch Session in the same way?
- I'd suggest deploying this to integration for a little while before releasing to prod, just to be sure.

BenjaminWCO:
> Logic looks sound to me, few thoughts:
> 
> 1. I prefer something like `AuthOrchSerializationServicesIntegrationTest` to `SessionServiceIntegrationTest` as there's nothing in the latter that says it involves Auth and Orch (particularly as it lives is in Auth code).
> 2. I don't think the comments in the new integration test are necessary (and not what we generally do style-wise).
> 3. Is there definitely no value in keeping the old integration test?
> 4. [This field ](https://github.com/govuk-one-login/authentication-api/pull/5256/commits/204daf577bd8cca25baf2c5aaf6ad0eca26ac5c1) was recently added to both Sessions, can it be removed from Orch Session in the same way?
> 5. I'd suggest deploying this to integration for a little while before releasing to prod, just to be sure.

1. Yes, that'd makes a lot of sense
2. The reason I added them was because it was a very artificial scenario just for the sake of demonstrating the behaviour with independent fields, so if I had made a mistake in the logic of the test, there wouldn't be a way of deciphering my intention. I don't mind removing them.
3. The old test is written with the explicit assumption that both session classes are identical. We could remove the use of browser session id as that doesn't exist on the auth session now, however at that point we're just only check the client session id in the old test, which is something the new test also checks. I don't think it really adds anything to keep it, but I we make it work if you wanted to keep it for the sake of caution.
4. Yes, I can add that to this PR.
5. Yeah, sounds good.

billrdunn (approved):

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
```diff
@@ -1,37 +1,93 @@
...
-                        .withBrowserSessionId("browser-session-id")
-                        .addClientSession("client-session-id");
-
-        var authSerializedSession = authObjectMapper.writeValueAsString(authSession);
-        var orchDeserialized =
-                OrchobjectMapper.readValue(
-                        authSerializedSession, uk.gov.di.orchestration.shared.entity.Session.class);
-
-        Assertions.assertTrue(new ReflectionEquals(orchSession).matches(orchDeserialized));
+    void authAndOrchSessionsShouldBeAbleToHaveDifferentFieldsWithoutOverwritingOneAnother() {
```
GHSwallow, line 48: I think these can be separated into multiple tests, so it's more readable and we won't need code comments. Eg 
1) create orch session, assert auth can read it
2) visa versa
3) create orch session, auth reads and edits some shared property (if there are any), assert orch can still read it and has the shared property updated
4) visa versa
5) orch creates session with browserSessionId (or any other property only orch has), auth reads and stores the session, assert orch still has the browserSessionId (and perhaps that the auth session doesn't)
6) visa versa
7) create orch session, auth can read it, then orch cycles sessionId, and auth can still read it with the new sessionId

What do you reckon?

BenjaminWCO: Not a bad idea.

BenjaminWCO: Done

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
```diff
@@ -1,37 +1,150 @@
...
+        var orchSession = orchSessionService.generateSession();
+        var sessionId = orchSession.getSessionId();
+        orchSession.setBrowserSessionId(BROWSER_SESSION_ID);
+        orchSessionService.storeOrUpdateSession(orchSession);
+        var authSession = authSessionService.getSession(sessionId).get();
+        authSession.addClientSession(CLIENT_SESSION_ID);
+        authSessionService.storeOrUpdateSession(authSession);
+        orchSession = orchSessionService.getSession(sessionId).get();
+        assertThat(
+                orchSession.getBrowserSessionId(), is(equalTo(orchSession.getBrowserSessionId())));
```
GHSwallow, line 103: I think one of these should be replaced with `BROWSER_SESSION_ID`. Currently asserting that it is itself

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
```diff
@@ -1,37 +1,150 @@
...
+        authSessionService.storeOrUpdateSession(authSession);
+        orchSession = orchSessionService.getSession(originalSessionId).get();
+        orchSessionService.updateWithNewSessionId(orchSession);
+        var newSessionId = orchSession.getSessionId();
+        authSessionService.getSession(newSessionId).get();
+        assertThat(authSession.getClientSessions(), contains(CLIENT_SESSION_ID));
+    }
+
+    @Test
+    void authCanResetSharedFieldsWithoutOverridingUnsharedFields() {
```
GHSwallow, line 137: Nice!

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/AuthOrchSerializationServicesIntegrationTest.java
BenjaminWCO: Haha, probably

GHSwallow (approved):
Approving for the test approach, and with Bill's approval on the logic ✅

PR #5156, opened 2024-08-30
AUT-3529: Added SMS code reauth attempt count

alhcomer:
## What

This PR implements the create or increment count logic into the verify code handler, incrementing the count for
ENTER_SMS_CODE in the authentication attempts table for a reauth journey when a user fails to enter the correct SMS MFA. If there is a count for a user having failed to enter the right sms mfa code and they then succeed, the count will be deleted.

A PR will follow implementing the corresponding logic for the auth app MFA in the verifyMfaCodeHandler.

## How to review

1. Code Review
2.  Deploy changes to authdev1
3.  Go through reauth journey successfully up to entering SMS MFA
4. Enter an incorrect MFA and see the count in the authdev1-authentication-attempt table
5. Enter the correct MFA code and see that the count clears

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -12,6 +12,7 @@ import uk.gov.di.authentication.frontendapi.entity.VerifyCodeRequest;
 import uk.gov.di.authentication.frontendapi.helpers.SessionHelper;
 import uk.gov.di.authentication.shared.domain.AuditableEvent;
 import uk.gov.di.authentication.shared.entity.CodeRequestType;
+import uk.gov.di.authentication.shared.entity.CountType;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
 import uk.gov.di.authentication.shared.entity.MFAMethodType;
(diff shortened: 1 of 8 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -244,6 +274,13 @@ private void processSuccessfulCodeRequest(
                     levelOfConfidence.getValue(),
                     clientService.isTestJourney(clientId, session.getEmailAddress()),
                     true);
+
+            if (configurationService.isAuthenticationAttemptsServiceEnabled()) {
```
BeckaL, line 278: We should only do this if this a reauth journey right?

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandlerTest.java
```diff
@@ -679,6 +690,46 @@ void shouldReturn204ForValidResetPasswordRequestUsingTestClient() {
...
+    @ParameterizedTest
+    @MethodSource("codeRequestTypes")
+    void shouldDeleteCountOnSuccessfulSMSCodeRequest(
+            CodeRequestType codeRequestType, JourneyType journeyType) {
+        when(codeStorageService.getOtpCode(EMAIL, MFA_SMS)).thenReturn(Optional.of(CODE));
+        when(configurationService.isAuthenticationAttemptsServiceEnabled()).thenReturn(true);
+
+        var result = makeCallWithCode(CODE, MFA_SMS.toString(), journeyType);
+
+        verify(authenticationAttemptsService, times(1))
```
BeckaL, line 702: Yeah here I was suspicious that although we're passing in journey type as a param, we're expecting this to be deleted in all cases. Had a look at the implementation and think we're missing a check

BeckaL (approved):

PR #5172, opened 2024-09-03
ATO-984 (3/5): Add previousSessionId claim to authorize request to Authentication

billrdunn:
## What

Include `previousSessionId` as a claim in authorize request to Authentication. If there is no `previousSessionId`, no claim is added.

## Why?

This is required in order to break session dependencies between Authentication and Orchestration. When Auth has its own session store, the hash key for the session will be `sessionId`. When `sessionId` gets updated, the record will need to be recreated, and in order to do this, Auth need to map the previous session ID to the new one (which will be extracted from the gs cookie).

## Testing

Has been deployed to sandpit / dev along with the changes below. On first journey `previousSessionId` is `null` in StartHandler. On second journey it is the previous value of `sessionId`.

## Related PRs

ATO-984 incremental deployment:
1. https://github.com/govuk-one-login/authentication-api/pull/5171
2. https://github.com/govuk-one-login/authentication-frontend/pull/2006
3. https://github.com/govuk-one-login/authentication-api/pull/5172
4. https://github.com/govuk-one-login/authentication-api/pull/5174
5. TODO

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -481,10 +481,10 @@ public class AuthorisationHandler
             updateAttachedSessionIdToLogs(session.getSessionId());
             LOG.info("Created session");
         } else {
-            var oldSessionId = session.getSessionId();
+            var previousSessionId = session.getSessionId();
             sessionService.updateSessionId(session);
             updateAttachedSessionIdToLogs(session.getSessionId());
-            LOG.info("Updated session id from {} - new", oldSessionId);
+            LOG.info("Updated session id from {} - new", previousSessionId);
         }
 
         Subject subjectId =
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -567,14 +567,15 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
         if (existingSession.isEmpty()) {
             updateAttachedSessionIdToLogs(session.getSessionId());
             LOG.info("Created session");
         } else {
-            var oldSessionId = session.getSessionId();
+            previousSessionId = Optional.of(session.getSessionId());
             sessionService.updateSessionId(session);
             updateAttachedSessionIdToLogs(session.getSessionId());
-            LOG.info("Updated session id from {} - new", oldSessionId);
+            LOG.info("Updated session id from {} - new", previousSessionId);
```
Ryan-Andrews99, line 578: Random question but what does passing an optional look like when logged? Just wondering if it actually logs out a meaningful value and not something weird

billrdunn: Good question - it looks like `"Updated session id from Optional[5RdN6mlCmeLp2Q-3s6FbSdS3Anc] - new"`

Ryan-Andrews99 (approved):

PR #5198, opened 2024-09-09
Aut 3530: check all reauth counts in reauth handler

BeckaL:
## What

Contains a refactor to the way we retrieve and check reauth counts, in order to:

* Make our code more flexible - this way, the calling code can decide what to do in the case of different exceeded counts, and later on it'll make life easier when we want to populate audit events with all counts.
* Make helpers contain static methods

Then, it updates the check reauth user handler to ensure that we're checking all reauth counts, and log a user out in the case of any that exceed the configured maximum.

## How to review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
@@ -192,9 +192,9 @@ public class StartService {
                                             authenticationAttemptsService.getCountsByJourney(
                                                     subjectId, JourneyType.REAUTHENTICATION))
                             .orElse(Map.of());
-            var helper = new ReauthAuthenticationAttemptsHelper(configurationService);
             var blockedCountTypes =
-                    helper.countTypesWhereUserIsBlockedForReauth(reauthCountTypesToCounts);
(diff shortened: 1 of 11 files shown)


alhcomer (approved):
this is great, enjoyed going through it. all makes sense to me

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -183,10 +185,19 @@ public UserStartInfo buildUserStartInfo(
 
         boolean isBlockedForReauth = false;
         if (configurationService.isAuthenticationAttemptsServiceEnabled()) {
-            isBlockedForReauth =
+            var reauthCountTypesToCounts =
```
andrew-moores, line 188: Should this be singular (`reauthCountTypesToCount`)?  Could it be renamed to `errorTypesWithCounts`?

andrew-moores (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
BeckaL: As discussed - will have a think about this as part of next PR 👍

PR #5207, opened 2024-09-10
ATO-966: turn on bsid cookie in integration and production

GHSwallow:
## What
Turns on cookie and metrics logging feature flag in integration and production

## How to review

1. Code Review

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -71,6 +71,8 @@ Conditions:
       !Equals [dev, !Ref Environment],
       !Equals [build, !Ref Environment],
       !Equals [staging, !Ref Environment],
+      !Equals [integration, !Ref Environment],
+      !Equals [production, !Ref Environment],
     ]
 
 Mappings:


template.yaml
```diff
@@ -71,6 +71,8 @@ Conditions:
       !Equals [dev, !Ref Environment],
       !Equals [build, !Ref Environment],
       !Equals [staging, !Ref Environment],
+      !Equals [integration, !Ref Environment],
+      !Equals [production, !Ref Environment],
```
isaac-GDS, line 75: Do we want it in Prod yet ? The plan was to do the tickets in ToDo column first to test before going into Prod.

GHSwallow: This is just the logging, so the only thing that's being turned on is the presence of this extra cookie - it won't change anything about the journey. Does anything else in ToDo need to go through before then? Perhaps https://govukverify.atlassian.net/browse/ATO-1013?

isaac-GDS: Ah okay - should be fine to go in as we aren't changing the journey. Think that ticket can be done after.

isaac-GDS (approved):

PR #5211, opened 2024-09-10
Aut 3530/reauth lockouts

andrew-moores:
## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -158,7 +158,11 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
                 return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1049);
             }
 
-            UserProfile userProfile = userProfileMaybe.get();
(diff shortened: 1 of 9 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -158,29 +158,8 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
-                                countsByJourney, configurationService);
-
-                if (!countTypesWhereBlocked.isEmpty()) {
-                    LOG.info(
-                            "Re-authentication locked due to {} counts exceeded.",
-                            countTypesWhereBlocked);
-                    return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1057);
-                }
-            }
+            if (checkErrorCountsForReauth(journeyType, userProfileMaybe))
```
BeckaL, line 161: I'm not suggesting to fix as part of this PR, but maybe for a follow up if we get to it, a more descriptive name for this might be "anyRetryCountsForReauthExceededOnReauthJourney" or something that indicates it's returning a bool. Also think might read better for me if the "if reauth journey" is done at this level, and it's clear we're only checking in the case of a reauth journey

BeckaL (approved):
Minor comment for future but doesn't need changing now

PR #5219, opened 2024-09-11
ATO-981 set is new account in userinfo

Ryan-Andrews99:
## What
- We're now setting the value of `isNewAccount` in the Auth userInfo lambda. This is a pre-requisite to splitting this value from the shared session store as currently Auth rely on Orchestration to update this in `AuthenticationCallBack` handler and `IPVCallBack` handler. We're setting it here as Auth need to start independently setting this value, and this point is closest to the end of the handoff between Auth and Orch, and closest to when it would be set by Orch.
- Soon we will co-set this value in the new Auth session store, before eventually migrating to the auth only store.

## How to review

For example:
- Deployed to sandpit and ran through a journey, confirmed that the session is updated in the userinfo handler

## Checklist

### DIFF
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
@@ -186,6 +186,8 @@ public class UserInfoHandler
                     accessToken.getValue());
         }
 
+        sessionService.storeOrUpdateSession(
+                userSession.setNewAccount(Session.AccountState.EXISTING));
         return generateApiGatewayProxyResponse(200, userInfo.toJSONString());
     }
 
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
@@ -14,6 +14,8 @@ import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
 import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.external.services.UserInfoService;
+import uk.gov.di.authentication.shared.entity.ErrorResponse;
+import uk.gov.di.authentication.shared.entity.Session;
 import uk.gov.di.authentication.shared.entity.token.AccessTokenStore;
 import uk.gov.di.authentication.shared.exceptions.AccessTokenException;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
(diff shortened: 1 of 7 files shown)


auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
```diff
@@ -77,18 +85,22 @@ void shouldReturn200WithUserInfoForValidRequestAndSetTokenStoreUsed()
...
 
         APIGatewayProxyResponseEvent response = userInfoHandler.userInfoRequestHandler(request);
+        Session updatedSession = testSession.setNewAccount(Session.AccountState.EXISTING);
 
         assertEquals(200, response.getStatusCode());
         assertTrue(
                 response.getBody()
                         .contains(String.format("\"sub\":\"%s\"", TEST_SUBJECT.getValue())));
 
+        verify(sessionService).getSessionFromRequestHeaders(request.getHeaders());
```
BeckaL, line 103: Do we need this assertion? This feels to me a bit too much like testing the internals of a function. We set up what this should return on line 92, that should be enough.

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
```diff
@@ -186,6 +186,8 @@ public APIGatewayProxyResponseEvent userInfoRequestHandler(APIGatewayProxyReques
                     accessToken.getValue());
         }
 
+        sessionService.save(userSession.setNewAccount(Session.AccountState.EXISTING));
+        LOG.info("Set account state to EXISTING");
```
BeckaL, line 190: Is this logging useful?

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
```diff
@@ -186,6 +186,8 @@ public APIGatewayProxyResponseEvent userInfoRequestHandler(APIGatewayProxyReques
                     accessToken.getValue());
         }
 
+        sessionService.save(userSession.setNewAccount(Session.AccountState.EXISTING));
```
BeckaL, line 189: Is the logic here that as long as the request is successful, any account (including new accounts) are marked as existing at this point?

Could you do some combo (maybe both :-)) of:
* writing a bit more about why this behaviour is useful in the commit message
* call it out as a specific scenario in a test (updates a new account to existing when user info handler request processed correctly?)

BeckaL (commented):
Nice! Thanks for an easily digestible review. A couple of comments / questions / suggestions below

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
Ryan-Andrews99: Yeah I see what you mean, feels a bit unnecessary, have removed!

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
Ryan-Andrews99: It was useful for testing, doesn't particularly tell you anything useful outside of "the code did its job" which is expected. Will remove 🙂

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
Ryan-Andrews99: I have added a separate test case for this and tried to elaborate a bit more in the commit message. Let me know if commit message is too context-y and not explain-y enough 😅

BeckaL (approved):
Lovely stuff!

billrdunn (dismissed):
A beautiful PR  👏

BeckaL (approved):

PR #5231, opened 2024-09-13
AUT-3601: In order to construct the REAUTH_SUCCESS audit event we nee…

andrew-moores:
…d to access the error counts table to report how many times a User presented an invalid credential before completing the re-authentication.

The last step of re-authentication is deemed to be the AuthenticationAuthCodeHandler but the error counts are deleted in the Verify Code handlers.

We will move the deletion of the error counts to the AuthenticationAuthCodeHandler.

## Checklist


BeckaL:
Closing this in favour of different approach where we preserve counts in the session, rather than change where the counts are wiped (changing where the counts are wiped opens up a whole can of worms in terms of silent login behaviour and other edge cases)

PR #5248, opened 2024-09-18
ATO-979/a: Set verifiedMfaMethodType in Auth session

billrdunn:
## Background
In order to break the shared session dependency between Auth and Orch, session data are being moved to an Auth owned Auth session table.

There are three parts to breaking the `verifiedMfaMethodType` dependency, this PR is part a:
- ATO-979/a: VerifyCode & VerifyMfaCode: Set in Auth session
- ATO-979/b: UserInfoHandler: Get from Auth session and include in userinfo response
- ATO-979/c: AuthenticationCallbackHandler: Take from userinfo response and set in Orch session

## What

This PR sets `verifiedMfaMethodType` in the Auth session, in the VerifyCode and VerifyMfaCode lambdas.

Note that we exceeded the quote of 20 policies per role, so as a workaround the read and write policies have been combined. Ticket to track requesting a quota increase: https://govukverify.atlassian.net/browse/ATO-1056

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -146,7 +146,8 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
             LOG.info("Processing request");
 
             var session = userContext.getSession();
-            Optional<String> authSessionId = authSessionService.getSessionIdFromRequestHeaders(input.getHeaders());
+            Optional<String> authSessionId =
+                    authSessionService.getSessionIdFromRequestHeaders(input.getHeaders());
             if (authSessionId.isEmpty()) {
                 LOG.warn("Auth session ID cannot be found");
                 return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
@@ -225,6 +226,7 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
 
             processSuccessfulCodeRequest(
                     session,
+                    authSessionId.get(),
                     codeRequest,
                     userContext,
                     userProfileMaybe.isPresent() ? userProfileMaybe.get().getSubjectID() : null,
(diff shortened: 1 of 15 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -339,6 +342,7 @@ private void processSuccessfulCodeRequest(
                     false);
             sessionService.storeOrUpdateSession(
                     session.setVerifiedMfaMethodType(MFAMethodType.SMS));
+            authSessionService.setVerifiedMfaMethodType(authSessionId, MFAMethodType.SMS);
```
mattvot, line 345: Wondering what we're going to do about migrating this value from the session table to the auth session table

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -292,6 +294,9 @@ private APIGatewayProxyResponseEvent verifyCode(
                                             .setVerifiedMfaMethodType(
                                                     codeRequest.getMfaMethodType()));
 
+                            authSessionService.setVerifiedMfaMethodType(
+                                    authSessionId, codeRequest.getMfaMethodType());
+
```
mattvot, line 299: Same migration related question as previous. Is something in hand already for when the line above this one will be removed?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
billrdunn: The plan is to set to both the old session and the new session for some period of time (greater than the session expiry). By that point the old and the new should contain the same data, so we can start getting from the new session. We may implement a fallback to get from the old session in case it's not found in the new session. Then we'll stop setting to the old session.

mattvot (approved):

mattvot (dismissed):

mattvot (approved):

PR #5250, opened 2024-09-18
Aut 3632/increment user submitted credential

BeckaL:
## What

Increments a new metric, UserSubmittedCredential, whenever a user is verifying a new email or phone number.

## How to review

1. Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
@@ -32,6 +32,7 @@ import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.serialization.Json.JsonException;
 import uk.gov.di.authentication.shared.services.AuditService;
 import uk.gov.di.authentication.shared.services.ClientService;
+import uk.gov.di.authentication.shared.services.CloudwatchMetricsService;
 import uk.gov.di.authentication.shared.services.CodeGeneratorService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 import uk.gov.di.authentication.shared.services.DynamoClientService;
@@ -52,6 +53,7 @@ import static uk.gov.di.authentication.shared.entity.ErrorResponse.ERROR_1002;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
 import static uk.gov.di.authentication.shared.helpers.ApiGatewayResponseHelper.generateEmptySuccessApiGatewayResponse;
+import static uk.gov.di.authentication.shared.helpers.FraudCheckMetricsHelper.incrementUserSubmittedCredentialIfNotificationSetupJourney;
 import static uk.gov.di.authentication.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
 import static uk.gov.di.authentication.shared.helpers.LocaleHelper.getUserLanguageFromRequestHeaders;
 import static uk.gov.di.authentication.shared.helpers.LocaleHelper.matchSupportedLanguage;
@@ -74,6 +76,7 @@ public class SendOtpNotificationHandler
     private final ClientService clientService;
     private final Json objectMapper = SerializationService.getInstance();
     private final AuditService auditService;
+    private final CloudwatchMetricsService cloudwatchMetricsService;
     private static final String GENERIC_500_ERROR_MESSAGE = "Internal server error";
 
     public SendOtpNotificationHandler(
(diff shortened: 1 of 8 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/EmailCheckMetricsHelper.java
```diff
@@ -0,0 +1,33 @@
...
+import uk.gov.di.authentication.shared.entity.JourneyType;
+import uk.gov.di.authentication.shared.entity.NotificationType;
+import uk.gov.di.authentication.shared.services.CloudwatchMetricsService;
+
+import java.util.List;
+import java.util.Map;
+
+import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.ENVIRONMENT;
+
+public class EmailCheckMetricsHelper {
```
mattvot, line 17: Would be worth renaming this, maybe to `FraudCheckMetricsHelper`, as this will be used for both phone and email

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
```diff
@@ -169,6 +175,12 @@ public APIGatewayProxyResponseEvent sendOtpRequestHandler(
             return generateApiGatewayProxyResponse(500, GENERIC_500_ERROR_MESSAGE);
         }
 
+        incrementUserSubmittedCredentialIfNotificationSetupJourney(
+                cloudwatchMetricsService,
+                JourneyType.ACCOUNT_MANAGEMENT,
+                sendNotificationRequest.getNotificationType().name(),
+                configurationService.getEnvironment());
```
mattvot, line 182: I think this could work here for now. I'm slightly hesitant just because we're not conditioning this if `sendNotificationRequest.getNotificationType()` is `VERIFY_EMAIL` or `VERIFY_PHONE_NUMBER`.

Should another notification type be supported by this lambda in the future, we may get some extra metrics recorded that don't make sense in context (i.e. they may not be about "user submitting a credential").

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SendNotificationHandler.java
```diff
@@ -295,6 +296,13 @@ private APIGatewayProxyResponseEvent handleNotificationRequest(
         sessionService.storeOrUpdateSession(
                 session.incrementCodeRequestCount(
                         request.getNotificationType(), request.getJourneyType()));
+
+        incrementUserSubmittedCredentialIfNotificationSetupJourney(
+                METRICS,
+                request.getJourneyType(),
+                request.getNotificationType().name(),
+                configurationService.getEnvironment());
```
mattvot, line 304: Sorry @BeckaL. Looking at this I've seen that I've not specified enough on this ticket, as `request.getJourneyType() == JourneyType.REGISTRATION` must also be `true` for this metric to be raised.

We'd also need to condition this on `notificationType == NotificationType.VERIFY_EMAIL || notificationType == NotificationType.VERIFY_PHONE_NUMBER` please, as there are other notification types that are handled by this lambda but that shouldn't have this metric raised for them.

It may be cleaner to execute `incrementUserSubmittedCredentialIfNotificationSetupJourney` within the conditions below that handle the different notification types (phone will need a condition on journey type added too).

shared/src/main/java/uk/gov/di/authentication/shared/helpers/EmailCheckMetricsHelper.java
```diff
@@ -0,0 +1,43 @@
...
+    public static void incrementUserSubmittedCredentialIfNotificationSetupJourney(
+            CloudwatchMetricsService metricsService,
+            JourneyType journeyType,
+            String notificationType,
+            String environment) {
+        if (notificationSetupTypes.contains(notificationType)) {
+            var credentialType =
+                    Objects.equals(notificationType, NotificationType.VERIFY_EMAIL.name())
+                            ? CredentialType.EMAIL_ADDRESS
+                            : CredentialType.PHONE_NUMBER;
```
mattvot, line 34: As we may find any random string is put in as the `notificationType` in the future, could we assert the string is either `VERIFY_EMAIL` or `VERIFY_PHONE_NUMBER` before using it?

I'm aware if we make a mistake in the future that we may fallback to miss-recording a the `CredentialType` dimension as `PHONE_NUMBER`.

mattvot (changes requested):
Thanks @BeckaL. Approach looks great. Nice use of a the helper. I've left a few of comments (and clarified where I missed something when scoping)

shared/src/main/java/uk/gov/di/authentication/shared/helpers/EmailCheckMetricsHelper.java
mattvot: Redacted, you are already doing this with `notificationSetupTypes.contains(notificationType)` in `EmailCheckMetricsHelper

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
mattvot: Redacted, you are already doing this with `notificationSetupTypes.contains(notificationType)` in `EmailCheckMetricsHelper

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SendNotificationHandler.java
mattvot: After discussion, we don't actually need to condition this on journey type and we already have a check of the notification type in the metrics helper

mattvot (approved):

PR #5255, opened 2024-09-20
AUT-3602: Add reauth_failed audit event to StartHandler

alhcomer:
## What

The event is emitted when a user has had too many pre-existing reauthentication attempts and is blocked from reauthentication.

Also refactored the start handler and service so that the counts from the authentication attempts service are only queried
once from the handler. The query in the start service has been removed with the handler query passing the isBlockedForReauth value through to the service to be user in the UserStartInfo.

## How to review

1. Code Review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -194,7 +194,6 @@ public class StartHandler
             var txmaAuditHeader =
                     getOptionalHeaderValueFromHeaders(
                             input.getHeaders(), TXMA_AUDIT_ENCODED_HEADER, false);
-            String internalSubjectIdForAuditEvent = AuditService.UNKNOWN;
             String internalCommonSubjectIdentifierForAuditEvent = AuditService.UNKNOWN;
             String phoneNumber = AuditService.UNKNOWN;
 
@@ -242,6 +241,7 @@ public class StartHandler
 
             StartResponse startResponse = new StartResponse(userStartInfo, clientStartInfo);
 
+            String internalSubjectIdForAuditEvent = AuditService.UNKNOWN;
             if (userStartInfo.isAuthenticated()) {
                 LOG.info(
                         "User is authenticated. Setting internalCommonSubjectId and internalSubjectId for audit event");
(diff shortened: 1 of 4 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -188,19 +186,48 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                    .map(UserProfile::getEmail)
+                                    .orElse(AuditService.UNKNOWN),
+                            IpAddressHelper.extractIpAddress(input),
+                            AuditService.UNKNOWN,
+                            extractPersistentIdFromHeaders(input.getHeaders()),
+                            txmaAuditHeader);
+
+            boolean isBlockedForReauth = false;
+            if (configurationService.isAuthenticationAttemptsServiceEnabled() && reauthenticate) {
+                isBlockedForReauth =
```
andrew-moores, line 217: If a User is blocked from re-authenticating and they are attempting to re-authenticate why do we allow them to continue here?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -214,36 +241,16 @@ public APIGatewayProxyResponseEvent handleRequest(
 
             StartResponse startResponse = new StartResponse(userStartInfo, clientStartInfo);
 
-            String internalSubjectIdForAuditEvent = AuditService.UNKNOWN;
-            String internalCommonSubjectIdentifierForAuditEvent = AuditService.UNKNOWN;
-
             if (userStartInfo.isAuthenticated()) {
                 LOG.info(
                         "User is authenticated. Setting internalCommonSubjectId and internalSubjectId for audit event");
```
andrew-moores, line 246: Does this log message have any value in production?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
andrew-moores: Makes sense.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -188,19 +186,48 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                    .orElse(AuditService.UNKNOWN),
+                            IpAddressHelper.extractIpAddress(input),
+                            phoneNumber,
+                            extractPersistentIdFromHeaders(input.getHeaders()),
+                            txmaAuditHeader);
+
+            boolean isBlockedForReauth = false;
+            if (configurationService.isAuthenticationAttemptsServiceEnabled() && reauthenticate) {
+                isBlockedForReauth =
+                        checkUserIsBlockedForReauthAndEmitFailureAuditEvent(
```
andrew-moores, line 218: This event is being emitted before the `AUTH_REAUTH_REQUESTED` event at line 259 which makes the order of the events difficult to follow.  Consider moving the code that emits the `AUTH_REAUTH_REQUESTED` to a point before this event is emitted.

andrew-moores (approved):

PR #5256, opened 2024-09-20
Aut 3743/preserve reauth counts in session

BeckaL:
## What

Populates the attempt counts in the AUTH_REAUTH_SUCCESS event by:

* Storing them in the user session upon wiping them from the code count store
* Reading them from the session in the auth code handler to populate the audit event
* Wiping them from the session once they've been used.

## How to review

1. Code Review

I've tested this, and you can too if you like!
1. Deploy to a test env (I've got authdev1 reserved)
1. Sign in
1. Start a reauth journey - purge the queue as per step below before continuing
1. In aws, purge the audit queue (authdev1-oidc-txma-audit-queue) so it's easier to isolate messages from your journey
1. Complete your reauth journey successfully, but entering a few incorrect details (not enough to fail reauth)
1. Poll for messages on the queue (tip: edit the polling settings so that max polled = 100 or so)
1. Search for the "REAUTH_AUTH_SUCCESS" message in the search bar on the polled messages
1. See that the relevant counts have been populated

I have repeated this test for sms and auth app, and also retested going through a reauth journey under the same account to ensure that the counts from one journey don't persist to the next.

## Checklist


alhcomer (dismissed):
Looks good from an auth perspective, I tested in authdev also

BeckaL:
Need to check whether we add to orch session too

BeckaL:
Added to both sessions and orch have confirmed they're happy, so removing do not merge label

billrdunn:
> Need to check whether we add to orch session too

LGTM on this point

I believe adding to both Session classes will be unnecessary soon due to work making the [sessions independent](https://github.com/govuk-one-login/authentication-api/pull/5133), but fine to play it safe for now and add to both

alhcomer (approved):

PR #5257, opened 2024-09-20
Bau/refactor check reauth users to not lookup so much

BeckaL:
## What

Previously, we were making multiple duplicate calls to look up counts from the database for reauth:

1. an initial call when the user has an account, to check if we should immediately log them out due to exceeded counts
2. a call when emitting the success audit event result, to populate the extensions
3. another when the user submits an incorrect code
4. another when the user submits an incorrect code and breaches the max retries exceeded limit

In this PR, we just do this lookup once - either at the beginning for a user who exists, or when emitting the failed event for a user who does not exist. In the case of an incorrect code, we increment both the database value and either the value in memory. 

This limits the number of database calls made.

## How to review

1. Code Review


BeckaL:
Parking this for now

PR #5265, opened 2024-09-23
ATO-1060: Add channel field to client registry which can be WEB or STRATEGIC_APP

billrdunn:
## What

Add `channel` field to client registry. The possible values are "WEB" (default) and "STRATEGIC_APP".

The handlers to register / update a client in the client registry have also been updated.

`channel` will be passed to Authentication via the authorize request. "STRATEGIC_APP" is a request for mobile view.

## Testing

Once this is deployed to integration, I'll test adding a new client and check it sets the value of `channel` to "WEB".


client-registry-api/src/test/java/uk/gov/di/authentication/clientregistry/services/ClientConfigValidationServiceTest.java
```diff
@@ -68,7 +70,8 @@ private static Stream<Arguments> registrationRequestParams() {
                         null,
                         null,
                         null,
-                        null),
+                        null,
+                        Channel.WEB.getValue()),
```
isaac-GDS, line 74: We should also test `channel = STRATEGIC_APP` for one of these or a new argument

isaac-GDS:
One small comment but then after that happy for it to get merged

isaac-GDS (dismissed):

isaac-GDS (approved):

PR #5266, opened 2024-09-23
ATO-1029: Set JAVA_TOOL_OPTIONS env var in all Orchestration lambdas

billrdunn:
## What

Add JAVA_TOOL_OPTIONS environment variable to all lambdas, with the value:
- `-XX:+TieredCompilation -XX:TieredStopAtLevel=1` in production
- `-XX:+TieredCompilation -XX:TieredStopAtLevel=1 '--add-reads=jdk.jfr=ALL-UNNAMED'` in all other envs.

This is required for updating the dynatrace lambda layer version.

Note that the production value is the default (and current) value.


ethanmills:
The other values are not required now we're on 17, which is why there were removed: https://aws.amazon.com/blogs/compute/java-17-runtime-now-available-on-aws-lambda/#:~:text=JVM%20configuration%20changes-,In%20the%20Java%2017%20runtime%2C%20the%20JVM%20flag%20for%20tiered%20compilation,.,-This%20is%20helpful

CarlyG55 (approved):

PR #5269, opened 2024-09-24
BAU: Add lambda function pruner for orchestration lambdas

CarlyG55:
This a function to remove old lambda versions from aws. Login with sso in the command line, then run this script with the number of versions you'd like to keep as the first argument, eg `bash scripts/lambda-function-pruner.sh 5`. The minimum number of versions to keep is 3.

## Checklist


ethanmills (approved):

andrew-moores:
@CarlyG55 do we still need this?  The auth deployments do the pruning for us now so we don't have lots of old versions of the lambdas.

PR #5270, opened 2024-09-24
AUT-3602: Emit reauth_failed audit event in VerifyCodeHandler

alhcomer:
## What

- Emits the AUTH_REAUTH_FAILED audit event from the VerifyCodeHandler when a user has exceeded the max number of authentication attempts for any count type.
- Also verified in tests that there are no interactions with the authentication attempts service when it is not turned on.

## How to review

1. Code Review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -514,12 +514,17 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
 
     private String getInternalCommonSubjectIdentifier(
             UserProfile userProfile, ClientRegistry client) {
-        var internalCommonSubjectIdentifier =
-                ClientSubjectHelper.getSubject(
-                        userProfile,
-                        client,
-                        authenticationService,
-                        configurationService.getInternalSectorUri());
-        return internalCommonSubjectIdentifier.getValue();
+        try {
+            var internalCommonSubjectIdentifier =
+                    ClientSubjectHelper.getSubject(
+                            userProfile,
+                            client,
+                            authenticationService,
+                            configurationService.getInternalSectorUri());
+            return internalCommonSubjectIdentifier.getValue();
+        } catch (RuntimeException e) {
+            LOG.info("Failed to derive Internal Common Subject Identifier. Defaulting to UNKNOWN.");
+            return AuditService.UNKNOWN;
+        }
     }
 }
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -173,13 +173,14 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
                     userContext.getClient().orElseThrow(() -> new ClientNotFoundException(session));
 
             Optional<UserProfile> userProfileMaybe = userContext.getUserProfile();
+            UserProfile userProfile = userProfileMaybe.orElse(null);
 
-            if (userProfileMaybe.isEmpty() && journeyType == JourneyType.REAUTHENTICATION) {
+            if (userProfile == null && journeyType == JourneyType.REAUTHENTICATION) {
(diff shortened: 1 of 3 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -263,7 +270,17 @@ private boolean checkErrorCountsForReauth(
                     ReauthAuthenticationAttemptsHelper.countTypesWhereUserIsBlockedForReauth(
                             countsByJourney, configurationService);
 
+            ClientRegistry client = userContext.getClient().orElse(null);
```
andrew-moores, line 275: What advantage does using orElse give if the client in the userContext is null?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -263,7 +270,17 @@ private boolean checkErrorCountsForReauth(
...
             if (!countTypesWhereBlocked.isEmpty()) {
+                String rpPairwiseId =
+                        getInternalCommonSubjectIdentifier(userProfileMaybe.get(), client);
```
andrew-moores: If this is called with client having a value of null then you will get a NullPointerException.  If we know client is null then there is no point making this call as we are in an error state.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -472,4 +489,15 @@ private JourneyType getJourneyType(
         }
         return journeyType;
     }
+
+    private String getInternalCommonSubjectIdentifier(
+            UserProfile userProfile, ClientRegistry client) {
+        var internalCommonSubjectIdentifier =
+                ClientSubjectHelper.getSubject(
```
andrew-moores: As mentioned above there is potential for escaped exceptions here.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -217,7 +218,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                     session,
                     codeRequest,
                     userContext,
-                    userProfileMaybe.isPresent() ? userProfileMaybe.get().getSubjectID() : null,
+                    userProfile != null ? userProfile.getSubjectID() : null,
```
andrew-moores, line 221: If the journey is reauth can we get to this point with a null subjectId?  Should we add this check to the guard at 161?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -197,7 +198,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                         codeRequest,
                         journeyType,
                         notificationType,
-                        userProfileMaybe,
+                        userProfile,
```
andrew-moores, line 201: We only need the subjectId in the `handleInvalidVerificationCode` method.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -330,7 +330,7 @@ private void processSuccessfulCodeRequest(
         int loginFailureCount =
                 codeStorageService.getIncorrectMfaCodeAttemptsCount(session.getEmailAddress());
         var clientSession = userContext.getClientSession();
-        var clientId = userContext.getClient().get().getClientID();
+        var clientId = userContext.getClient().map(ClientRegistry::getClientID).orElse(null);
```
andrew-moores, line 333: What advantage does this give over simply calling getClient when the client is null?  Similar comments as previously about proceeding when the client is null.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
alhcomer: Rather than returning an optional we get null

andrew-moores (dismissed):

BeckaL (approved):

BeckaL (approved):

PR #5272, opened 2024-09-24
AUT-3751: Align check for max number of OTPs sent for sign-in and re-…

andrew-moores:
…auth journeys.

Fix a defect that wasn't allowing the User to have 6 attempts to enter a valid security code for a reauth journey.  Set the default value for "max retries" to 6 which reflects the business requirements better.  Made sure that the lambdas had guard conditions to check if the User has exceeded the maximum number of retries before processing the request and after the request has been processed in case a failure now takes the User over the limit.  The rule is that for lambdas where the User is requesting something they will get maximum of 5 requests, i.e for requesting a password reset, and for lambdas where a value is presented for validation they can present a maximum of 6 incorrect values.

## What

Standardised on using a single value from the ConfigurationService to check maximum number allowed of times a security code can be requested (see the password reset request journey).  Tidied up some of the naming to clarify how the code is meeting the requirements.

## How to review

1. Code Review
2. Test the re-auth journeys manually or using the acceptance tests.
3. Test the password reset journey manually and confirm the correct number of emails are received.

## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
@@ -221,7 +221,7 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
                         auditContext);
 
                 if (userHasExceededAllowedAttemptsForReauthenticationJourney(
-                        journeyType, subjectId, auditContext, userProfile, client)) {
+                        journeyType, subjectId, auditContext, maybeRpPairwiseId)) {
                     return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1057);
                 }
                 return generateApiGatewayProxyErrorResponse(400, errorResponse.get());
@@ -256,11 +256,10 @@ public class VerifyCodeHandler extends BaseFrontendHandler<VerifyCodeRequest>
             JourneyType journeyType,
             String subjectId,
             AuditContext auditContext,
-            UserProfile userProfile,
-            ClientRegistry client) {
+            Optional<String> maybeRpPairwiseId) {
         return journeyType == JourneyType.REAUTHENTICATION
                 && checkReauthErrorCountsAndEmitReauthFailedAuditEvent(
-                        journeyType, subjectId, auditContext, userProfile, client);
+                        journeyType, subjectId, auditContext, maybeRpPairwiseId);
     }
 
     private Optional<String> getCode(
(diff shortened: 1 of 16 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
```diff
@@ -197,6 +197,11 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                         errorResponse.get(),
                         session,
                         auditContext);
+
+                if (journeyType == JourneyType.REAUTHENTICATION
+                        && checkErrorCountsForReauth(journeyType, userProfileMaybe)) {
+                    return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1057);
```
BeckaL, line 203: A note that we should also emit the reauth failed audit event here (and in the other handler). I think it might be easier to come back to this once Aidan's PRs are in

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -252,7 +253,15 @@ private APIGatewayProxyResponseEvent verifyCode(
         sessionService.storeOrUpdateSession(session);
 
         return errorResponseMaybe
-                .map(response -> generateApiGatewayProxyErrorResponse(400, response))
+                .map(
+                        response -> {
```
BeckaL, line 257: So, if possible and if it doesn't end up being a rabbit hole, it'd be good to generate this response where the other error responses are generated, so that we don't end up with this specific reauth logic here. Same with where the counts get incremented. 

This should hopefully mean that if we get to refactoring this stuff to put things in more sensible places, all the count and error logic is in one place.

If it ends up being a rabbit hole, happy to approve as is on the basis that we're going to think about how to refactor the mfa stuff.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
BeckaL: Also same comment as on the VerifyMfaCodeHandler - would be nice to see if we can get this logic in the same place as the other error codes, while acknowledging that in general we'd like to do a refactor to get this stuff done in less surprising places!

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
andrew-moores: Let's defer this until the refactor.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandler.java
andrew-moores: Let's defer anymore change on this until the refactor.

alhcomer (approved):

BeckaL (dismissed):
All the code changes look sensible to me, some nice refactorings / clarifications of existing code.

BeckaL (dismissed):

alhcomer (approved):

PR #5280, opened 2024-09-25
Ato 981 set is new account in new session

Ryan-Andrews99:
## Checklist


shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
```diff
@@ -109,6 +110,30 @@ public Optional<AuthSessionItem> getSession(String sessionId) {
...
+
+        return sessionId
+                .flatMap(InputSanitiser::sanitiseBase64)
+                .flatMap(
+                        id -> {
+                            try {
+                                return getSession(id);
+                            } catch (Exception e) {
+                                LOG.error("Failed to retrieve session for id: {}", id);
+                                throw e;
```
billrdunn, line 132: Maybe define and throw a dedicated exception (eg. `AuthSessionServiceException`) instead?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
```diff
@@ -107,6 +108,13 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
 
         attachSessionIdToLogs(userContext.getSession());
 
+        var optionalAuthSessionItem =
+                authSessionService.getSessionFromRequestHeaders(input.getHeaders());
+        if (optionalAuthSessionItem.isEmpty()) {
+            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
```
billrdunn, line 114: Is this defo a 400? What if the request is valid but the session can't be found for another reason? 
Maybe return 400 for missing Session-Id and 500 for session not found?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/LoginHandler.java
```diff
@@ -156,6 +157,15 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
             LoginRequest request,
             UserContext userContext) {
 
+        var optionalAuthSession =
+                authSessionService.getSessionFromRequestHeaders(input.getHeaders());
+
+        if (optionalAuthSession.isEmpty()) {
```
billrdunn, line 163: Similar comment to with the other handler - should this be 400 when Session-Id is missing, 500 otherwise?

auth-external-api/src/main/java/uk/gov/di/authentication/external/lambda/UserInfoHandler.java
```diff
@@ -130,6 +131,17 @@ public APIGatewayProxyResponseEvent userInfoRequestHandler(APIGatewayProxyReques
...
         }
 
+        AuthSessionItem authSession;
+
+        Optional<AuthSessionItem> optionalAuthSession =
+                authSessionService.getSessionFromRequestHeaders(input.getHeaders());
+
+        if (optionalAuthSession.isEmpty()) {
+            LOG.warn("Session cannot be found");
+            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
```
billrdunn, line 141: Same comment as previous. Also any reason we are logging here but not in the other handlers?

Ryan-Andrews99: Yeah not sure why this logging is here, will remove because I realise it could be a bit misleading if the sessionId header is not present

shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
Ryan-Andrews99: Nice, didn't know this exception existed, have implemented this!

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
billrdunn: Good points, I agree with you now 😄

billrdunn (approved):

billrdunn:
Approved on behalf of Orchestration, needs a review from Authentication too 🙂

alhcomer (approved):
LGTM

PR #5330, opened 2024-10-04
Ato 981 set is new account from auth user info response

Ryan-Andrews99:
## What:
- This sets isNewAccount in the authentication callback handler using the UserInfo response from auth, regardless of whether this is an auth or identity journey. This is needed because when a user is returned from the identity journey (IPV callback handler) we need to emit metrics on this session field, and we can no longer rely on auth setting it.
- 
## How to review
Deploy to a dev environment 
   - Run through a journey, check all is fine 
   
## Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -80,7 +80,6 @@ import static uk.gov.di.authentication.oidc.domain.OrchestrationAuditableEvent.A
 import static uk.gov.di.orchestration.shared.conditions.DocAppUserHelper.isDocCheckingAppUserWithSubjectId;
 import static uk.gov.di.orchestration.shared.conditions.IdentityHelper.identityRequired;
 import static uk.gov.di.orchestration.shared.domain.RequestHeaders.SESSION_ID_HEADER;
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -354,6 +353,8 @@ public APIGatewayProxyResponseEvent handleRequest(
 
                 Boolean newAccount = userInfo.getBooleanClaim("new_account");
                 AccountState accountState = newAccount ? AccountState.NEW : AccountState.EXISTING;
+
+                sessionService.storeOrUpdateSession(userSession.setNewAccount(accountState));
```
isaac-GDS, line 357: Does `userSession.setNewAccount(accountState)` return back `accountState` . I would have though that we would want `sessionService.storeOrUpdateSession(accountState)`

billrdunn (approved):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
Ryan-Andrews99: SetNewAccount modifies the session object and returns it on one go, the `sessionService.storeOrUpdateSession` takes a session object as it's argument. The service is separate from the individual session so we need to pass the updated session object here and can't just pass the account state

isaac-GDS (approved):

PR #5339, opened 2024-10-07
BAU: refactor account management request objects to records

alhcomer:
## What

BAU: refactor account management request objects to records

## How to review

1. Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/AuthenticateRequest.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/AuthenticateRequest.java
@@ -3,24 +3,5 @@ package uk.gov.di.accountmanagement.entity;
 import com.google.gson.annotations.Expose;
 import uk.gov.di.authentication.shared.validation.Required;
 
-public class AuthenticateRequest {
-
-    @Expose @Required private String email;
-
-    @Expose @Required private String password;
-
-    public AuthenticateRequest() {}
-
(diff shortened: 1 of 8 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/entity/UpdateEmailRequest.java
```diff
@@ -4,38 +4,17 @@
...
+        @Expose @SerializedName("existingEmailAddress") @Required String existingEmailAddress,
+        @Expose @SerializedName("replacementEmailAddress") @Required String replacementEmailAddress,
+        @Expose @Required String otp) {
 
     public UpdateEmailRequest(
             String existingEmailAddress, String replacementEmailAddress, String otp) {
-        this.existingEmailAddress = existingEmailAddress.toLowerCase();
-        this.replacementEmailAddress = replacementEmailAddress.toLowerCase();
+        this.existingEmailAddress =
+                existingEmailAddress != null ? existingEmailAddress.toLowerCase() : null;
```
BeckaL, line 15: Is there any particular reason we added this null check in here? (Thoughts being I'd rather not have assumptions that things could be null, I'd rather use optionals in all nullable cases)

BeckaL (commented):
Looks generally good to me! lovely line count removal too :-D One question / comment in there, but otherwise LGTM. I think because the account management api isn't exercised in acceptance tests, we should probably do a cautious test of some account management stuff in staging before we let it go to prod, but I think should all be fine.

PR #5326, opened 2024-10-03
ATO-1076: Create / update Orch session

billrdunn:
## What

When there is no existing session, create a new Orch session and add to the Orch session Dynamo table.
When there is an existing session, update Orch session with a new session ID.

Review commit by commit

## Manual testing
Tested in dev:
- On first login, a new Orch session is saved to dynamo with same sessionId as in shared session
- On return, sessionId is updated in dynamo
- Doc app testing in progress

PR #5343, opened 2024-10-07
ATO-922: set email in orch session

GHSwallow:
## What
- Adds the email property to the orch session (needed for client registry lookups)
- Sets the email property in the Authentication Callback Handler to the email in the user info response

## How to review
1. Code Review


billrdunn (commented):
Looks good although there won't be an Orch session to add to until this is deployed: https://github.com/govuk-one-login/authentication-api/pull/5326

and you'll need to add policies to AuthenticationCallbackFunction to read and write from Orch session table.

PR #5356, opened 2024-10-09
Aut 2257: Fix index out of bounds exception

BeckaL:
## What

We've very occassionally been hitting errors `java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0` as a result of a .get(0) in our dynamo service, when a user profile is not found for a subject id.

This fixes this code by not assuming that the list we're calling .get(0) on has any items. This will mean that we get a more descriptive error (no user profile found) rather than this exception. It won't stop us throwing exceptions in this case.

Note that I attempted to reproduce this in tests but didn't manage to since I think the situation that produces this is very edge case-y (backed up by the very occassionnal nature of these errors).

## How to review

1. Code Review


andrew-moores:
If we now deal with this error in a bau fashion have we lost information about a user that might need a data fix to use our service?

BeckaL:
> If we now deal with this error in a bau fashion have we lost information about a user that might need a data fix to use our service?

I don't think so. We still error and that will flag an alert in the same way it does now. The error is more descriptive now though, and less obviously a programming error

mattvot (approved):

PR #5379, opened 2024-10-14
ATO-922: add logging to verify if session email is null in DocApp

GHSwallow:
## What
The email is added to the session in `CheckUserExistsHandler` in authentication, so in a doc app journey, this should be null. If it is, we can remove the use of email in the doc app journey.

## How to review
1. Code Review


billrdunn (approved):
What's the plan for monitoring the logged value? At what point can you confidently say if it's always null or not?

PR #5380, opened 2024-10-14
AUT-3800: Add values required by acceptance tests into SSM

whi-tw:
## What

Create SSM parameters with values required by the acceptance tests.

This allows us to provide the acceptance tests with the actual configured values,
rather than manually duplicating the configured values in each environment

## How to review

- Code review
- test deploy in authdev / sandpit environments and ensure that the right values are created.


BeckaL (approved):
Have deployed and seen the right variables be created, code looks good to me as far as I can tell

BeckaL (approved):

PR #5384, opened 2024-10-14
BAU: update OrchSessionItem item to always require a sessionId

GHSwallow:
## What
sessionId is the primary key of the table, so any OrchSessionItem must have a sessionId associated with it. By removing the OrchSessionItem constructor with no arguments, we insist that any OrchSessionItem will have a sessionId.

## How to review
1. Code Review


billrdunn (approved):
LGTM
I'd deploy to dev and run through journeys with and without a session, then when deploying leave in integration for a while before releasing to prod. Can't be too careful :)

PR #5391, opened 2024-10-15
ATO-1049: Adjust alerting when AIS fails open

billrdunn:
## What

A pager duty (P1) alarm shouldn't be triggered in the case that communication with Account Intervention Service fails, and the "abort on error" flag is false (ie. assuming no intervention).

This PR defines a custom log level (AIS_FAIL_OPEN) which a metric is set up on so we can have finer control over AIS alarms, and only send alerts to slack.

An alarm is also renamed to not include "P1" in the case that AIS fails open.

## Testing

AIS is only set up in staging and higher envs, so this needs to be deployed there to test. 

The plan is to force calls to AIS to fail by using an incorrect URI for AIS and check that the alarms are triggered as intended. I don't know how this can be fully tested as forcing AIS errors in prod isn't an option - suggestions welcome.

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
@@ -1,5 +1,6 @@
 package uk.gov.di.orchestration.shared.services;
 
+import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.orchestration.audit.AuditContext;
@@ -27,6 +28,7 @@ import static uk.gov.di.orchestration.shared.services.AuditService.MetadataPair.
 public class AccountInterventionService {
 
     private static final Logger LOG = LogManager.getLogger(AccountInterventionService.class);
+    private static final Level AIS_FAIL_OPEN = Level.forName("AIS_FAIL_OPEN", 200);
     private final HttpClient httpClient;
     private final URI accountInterventionServiceURI;
     private final AuditService auditService;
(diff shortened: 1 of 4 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -141,12 +143,16 @@ private AccountIntervention handleException(Exception e) {
...
-                    "Problem communicating with Account Intervention Service", e);
+                    "Problem communicating with Account Intervention Service. Aborting user journey",
+                    e);
         }
-        LOG.error(
-                "Problem communicating with Account Intervention Service. Assuming no intervention. ",
-                e);
+        LOG.atLevel(AIS_FAIL_OPEN)
+                .log(
+                        "Problem communicating with Account Intervention Service. Assuming no intervention");
```
CarlyG55, line 155: I'm not sure how I feel about this - it feels a bit unusual... could we create some kind of metric and alert on that?

CarlyG55: In fact is there a need for this? Can we not just log at error level, and then just use the flags you've already got above to decide if it's p1 & goes to pagerduty?

CarlyG55: Would still be nice to use a metric for this though, means we're not just alarming on all cases of error logs in this lambda

billrdunn: > In fact is there a need for this? Can we not just log at error level, and then just use the flags you've already got above to decide if it's p1 & goes to pagerduty?

I don't think this can work - as long as there's logs at level ERROR in the processing identity lambda in prod we will trigger pager duty. It can't distinguish whether it's failing open or aborting on error.

Unless... we use a different sort of pattern matching for the AIS fail open metric, like matching a particular string in the logs. But this feels quite brittle (easy to accidentally change the log message and break it). The solution here seems more resilient and neater (to me anyway 🙂)

billrdunn:
Closed as superseded by https://github.com/govuk-one-login/authentication-api/pull/5391

PR #5395, opened 2024-10-16
ATO-1049: Improvements to Account Intervention Service alerting

billrdunn:
## What

- A pager duty (P1) alarm shouldn't be triggered in the case that communication with Account Intervention Service fails open (ie. assumes no intervention). To ensure this, a dimension has been added to the AISException metric, which is used to determine the type of alarm

- `env-P1-account-interventions-alarm` alarm is renamed to not include "P1" in the case that AIS fails open. This alarm shouldn't be triggered, as its metric monitors a deprecated log group (orchestration_redirect_lambda_log_group).

## What's not included

- Setting up a pager duty alarm for the P1 AIS error
- Adding alarms for the other lambdas which call AIS (IpvCallback and ProcessingIdentity). This isn't too bad as AuthenticationCallback calls AIS first in all journeys.

These will form part of a wider piece of work to audit and improve the handling of alarms in Orchestration.

## Testing
Forced the alarms in dev and functioning as intended:

image
image

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
@@ -141,11 +141,13 @@ public class AccountInterventionService {
                 configurationService.getAccountInterventionsErrorMetricName(),
                 Map.of("Environment", configurationService.getEnvironment()));
         if (accountInterventionsActionEnabled && acountInterventionsAbortOnError) {
-            throw new AccountInterventionException(
-                    "Problem communicating with Account Intervention Service", e);
+            String errorMessage =
+                    "Problem communicating with Account Intervention Service. Aborting user journey.";
+            LOG.error(errorMessage, e);
+            throw new AccountInterventionException(errorMessage, e);
         }
-        LOG.error(
-                "Problem communicating with Account Intervention Service. Assuming no intervention. ",
+        LOG.warn(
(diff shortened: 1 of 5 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
```diff
@@ -139,13 +139,19 @@ private static AuditContext addStatusMetadata(
...
+                        "Environment",
+                        configurationService.getEnvironment(),
+                        "AbortOnError",
+                        String.valueOf(acountInterventionsAbortOnError)));
         if (accountInterventionsActionEnabled && acountInterventionsAbortOnError) {
-            throw new AccountInterventionException(
-                    "Problem communicating with Account Intervention Service", e);
+            String errorMessage =
+                    "Problem communicating with Account Intervention Service. Aborting user journey.";
+            LOG.error(errorMessage, e);
```
CarlyG55, line 150: This does mean that the other alarm in this lambda is also probably going to go off at the same time if we P1 - but that's not the end of the world, and I think we can look into tidying that up when we pick up further alarms work

CarlyG55 (approved):

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AccountInterventionService.java
billrdunn: Yeah it will, better too many alarms that too few 😅 . Another thing to look at in the general alarms audit.

PR #5399, opened 2024-10-16
BAU: refactor verify mfa code handler

alhcomer:
## What

- Moved the error handling and audit event emission out of the processCodeSession method. This means that error handling (predominantly) happens in one place, followed by processing a successful code request, followed by audit event emission. This gives the processSuccessfulCodeSession (successor to processCodeSession) method a single purpose and makes the logic easier to follow..
- Stopped passing unnecessary audit context through to verify code method.
- Replaced switch with enhanced switch in the MfaCodeProcessorFactory
- Throw and catch exception when the client is not present in the VerifyMfaCodeHandler
- Delete unused empty constructor in the VerifyMfaCodeRequest

## How to review

1. Code Review
2. Deploy to dev and run through auth app journey successfully

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -287,6 +287,8 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
             AuditContext auditContext,
             Optional<String> maybeRpPairwiseId) {
 
+        var session = userContext.getSession();
+
         var mfaCodeProcessor =
                 mfaCodeProcessorFactory
                         .getMfaCodeProcessor(
@@ -298,14 +300,6 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
             return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1002);
         }
 
-        var errorResponseMaybe = mfaCodeProcessor.validateCode();
-
-        if (errorResponseMaybe.filter(ErrorResponse.ERROR_1041::equals).isPresent()) {
-            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1041);
-        }
-
-        var session = userContext.getSession();
-
         if (JourneyType.PASSWORD_RESET_MFA.equals(codeRequest.getJourneyType())) {
             SessionHelper.updateSessionWithSubject(
                     userContext,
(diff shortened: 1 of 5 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -171,7 +172,18 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
         UserProfile userProfile = userProfileMaybe.orElse(null);
-        var maybeRpPairwiseId = getRpPairwiseId(userProfile, userContext.getClient());
+        ClientRegistry client;
+        try {
+            client =
+                    userContext
+                            .getClient()
+                            .orElseThrow(
+                                    () -> new ClientNotFoundException(userContext.getSession()));
+        } catch (ClientNotFoundException e) {
```
BeckaL, line 182: Is there any point throwing an exception here to immediately catch it? (In general, I'm kind of anti throwing exceptions although we've got a way to go in our codebase to get rid of them)

An alternative that I think sonar would be happy with:

```
        if (userContext.getClient().isEmpty()) {
            LOG.warn("Client not found");
            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1015);
        }
        var client = userContext.getClient().get();
```

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -315,15 +309,58 @@ private APIGatewayProxyResponseEvent verifyCode(
...
-                input,
-                userContext,
-                subjectId,
-                codeRequest,
-                mfaCodeProcessor,
-                maybeRpPairwiseId);
+        var errorResponseMaybe = mfaCodeProcessor.validateCode();
+
+        if (errorResponseMaybe.isPresent()) {
+            var errorResponse = errorResponseMaybe.get();
```
BeckaL, line 326: I wonder here about introducing a "process error response" method

BeckaL (commented):
Nice set of refactorings. A couple of comments / suggestions.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
alhcomer: I had a go at that, but I found that with ERROR_1041 returning an APIGatewayProxyResponseEvent while the other errors only cause interactions with services it looks clunkier and harder to read. I might be missing something though so would be happy to hear if you have a better idea on how to do it

BeckaL: Fair enough, I think my inclination would be to try a method that returns an optional gateway response, but I haven't seen how clunky that is :-) Happy to leave as is

PR #5410, opened 2024-10-18
BAU: Add `Environment` dimension to all metrics by default

mrwilson:
We are currently configuring this at every call site.


ethanmills (approved):
Are there any metrics that don't currently have this dimension?

mrwilson:
Closing because it was far behind enough that it's easier to just do it again.

PR #5416, opened 2024-10-21
ATO-922: add rp pairwise id to orch session

GHSwallow:
## What
Adds rpPairwiseId to the Orch session. This field is already being sent in the auth userinfo response. This is needed because in `AuthCodeHandler`, there is a function `getRpPairwiseId`, which uses the userProfile table to get the `rpPairwiseId`. We instead want to get it from the orch session here, so we can remove orch's use of the userProfileTable

## How to review


GHSwallow:
Tested on dev. Journey is unchanged as expected, and the test logs `is rpPairwiseId attached to orch session` are returning `true` as expected.

PR #5421, opened 2024-10-22
ATO-1104: Fix broken Auth session expiry

billrdunn:
## Issue

Auth DynamoDB sessions aren't expiring and being deleted. This is because the `ttl` `attribute_name`, defined in the `auth_session_table` terraform resource, doesn't match any AuthSessionItem attribute.

## Fix

1. Introduce a new "TimeToLive" attribute which matches the existing terraform. Set this value on session creation. Set the exisiting "ttl" attribute to 0.
2. Modify the `getSession` logic in AuthSessionService to filter on the exisiting "ttl" attribute if it's non-zero, otherwise filter on the new "TimeToLive" attribute.
3. Wait some time (> session expiry) so that all live sessions have the new "TimeToLive" attribute.
4. Deprecate "ttl" in code.
5. Remove all sessions in DynamoDB that have a value set for "ttl" - will likely need to write a script for this.

This PR covers 1 and 2 above. This will be carefully tested in lower envs before releasing to prod.

## How to review
1. Review plan for fix above
2. Check code matches plan steps 1 and 2

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/services/AuthSessionServiceIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/services/AuthSessionServiceIntegrationTest.java
@@ -3,18 +3,24 @@ package uk.gov.di.authentication.services;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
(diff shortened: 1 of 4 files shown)


mattvot (approved):

Ryan-Andrews99 (approved):

billrdunn:
Closing as testing in dev showed session expiry for users mid-way through a journey when this change is deployed

PR #5426, opened 2024-10-22
ATO-922: verify internalCommonSubjectIdentifier and internalPairwiseI…

GHSwallow:
## What
tl;dr: Temporary logs to verify that these two have the same value.

In UserInfoService, the definition of internalPairwiseId is the **subject** of the UserInfo request. Then in LogoutService, createAuditUser uses getInternalCommonSubjectIdentifier(), which is set in updateSessionWithSubject() with the value returned from getSubjectWithSectorIdentifier() which has exactly the same definition as internalPairwiseId! This logging is to verify that these two are in fact the same value. If they are, then we don't need to get the subject from the userInfo response, and can instead just get `internalPairwiseId` from `userInfo.getSubject()` in AuthenticaionCallbackHandler, add it to the orchSession and call that when we want i`nternalCommonSubjectIdentifier`. 

## How to review
1. Code Review. No impact, just test logs.


GHSwallow:
Verified in dev that internalCommonSubjectIdentifier is the same thing as internalPairwiseId

PR #5427, opened 2024-10-22
ATO-922: set internal subject id in orch session

GHSwallow:
## What
Adds internalSubjectId to the orch session. This is needed in LogoutService, which currently takes the subjectId and builds the internalSubjectId again (named as subjectWithSectorIdentifier). By having this in the orch session, we can skip having to rebuild it, and also consolidate the name of this value to hopefully be named the same thing (internalSubjectId) everywhere.

Code does not affect any journey's. It just sets it in the orch session with some logging to verify the value is defined.

## How to review
1. Code Review


billrdunn (approved):
LGTM
In future can you add a few sentences to each commit message just to explain why each change is required? [Here](https://www.annashipman.co.uk/jfdi/good-pull-requests.html)'s a nice guide for general good practice. There might be some redundancy with the PR comment but that's fine :)

GHSwallow:
Because internalCommonSubjectIdentifier is the same thing as internalPairwiseId, we don't need this PR, as internalCommonSubjectIdentifier is being added already here https://github.com/govuk-one-login/authentication-api/pull/5396

PR #5429, opened 2024-10-23
ATO-1088: Set authenticated flag in all cases a user is handed back from Auth

Ryan-Andrews99:
## What
- We now set the user as authenticated in the session when they are handed back from Auth regardless of whether they are on an identity or auth-only journey. 
- The practical effect of this on a User's journey is that now if they abandon an identity journey from a fresh session, they will pass through Auth as logged in when they return if their session has not expired. The current behaviour is that they would be prompted to log in before being passed to back IPV.

- There is a second part of this change which involves removing the redundant setting this field in IPV callback handler in this PR https://github.com/govuk-one-login/authentication-api/pull/5430, however this change is the part that will have a user facing change. 

## How to review
- Deploy to a dev environment
    - Check that silent login is not affected by this 
    - There is not much testing that can be done in dev as IPV is not hooked up
    - Check with IPV before merging

- Before deploying to Prod:
    - Pause promotion from staging  
    - Run through an identity journey and abandon it when handed over to IPV 
    - Return to the stub and request a new identity journey 
    - Observe skipping auth
    - Check that the handover back from IPV -> Orch works fine
## Checklist

## Related PRs
- PR to remove redundant setting: https://github.com/govuk-one-login/authentication-api/pull/5430

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
@@ -152,7 +152,8 @@ public class AuthenticationCallbackHandlerIntegrationTest extends ApiGatewayHand
     }
 
     @Test
-    void shouldStoreUserInfoAndRedirectToRpWhenSuccessfullyProcessedCallbackResponse() {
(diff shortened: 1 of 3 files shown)


CarlyG55:
Could you fix the code smell please? Happy to approve once that is done

CarlyG55 (approved):

PR #5434, opened 2024-10-25
ATO-1118: add phone number to all auth userinfo responses

GHSwallow:
## What
Orch needs phone number to add to the orch session, so eg in IPVCallbackHandler, the phone number can be used without using the UserProfile table. This is an internal endpoint, so although phone number is always sent back, the RP to Orch UserInfo response will still use the set of requested claims to determine what to send to the RP.

## How to review
1. Code Review
2. Tested in dev


GHSwallow:
Taking a slightly different approach: instead of adding to all UserInfo responses in auth, we specify that we always want phone number for only identity journeys on Orch's side https://github.com/govuk-one-login/authentication-api/pull/5441

PR #5441, opened 2024-10-28
ATO-1118: add phone number to auth UserInfo response for all identity journeys

GHSwallow:
## What
Orch needs access to a user's phone number only within identity journeys, so following the pattern of email and local_account_id, we add phone number as a claim for all identity journeys.

I've decided to follow this pattern, instead of adding phone number to all UserInfo requests, as we explicitly only need it in identity journeys.

This PR also adds a low-level trust journey test to verify that specifically for high-trust identity journeys, we add in a number of scopes, which now includes phone number.

## How to review
1. Code Review

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -883,6 +883,7 @@ public class AuthorisationHandler
             // Email required for ID journeys for use in Face-to-Face flows
             claimsSet.add(AuthUserInfoClaims.EMAIL);
             claimsSet.add(AuthUserInfoClaims.EMAIL_VERIFIED);
+            claimsSet.add(AuthUserInfoClaims.PHONE_NUMBER);
         }
         if (amScopePresent) {
             LOG.info("am scope is present. Adding the public_subject_id claim");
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -883,6 +883,7 @@ private Optional<OIDCClaimsRequest> constructAdditionalAuthenticationClaims(
             // Email required for ID journeys for use in Face-to-Face flows
             claimsSet.add(AuthUserInfoClaims.EMAIL);
             claimsSet.add(AuthUserInfoClaims.EMAIL_VERIFIED);
+            claimsSet.add(AuthUserInfoClaims.PHONE_NUMBER);
```
Ryan-Andrews99, line 886: Small thing but we're adding the `PHONE_NUMBER` claim twice in the case `if (phoneScopePresent)`. Should we remove the second instance of  `claimsSet.add(AuthUserInfoClaims.PHONE_NUMBER)` given it's now always present

GHSwallow: This change only adds the phone number claim if identityRequired is true, so we still need the `phoneScopePresent` check to add phone number if requested (in the same way the email claim is added twice in the case of an identity journey which also requires identity)

Ryan-Andrews99: 🤦 - ah yeah I can see that. Didn't expand the git diff fully

Ryan-Andrews99 (approved):

PR #5451, opened 2024-10-29
ATO-1129: add email and verified MFA claims to all userinfo requests

GHSwallow:
## What
We need these claims in every UserInfo response, so we always adds these claims.

Currently, we always add these within the auth UserInfo response, but (taken from the ticket description),
> _For authentication to be truly separate from orch and for their interaction to be a by-the-book oidc flow, auth should not be at all aware of what orch is using the data for - it should just provide whatever data the consuming service asks for (and is allowed). Currently, authentication code has been changed to send back email and verified mfa method. Instead, we should change orch code to always request these claims. Else, the two services remain coupled._

This shouldn't make any functional difference, as currently, both of these are always sent back (auth code has been changed such that this is the case). Instead of migrating all at once, I've updated orch's code and added some logging in authentication to verify that the request always contains these claims. Once this is the case, I can safely add back in authentications "if claim is present" logic for email and verified MFA type.

## How to review
1. Code Review

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -876,12 +876,14 @@ public class AuthorisationHandler
         var emailScopePresent = requestedScopesContain(OIDCScopeValue.EMAIL, authenticationRequest);
 
         var claimsSet = new HashSet<AuthUserInfoClaims>();
+        claimsSet.add(AuthUserInfoClaims.EMAIL);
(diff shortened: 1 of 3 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -901,7 +902,6 @@ private Optional<OIDCClaimsRequest> constructAdditionalAuthenticationClaims(
         }
         if (emailScopePresent) {
             LOG.info("email scope is present. Adding the email and email_verified claim");
```
billrdunn, line 904: This'll need updating

billrdunn (changes requested):
Could you add a test which ensures the claims are always present?

billrdunn (approved):

PR #5398, opened 2024-10-16
ATO-983: Copy InternalCommonSubjectIdentifier From UserInfo Response Subject To OrchSession (Orch ICSID 1/3)

BenjaminWCO:
## What

Copy Internal Common Subject Identifier From User Info Subject To Orch Session

## How to review

1. Code Review

## Checklist

- [x] Impact on orch and auth mutual dependencies has been checked.

Next PR: https://github.com/govuk-one-login/authentication-api/pull/5452

PR #5452, opened 2024-10-29
ATO-983: Get InternalCommonSubjectIdentifier From OrchSession (Orch ICSID 2/3)

ghost:
## What

Use Internal Common Subject Identifier From Orch Session

## How to review

1. Code Review

## Checklist

Previous PR: https://github.com/govuk-one-login/authentication-api/pull/5398
Next PR:  https://github.com/govuk-one-login/authentication-api/pull/5453

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IdentityProgressFrontendHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IdentityProgressFrontendHandler.java
@@ -14,6 +14,7 @@ import uk.gov.di.authentication.ipv.entity.IdentityProgressResponse;
 import uk.gov.di.authentication.ipv.entity.IdentityProgressStatus;
 import uk.gov.di.orchestration.audit.TxmaAuditUser;
 import uk.gov.di.orchestration.shared.entity.ErrorResponse;
+import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
 import uk.gov.di.orchestration.shared.lambda.BaseOrchestrationFrontendHandler;
 import uk.gov.di.orchestration.shared.serialization.Json;
 import uk.gov.di.orchestration.shared.services.AuditService;
@@ -22,6 +23,7 @@ import uk.gov.di.orchestration.shared.services.ClientSessionService;
 import uk.gov.di.orchestration.shared.services.CloudwatchMetricsService;
 import uk.gov.di.orchestration.shared.services.ConfigurationService;
 import uk.gov.di.orchestration.shared.services.DynamoIdentityService;
+import uk.gov.di.orchestration.shared.services.OrchSessionService;
 import uk.gov.di.orchestration.shared.services.SessionService;
 import uk.gov.di.orchestration.shared.state.OrchestrationUserSession;
 
@@ -39,6 +41,7 @@ public class IdentityProgressFrontendHandler extends BaseOrchestrationFrontendHa
     private final AuditService auditService;
     private final CloudwatchMetricsService cloudwatchMetricsService;
     private final AuthenticationUserInfoStorageService userInfoStorageService;
+    private final OrchSessionService orchSessionService;
 
     private static final Logger LOG = LogManager.getLogger(IdentityProgressFrontendHandler.class);
 
@@ -49,6 +52,7 @@ public class IdentityProgressFrontendHandler extends BaseOrchestrationFrontendHa
         this.cloudwatchMetricsService = new CloudwatchMetricsService();
         this.userInfoStorageService =
                 new AuthenticationUserInfoStorageService(configurationService);
+        this.orchSessionService = new OrchSessionService(configurationService);
     }
 
     public IdentityProgressFrontendHandler() {
@@ -61,6 +65,7 @@ public class IdentityProgressFrontendHandler extends BaseOrchestrationFrontendHa
             AuditService auditService,
             CloudwatchMetricsService cloudwatchMetricsService,
             SessionService sessionService,
+            OrchSessionService orchSessionService,
             AuthenticationUserInfoStorageService userInfoStorageService,
             ClientSessionService clientSessionService) {
         super(configurationService, sessionService, clientSessionService);
@@ -68,6 +73,7 @@ public class IdentityProgressFrontendHandler extends BaseOrchestrationFrontendHa
         this.auditService = auditService;
         this.cloudwatchMetricsService = cloudwatchMetricsService;
         this.userInfoStorageService = userInfoStorageService;
+        this.orchSessionService = orchSessionService;
     }
 
     @Override
@@ -83,8 +89,14 @@ public class IdentityProgressFrontendHandler extends BaseOrchestrationFrontendHa
             OrchestrationUserSession userSession) {
         LOG.info("IdentityProgress request received");
         try {
-            var internalCommonSubjectIdentifier =
-                    userSession.getSession().getInternalCommonSubjectIdentifier();
+            var orchSessionOptional = orchSessionService.getSessionFromRequestHeaders(input.getHeaders());
+            if (orchSessionOptional.isEmpty()) {
+                return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
+            }
+
+            var orchSession = orchSessionOptional.get();
+            var internalCommonSubjectIdentifier = orchSession.getInternalCommonSubjectId();
+
 
             AuthenticationRequest authenticationRequest;
             try {
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java
@@ -13,6 +13,7 @@ import uk.gov.di.authentication.ipv.entity.ProcessingIdentityStatus;
 import uk.gov.di.orchestration.audit.AuditContext;
 import uk.gov.di.orchestration.shared.entity.AccountIntervention;
 import uk.gov.di.orchestration.shared.entity.ClientRegistry;
+import uk.gov.di.orchestration.shared.entity.ErrorResponse;
 import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
 import uk.gov.di.orchestration.shared.entity.UserProfile;
 import uk.gov.di.orchestration.shared.helpers.ClientSubjectHelper;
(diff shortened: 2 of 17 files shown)


GHSwallow (commented):
Haven't done a full review yet, but figure these comments are something to get started on:)

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -100,47 +101,57 @@ private APIGatewayProxyResponseEvent generateLogoutResponse(
...
-            clientSessionService.deleteStoredClientSession(clientSessionId);
-        }
-        LOG.info("Deleting Session");
-        sessionService.deleteStoredSession(session.getSessionId());
+    private void destroySessions(
+            Optional<Session> session,
+            Optional<OrchSessionItem> orchSession,
+            Optional<String> clientId,
+            Optional<AccountIntervention> accountIntervention) {
+        session.ifPresent(
```
GHSwallow, line 109: I'm not a big fan of this. Only in `handleLogout` are the session and orchSession optional, not in the other 2, and it means here, we end up with a lot of nesting.

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
```diff
@@ -206,7 +212,7 @@ void successfullyReturnsClientLogoutResponse() {
         verify(backChannelLogoutService)
                 .sendLogoutMessage(
                         argThat(withClientId("client-id")), eq(EMAIL), eq(INTERNAL_SECTOR_URI));
-        verify(cloudwatchMetricsService).incrementLogout(Optional.of(CLIENT_ID));
+        verify(cloudwatchMetricsService).incrementLogout(Optional.of(CLIENT_ID), Optional.empty());
```
GHSwallow, line 215: why add this `Optional.empty()` argument? There's already a method to handle `intervention` being not Optional.empty(). Same for all the cases in this file.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java
```diff
@@ -211,9 +219,20 @@ private APIGatewayProxyResponseEvent performIntervention(
             ClientRegistry client,
             AccountIntervention intervention)
             throws Json.JsonException {
+        var orchSessionOptional =
+                orchSessionService.getSessionFromRequestHeaders(input.getHeaders());
+        if (orchSessionOptional.isEmpty()) {
+            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
+        }
```
GHSwallow, line 226: I'm fairly certain right now, this will break. I tested the orch session permissions for this, and it's trying to get info from a table called orchSessionTable in Auth, not in Orch. This is a fix I need to do for destroy orch session to work properly, but that work blocks this.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/entity/LogoutRequestTest.java
```diff
@@ -65,12 +68,13 @@ class LogoutRequestTest {
             URI.create("https://di-authentication-frontend.london.cloudapps.digital/signed-out");
     private static final URI CLIENT_LOGOUT_URI = URI.create("http://localhost/logout");
     private SignedJWT signedIDToken;
-    private static final Subject SUBJECT = new Subject();
+    private static final Subject INTERNL_COMMON_SUBJECT_ID = new Subject();
```
GHSwallow, line 71: Spelling mistake: `INTERNL_COMMON_SUBJECT_ID` -> `INTERNAL_COMMON_SUBJECT_ID`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/entity/LogoutRequestTest.java
```diff
@@ -65,12 +68,13 @@ class LogoutRequestTest {
...
     private static final String EMAIL = "joe.bloggs@test.com";
     private final ClientRegistry clientRegistry = createClientRegistry();
     private String idTokenHint;
     private String rpPairwiseId;
     private uk.gov.di.orchestration.shared.entity.Session session;
```
GHSwallow, line 76: Not your code, but can this be changed so the type is just Session instead of the full path to the module?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/entity/LogoutRequestTest.java
```diff
@@ -421,8 +460,11 @@ private uk.gov.di.orchestration.shared.entity.Session generateSession() {
                 .addClientSession(CLIENT_SESSION_ID);
     }
 
-    private void generateSessionFromCookie(uk.gov.di.orchestration.shared.entity.Session session) {
+    private void generateSessionFromCookie(
+            uk.gov.di.orchestration.shared.entity.Session session, OrchSessionItem orchSession) {
```
GHSwallow, line 464: This too can be updated once the module is imported.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandlerTest.java
```diff
@@ -254,6 +254,7 @@ void shouldGenerateSuccessfulAuthResponseAndUpliftAsNecessary(
                 ClientSubjectHelper.calculatePairwiseIdentifier(
                         SUBJECT.getValue(), "test.account.gov.uk", SaltHelper.generateNewSalt());
         session.setInternalCommonSubjectIdentifier(expectedCommonSubject);
+        orchSession.setInternalCommonSubjectId(expectedCommonSubject);
```
GHSwallow, line 257: Not your code, but I'm not a huge fan of this name, given we're using it to set in the sessions. Also, it should have `internalCommonSubjectId` in the name. A big grievance with the current state of of authentication-api is the names of the subjects, so I think it's good to address discrepancies as we go.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandlerTest.java
```diff
@@ -94,6 +94,10 @@ class AuthenticationCallbackHandlerTest {
                     .setVerifiedMfaMethodType(MFAMethodType.EMAIL)
                     .setAuthenticated(false)
                     .setCurrentCredentialStrength(null);
+    private static final OrchSessionItem ORCH_SESSION =
```
GHSwallow, line 97: Neither the session nor the client session use screaming snake case. Either update them to use it, or do camelCase for orch session. I think the latter, as that's what other tests do.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/LogoutHandlerTest.java
```diff
@@ -205,28 +215,32 @@ void shouldNotThrowWhenTryingToDeleteClientSessionWhichHasExpired() {
...
     }
 
     private uk.gov.di.orchestration.shared.entity.Session generateSession() {
         return new uk.gov.di.orchestration.shared.entity.Session(SESSION_ID)
                 .addClientSession(CLIENT_SESSION_ID);
     }
 
-    private void generateSessionFromCookie(uk.gov.di.orchestration.shared.entity.Session session) {
+    private void generateSessionFromCookie(
+            uk.gov.di.orchestration.shared.entity.Session session, OrchSessionItem orchSession) {
```
GHSwallow, line 240: Same here with the full module path for Session

GHSwallow (commented):
Although this is getting internalCommomSubjectId from orchSession, it also removes the session getters. I think in test, deleting all session.setInternalCommonSubjectId() would prove that we are only use the orch version, and that the tests still pass. What do you think?

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java
GHSwallow, line 1: This lambda is a wip that isn't currently in use. Still, it will need orch session access when we update the IaC. I think worth adding this to the `.tf` file (it hasn't been migrated to CloudFormation yet), even if only as a comment that it will need orch session read access. It should also eventually need delete access, as an account intervention can happen here, which triggers a logout.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IdentityProgressFrontendHandler.java
```diff
@@ -84,7 +90,10 @@ public APIGatewayProxyResponseEvent handleRequestWithUserSession(
         LOG.info("IdentityProgress request received");
         try {
             var internalCommonSubjectIdentifier =
-                    userSession.getSession().getInternalCommonSubjectIdentifier();
+                    orchSessionService
+                            .getSession(userSession.getClientSessionId())
```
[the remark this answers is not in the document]
BenjaminWCO: No idea what I was trying to do here

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
BenjaminWCO: I agree it's ugly, but personally I think this is fine as a stopgap until we delete the session logic. We could add a separate destroyOrchSession method, but it might be easy to delete the deleteSession method without moving over the client session logic.

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/LogoutServiceTest.java
BenjaminWCO: It's due to the refactor of the destroySession method. However it's really a symptom of using `verify` as we're leaking implementation details into the unit test. Ideally we'd be using a fake instead of a mock and we could query the increment value.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java
GHSwallow: commented on the wrong file 😅

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IdentityProgressFrontendHandler.java
GHSwallow, line 1: This lambda is a wip that isn't currently in use. Still, it will need orch session access when we update the IaC. I think worth adding this to the .tf file (it hasn't been migrated to CloudFormation yet), even if only as a comment that it will need orch session read access. It should also eventually need delete access, as an account intervention can happen here, which triggers a logout.

PR #5465, opened 2024-10-31
BAU: Update doc app stub key id

CarlyG55:
## What

DCMAW are updating their stub, we need to update the key id we are using
## Checklist


billrdunn (approved):

billrdunn (commented):
Is it right that the build key is used in dev?

CarlyG55:
@billrdunn I think it's due to how the stubs are set up - I think identity just have build and production stubs, and since it's just a stub it doesn't matter too much to hook up dev to build. Good spot though

PR #5470, opened 2024-10-31
BAU: bump alarm threshold for the email notification dlq

alhcomer:
## What

It is actually expected that some sqs messages end up in the dlq as Notify may not be able to send messages to some numbers (e.g. international numbers), for example. As such, di-2nd-line should not be alerted for low numbers of messages in the dlq. I've created a ticket to create a tiered alerting system (AUT-3854) where a lower threshold sends an alarm to auth for investigation, and a higher threshold alerts to di-2nd-line. For the time being, this commit bumps the threshold so we don't send alarms to 2nd line unnecessarily.

## How to review

1. Code Review


andrew-moores:
I think it will need to be higher than 10 as that wouldn't have prevented the alarm we saw earlier where there were 12 events on the DLQ.

PR #5480, opened 2024-11-04
Ato 1088 set autnethicated flag for all journeys

Ryan-Andrews99:
## What
- Sets up a feature flag `SET_AUTHENTICATED_FOR_IPV` which controls whether we enable Authenticated in the session once a User is handed back from Auth in all journeys
- Feature flag is set to true for `dev` , `build` and `staging` but turned off for `integration` and `production`

## How to review
- Deploy to dev, I did sandpit:
   - Run through a journey - done
   - Run Auth acceptance tests against the deployed environment - Ran against sandpit, did have 9 failing tests but these also failed when ran on main in sandpit so writing off as noise in sandpit. These were all lockout tests which failed at the  6 wrong MFA code step

## Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
@@ -414,6 +414,10 @@ public class ConfigurationService implements BaseLambdaConfiguration, AuditPubli
         }
     }
 
+    public boolean isAuthenticatedFlagForIpvEnabled() {
+        return getFlagOrFalse("SET_AUTHENTICATED_FLAG_FOR_IPV");
+    }
+
     private Map<String, String> getSsmRedisParameters() {
         if (ssmRedisParameters == null) {
             var getParametersRequest =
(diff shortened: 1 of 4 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
```diff
@@ -414,6 +414,10 @@ public Optional<String> getIPVCapacity() {
         }
     }
 
+    public boolean setAuthenticatedFlagForIpv() {
```
billrdunn, line 417: Bit of a nitpick but reads a bit odd to have a method named `set` which actually does getting. Going to be removed soon anyway so not a blocker to merge

billrdunn (dismissed):

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
Ryan-Andrews99: Can rename to `IsAuthenticatedFlagForIpvEnabled`

Ryan-Andrews99: Have renamed to the suggested

CarlyG55 (approved):

billrdunn (approved):

PR #5482, opened 2024-11-04
ATO-1134: Destroy Orch Session on Sign Out

GHSwallow:
## What
Destroys orchSession from the DynamoDb table on logout.

## How to review
1. Code Review
2. Test in dev
3. Once deployed, test in staging that the account interventions work as expected


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchSessionService.java
```diff
@@ -158,6 +158,10 @@ public Optional<OrchSessionItem> getSessionFromRequestHeaders(Map<String, String
         return sanitiseBase64(sessionId).flatMap(id -> getSession(sessionId));
     }
 
+    public void deleteStoredOrchSession(String sessionId) {
```
billrdunn, line 161: We seem to just refer to the Orch session as "session" in the context of OrchSessionService. So maybe just `deleteSession()`?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchSessionService.java
```diff
@@ -158,6 +158,10 @@ public Optional<OrchSessionItem> getSessionFromRequestHeaders(Map<String, String
...
+        delete(sessionId);
```
billrdunn, line 162: Is there any value wrapping this in a try / catch like we do with the other calls to dynamo?

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/OrchSessionServiceTest.java
```diff
@@ -144,6 +145,15 @@ void shouldNotRetrieveSessionWithMissingHeader() {
         assertTrue(session.isEmpty());
     }
 
+    @Test
+    void shouldDeleteSessionIdFromDatabase() {
```
billrdunn, line 149: Typo? Should it be `shouldDeleteSessionFromDatabase()`?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -120,6 +120,8 @@ private void destroySessions(Session session) {
         }
         LOG.info("Deleting Session");
         sessionService.deleteStoredSession(session.getSessionId());
+        LOG.info("Deleting Orch Session");
+        orchSessionService.deleteStoredOrchSession(session.getSessionId());
```
billrdunn, line 124: What's the plan when the shared session gets deprecated? The Orch session will at some point need to be passed  to `destroySessions()` to get `sessionId`, right?

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
```diff
@@ -309,7 +310,7 @@ void shouldLogoutAndRedirectToBlockedPageWhenAccountStatusIsBlocked()
                         constructQueryStringParameters());
 
         assertRedirectToBlockedPage(response);
-        assertOrchSessionIsUpdatedWithUserInfoClaims();
```
billrdunn, line 312: I assume these were removed because when the Orch session gets deleted these fail? Even though the Orch session does get updated with userinfo claims before it's deleted

GHSwallow: Yea exactly: the tests fail because we can't get the orch session at this point, so `orchSessionExtension.getSession(SESSION_ID)` throws

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
GHSwallow: My thinking is that this was the minimum change required to get this through asap. For sure, eventually we want to add orchSession to the `destroySessions` interface, but we need that for the clientSessions and email as well. So I'll be updating the interface for email, and at that point, the change to use orchSession.getSessioId() will be a simple change.

billrdunn: Sounds good to me!

billrdunn (approved):

billrdunn (approved):

PR #5487, opened 2024-11-05
Ato 975/add current credential strength to user info response

isaac-GDS:
## What

Last steps on the Auth side for the current credential strength:

- Sets`currentCredentialStrength` as `medium` on the `verifyMfaCode` handler in the auth session
- Takes the `currentCredentialStrength` from the auth session and passes it through the userInfo response

Run acceptance test
Tested on Sandpit

### DIFF
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
@@ -113,12 +113,16 @@ public class UserInfoService {
             userInfo.setClaim(
                     AuthUserInfoClaims.VERIFIED_MFA_METHOD_TYPE.getValue(),
                     authSession.getVerifiedMfaMethodType());
+            LOG.info("verified_mfa value: {}", authSession.getVerifiedMfaMethodType());
         }
         if (accessTokenInfo
                 .getClaims()
                 .contains(AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue())) {
             userInfo.setClaim(
                     AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue(),
+                    authSession.getCurrentCredentialStrength());
+            LOG.info(
+                    "current_credential_strength value: {}",
                     authSession.getCurrentCredentialStrength());
         }
     }
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
@@ -114,6 +114,13 @@ public class UserInfoService {
                     AuthUserInfoClaims.VERIFIED_MFA_METHOD_TYPE.getValue(),
                     authSession.getVerifiedMfaMethodType());
         }
+        if (accessTokenInfo
+                .getClaims()
+                .contains(AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue())) {
+            userInfo.setClaim(
+                    AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue(),
+                    authSession.getCurrentCredentialStrength());
+        }
     }
 
     private static String bytesToBase64(ByteBuffer byteBuffer) {
--- a/auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
+++ b/auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
@@ -14,6 +14,7 @@ import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.external.domain.AuthExternalApiAuditableEvent;
 import uk.gov.di.authentication.external.services.UserInfoService;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
+import uk.gov.di.authentication.shared.entity.CredentialTrustLevel;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.MFAMethodType;
 import uk.gov.di.authentication.shared.entity.Session;
@@ -61,6 +62,9 @@ class UserInfoHandlerTest {
     private final Session testSession = new Session(sessionId);
     private final AuthSessionItem authSession = new AuthSessionItem().withSessionId(sessionId);
     private final SerializationService objectMapper = SerializationService.getInstance();
+    private final String testVerifiedMfaMethodType = MFAMethodType.AUTH_APP.getValue();
+    private final CredentialTrustLevel testCurrentCredentialStrength =
+            CredentialTrustLevel.MEDIUM_LEVEL;
 
     @BeforeEach
     public void setUp() {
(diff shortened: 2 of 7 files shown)


auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
```diff
@@ -102,6 +102,9 @@ private void addClaimsFromSession(AuthSessionItem authSession, UserInfo userInfo
         userInfo.setClaim(
                 AuthUserInfoClaims.VERIFIED_MFA_METHOD_TYPE.getValue(),
                 authSession.getVerifiedMfaMethodType());
+        userInfo.setClaim(
+                AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue(),
+                authSession.getCurrentCredentialStrength());
```
GHSwallow, line 107: Rather than always setting the claim, orch should always request it as outlined in this ticket
https://govukverify.atlassian.net/browse/ATO-1129

This is why you're getting the merge conflicts with main I expect - I've just merged the changes that implements this on auth's side.

isaac-GDS: Feels a bit strange for this as from an orch perspective we dont care about the currentCredentialStrength . We just hold on to it. 

But if this is the pattern for things to be passed onto orch will implement.

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
```diff
@@ -88,6 +85,8 @@ public void setUp() {
         TEST_SUBJECT_USER_INFO.setPhoneNumber("0123456789");
         TEST_SUBJECT_USER_INFO.setClaim(
                 "verified_mfa_method_type", MFAMethodType.AUTH_APP.getValue());
+        TEST_SUBJECT_USER_INFO.setClaim(
+                "current_credential_strength", CredentialTrustLevel.MEDIUM_LEVEL);
```
GHSwallow, line 89: I don't think this is doing anything. When I comment it out, the tests still pass. Could you add some assertion if this is necessary here?

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
```diff
@@ -294,6 +292,7 @@ private void withAuthSessionNewAccount() {
                         .getSession(TEST_SESSION_ID)
                         .get()
                         .withAccountState(AuthSessionItem.AccountState.NEW)
-                        .withVerifiedMfaMethodType(MFAMethodType.AUTH_APP.getValue()));
+                        .withVerifiedMfaMethodType(MFAMethodType.AUTH_APP.getValue())
+                        .withCurrentCredentialStrength(CredentialTrustLevel.MEDIUM_LEVEL));
```
GHSwallow, line 296: This also doesn't seem to be being tested. Also, question: if this is a NewAccount, then would their currentCredentialStrength always be null at this point?

GHSwallow (commented):
A few things to think about. Looks good, but I'll leave the approving to Auth:)

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
```diff
@@ -13,10 +13,7 @@
 import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.external.domain.AuthExternalApiAuditableEvent;
 import uk.gov.di.authentication.external.services.UserInfoService;
-import uk.gov.di.authentication.shared.entity.AuthSessionItem;
-import uk.gov.di.authentication.shared.entity.ErrorResponse;
-import uk.gov.di.authentication.shared.entity.MFAMethodType;
-import uk.gov.di.authentication.shared.entity.Session;
+import uk.gov.di.authentication.shared.entity.*;
```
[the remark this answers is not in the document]
BeckaL: Ah yes there's official GDS guidance on avoiding wildcard imports and how to set up intellij from stopping doing it automatically: https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages/java.html#imports

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
```diff
@@ -12,10 +12,7 @@
 import software.amazon.awssdk.core.SdkBytes;
 import uk.gov.di.authentication.external.domain.AuthExternalApiAuditableEvent;
 import uk.gov.di.authentication.external.lambda.UserInfoHandler;
-import uk.gov.di.authentication.shared.entity.AuthSessionItem;
-import uk.gov.di.authentication.shared.entity.ErrorResponse;
-import uk.gov.di.authentication.shared.entity.MFAMethodType;
-import uk.gov.di.authentication.shared.entity.UserProfile;
+import uk.gov.di.authentication.shared.entity.*;
```
BeckaL, line 15: Same here, avoid wildcard imports

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
isaac-GDS: It isn't needed here since its all tested in the `userInfoService` . Was following a pattern of an example userInfo. Think only the `email` and `phone number` is needed. Can remove both the `verified_mfa_method_type` and `current_credential_strength` claims.

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
isaac-GDS: Same as above where following the pattern of including an example but don't know if the claim needs to be specifically tested here.

isaac-GDS: If its a new account it would almost always be non-null as for NewAccount i believe verification always goes through the verifyMfaCode lambda which sets the current credential value.

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
GHSwallow: It isn't used anywhere, but we do expect it to be on the response. The `UserInfoServiceTest` tests that we conditionally add claims, but this file tests the full handler's response. So we can, for example, just add 
```
assertTrue(
                response.getBody()
                        .contains(String.format("\"current_credential_strength\":\"MEDIUM_LEVEL\"")));
```
to verify this is on the response on the first test in this file. Or perhaps better still would be testing the whole response body as a JSON object. What do you reckon?

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
GHSwallow: I think I agree here, that nothing new is being tested. I personally would read this line as saying, "it's important for the authSession to have this property for this test". Certainly if in 6 months I go to uses of the `withCurrentCredentialStrength` method. So I'd either remove it, or add an assertion. Happy with which ever you decide on this one:)

auth-external-api/src/test/java/uk/gov/di/authentication/external/lambda/UserInfoHandlerTest.java
isaac-GDS: I think I do want it in so therefore we should assert something on it. We mock the userInfoService so all we are testing is that the object gets passed through into the response. We don't check the object apart from email and phone number. But I think I prefer that over nothing. Will make the change.

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthExternalApiUserInfoIntegrationTest.java
isaac-GDS: Cool I'll make one of those changes :)

GHSwallow (approved):
lgtm:)

PR #5488, opened 2024-11-05
ATO-975: add currentCredentialStrength claim to all userinfo requests

isaac-GDS:
## What

Following https://govukverify.atlassian.net/browse/ATO-1129 , adding the currentCredentialStrength to always be requested in the userInfo.

This is to allow auth to be able to send back the currentCredentialStrength from to userInfo to orch.

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
@@ -339,7 +339,8 @@ class AuthorisationHandlerTest {
         }
 
         @Test
-        void shouldRedirectToLoginWhenUserHasNoExistingSessionWithSignedAndEncryptedJwtInBody() {
+        void shouldRedirectToLoginWhenUserHasNoExistingSessionWithSignedAndEncryptedJwtInBody()
+                throws com.nimbusds.oauth2.sdk.ParseException, ParseException {
             var orchClientId = "orchestration-client-id";
             when(configService.getOrchestrationClientId()).thenReturn(orchClientId);
             when(orchestrationAuthorizationService.getSignedAndEncryptedJWT(any()))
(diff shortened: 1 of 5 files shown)


billrdunn (commented):
LGTM, only thing I'd suggest is to add some verification logging to follow the same process @GHSwallow is using

https://github.com/govuk-one-login/authentication-api/blob/86642f608b4108719b657088f820b2d63bea37fb/auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java#L65

isaac-GDS:
@billrdunn Shouldn't need the logging as planning on merging [this](https://github.com/govuk-one-login/authentication-api/pull/5487) after. This should be fine as we aren't using the credential strength value in orch yet.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -367,7 +367,7 @@ void shouldRedirectToLoginWhenUserHasNoExistingSessionWithSignedAndEncryptedJwtI
             var captor = ArgumentCaptor.forClass(JWTClaimsSet.class);
             verify(orchestrationAuthorizationService).getSignedAndEncryptedJWT(captor.capture());
             assertEquals(
-                    "{\"userinfo\":{\"email_verified\":null,\"verified_mfa_method_type\":null,\"email\":null}}",
+                    "{\"userinfo\":{\"email_verified\":null,\"current_credential_strength\":null,\"verified_mfa_method_type\":null,\"email\":null}}",
```
GHSwallow, line 370: Not your code, but these tests are quite flaky. I must've missed updating this one when I made the change. Can you replace this with something like the others have, where we compare JSON objects rather than strings?

GHSwallow (commented):
My take on adding logging is that it's been really helpful to uncover cases we're not expecting, and verifying our assumptions. Eg I know that email is always defined on orch, so I can confidently use it. But eg they also showed `verified_mfa_method_type` is not always defined, so we need some null safety. I was being extra careful because the values were in use, but this one isn't yet. What do you reckon?

isaac-GDS:
Will add logging to check for null when running through different journeys

GHSwallow (approved):
lgtm:)

PR #5490, opened 2024-11-06
ATO-1096: Add auth time to Orch session

billrdunn:
## What

As part of the max_age initiative, Orch are to begin setting `authTime` in the Orch session, at the time that the user is authenticated (ie. when the `authenticated` session field goes from `false` to `true`).

Currently `authenticated` is set like this in a number of places, but there is ongoing work ([ATO-1088](https://govukverify.atlassian.net/browse/ATO-1088)) to set it just once in AuthenticationCallback (under the feature flag `SET_AUTHENTICATED_FLAG_FOR_IPV`). To avoid this ongoing work being a blocker, this PR considers only the place that the `authenticated` flag will be set once the flag is live in production.

There is another case that Orch need to set `authTime`, which is when the user has had their credential strength uplifted (eg. from LOW to MEDIUM), which happens when a user with an existing session authenticates with a higher trust level on a second journey.

## Testing
- [x] Run through affected journeys in dev with flag on
- [x] Run through affected journeys in dev with flag off
- [x] Run Auth acceptance tests against dev with flag on
- [x] Run Auth acceptance tests against dev with flag off

[ATO-1088]: https://govukverify.atlassian.net/browse/ATO-1088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -399,17 +399,18 @@ public class AuthenticationCallbackHandler
 
                 CredentialTrustLevel lowestRequestedCredentialTrustLevel =
                         VectorOfTrust.getLowestCredentialTrustLevel(clientSession.getVtrList());
-                long authTime = NowHelper.now().toInstant().getEpochSecond();
                 if (setAuthenticatedFlagForIPV) {
-                    if (userSession.isAuthenticated()) {
-                        if (lowestRequestedCredentialTrustLevel.compareTo(
-                                        userSession.getCurrentCredentialStrength())
-                                > 0) {
-                            orchSessionService.updateSession(orchSession.withAuthTime(authTime));
-                        }
-                    } else {
+                    long authTime = NowHelper.now().toInstant().getEpochSecond();
+                    boolean userHasUplifted =
+                            userSession.getCurrentCredentialStrength() != null
+                                    && lowestRequestedCredentialTrustLevel.compareTo(
+                                                    userSession.getCurrentCredentialStrength())
+                                            > 0;
+                    if (!userSession.isAuthenticated()) {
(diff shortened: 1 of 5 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -384,11 +384,21 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        if (lowestRequestedCredentialTrustLevel.compareTo(
+                                        userSession.getCurrentCredentialStrength())
+                                > 0) {
+                            orchSessionService.updateSession(orchSession.withAuthTime(authTime));
+                        }
+                    } else {
+                        sessionService.storeOrUpdateSession(
+                                userSession.setNewAccount(accountState).setAuthenticated(true));
+                        orchSessionService.updateSession(orchSession.withAuthTime(authTime));
+                    }
```
Ryan-Andrews99, line 401: Do you think there is much value in writing this a bit inverted like this:

```
 long authTime = NowHelper.now().toInstant().getEpochSecond();
                if (setAuthenticatedFlagForIPV) {

                    if(!userSession.isAuthenticated()){
                        orchSessionService.updateSession(orchSession.withAuthTime(authTime));
                    } else if (userSession.isAuthenticated() &&  isUpliftJourney){
                        orchSessionService.updateSession(orchSession.withAuthTime(authTime));
                    }
                } else {
                    sessionService.storeOrUpdateSession(userSession.setNewAccount(accountState));
                }
```

billrdunn: Have updated in the latest commit

Ryan-Andrews99 (approved):

billrdunn:
Closing as superseded by https://github.com/govuk-one-login/authentication-api/pull/5531

PR #5504, opened 2024-11-11
Ato 975/setters current credential strength orch session

isaac-GDS:
##What

Set Current Credential Strength in:
 - AuthenticationCallback
 - AuthCode

It still uses the currentCredentialStrength Value from the old session . Will do the getter in another PR.

 - AuthCode is behind feature flag.
 - Update policy for authCode.


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -365,19 +370,28 @@ private AuthorizationCode generateAuthCode(
...
             session.setCurrentCredentialStrength(lowestRequestedCredentialTrustLevel);
         }
+
+        if (configurationService.isCurrentCredentialStrengthInOrchSessionEnabled()
+                && (isNull(currentCredentialStrength)
+                        || lowestRequestedCredentialTrustLevel.compareTo(currentCredentialStrength)
+                                > 0)) {
+            orchSessionService.updateSession(
+                    orchSession.withCurrentCredentialStrength(lowestRequestedCredentialTrustLevel));
+        }
```
Ryan-Andrews99, line 393: Would it be neater to nest these check so we don't need to make the same comparison twice?

isaac-GDS: Deliberately kept this separate as in the next PR we will be using the orchSession to the do comparison.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -370,19 +375,27 @@ private AuthorizationCode generateAuthCode(
...
         if (isNull(session.getCurrentCredentialStrength())
                 || lowestRequestedCredentialTrustLevel.compareTo(
                                 session.getCurrentCredentialStrength())
                         > 0) {
             session.setCurrentCredentialStrength(lowestRequestedCredentialTrustLevel);
         }
+        if (isNull(currentCredentialStrength)
+                || lowestRequestedCredentialTrustLevel.compareTo(currentCredentialStrength) > 0) {
+            orchSessionService.updateSession(
+                    orchSession.withCurrentCredentialStrength(lowestRequestedCredentialTrustLevel));
```
GHSwallow, line 397: shared session isn't saved at this point, but later in code just before the handler returns. This orchSession save happens in a different `try` block even, so orchSession could get updated while the shared session doesn't get updated. It also means orchSession gets updated with a currentCredentialStrength even if this handler fails in the last `try` block point. Perhaps best to save at the same point as shared session saves here I reckon.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -512,6 +512,17 @@ public APIGatewayProxyResponseEvent handleRequest(
                                 > 0) {
                     userSession.setCurrentCredentialStrength(lowestRequestedCredentialTrustLevel);
                 }
+                var currentCredentialStrength =
+                        userInfo.getStringClaim(
```
GHSwallow, line 516: It looks like getClaim and getStringClaim do basically the same thing but different. However, we haven't used getStringClaim anywhere. Is there a reason here to use this one? Otherwise I think better to keep with getClaim so we can easily find all uses of it

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -512,6 +512,17 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue());
+                if (isNull(currentCredentialStrength)
+                        || lowestRequestedCredentialTrustLevel.compareTo(
+                                        CredentialTrustLevel.valueOf(currentCredentialStrength))
+                                > 0) {
+                    orchSessionService.updateSession(
+                            orchSession.withCurrentCredentialStrength(
+                                    lowestRequestedCredentialTrustLevel));
```
GHSwallow, line 524: This saves earlier than sharedSession, but in the same `try` block. I'm pretty sure that means our sharedSession and orchSession could go out of sync if the code between fails (eg if `generateAndSaveAuthorisationCode` fails). I think best to have the save be the last action, since we only want to save the new state if everything succeeds.

template.yaml
```diff
@@ -2700,8 +2700,7 @@ Resources:
         - !Ref UserProfileTableWriteAccessPolicy
         - !Ref TxmaQueueSendPermissionPolicy
         - !Ref RedisParametersAccessPolicy
-        - !Ref OrchSessionTableReadAccessPolicy
-        - !Ref OrchSessionTableWriteAccessPolicy
+        - !Ref OrchSessionTableReadAndWriteAccessPolicy
```
GHSwallow, line 2714: Better to keep reading and writing as separate policies (unless the limit of 20 policies is being hit)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -384,14 +384,15 @@ private AuthorizationCode generateAuthCode(
...
                 || lowestRequestedCredentialTrustLevel.compareTo(currentCredentialStrength) > 0) {
+            session.setCurrentCredentialStrength(lowestRequestedCredentialTrustLevel);
+        }
+
+        if (configurationService.isCurrentCredentialStrengthInOrchSessionEnabled()
+                && (isNull(currentCredentialStrength)
+                        || lowestRequestedCredentialTrustLevel.compareTo(currentCredentialStrength)
+                                > 0)) {
             orchSessionService.updateSession(
                     orchSession.withCurrentCredentialStrength(lowestRequestedCredentialTrustLevel));
```
GHSwallow, line 397: Open question here about the way we use feature flags: do we need these setters to be feature flagged? I would've thought feature flagging would be more pertinent for getters, or where we're using the getters. Is there some particular testing you want to do at staging with this piece of work?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
isaac-GDS: Think they basically do the same - just a bit more tidy. We do use it elsewhere in the repo. Not consistent in the handler but think we should push to switch it over to `getClaimString` as we do use it elsewhere.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
isaac-GDS: I think I do agree with what you were saying but also hesitant as with this save ordering I was trying to keep to the old code as much as possible. In case ordering was important due to failures.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
isaac-GDS: I do agree with that there shouldn't be any change. Not expecting any but being careful and following the pattern as keeping the setters and getters separated so therefore keeping the testing the same. Its more just checking nothing broke as I would have probably paused the pipeline to run an identity journey. Don't think its too much more as it will be used in the getter under the same flag.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
isaac-GDS: Good point. Agree.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
GHSwallow: When I search the codebase, I can't see any uses of getClaimString. Where are you seeing it be used? 

I do agree though. There are instances of using `getClaim(claimName, String.class)` which is the same, but `getClaimString` is cleaner of course. Might be an easy strong-typing win; a 1 pointer for the backlog perhaps.

template.yaml
GHSwallow: I think the granularity means that we can easily add or remove part of a policy, without risk of affecting a policy the lambda does need. Eg if we were to remove a handler's access to write, but needed it to keep reading, we could just remove write, rather than replacing read-write with read. The reason to use read-write was because we hit the maximum number of policies.

GHSwallow: Also, can you add this policy in a separate PR as a one-line change? Talking with Carly about my work, there's a small risk that the lambda could get deployed before it has this new policy on it, meaning the lambda will briefly fail (for a matter of minutes at most, but still). It's just safer to add the policy then start using it

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
isaac-GDS: It used for the JWTclaims set, claimsSet in the AuthoriseAccessTokenHandler and in a few tests. Its not too much. I guess should we leave this as `getStringClaim` as a reminder and do the rest in another ticket . Or keep it consistent for now and switch it back to `getClaim`.

template.yaml
isaac-GDS: Sure - Ill add it on another PR and merge it first.

isaac-GDS: Just realised if keeping the granularity no change needs to happen here. Will just remove.

PR #5511, opened 2024-11-12
ATO-1147: add resource policy for cross-account access

GHSwallow:
## What
Adds a [resource-based policy](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/rbac-create-table.html) on the OrchSession table in order to grant access to it from the IdentityProcessingHandler in Auth for the purpose of logouts.

## How to review
1. Code Review
2. Deploy and see if the policy is attached as intended


template.yaml
```diff
@@ -417,6 +417,31 @@ Resources:
       Tags:
         - Key: Name
           Value: OrchSessionTable
+      ResourcePolicy:
+        PolicyDocument:
+          Version: 2012-10-17
+          Statement:
+            - Sid: AllowOrchsessionTableAccessForLogout
```
billrdunn, line 424: Not a blocker but I think this follows how we generally name an `Sid`
```suggestion
            - Sid: AllowOrchSessionTableCrossAccountDeleteAccess
```

template.yaml
```diff
@@ -417,6 +417,31 @@ Resources:
...
+              Effect: Allow
+              Action:
+                - dynamodb:DescribeTable
+                - dynamodb:DeleteItem
+              Resource: "*"
```
billrdunn, line 429: Should this be `Resource:  !GetAtt OrchSessionTable.Arn`?

billrdunn (commented):
Can this be tested?

template.yaml
GHSwallow: Because it's a resource policy and is attached to the orch session table, that gives a circular dependency error. `Resource: "*"` (I _believe_) just means every resource, which in this case only means the dynamoDb table.

It seems I can have no `Resource` property. I'll push that change and see if that works in testing

Ryan-Andrews99: [Does a `DependsOn` attribute help fix the circular dependency? ](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-attribute-dependson.html)

Ryan-Andrews99: Wait that wont work, as is, but maybe if the resource policy isn't defined inline

template.yaml
```diff
@@ -393,6 +393,24 @@ Resources:
             Condition:
               ArnLike:
                 kms:EncryptionContext:aws:dynamodb:table/arn: !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*
+          - !If
+            - IsNotDevEnvironment
```
CarlyG55, line 397: Why not dev?

template.yaml
CarlyG55: Yeah it looks a little strange but this is the correct way of writing it

template.yaml
GHSwallow: Because Auth stub Orch for all their dev environments, so none of Auth's lambdas will access Orch's dev environment, and this limits policies to only add access iff Auth needs it

CarlyG55: Auth sandpit links up to orch dev :)

CarlyG55 (approved):

GHSwallow:
Deployed to dev and tested that the `gds-di-development-admin` account can describe-table, get-item and delete-item via the CLI.

PR #5512, opened 2024-11-12
ATO-1147: add IAM policy to access orch session

GHSwallow:
## What
Adds policy for ProcessingIdentityHandler to access the Orch Session table on Orch AWS accounts for logout (resource policy for the DynamboDb table here https://github.com/govuk-one-login/authentication-api/pull/5511)

## How to review

1. Code Review
2. Deployed to dev (along with https://github.com/govuk-one-login/authentication-api/pull/5512 to sandpit). I was able to describe the table when I sign into the Auth dev account from command line. I also created a lambda in Auth dev and verified that I wasn't able to access the orch session table. I then added the policies in https://github.com/govuk-one-login/authentication-api/pull/5512, and verified I was able to both describe and delete an item from the orch session table.

### DIFF
--- a/ci/terraform/oidc/authdev1.tfvars
+++ b/ci/terraform/oidc/authdev1.tfvars
@@ -59,6 +59,7 @@ reauth_enter_sms_code_count_ttl           = 120
 
 orch_client_id  = "orchestrationAuth"
 orch_account_id = "816047645251"
+is_orch_stubbed = true
 
 contra_state_bucket = "di-auth-development-tfstate"
 
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -732,6 +732,46 @@ data "aws_iam_policy_document" "dynamo_auth_session_read_write_policy_document" 
   }
 }
 
+data "aws_iam_policy_document" "dynamo_orch_session_encryption_key_cross_account_decrypt_policy_document" {
+  statement {
+    sid    = "AllowOrchSessionEncryptionKeyCrossAccountDecryptAccess"
+    effect = "Allow"
+    actions = [
+      "kms:Decrypt",
+    ]
+    resources = [
+      var.orch_session_table_encryption_key_arn,
+    ]
+  }
+}
+
+data "aws_iam_policy_document" "dynamo_orch_session_cross_account_read_access_policy_document" {
+  statement {
+    sid    = "AllowOrchSessionCrossAccountReadAccess"
+    effect = "Allow"
+    actions = [
+      "dynamodb:DescribeTable",
+      "dynamodb:Get*",
+    ]
+    resources = [
+      "arn:aws:dynamodb:eu-west-2:${var.orch_account_id}:table/${var.orch_environment}-Orch-Session",
+    ]
+  }
+}
+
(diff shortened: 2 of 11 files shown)


ci/terraform/oidc/authdev1.tfvars
```diff
@@ -57,8 +57,9 @@ reauth_enter_password_count_ttl           = 120
 reauth_enter_auth_app_code_count_ttl      = 120
 reauth_enter_sms_code_count_ttl           = 120
 
-orch_client_id  = "orchestrationAuth"
-orch_account_id = "816047645251"
+orch_client_id   = "orchestrationAuth"
+orch_account_id  = "816047645251"
+orch_environment = "dev"
```
CarlyG55, line 62: Comment to myself: what do we do here, as it's not linked up to orch

CarlyG55: Okay, maybe we have another flag which is to flag whether orch is stubbed or not, and if it is, we don't create the resources that rely on orch_environment?

alhcomer (approved):

ci/terraform/oidc/processing-identity.tf
```diff
@@ -1,3 +1,32 @@
...
+  environment = var.environment
+  role_name   = "ipv-processing-identity-role-with-orch-session-access"
+  vpc_arn     = local.authentication_vpc_arn
+
+  policies_to_attach = var.is_orch_stubbed ? local.base_policies : concat(local.base_policies, [
+    aws_iam_policy.dynamo_orch_session_encryption_key_policy.arn,
+    aws_iam_policy.dynamo_orch_session_logout_policy.arn,
+  ])
+}
+
```
CarlyG55, line 29: This is nice, but will cause us troubles in the future if we have to add more policies. Can we just create a role that has all of the normal policies, and the two orch session ones, and use is_orch_stubbed as a flag on whether to use the old one or the new one?

ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -957,3 +985,19 @@ resource "aws_iam_policy" "dynamo_auth_session_delete_policy" {
...
 }
+
+resource "aws_iam_policy" "dynamo_orch_session_logout_policy" {
+  name_prefix = "dynamo-orch-session-logout-policy"
+  path        = "/${var.environment}/oidc-shared/"
+  description = "IAM policy for managing delete and describe permissions to the orch session table"
+
+  policy = data.aws_iam_policy_document.dynamo_orch_session_logout_policy_document.json
+}
+
```
CarlyG55, line 996: Oh the other comment I got halfway though writing earlier - can we label these policies by what permissions they give, rather than what they're used for?

ci/terraform/oidc/processing-identity.tf
GHSwallow: ah good point

CarlyG55 (dismissed):

GHSwallow:
Deployed to dev and tested that the `gds-di-development-admin` account can describe-table, get-item and delete-item via the CLI.

GHSwallow:
Deployed to sandpit with https://github.com/govuk-one-login/authentication-api/pull/5511 merged. Created a test lambda on `gds-di-development` with all 3 of the new policies, and verified that the lambda could `describeTable`, `get` a table entry and `delete` a table entry.

PR #5514, opened 2024-11-13
AUT-3845: Enable snapstart for AIS stub

mattvot:
## What

The AIS stub is causing acceptance tests to fail when run in parallel. This happened because the stub lambda cold started another instance during the run, meaning the response took longer than the client configured timeout.

Here we remove the provisioned concurrency for the stub lambda, replacing it with snapstart. We also switch the architecture for this lambda to arm64 for the snapstart support.

## How to review

1. Code Review
1. Deploy to sandpit with `./deploy-dev.sh sandpit -b -c -s -o -i`
1. Run though sign in journey
1. See no error

## Checklist

### DIFF
--- a/ci/terraform/modules/endpoint-lambda/variables.tf
+++ b/ci/terraform/modules/endpoint-lambda/variables.tf
@@ -149,3 +149,13 @@ variable "dynatrace_secret" {
   })
   sensitive = true
 }
+
+variable "snapstart" {
+  type    = bool
+  default = false
+}
+
+variable "architectures" {
+  type    = list(string)
+  default = ["x86_64"]
+}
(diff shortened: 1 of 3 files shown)


ci/terraform/modules/endpoint-lambda/variables.tf
```diff
@@ -149,3 +149,13 @@ variable "dynatrace_secret" {
   })
   sensitive = true
 }
+
+variable "snapstart" {
+  type    = bool
+  default = false
+}
```
ethanmills, line 156: I wonder if it would be clearer to make the exclusivity explicit, rather than the changes to concurrency config you've made above?
```suggestion
variable "snapstart" {
  type    = bool
  default = false
  validation {
    condition = var.provisioned_concurrency == 0 || !var.snapstart
    error_message = "Snapstart and provisioned concurrency cannot be used together"
  }
}
```

whi-tw: issue: can't refer to other variables within the condition

PR #5529, opened 2024-11-14
ATO-1182: Refactor sessions setting in AuthenticationCallback

billrdunn:
## What

While working on setting auth_time in AuthenticationCallback, it was becoming a bit complex and hard to read. For maintainability, this PR contains two small refactors:

1. Where possible, separate setting to the session objects, and updating the database.
2. Use private method to separate logic from main handler.

## Testing
No changes to tests required, so I'm confident this is OK, but I will

- [x] Run through auth journey in dev

PR #5531, opened 2024-11-14
ATO-1096: Add auth time to Orch session

billrdunn:
## What

As part of the max_age initiative, Orch are to begin setting `authTime` in the Orch session, at the time that the user is authenticated (ie. when the `authenticated` session field goes from `false` to `true`).

There is another case that Orch need to set `authTime`, which is when the user has had their credential strength uplifted (eg. from LOW to MEDIUM), which happens when a user with an existing session authenticates with a higher trust level on a second journey.

This PR is based off two others and so PRs need to be merged in this order:
1. https://github.com/govuk-one-login/authentication-api/pull/5522
2. https://github.com/govuk-one-login/authentication-api/pull/5529
3. https://github.com/govuk-one-login/authentication-api/pull/5531

## Testing
- [ ] Run through affected journeys in dev 
- [ ] Run Auth acceptance tests against dev

[ATO-1088]: https://govukverify.atlassian.net/browse/ATO-1088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthenticationCallbackHandlerIntegrationTest.java
@@ -700,6 +700,49 @@ public class AuthenticationCallbackHandlerIntegrationTest extends ApiGatewayHand
                                 constructQueryStringParameters()));
     }
 
+    @Test
+    void shouldSetAuthTimeWhenUserHasBeenUplifted() throws Json.JsonException {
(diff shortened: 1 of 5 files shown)


Ryan-Andrews99 (approved):

CarlyG55 (dismissed):

billrdunn:
Closing as superseded by https://github.com/govuk-one-login/authentication-api/pull/5583

PR #5549, opened 2024-11-19
BAU: make value a string in terraform

isaac-GDS:
## What

Running into error when trying to deploy sandpit. Fix is just to update int value to string.


billrdunn (approved):

isaac-GDS:
https://github.com/govuk-one-login/authentication-api/pull/5550#pullrequestreview-2445374299 - Same changes

PR #5556, opened 2024-11-20
ATO-1229: Set `UpliftRequired` in Auth session

billrdunn:
## What

When uplift is required, set it in the Auth session. This value will be passed to Orch in the userinfo response.

## Testing

- [x] uplift and non-uplift journey in dev sets value appropriately

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -259,6 +259,13 @@ public class StartHandler
                             isBlockedForReauth,
                             isUserAuthenticatedWithValidProfile);
 
+            var authSession = authSessionService.getSession(session.getSessionId());
+            authSession.ifPresent(
+                    authSessionItem ->
+                            authSessionService.updateSession(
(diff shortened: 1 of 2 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -259,6 +259,14 @@ public APIGatewayProxyResponseEvent handleRequest(
                             isBlockedForReauth,
                             isUserAuthenticatedWithValidProfile);
 
+            if (userStartInfo.isUpliftRequired()) {
+                var authSession = authSessionService.getSession(session.getSessionId());
+                authSession.ifPresent(
+                        authSessionItem ->
+                                authSessionService.updateSession(
```
billrdunn, line 266: I'm aware that this updates the value in the Auth session at the point that _uplift is determined to be required_, not at the point that _uplift actually happens_. I think this is fine, but can we be certain that the two will never differ?

mattvot (approved):

billrdunn:
Closed as superseded by https://github.com/govuk-one-login/authentication-api/pull/5575

PR #5559, opened 2024-11-20
ATO-1230: Return uplift_required in userinfo response

billrdunn:
## What

Include uplift_required in userinfo response if Orch have requested it (Orch will [always request it](https://github.com/govuk-one-login/authentication-api/pull/5558)). 
If true, Orch will set auth time in the Orch session

## Testing

- [x] in sandpit, userinfo response includes `uplift_required=false` on a no-2FA journey. Then without logging out, `uplift_required=true` on a 2FA journey


mattvot (approved):

billrdunn:
Closed as superseded by https://github.com/govuk-one-login/authentication-api/pull/5580

PR #5567, opened 2024-11-22
BAU: Refactor Auth session setting in Start handler (2nd attempt)

billrdunn:
_This is a second attempt at deploying as [the previous](https://github.com/govuk-one-login/authentication-api/pull/5561) was reverted_

## What
The `addOrUpdateSessionId` method is specifically for updating the sessionId of the Auth session. Instead the Auth session is built with certain fields (including currentCredentialStrength) and then updated in Dynamo later in the handler.

Feature flag implementation as the previous deploy caused errors in StartHandler in prod.

## Testing

- [x] currentCredentialStrength still set properly in sandpit-auth-session table

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -209,8 +209,12 @@ public class StartHandler
                     Optional.ofNullable(startRequest.previousSessionId());
             CredentialTrustLevel currentCredentialStrength =
                     startRequest.currentCredentialStrength();
-            authSessionService.addOrUpdateSessionId(
-                    previousSessionId, session.getSessionId(), currentCredentialStrength);
+            if (configurationService.isStartSessionRefactorEnabled()) {
+                authSessionService.addOrUpdateSessionId(previousSessionId, session.getSessionId());
+            } else {
+                authSessionService.addOrUpdateSessionId(
+                        previousSessionId, session.getSessionId(), currentCredentialStrength);
+            }
 
             var clientSessionId =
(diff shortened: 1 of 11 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -209,8 +210,21 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                                .orElseThrow(
+                                        () ->
+                                                new AuthSessionException(
+                                                        "Auth session not found in DynamoDB"));
+                authSession.setCurrentCredentialStrength(startRequest.currentCredentialStrength());
+                authSessionService.updateSession(authSession);
+            } else {
+                authSessionService.addOrUpdateSessionId(
+                        previousSessionId, session.getSessionId(), currentCredentialStrength);
+            }
```
GHSwallow, line 227: Could you explain why we need to write to the table twice? I agree that `addOrUpdateSessionId` as a name doesn't encapsulate that the session properties are also updated, but it could be renamed. Perhaps eg `updateSessionAndSessionId(previousSessionId, authSession)`. And then we don't have to build a new authSession within the method - we can just pass through the session we want to update it to.

billrdunn: We could rename to what you suggest, but the complication comes when `upliftRequired` also needs setting in StartHandler (and potentially more session fields in future). My plan was to first update the sessionId, then update all other session fields in one go (if possible). We may find in future that we can update sessionId and all the fields at once which would be nice, but I didn't want to assume that right now

billrdunn:
Closing as not required now

PR #5573, opened 2024-11-25
AUT-3906: Add cost allocation tags to authentication-api

whi-tw:
## What

Add Cost Allocation tags to all components

I've updated the `default_tags` block on the providers to the following:
```
  default_tags {
    tags = {
      Environment = var.environment
      Owner       = "di-orchestration@digital.cabinet-office.gov.uk" / "di-authentication@digital.cabinet-office.gov.uk"
      Product     = "GOV.UK Sign In"
      System      = "${terraform-module-name}"

      application = "${terraform-module-name}"
    }
  }
```

This means that all the resources created by this terraform will now be correctly labeled and easily identified through the cost explorer, so we can better gauge what's costing what.

Note: `environment` => `Environment`. I would have left the two, but labels are case sensitive except for one specific instance, so it's easier to just keep one of these, otherwise we get flip-flopping resources.

## How to review

1. Deploy to a dev environment
2. Check some arbitrary resources have the expected tags

### DIFF
--- a/ci/terraform/modules/endpoint-lambda/variables.tf
+++ b/ci/terraform/modules/endpoint-lambda/variables.tf
@@ -74,6 +74,15 @@ variable "extra_tags" {
   description = "Extra tags to apply to resources"
 }
 
+locals {
+  extra_tags = merge(
+    var.extra_tags,
+    {
+      Service = var.endpoint_name
+    }
+  )
+}
+
 variable "cloudwatch_key_arn" {
   type        = string
   description = "The ARN of the KMS key to use log encryption"
--- a/ci/terraform/oidc/mfa-reset-jar-jwk.tf
+++ b/ci/terraform/oidc/mfa-reset-jar-jwk.tf
@@ -1,3 +1,7 @@
+locals {
+  reverification_jwk_json_endpoint_name = "reverification-jwk.json"
+}
+
 module "mfa_reset_jar_signing_jwk_role" {
   source      = "../modules/lambda-role"
   environment = var.environment
@@ -7,12 +11,15 @@ module "mfa_reset_jar_signing_jwk_role" {
   policies_to_attach = [
     aws_iam_policy.mfa_reset_jar_kms_signing_jwk_policy.arn
   ]
+  extra_tags = {
+    Service = local.reverification_jwk_json_endpoint_name
+  }
 }
 
 module "mfa_reset_jar_signing_jwk" {
   source = "../modules/endpoint-module"
 
-  endpoint_name   = "reverification-jwk.json"
+  endpoint_name   = local.reverification_jwk_json_endpoint_name
   path_part       = "reverification-jwk.json"
   endpoint_method = ["GET"]
   environment     = var.environment
--- a/ci/terraform/oidc/site.tf
+++ b/ci/terraform/oidc/site.tf
@@ -37,7 +37,12 @@ provider "aws" {
 
   default_tags {
     tags = {
-      environment = var.environment
+      Environment = var.environment
+      Owner       = "di-authentication@digital.cabinet-office.gov.uk"
+      Product     = "GOV.UK Sign In"
+      System      = "oidc-api"
+      # Don't set `Service` by default as it's not always applicable
+
       application = "oidc-api"
     }
   }
--- a/ci/terraform/oidc/token.tf
+++ b/ci/terraform/oidc/token.tf
@@ -1,3 +1,7 @@
+locals {
+  oidc_token_endpoint_name = "token"
+}
+
 module "oidc_token_role" {
   source      = "../modules/lambda-role"
   environment = var.environment
@@ -16,6 +20,9 @@ module "oidc_token_role" {
     local.user_credentials_encryption_policy_arn,
     local.user_profile_encryption_policy_arn
   ]
+  extra_tags = {
+    Service = local.oidc_token_endpoint_name
+  }
 }
 
 data "aws_iam_policy_document" "kms_signing_policy_document" {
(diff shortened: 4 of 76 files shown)


ci/terraform/oidc/mfa-reset-jar-jwk.tf
```diff
@@ -7,6 +7,9 @@ module "mfa_reset_jar_signing_jwk_role" {
   policies_to_attach = [
     aws_iam_policy.mfa_reset_jar_kms_signing_jwk_policy.arn
   ]
+  extra_tags = {
+    Service = "reverification-jwk.json"
```
andrew-moores, line 11: What is the significance of the .json extension to the name of the Service?

andrew-moores, line 11: Why isn't the name here `mfa-reset-jar-jwk`.

andrew-moores (commented):
When should we be able to see these new tags in Cost Explorer.  We deployed oidc to sandpit but couldn't see any of the tags yet.

whi-tw:
> When should we be able to see these new tags in Cost Explorer.  We deployed oidc to sandpit but couldn't see any of the tags yet.

From what I can tell, [~24 hours](https://aws.amazon.com/aws-cost-management/aws-cost-explorer/#:~:text=The%20current%20month's%20data%20is,least%20once%20every%2024%20hours.)

It should start to trickle in over time though. Maybe there will be something in a couple hours. This was deployed in authdev1 until about 24 hours ago, if that's a reasonable 'it works' check?

ci/terraform/modules/endpoint-lambda/variables.tf
```diff
@@ -74,6 +74,15 @@ variable "extra_tags" {
...
 }
 
+locals {
+  extra_tags = merge(
+    var.extra_tags,
+    {
+      Service = var.endpoint_name
+    }
+  )
+}
```
andrew-moores, line 84: Its not clear what is happening here, what is the difference between var.extra_tags and local.extra_tags in preceding changes.

ci/terraform/oidc/site.tf
```diff
@@ -37,7 +37,12 @@ provider "aws" {
 
   default_tags {
     tags = {
-      environment = var.environment
+      Environment = var.environment
+      Owner       = "di-orchestration@digital.cabinet-office.gov.uk"
```
andrew-moores, line 41: Is the owner `di-orchestration` or is it auth as the frontend-api is deployed via this terraform (and not sure if orchs lambdas are still deployed by terraform)?

ci/terraform/oidc/token.tf
```diff
@@ -16,6 +16,9 @@ module "oidc_token_role" {
     local.user_credentials_encryption_policy_arn,
     local.user_profile_encryption_policy_arn
   ]
+  extra_tags = {
+    Service = "token"
```
andrew-moores, line 20: Should this be `oidc-token` see `auth-token` previously defined.  Similar comment about the following names that could apply to several of the apis.

ci/terraform/oidc/wellknown.tf
```diff
@@ -4,6 +4,9 @@ module "openid_configuration_role" {
   role_name   = "openid-configuration"
   environment = var.environment
   vpc_arn     = local.authentication_vpc_arn
+  extra_tags = {
+    Service = "openid-configuration"
```
andrew-moores, line 8: Not clear to us how someone is going to use Cost Explorer and know that `openid-configuration` is the OAuth well known endpoint.

ci/terraform/shared/site.tf
```diff
@@ -59,7 +64,12 @@ provider "aws" {
 
   default_tags {
```
andrew-moores, line 65: Is this duplicated with above?  What is the backend that sits behind this and is it shared with Orch?

ci/terraform/test-services/delete-synthetics-user.tf
```diff
@@ -9,6 +9,9 @@ module "test_services_api_delete-synthetics-user_role" {
     aws_iam_policy.dynamo_test_services_user_delete_access_policy.arn,
     module.test_services_txma_audit.access_policy_arn,
   ]
+  extra_tags = {
+    Service = "synthetics-user"
```
andrew-moores, line 13: `delete-synthetics-user`?

ci/terraform/modules/endpoint-lambda/variables.tf
whi-tw: resources don't get 'Service' by default, so we add a value for 'Service' based on the endpoint name. Merging this one extra tag with `var.extra_tags` and then using the combination instead of the bare variable is the simplest way to keep both the specifically defined extra_tags and the additional service one.

ci/terraform/oidc/site.tf
whi-tw: entirely reasonable. I'll switch this out.

ci/terraform/oidc/token.tf
whi-tw: same as before: because module tag = endpoint_name, role tag has to be the same. Pulled out to variable here too.

ci/terraform/oidc/wellknown.tf
whi-tw: as above

ci/terraform/shared/site.tf
whi-tw: yeah, this is so we can deploy stuff in `us-east-1`: we need a separate aws provider instance to achieve that.

I've kept the tags the same for simplicity, it may be that these are separate IRL, but unsure / this feels a bit too 'in the weeds' for this first-brush attempt of adding these tags. Happy to update if y'all want though.

ci/terraform/test-services/delete-synthetics-user.tf
whi-tw: as above

andrew-moores (approved):

PR #5574, opened 2024-11-25
BAU: Rename AuthSessionService method to better reflect its usage

billrdunn:
## What
1. The method was named as if it was just updating the sessionId, but it actually updates currentCrendentialStrength too, so it has been renamed for clarity.
2. Remove unnecessary variable declarations.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
@@ -40,7 +40,7 @@ public class AuthSessionService extends BaseDynamoService<AuthSessionItem> {
         this.configurationService = configurationService;
     }
 
-    public void addOrUpdateSessionId(
+    public void updateSessionIncludingSessionId(
             Optional<String> previousSessionId,
             String newSessionId,
             CredentialTrustLevel currentCredentialStrength) {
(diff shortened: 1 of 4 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
```diff
@@ -40,7 +40,7 @@ public AuthSessionService(
         this.configurationService = configurationService;
     }
 
-    public void addOrUpdateSessionId(
+    public void updateSessionIncludingSessionId(
```
GHSwallow, line 43: I think perhaps `addOrUpdateSessionIncludingSessionId` might be better here. In the case that an `oldItem` can't be found, it will create a new one, with the log message 
`"New item added to Auth session table"`
which I don't think is expected from a method which says it updates a session.

This is definitely a tricky one to name though. I think this method might be doing too much to be honest. But I think `addOrUpdateSessionIncludingSessionId` does capture all that it does. 

What do you reckon?

billrdunn: Yep sure I can go with that 👍

billrdunn:
Closed as superseded by https://github.com/govuk-one-login/authentication-api/pull/5575

PR #5590, opened 2024-11-27
ATO-1248: Add AuthSession To UserContext

ghost:
## What
Added `AuthSession` to `UserContext`. The old `Session` lives on the `UserContext` so this will help us switch over to the `AuthSession` with less refactoring.

## How to review

Code Review

## Notes
The auth session has been added as an optional field for the time being. This is related to how the auth session is currently initialised/updated in the start handler where the user context is used to construct one of the inputs. With a bit of refactoring we should be able to make it non-optional once the start handler is no longer dependant on the session.

### DIFF
--- a/ci/terraform/oidc/account-interventions.tf
+++ b/ci/terraform/oidc/account-interventions.tf
@@ -15,7 +15,8 @@ module "frontend_api_account_interventions_role" {
     module.oidc_txma_audit.access_policy_arn,
     ], local.deploy_ticf_cri_count == 1 ? [
     aws_iam_policy.ticf_cri_lambda_invocation_policy[0].arn,
-  ] : [])
+    ] : [],
+  [aws_iam_policy.dynamo_auth_session_read_policy.arn])
 }
 
 module "account_interventions" {
--- a/shared/src/main/java/uk/gov/di/authentication/shared/state/UserContext.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/state/UserContext.java
@@ -1,5 +1,6 @@
 package uk.gov.di.authentication.shared.state;
 
+import uk.gov.di.authentication.shared.entity.AuthSessionItem;
 import uk.gov.di.authentication.shared.entity.ClientRegistry;
 import uk.gov.di.authentication.shared.entity.ClientSession;
 import uk.gov.di.authentication.shared.entity.Session;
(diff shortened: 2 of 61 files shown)


Ryan-Andrews99 (commented):
Have looked through the first few of the handlers and most of them don't have the read permissions for the auth session table. All the handlers which inherit from the BaseFrontendHandler will need this as it is querying the table to check if there is an auth session.

Also would it be possible to maybe break this PR up into several commits which apply the changes to each lambda. Just makes this PR less scary to review 😄

Ryan-Andrews99 (approved):
Approved awaiting auth review

shared/src/main/java/uk/gov/di/authentication/shared/state/UserContext.java
```diff
@@ -45,6 +49,10 @@ public Session getSession() {
         return session;
     }
 
+    public AuthSessionItem getAuthSession() {
+        return authSession.orElseThrow(() -> new RuntimeException("Auth session not set."));
```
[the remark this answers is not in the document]
BenjaminWCO: This is a good question. I actually made a mistake and was calling `buildUserContext` with a `null` auth session from the start handler instead of `optional.empty` (fixed now). From the other handlers, it will always be present as the base frontend handler returns a http error if the auth session isn't present. The reason for making it optional is really just a stopgap, as if you look in the main handle method, the auth session is constructed with  _upliftRequired_ as a parameter, but  _upliftRequired_ is computed off of the user context, which otherwise requires an auth session. The construction loop only effects this handler and having discussed it with Bill we thought it's something to be considered in a future refactor.

ci/terraform/oidc/account-interventions.tf
```diff
@@ -13,6 +13,7 @@ module "frontend_api_account_interventions_role" {
     aws_iam_policy.dynamo_client_registry_read_access_policy.arn,
     aws_iam_policy.redis_parameter_policy.arn,
     module.oidc_txma_audit.access_policy_arn,
+    aws_iam_policy.dynamo_auth_session_read_policy.arn,
```
GHSwallow, line 16: This (and all new policies) needs to be added at the end of the list of policies in auth's roles. If the order is changed, the role is rewritten while live (not ideal infrastructure), meaning we'd have brief downtime on this lambda while it's permissions were updated. Concat can take multiple arguments, so it's best to have this added after the `ticf_cri_lambda_invocation_policy` just below.

BenjaminWCO: done

PR #5591, opened 2024-11-27
BAU/fix resource tagging

whi-tw:
## What

Use provider default_tags, rather than passing tags around like fools

## Checklist


ci/terraform/delivery-receipts/site.tf
```diff
@@ -27,15 +30,16 @@ provider "aws" {
...
 
-locals {
-  // Using a local rather than the default_tags option on the AWS provider, as the latter has known issues which produce errors on apply.
-  default_tags = var.use_localstack ? null : {
-    environment = var.environment
-    application = "delivery-receipts-api"
+  default_tags {
+    tags = {
+      environment = var.environment
+      application = "account-management-api"
```
mattvot, line 37:
```suggestion
      application = "delivery-receipts-api"
```

ci/terraform/modules/api-gateway/variables.tf
```diff
@@ -7,7 +7,7 @@ variable "api_gateway_name" {
   type        = string
 }
 
-variable "tags" {
+variable "extra_tags" {
   description = "The tags to apply to the created resources"
```
mattvot, line 11: Does `extra_tags` exists as it adds additional tags to the resource on top of the other set of default tags?

ci/terraform/modules/public-api-gateway/variables.tf
```diff
@@ -51,3 +51,8 @@ variable "waf_arns" {
   description = "The ARNs of any WAFs to attach to the API Gateway"
   type        = list(string)
 }
+
+variable "domain_name" {
```
mattvot, line 55: Is this use in [this file](https://github.com/govuk-one-login/authentication-api/blob/03f1103df1f7985ef6363a07b7b7d529081f6187/ci/terraform/modules/public-api-gateway/domain-mapping.tf) the reason to add this. Should we expect an error or warning without this variable declared?

mattvot (commented):
Thanks @whi-tw, just left a few comments

ci/terraform/delivery-receipts/site.tf
whi-tw: Oops! fixed in a fixup

ci/terraform/modules/api-gateway/variables.tf
whi-tw: yeah, that's exactly right.

I'll update the description of the variable to make this more obvious!

ci/terraform/modules/public-api-gateway/variables.tf
whi-tw: Yeah, that is correct.

I don't believe this module is actually in use anywhere, it's there mostly because we have a private one, so it makes sense to also have the `public` version of the module implemented.

When creating the module, apparently I forgot to add this variable to `variables.tf` and so when modifying the `extra_tags` variable, I had a red squiggly, so fixed this variable while here, to avoid any validation errors.

Happy to drop the commit that adds this variable if it's creating noise in the context of this PR

mattvot: Thanks, makes sense. No need to drop from this PR 👍

mattvot (dismissed):

mattvot (approved):

mattvot (approved):

PR #5592, opened 2024-11-27
BAU: It's Never Localstack

whi-tw:
## What

It's never localstack. Remove all references to localstack

## How to review

1. Deploy to dev
2. Check the deploy worked


mattvot (commented):
Changes look good to me. Thanks for going though this. Just need to test deployments now

mattvot (approved):

PR #5602, opened 2024-11-29
Ato 975/turn on current credential strength auth code setting

isaac-GDS:
## What

Turn on setting current credential strength in the orch session in Prod and Int.

This shouldn't affect any flows or behaviours.


isaac-GDS:
Tested Success identity journey and return Code in staging. Working as expected.

PR #5604, opened 2024-11-29
BAU: reduce lamba policies

whi-tw:
## What

Previously, we created many policies, and then attached them all to the
individual roles.

This meant that we often bumped up against the 20 policy limit for a
role. It also took a long time to run the terraform apply, as it had to
check MANY policies for changes.

This change introduces a replacement for the lambda-role module, which
takes in a map of policy documents lists and combines them into a
reduced number of policies specific to the role.

eg.

```hcl
module "lambda-role-policy-reduction" {
  source = "./modules/lambda-role-policy-reduction"

  role_name = "my-role"
  environment = "dev"
  default_tags = var.default_tags
  vpc_arn = var.vpc_arn

  policy_documents_to_attach = {
    "dynamodb" = [
      data.aws_iam_policy_document.dynamodb_write_policy.json,
      data.aws_iam_policy_document.dynamodb_read_policy.json,
    ],
    "s3" = [
      data.aws_iam_policy_document.s3_list_policy.json,
      data.aws_iam_policy_document.s3_read_policy.json,
      data.aws_iam_policy_document.s3_write_policy.json,
    ],
    "sqs" = [
      data.aws_iam_policy_document.sqs_policy.json,
    ],
    "redis" = [
      data.aws_iam_policy_document.redis_read_policy.json,
      data.aws_iam_policy_document.redis_write_policy.json,
    ],
  }
}
```
This would create and attach 5 policies:
1. base: Contains the logging, networking and xray policies that are
    common to all roles.
2. dynamodb: Contains the dynamodb policies
3. s3: Contains the s3 policy
4. sqs: Contains the sqs policy
5. redis: Contains the redis policy

previously, this would have created and attached 11 policies to the role
ie. 1 more than the limit.

## How to review

- Code Review

## Before Merging

- remove POC implementation

### DIFF
--- a/ci/terraform/modules/lambda-role-policy-reduction/role.tf
+++ b/ci/terraform/modules/lambda-role-policy-reduction/role.tf
@@ -34,3 +34,7 @@ resource "aws_iam_role_policy_attachment" "provided_policies" {
     aws_iam_role.lambda_role
   ]
 }
+resource "aws_iam_role_policy_attachments_exclusive" "exclusively_managed_policies" {
+  role_name   = aws_iam_role.lambda_role.name
+  policy_arns = aws_iam_policy.iam-policy[*].arn
+}


ci/terraform/modules/lambda-role-policy-reduction/variables.tf
```diff
@@ -0,0 +1,33 @@
...
+  type        = map(list(string))
+  default     = {}
+  description = "Policy documents to combine and attach to the role"
+
+  validation {
+    condition     = !contains(keys(var.policy_documents_to_attach), "base")
+    error_message = "policy_documents_to_attach cannot contain a key named 'base'"
+  }
+  validation {
+    condition     = length(keys(var.policy_documents_to_attach)) <= 20
```
mattvot, line 25: Off by 1 here. We already have 1 policy defined, `base`

ci/terraform/modules/lambda-role-policy-reduction/policy.tf
```diff
@@ -0,0 +1,73 @@
...
+}
+
+locals {
+  policy_documents_to_attach = merge({
+    base = concat([
+      data.aws_iam_policy_document.logging_policy_document.json,
+      data.aws_iam_policy_document.endpoint_xray_policy.json,
+      ],
+    var.vpc_arn == "" ? [] : [data.aws_iam_policy_document.networking_policy_document[0].json]),
+  }, var.policy_documents_to_attach)
```
mattvot, line 57: Do the order of these policies matter? i.e. if the json object keys are not returned in the same order, will that cause AWS/terraform to fail to deploy the change (see previous conversations about changes to policy ordering messing up a deployment)?

ci/terraform/modules/lambda-role-policy-reduction/policy.tf
```diff
@@ -0,0 +1,73 @@
...
+}
+
+data "aws_iam_policy_document" "iam-policy-document" {
+  for_each                = local.policy_documents_to_attach
+  source_policy_documents = each.value
```
ethanmills, line 62: Do we need to worry about this?

> Statements defined in source_policy_documents must have unique `sid`s

ethanmills:
What does the plan look like if I add a policy doc to e.g. the `sqs` list? Is it a modify in place or a delete and replace?

ci/terraform/modules/lambda-role-policy-reduction/policy.tf
whi-tw: @ethanmills thanks for taking a look - I was going to reach out for input.

Yeah, that's a super valid point I hadn't considered. If we go ahead with this approach, I'll add a validation step in to ensure uniqueness.

PR #5608, opened 2024-12-02
ATO-1099: Update session when max age has expired

billrdunn:
## What
There is a specific set of updates required to the Orch session in relation to max age when all the following conditions are true:
- There is an existing Orch session
- Max age is enabled for the client
- Max age has expired

The updates to the Orch session are:
- Update the session ID of the existing (previous) Orch session. The session ID is updated as a security measure. Note that the previous session is not deleted (contrary to the non max age session update journey) because it is required after Authentication. The value of internalCommonSubjectIdentifier from the previous session will be compared against the value in the userinfo response, to confirm that the reauthenticated user is the same as the original user. Also the TTL of the previous session is bumped to ensure it is accessible after Authentication.
- Update the existing Orch session with authenticated=false (to force reauthentication) and add the new session ID for the previous session (to query ICSI from the previous session as described above).

Implementation has been feature flagged.

## Testing
Successfully run through the following journeys in dev, with both query parameter and request object requests:
- max age not set, sign in 2FA journey, claims returned
- max age not set, no sign in journey, claims returned
- max age 0 forces reauthentication
- max age 1234 does not force reauthentication
- max age 5 forces reauthentication
- max age -3 fails with appropriate error message
- max age “abc” fails with appropriate error message

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -609,26 +609,69 @@ public class AuthorisationHandler
                     authenticationRequest.getClientID().getValue(),
                     user);
         }
-        var session = existingSession.orElseGet(sessionService::generateSession);
-        attachSessionIdToLogs(session);
 
+        Session session;
+        OrchSessionItem orchSession;
         Optional<String> existingSessionId = existingSession.map(Session::getSessionId);
-        if (existingSession.isEmpty()) {
-            updateAttachedSessionIdToLogs(session.getSessionId());
-            LOG.info("Created session");
+
+        if (existingSession.isEmpty() || existingOrchSessionOptional.isEmpty()) {
+            session = sessionService.generateSession();
+            var newSessionId = session.getSessionId();
+            orchSession = createNewOrchSession(newSessionId);
+            LOG.info("Created session with id: {}", newSessionId);
         } else {
-            existingSessionId = Optional.of(session.getSessionId());
-            sessionService.updateWithNewSessionId(session);
-            updateAttachedSessionIdToLogs(session.getSessionId());
-            LOG.info("Updated session id from {} - new", existingSessionId);
+            var maxAgeParam = getMaxAge(authenticationRequest);
+            boolean isMaxAgeSupported =
+                    configurationService.supportMaxAgeEnabled() && client.getMaxAgeEnabled();
+            final long timeNow = NowHelper.now().toInstant().getEpochSecond();
+
+            if (maxAgeParam.isPresent()
+                    && isMaxAgeSupported
+                    && maxAgeExpired(
+                            existingOrchSessionOptional.get().getAuthTime(),
+                            maxAgeParam,
+                            timeNow)) {
+                var newSessionIdForPreviousSession = IdGenerator.generate();
+                var newSessionId = IdGenerator.generate();
+                session =
+                        updateSharedSessionDueToMaxAgeExpiry(
+                                existingSession.get(),
+                                newSessionIdForPreviousSession,
+                                newSessionId);
+                orchSession =
+                        updateOrchSessionDueToMaxAgeExpiry(
+                                newSessionId,
+                                existingOrchSessionOptional.get(),
+                                timeNow,
+                                newSessionIdForPreviousSession);
+
+                LOG.info(
+                        "Updated previous Orch and shared session ID due to max age expiry. Session ID updated from {} to {}",
+                        existingOrchSessionOptional.get().getSessionId(),
+                        newSessionIdForPreviousSession);
+
+                LOG.info(
+                        "Created new Orch and shared sessions with session ID {} due to max age expiry",
+                        newSessionId);
+
+            } else {
+                var sessionToBeUpdated = existingSession.get();
+                sessionService.updateWithNewSessionId(sessionToBeUpdated);
+                session = sessionToBeUpdated;
+
+                orchSession =
+                        updateOrchSession(
+                                session.getSessionId(), existingOrchSessionOptional.get(), timeNow);
+
+                LOG.info(
+                        "Updated existing session ID from {} to {}",
+                        orchSession.getSessionId(),
+                        session.getSessionId());
+            }
         }
 
-        String newSessionId = session.getSessionId();
-        var orchSession =
-                existingOrchSessionOptional.isPresent()
-                        ? updateExistingOrchSession(newSessionId, existingOrchSessionOptional.get())
-                        : createNewOrchSession(newSessionId);
-        attachOrchSessionIdToLogs(newSessionId);
+        attachSessionIdToLogs(session);
+        attachOrchSessionIdToLogs(orchSession.getSessionId());
 
         user = user.withSessionId(session.getSessionId());
         auditService.submitAuditEvent(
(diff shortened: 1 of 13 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -685,6 +718,33 @@ private OrchSessionItem createNewOrchSession(String sessionId) {
         return newOrchSessionItem;
     }
 
+    private long getMaxAge(AuthenticationRequest authRequest) {
+        int maxAge;
+        try {
+            maxAge = authRequest.getMaxAge();
+        } catch (Exception e) {
+            LOG.error(
```
Ryan-Andrews99, line 726: Do we want this to be a warn instead of an error? Assuming this throwing is RP provided a junk value, in which case we could end up with noisy error alarms

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/QueryParamsAuthorizeValidator.java
```diff
@@ -137,6 +137,11 @@ public Optional<AuthRequestError> validate(AuthenticationRequest authRequest) {
                             redirectURI,
                             state));
         }
+        if (!client.getMaxAgeEnabled() && authRequest.getMaxAge() != -1) {
+            LOG.warn(
+                    "Max age present in auth request but not enabled for client, so max age will be ignored. Client ID: {}",
+                    client.getClientID());
+        }
```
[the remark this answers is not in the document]
billrdunn: Actually nimbus throws in `AuthenticationRequest.parse(requestParameters)` before we do this validation. But yes we can catch the exception there and return something more useful to the RP.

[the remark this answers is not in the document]
billrdunn: Turns out we already return INVALID_REQUEST when the request params can't be parsed, so nothing to do here I don't think

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -624,10 +625,21 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                    && maxAgeExpired(
+                            previousOrchSession.getAuthTime(), getMaxAge(authenticationRequest))) {
+                orchSession = updateOrchSessionDueToMaxAgeExpiry(newSessionId, previousOrchSession);
+            } else {
+                orchSession = updateOrchSession(newSessionId, previousOrchSession);
+            }
+        } else {
+            orchSession = createNewOrchSession(newSessionId);
+        }
+
```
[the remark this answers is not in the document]
billrdunn: Works for me! Have updated

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -685,6 +718,33 @@ private OrchSessionItem createNewOrchSession(String sessionId) {
...
+        int maxAge;
+        try {
+            maxAge = authRequest.getMaxAge();
+        } catch (Exception e) {
+            LOG.error(
+                    "Failed to retrieve max_age from auth request. Assuming value of -1 meaning 'not specified'.");
+            return -1;
+        }
+        return maxAge;
+    }
```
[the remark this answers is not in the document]
billrdunn: hmm not sure what you mean

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
billrdunn: RP providing a junk value should be caught earlier on. I can't imagine a case when this exception would be thrown but thought it would be good to add it just in case. Error logging so we're aware if max age isn't functioning as intended

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -658,24 +670,45 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                newSessionIdForPreviousSession);
+
+        OrchSessionItem newSession =
+                new OrchSessionItem(previousSession)
+                        .withSessionId(newSessionId)
+                        .withTimeToLive(timeNow + configurationService.getSessionExpiry())
+                        .withAuthenticated(false)
+                        .withPreviousSessionId(newSessionIdForPreviousSession);
+        orchSessionService.addSession(newSession);
+        LOG.info("Created new Orch session with session ID {}", newSessionId);
```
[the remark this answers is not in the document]
billrdunn: Updated :)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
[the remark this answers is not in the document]
Ryan-Andrews99: Hmm yeah bit of an unclear comment, sorry! 😅 I'm not really a fan of the nimbus method `getMaxAge` returning -1 when not present  and we're kind of leaning into it here. Was thinking if we used an optional, the output is a bit clearer/ doesn't require context that nimbus does something weird
```
 private Optional<Long> getMaxAge(AuthenticationRequest authRequest) {
        Optional<Long>  maxAge = Optional.empty();
        try {
            var maxAgeParam = authRequest.getMaxAge();
            //Nimbus returns -1 if maxAge parameter is not present
            if (maxAgeParam != -1) {
                maxAge = Optional.of((long) maxAgeParam);
            }
        } catch (Exception e) {
            LOG.error(
                    "Failed to parse MaxAge param form Auth request, assuming no max age parameter ");
        }
        return  maxAge;
    }
```

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
Ryan-Andrews99: Ah makes sense

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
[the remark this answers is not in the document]
billrdunn: Ah yeah i see what you mean, LGTM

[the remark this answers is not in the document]
billrdunn: Updated, changed a couple of things in your example. Also updated `maxAgeExpired()` 

and `if (authTime > timeNow)` check is new too

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/QueryParamsAuthorizeValidator.java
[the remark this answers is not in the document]
CarlyG55: We should, yes!

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -658,31 +680,85 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                        .withAuthenticated(false)
+                        .withPreviousSessionId(newSessionIdForPreviousSession);
+        orchSessionService.addSession(newSession);
+        LOG.info("Created new Orch session with session ID {} due to max age expiry", newSessionId);
+        return newSession;
+    }
+
+    private Optional<Long> getMaxAge(AuthenticationRequest authRequest) {
+        Optional<Long> maxAge = Optional.empty();
+        try {
```
CarlyG55, line 732: I wonder if we want to be pulling these methods out this file, as it's getting so long. I created an `AuthorisationHelper` file a little while back - might be nice to move them into there?

billrdunn: As discussed, will add this to tech debt in order to get this deployed and unblock other max age work

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/QueryParamsAuthorizeValidator.java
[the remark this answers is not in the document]
billrdunn: I thought this would Just Work but it doesn't, so looking into it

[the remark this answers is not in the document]
billrdunn: Works now :)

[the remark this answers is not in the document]
Ryan-Andrews99: Nice!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/Session.java
```diff
@@ -53,6 +53,22 @@ public Session(String sessionId) {
         initializeCodeRequestMap();
     }
 
+    public Session(Session session, String newSessionId) {
```
billrdunn, line 56: Not totally convinced on whether this is needed. Seems like it will be only used for one specific case. Could all these fields just be set in `copySessionForMaxAge` instead?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -609,26 +609,70 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
                     authenticationRequest.getClientID().getValue(),
                     user);
         }
-        var session = existingSession.orElseGet(sessionService::generateSession);
-        attachSessionIdToLogs(session);
 
+        Session session;
+        OrchSessionItem orchSession;
+        var maxAgeParam = getMaxAge(authenticationRequest);
```
billrdunn, line 615: Could this be in the `else`? As there's no point calculating it if the session is empty

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -658,31 +702,103 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                        .withTimeToLive(timeNow + configurationService.getSessionExpiry())
+                        .withCurrentCredentialStrength(null)
+                        .withAuthenticated(false)
+                        .withPreviousSessionId(newSessionIdForPreviousSession);
+        orchSessionService.addSession(newSession);
+
+        return newSession;
+    }
+
+    private Session updateSharedSessionDueToMaxAge(
```
billrdunn, line 751: `updateSharedSessionDueToMaxAgeExpiry` to match the other method?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -609,26 +609,70 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                                newSessionIdForPreviousSession,
+                                newSessionId);
+                orchSession =
+                        updateOrchSessionDueToMaxAgeExpiry(
+                                newSessionId,
+                                existingOrchSessionOptional.get(),
+                                timeNow,
+                                newSessionIdForPreviousSession);
+
+                LOG.info(
```
billrdunn, line 647: Does this logging cover both the shared and orch sessions? In which case might be better to use `newSessionId` instead of `orchSession.getSessionId()` and update the messages to make that clear

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -609,26 +609,70 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
...
+                        newSessionIdForPreviousSession);
+
+                LOG.info(
+                        "Created new session with session ID {} due to max age expiry",
+                        orchSession.getSessionId());
+
+            } else {
+                var sessionToBeUpdated = existingSession.get();
+                sessionService.updateWithNewSessionId(sessionToBeUpdated);
+                session = sessionToBeUpdated;
```
billrdunn, line 659: I'm OK with this but for readability feels a little odd to be using `sessionToBeUpdated` after it's been updated. Could replace `sessionToBeUpdated` with `session` on lines after 659

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
Ryan-Andrews99: Ooh yeah you're right

Ryan-Andrews99: Have moved this and the timeNow variables into the else block

isaac-GDS (dismissed):

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/Session.java
Ryan-Andrews99: Hmm would you prefer a _plain_ copy constructor with the fields being modified in the method instead?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
Ryan-Andrews99: Think I've updated this to do what you've described

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
Ryan-Andrews99: Updated this

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/Session.java
Ryan-Andrews99: Ah yeah I was um-ing and ah-ing about this bit - happy to change!

CarlyG55 (approved):

billrdunn (commented):
I can't approve as I'm a contributor but your changes LGTM

isaac-GDS (approved):

PR #5613, opened 2024-12-02
ATO-1255: Update sandpit client reg entry to have MaxAgeEnabled=true

billrdunn:
_second attempt after the first PR (https://github.com/govuk-one-login/authentication-api/pull/5609) was reverted_
## What

Update sandpit client reg entry to have MaxAgeEnabled=true. This will allow testing of the max age initiative in dev / sandpit without always having to manually update the client reg entry.


Ryan-Andrews99 (approved):
LGTM

PR #5623, opened 2024-12-04
ATO-975: update uplift to use the current credetial strength from the start request.

isaac-GDS:
##What

Pass the current credential strength from the start request instead of using the userContext.

This will be the final change to allow all the current credential strength be moved to split sessions.

Tested in sandpit
Auth Acceptance tests pass
Tested it deploys to authDev


billrdunn (approved):

alhcomer (approved):
LGTM

PR #5631, opened 2024-12-05
BAU: Update tags for 'core' data sources

whi-tw:
## What

We've changed the tagging for 'core', so when we search for resources by tag, we need to use these new tag values.

## How to review

1. Ensure core from https://github.com/alphagov/di-infrastructure/pull/607 is deployed in an environment
2. Deploy this pr
3. If it doesn't error, all is good

## Related PRs

Required to deploy at same time as https://github.com/alphagov/di-infrastructure/pull/607


mattvot (approved):
I've not deployed this, though changes LGTM

PR #5643, opened 2024-12-09
ATO-1100: copy client sessions

Ryan-Andrews99:
## What: 

- Adds max age handling to the authentication callback handler behind a feature flag
- This adds handling for the case in which the internalCommonSubjectIds match (ie we return the same user) after a max age journey
- This involves re-attatching the client sessions to their shared session 
- We also remove the previous session id from their session as this prevents subsequent journeys from having the same check the same check - we only want to do this if they have been on a max age journey

## How to review
- Deploy to dev and test:
- A journey without max age (establish an existing session)
- A journey forcing max age re-auth (max_age =0) 
- Observe logging around same user signed in 
- Do the same as above but sign in as different user
## Checklist

## Related PRs: 
- Branches off of https://github.com/govuk-one-login/authentication-api/pull/5608 so must be merged first

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -400,6 +400,12 @@ public class AuthenticationCallbackHandler
                     orchSession.setAuthTime(NowHelper.now().toInstant().getEpochSecond());
                 }
 
+                if (configurationService.supportMaxAgeEnabled()
+                        && Objects.nonNull(orchSession.getPreviousSessionId())) {
+                    LOG.info("Previous session id is present - handling max age");
+                    handleMaxAgeSession(orchSession, userSession);
+                }
+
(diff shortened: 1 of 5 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -789,4 +795,40 @@ private void addClaimsToOrchSession(OrchSessionItem orchSession, UserInfo userIn
...
+                    .getClientSessions()
+                    .forEach(currentSharedSession::addClientSession);
+
+        } else {
+            LOG.info(
+                    "Previous OrchSession InternalCommonSubjectId does not Auth UserInfo response");
+            // TODO: ATO-1101: Send backchannel logouts + audit events
+
+        }
+        // Remove previousSessionId to prevent a max_age doom loop
```
isaac-GDS, line 831: To be removed

isaac-GDS (approved):

isaac-GDS (approved):

PR #5644, opened 2024-12-09
BAU: Migrate all lambdas to endpoint-module-v2

whi-tw:
## What

`endpoint-module` is quite inefficient, as it does a number of data queries internally. Additionally, all of its' logic is implemented in `endpoint-lambda`.

I've migrated all our lambdas over to use `endpoint-module-v2`.

This new module has the following benefits:

- It does not do data queries internally, instead requiring these values to be passed in. This reduces the time for apply, as we will no longer needlessly re-request the same information over and over again
- It uses `endpoint-lambda` as the lambda-creation engine, which means we have configuration in only one place
- It exposes the endpoints required to create a openapi-backed API Gateway in future.

Additionally, I've stopped looking up the account alias and the slack SNS topic within each component, instead retrieving them from shared state. This further reduces the number of `data` lookups.

Finally, I've shifted `dynatrace.tf` out of the individual components, and into the parent directory, and have symlinked it in to the individual directories, removing the possibility of different modules getting out of sync.
This is the same pattern as we use with `versions.tf`

## How to review

- Code Review
- Deploy it to a dev environment and check it works okay


ci/terraform/modules/endpoint-lambda/alerts.tf
```diff
@@ -17,6 +17,13 @@ resource "aws_cloudwatch_log_metric_filter" "lambda_error_metric_filter" {
   }
 }
 
+locals {
+  base_error_alarm_description      = "${var.lambda_log_alarm_threshold} or more errors have occurred in the ${var.environment} ${var.endpoint_name} lambda.ACCOUNT: ${var.account_alias}"
+  error_alarm_description           = var.runbook_link == null ? local.base_error_alarm_description : "${local.base_error_alarm_description}. Runbook: ${var.runbook_link}"
+  base_error_rate_alarm_description = "Lambda error rate of ${var.lambda_log_alarm_error_rate_threshold} has been reached in the ${var.environment} ${var.endpoint_name} lambda.ACCOUNT: ${var.account_alias}"
+  error_rate_alarm_description      = var.runbook_link == null ? local.base_error_rate_alarm_description : "${local.base_error_rate_alarm_description}. Runbook: ${var.runbook_link}"
+}
+
```
pskushwaha1, line 26: These Locals are already defined  on line no 1 Throwing exception

│ A local value named "base_error_alarm_description" was already defined at
│ ../modules/endpoint-lambda/alerts.tf:2,3-187. Local value names must be
│ unique within a module.

whi-tw: oops. fixed

ci/terraform/auth-external-api/shared.tf
```diff
@@ -32,4 +32,7 @@ locals {
   redis_ssm_parameter_policy                  = data.terraform_remote_state.shared.outputs.redis_ssm_parameter_policy
   authentication_oidc_redis_security_group_id = data.terraform_remote_state.shared.outputs.authentication_oidc_redis_security_group_id
   redis_key                                   = "session"
+
```
mattvot, line 35: `local.dynatrace_secret` is referenced in `ci/terraform/auth-external-api/token.tf` but not in this list of locals in `ci/terraform/auth-external-api/shared.tf`. Am I missing something?

ci/terraform/account-management/shared.tf
```diff
@@ -19,4 +19,7 @@ locals {
   client_registry_encryption_key_arn          = data.terraform_remote_state.shared.outputs.client_registry_encryption_key_arn
   user_profile_kms_key_arn                    = data.terraform_remote_state.shared.outputs.user_profile_kms_key_arn
   email_check_results_encryption_policy_arn   = data.terraform_remote_state.shared.outputs.email_check_results_encryption_policy_arn
+
+  slack_event_sns_topic_arn = data.terraform_remote_state.shared.outputs.slack_event_sns_topic_arn
```
mattvot, line 23: As previous, potentially the same missing `dynatrace_secret` here too.

ci/terraform/delivery-receipts/shared.tf
```diff
@@ -14,4 +14,7 @@ locals {
   bulk_user_email_table_encryption_key_arn = data.terraform_remote_state.shared.outputs.bulk_user_email_table_encryption_key_arn
   user_profile_encryption_policy_arn       = data.terraform_remote_state.shared.outputs.user_profile_encryption_policy_arn
   user_profile_kms_key_arn                 = data.terraform_remote_state.shared.outputs.user_profile_kms_key_arn
+
+  slack_event_sns_topic_arn = data.terraform_remote_state.shared.outputs.slack_event_sns_topic_arn
```
mattvot, line 18: As previous, potentially the same missing `dynatrace_secret` here too.

Fairly confident I'm missing something in my understanding of terraform locals.

ci/terraform/auth-external-api/shared.tf
mattvot: Ok, I think I understand a bit more now. You can reference multiple locals blocks in different terraform files within the same module and they all get merged into one set of locals available across the module.

So in this case, what I see is missing dynatrace_secret is actually in the symlinked ci/terraform/dynatrace.tf file

ci/terraform/account-management/shared.tf
mattvot: https://github.com/govuk-one-login/authentication-api/pull/5644#discussion_r1903942594

ci/terraform/delivery-receipts/shared.tf
mattvot: https://github.com/govuk-one-login/authentication-api/pull/5644#discussion_r1903942594

mattvot (commented):
Looked through the PR and happy with nearly all the code changes here. Just getting myself acquainted with the differences between endpoint module v1 and v2 as defined.

mattvot:
OK, I've gone though a diff of the `endpoint-module` and `endpoint-module-v2` and understand the changes. Looks OK to me

pskushwaha1 (approved):

mattvot (approved):

PR #5659, opened 2024-12-11
ATO-1245: refactor session setting in start

Ryan-Andrews99:
## What: 

- Refactors how the session is setup in the StartHandler to remove the use of `addOrUpdateSessionIncludingSessionId` 
- Replaces its usage in the StartHandler with `getUpdatedPreviousSessionOrReturnNew` and removes it's usage in the integration tests.
## How to review

-  Deploy to dev env (sandpit)
   - Run auth acceptance tests against changes (passed)

## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -17,6 +17,7 @@ import uk.gov.di.authentication.frontendapi.helpers.ReauthMetadataBuilder;
 import uk.gov.di.authentication.frontendapi.services.StartService;
 import uk.gov.di.authentication.shared.domain.CloudwatchMetrics;
 import uk.gov.di.authentication.shared.entity.*;
+import uk.gov.di.authentication.shared.exceptions.ClientNotFoundException;
 import uk.gov.di.authentication.shared.helpers.DocAppSubjectIdHelper;
 import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.ReauthAuthenticationAttemptsHelper;
@@ -36,7 +37,6 @@ import uk.gov.di.authentication.shared.services.SessionService;
 
 import java.util.ArrayList;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 
@@ -172,6 +172,18 @@ public class StartHandler
                                         CLIENT_SESSION_ID_HEADER,
                                         configurationService.getHeadersCaseInsensitive()));
             }
+            var client = startService.getClient(clientSession.get());
+            var upliftRequired =
+                    startService.isUpliftRequired(
+                            clientSession.get(), client, startRequest.currentCredentialStrength());
+
+            var authSession =
+                    authSessionService.getUpdatedPreviousSessionOrCreateNew(
+                            Optional.ofNullable(startRequest.previousSessionId()),
+                            session.getSessionId(),
+                            startRequest.currentCredentialStrength());
+
+            authSessionService.addSession(authSession.withUpliftRequired(upliftRequired));
 
             var userContext = startService.buildUserContext(session, clientSession.get());
 
@@ -203,18 +215,6 @@ public class StartHandler
                     userContext.getUserProfile().map(UserProfile::getSubjectID);
             Optional<String> maybeInternalCommonSubjectIdentifier =
                     Optional.ofNullable(session.getInternalCommonSubjectIdentifier());
-
-            CredentialTrustLevel currentCredentialStrength =
-                    startRequest.currentCredentialStrength();
-
-            boolean upliftRequired =
-                    startService.isUpliftRequired(userContext, currentCredentialStrength);
-
-            authSessionService.addOrUpdateSessionIncludingSessionId(
-                    Optional.ofNullable(startRequest.previousSessionId()),
-                    session.getSessionId(),
-                    currentCredentialStrength,
-                    upliftRequired);
 
             var clientSessionId =
                     getHeaderValueFromHeaders(
@@ -295,7 +295,7 @@ public class StartHandler
             var errorMessage = "Unable to serialize start response";
             LOG.error(errorMessage, e);
             return generateApiGatewayProxyResponse(400, errorMessage);
-        } catch (NoSuchElementException e) {
+        } catch (ClientNotFoundException e) {
             return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1015);
         } catch (ParseException e) {
             return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1038);
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
@@ -226,16 +226,16 @@ public class StartService {
     }
 
     public boolean isUpliftRequired(
-            UserContext userContext, CredentialTrustLevel currentCredentialStrength) {
-        if (DocAppUserHelper.isDocCheckingAppUser(userContext)
+            ClientSession clientSession,
+            ClientRegistry client,
+            CredentialTrustLevel currentCredentialStrength) {
+        if (DocAppUserHelper.isDocCheckingAppUser(
+                        clientSession.getAuthRequestParams(), Optional.of(client))
                 || Objects.isNull(currentCredentialStrength)) {
             return false;
         }
         return (currentCredentialStrength.compareTo(
-                        userContext
-                                .getClientSession()
-                                .getEffectiveVectorOfTrust()
-                                .getCredentialTrustLevel())
+                        clientSession.getEffectiveVectorOfTrust().getCredentialTrustLevel())
                 < 0);
     }
 
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
@@ -28,7 +28,6 @@ import java.net.URI;
 import java.util.Collection;
 import java.util.List;
 import java.util.Map;
-import java.util.NoSuchElementException;
 import java.util.Objects;
 import java.util.Optional;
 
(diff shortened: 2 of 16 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -362,4 +366,40 @@ private void emitReauthRequestedObservability(
                 CloudwatchMetrics.REAUTH_REQUESTED.getValue(),
                 Map.of(ENVIRONMENT.getValue(), configurationService.getEnvironment()));
     }
+
+    private AuthSessionItem getUpdatedPreviousSessionOrCreateNew(
```
BeckaL, line 370: Drive by question: what's the value of having this method here rather than in the Auth Session Service? I kind of liked that it was tidied away there

Ryan-Andrews99: The method is sort of the same as the previous one except that it returns an updated session object and doesn't immediately write to the table (allowing it to be updated before writing in a separate call). The main reason I moved it out of the service is that it felt very Start handler specific and I didn't want it being reused (like `addOrUpdateSessionId` was in the integration tests). Not super wedded either way so happy to move back into the service if you'd prefer

Ryan-Andrews99: Came back to this a bit late, but think this method is actually better in the service so have updated this!

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -79,79 +78,55 @@ void updateSessionThrowsAnyDynamoExceptions() {
...
-        verify(table).putItem(captor.capture());
-        AuthSessionItem updatedItem = captor.getValue();
+        var updatedSession =
+                authSessionService.getUpdatedPreviousSessionOrCreateNew(
+                        Optional.of(SESSION_ID), NEW_SESSION_ID, null);
 
-        assertThat(updatedItem.getSessionId(), is(NEW_SESSION_ID));
-        assertTrue(updatedItem.getTimeToLive() > Instant.now().getEpochSecond());
+        assertThat(updatedSession.getSessionId(), is(NEW_SESSION_ID));
+        assertTrue(updatedSession.getTimeToLive() > Instant.now().getEpochSecond());
```
GHSwallow, line 102: I know this isn't new code, but the existingSession would also pass this assertion. I guess it's just making sure the session is still valid which isn't a bad thing.

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -79,79 +78,55 @@ void updateSessionThrowsAnyDynamoExceptions() {
...
-
-        assertThat(savedItem.getSessionId(), is(NEW_SESSION_ID));
-        assertTrue(savedItem.getTimeToLive() > Instant.now().getEpochSecond());
+        verifyNoInteractions(dynamoDbClient);
+        assertThat(newSession.getSessionId(), is(NEW_SESSION_ID));
+        assertTrue(newSession.getTimeToLive() > Instant.now().getEpochSecond());
     }
 
     @Test
     void shouldPutAndDeleteSessionWhenUpdatingSessionId() {
```
GHSwallow, line 94: this is no longer `shouldPut`, just `shouldDelete`

GHSwallow (commented):
looks great! 🚀 Mostly nits tbh, but a few important questions

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -45,6 +45,13 @@ void setup() {
         authSessionService = new AuthSessionService(dynamoDbClient, table, configurationService);
     }
 
+    @Test
+    void generatesANewSessionWithExpectedDefaults() {
+        var session = authSessionService.generateNewAuthSession(SESSION_ID);
+        assertEquals(SESSION_ID, session.getSessionId());
+        assertEquals(AuthSessionItem.AccountState.UNKNOWN, session.getIsNewAccount());
```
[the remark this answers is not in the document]
Ryan-Andrews99: Good point - will add to the tests in this file

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -79,79 +78,55 @@ void updateSessionThrowsAnyDynamoExceptions() {
...
+                authSessionService.getUpdatedPreviousSessionOrCreateNew(
+                        Optional.empty(), NEW_SESSION_ID, null);
 
-        ArgumentCaptor<AuthSessionItem> captor = ArgumentCaptor.forClass(AuthSessionItem.class);
-        verify(table).putItem(captor.capture());
-        AuthSessionItem savedItem = captor.getValue();
-
-        assertThat(savedItem.getSessionId(), is(NEW_SESSION_ID));
-        assertTrue(savedItem.getTimeToLive() > Instant.now().getEpochSecond());
+        verifyNoInteractions(dynamoDbClient);
```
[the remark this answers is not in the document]
Ryan-Andrews99: Nope we're not providing a previousSessionId so as far as the code is concerned there is no previous session to gert/delete and it returns a new session object instead

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -79,79 +78,55 @@ void updateSessionThrowsAnyDynamoExceptions() {
     }
 
     @Test
-    void shouldAddNewSessionWhenNoPreviousSessionGiven() {
```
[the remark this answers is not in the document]
Ryan-Andrews99: That is a good point, have amended!

shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
```diff
@@ -59,50 +59,37 @@ public void addSession(AuthSessionItem authSessionItem) {
...
+            LOG.info(
+                    "Existing Auth session updated from previousSessionId: {}, sessionId: {}",
+                    previousSessionId,
+                    newSessionId);
+
+            return updatedSession;
+        } else {
+            LOG.info("New Auth session item created with sessionId: {}", newSessionId);
+            return generateNewAuthSession(newSessionId)
+                    .withCurrentCredentialStrength(currentCredentialStrength);
```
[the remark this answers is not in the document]
Ryan-Andrews99: Good spot on the try/catch, have reintroduced. The first point not massively sure, I do have a preference for these two steps being in one method call as it means we don't have to handle this `isPreviousSessionThere` question twice. I'm not convinced deleting and saving later on is worthwhile especially as if we're erroring at this step, I would imagine something quite odd was happening

shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
Ryan-Andrews99: Hmm good point, not sure the best assertions to add here. Have added a couple of more assertions but not sure what the best thing to do here is

integration-tests/src/test/java/uk/gov/di/authentication/services/AuthSessionServiceIntegrationTest.java
```diff
@@ -75,24 +83,28 @@ void shouldStoreAnUpdatedSession() {
     }
 
     @Test
-    void shouldRetainAPreviousSessionsValuesWhenSessionIdIsUpdated() {
+    void shouldReturnAPreviousSessionWithRetainedValues() {
         withPreviousSession();
 
         var previousSession = authSessionExtension.getSession(PREVIOUS_SESSION_ID).get();
         previousSession.setIsNewAccount(AuthSessionItem.AccountState.EXISTING);
         authSessionExtension.updateSession(previousSession);
```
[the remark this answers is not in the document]
Ryan-Andrews99: Yep seems sensible to me, have updated!

shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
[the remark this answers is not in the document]
GHSwallow: What about if you instead refactored `isUpliftRequired` so that instead of taking in the whole UserContext object, it just takes in ClientSession? Or even just `credentialTrustLevel`? That way, we could get `isUpliftRequired` higher up in this handler, then update the session, then build the UserContext with that session?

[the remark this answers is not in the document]
Ryan-Andrews99: Have pushed up some changes which do some small refactoring like we discussed, lemme know what you think 😄

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -239,6 +235,28 @@ public boolean isUpliftRequired(
...
+                        clientSession.getEffectiveVectorOfTrust().getCredentialTrustLevel())
+                < 0);
+    }
+
+    public ClientRegistry getClient(ClientSession clientSession) {
+        var clientId =
+                clientSession.getAuthRequestParams().get(CLIENT_ID_PARAM).stream()
+                        .findFirst()
+                        .orElseThrow();
+        return clientService.getClient(clientId).orElseThrow();
```
GHSwallow, line 257: I think this should have more to it that just throwing, eg a specific exception, a message etc. Looks like there's already an exception `ClientNotFoundException`. Tbh though, this is not your code, and I'm getting a bit caught up in cleaning up the code as we go, so I'm happy for you to leave it as is.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -239,6 +235,28 @@ public boolean isUpliftRequired(
                 < 0);
     }
 
+    public boolean isUpliftRequired(
```
GHSwallow, line 238: The above method with signature `isUpliftRequired(UserContext userContext, CredentialTrustLevel currentCredentialStrength)` can use this method to save repeated code

integration-tests/src/test/java/uk/gov/di/authentication/services/AuthSessionServiceIntegrationTest.java
```diff
@@ -106,15 +116,8 @@ void shouldGetSessionFromRequestHeaders() {
...
-
-    private void withPreviousSession() {
-        authSessionExtension.addSession(Optional.empty(), PREVIOUS_SESSION_ID);
-    }
-
-    private void withUpdatedPreviousSession() {
-        authSessionExtension.addSession(Optional.of(PREVIOUS_SESSION_ID), SESSION_ID);
+    private AuthSessionItem withStoredSession(String sessionId) {
+        authSessionExtension.addSession(sessionId);
+        return authSessionExtension.getSession(sessionId).get();
```
GHSwallow, line 121: I love this, it's so much cleaner!

GHSwallow (approved):
Only one comment to action, maybe. As I said, I might be getting too bogged down in cleaning up the code, I leave it to you to decide whether it's worth actioning. Approving if not:)

GHSwallow (approved):

PR #5679, opened 2024-12-17
ATO-1036: Update sequence diagrams to show AIS calls

rpjgds:
## What

Sequence diagrams updated to show where AIS is called in the auth-only and identity journeys.

## How to review

Has the happy-path AIS call been documented correctly?
Do you agree that AIS calls should be represented in these happy-path diagrams?

## Checklist

N/A

### DIFF
--- a/docs/diagrams/orchestration/auth-only/auth-only.mmd
+++ b/docs/diagrams/orchestration/auth-only/auth-only.mmd
@@ -2,6 +2,7 @@ sequenceDiagram
     participant rp as Relying Party
     participant orch_api as Orchestration OIDC API
     participant auth as Authentication
+    participant ais as AIS
     rp ->> orch_api : OIDC Authentication request to /authorize via browser (AuthorisationHandler.java)
     orch_api  ->> orch_api : Validates Authentication request, creates session cookie
     orch_api ->> orch_api : Generates a signed and encrypted OAuth 2.0 Authorization request
@@ -12,6 +13,7 @@ sequenceDiagram
     auth ->> orch_api : Token response containing access token
     orch_api ->> auth : Userinfo request containing access token to /userinfo (UserInfoHandler.java)
     auth ->> orch_api : Userinfo response
+    orch_api -->> ais : Check for interventions
     orch_api ->> orch_api : Generates and store Authorization Code with 5 min TTL
     orch_api ->> rp : 302 redirect containing Authentication Response to the RPs redirect URL
     rp ->> orch_api : Token request to /token (TokenHandler.java)
@@ -19,4 +21,4 @@ sequenceDiagram
     orch_api ->> rp : Token response containing access and id tokens
     rp ->> orch_api : UserInfo request to /userinfo (UserInfoHandler.java)
     orch_api ->> orch_api : Validates Access Token
-    orch_api ->> rp : UserInfo response containing user information 
+    orch_api ->> rp : UserInfo response containing user information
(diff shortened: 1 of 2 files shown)


docs/diagrams/orchestration/auth-only/auth-only.mmd
```diff
@@ -12,11 +13,12 @@ sequenceDiagram
     auth ->> orch_api : Token response containing access token
     orch_api ->> auth : Userinfo request containing access token to /userinfo (UserInfoHandler.java)
     auth ->> orch_api : Userinfo response
+    orch_api <<->> ais : Check for interventions
```
billrdunn, line 16: This symbol `<<->>` doesn't render properly for me locally

docs/diagrams/orchestration/identity/identity.mmd
```diff
@@ -5,6 +5,7 @@ sequenceDiagram
     participant auth as Authentication
     participant ipv as IPV Core
     participant spot as SPOT
+    participant ais as AIS
```
billrdunn, line 8: There's one futher call to AIS needed in this diagram - after the response from IPV

docs/diagrams/orchestration/identity/identity.mmd
```diff
@@ -17,6 +18,7 @@ sequenceDiagram
     auth ->> orch_api : Token response containing access token
     orch_api ->> auth : UserInfo request containing access token to /userinfo
     auth ->> orch_api : UserInfo response
+    orch_api <<->> ais : Check for interventions
```
billrdunn, line 21: Same comment as above

docs/diagrams/orchestration/identity/identity.mmd
```diff
@@ -33,6 +35,7 @@ sequenceDiagram
     spot -) orch_api : Return a signed Core Identity JWT
     orch_api ->> orch_api : Store signed Core Identity JWT
     auth ->> orch_api : Poll for updates
+    orch_api <<->> ais : Check for interventions
```
billrdunn, line 38: Same as above

docs/diagrams/orchestration/auth-only/auth-only.mmd
rpjgds: I made them all a dotted single ended line, the double-ender doesn't render in markdown <11.

docs/diagrams/orchestration/identity/identity.mmd
billrdunn: Ah yes you're right - we only call getAccountIntervention in IPVCallback if there's an error and we're not going to SPOT, so non happy-path

docs/diagrams/orchestration/identity/identity.mmd
```diff
@@ -33,6 +35,7 @@ sequenceDiagram
     spot -) orch_api : Return a signed Core Identity JWT
     orch_api ->> orch_api : Store signed Core Identity JWT
     auth ->> orch_api : Poll for updates
+    orch_api -->> ais : Check for interventions
```
billrdunn, line 38: Should this come after the COMPLETED response?

rpjgds: It looks like Orch check for interventions before returning the COMPLETED response back to Auth.

The response is here https://github.com/govuk-one-login/authentication-api/blob/7f26ad1d332654ec038ad26ba51591a3c868c6fa/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/ProcessingIdentityHandler.java#L194 - AIS is checked on L181 above

billrdunn (approved):

PR #5683, opened 2024-12-18
ATO-1097: Return auth_time in ID token

billrdunn:
## What

As part of the max age initiative, Orch need to return the auth_time claim in the ID token to RPs. The claim is retrieved from the auth code exchange data store, and then added to all token responses (not just when the RP includes max age in the authorize request).

Implementation is feature flagged and enabled in all envs except production. As this change is RP facing, it needs to be tested in integration and RPs made aware of the changes before releasing to production.

### DIFF
--- a/orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/IntegrationTest.java
+++ b/orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/IntegrationTest.java
@@ -346,5 +346,10 @@ public class IntegrationTest {
         public boolean supportMaxAgeEnabled() {
             return true;
         }
+
+        @Override
+        public boolean isReturnAuthTimeInIdTokenEnabled() {
+            return false;
+        }
     }
 }
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenService.java
@@ -92,7 +92,8 @@ public class TokenService {
             boolean isDocAppJourney,
             JWSAlgorithm signingAlgorithm,
             String journeyId,
-            String vot) {
+            String vot,
+            Long authTime) {
         List<String> scopesForToken = authRequestScopes.toStringList();
         AccessToken accessToken =
                 segmentedFunctionCall(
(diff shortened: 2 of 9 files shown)


orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/IntegrationTest.java
```diff
@@ -341,5 +341,10 @@ public Optional<String> getIPVCapacity() {
         public boolean isDestroyOrchSessionOnSignOutEnabled() {
             return true;
         }
+
+        @Override
+        public boolean isReturnAuthTimeInIdTokenEnabled() {
+            return false;
```
billrdunn, line 347: Note I've deliberately set this `false` here and in unit tests to match production. When the feature flag is enabled in prod the tests will need to be updated.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenService.java
```diff
@@ -249,7 +249,9 @@ private SignedJWT generateIDToken(
             idTokenClaims.setClaim("vot", vot);
         }
         idTokenClaims.setClaim("vtm", trustMarkUri.toString());
-        idTokenClaims.setClaim("auth_time", authTime);
+        if (configService.isReturnAuthTimeInIdTokenEnabled()) {
+            idTokenClaims.setClaim("auth_time", authTime);
+        }
```
Ryan-Andrews99, line 254: Do we not want to put this inside the ` if (!isDocAppJourney)` check above, as we're not returning this claim for docapp users are we?

template.yaml
```diff
@@ -72,6 +72,13 @@ Conditions:
       !Equals [build, !Ref Environment],
       !Equals [staging, !Ref Environment],
     ]
+  IsReturnAuthTimeInIdTokenEnabled:
+    !Or [
+      !Equals [dev, !Ref Environment],
+      !Equals [build, !Ref Environment],
+      !Equals [staging, !Ref Environment],
+      !Equals [integration, !Ref Environment],
```
Ryan-Andrews99, line 80: Do we want to enable this in integration straight away? Just considering it's an RP facing environment so might want to give them some heads up before we enable it. What do you think?

billrdunn: I was under the impression we'd go straight to integration but I'll check

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenService.java
billrdunn: Shouldn't matter as if it's a Doc App journey then auth_time is `null` but I'll do it anyway

billrdunn: Done

Ryan-Andrews99 (approved):

PR #5715, opened 2025-01-07
AUT-3814: Script to create access token for hitting the account management API in lower envs

mattvot:
## What

This new script, run locally, will create auth codes for lower environment user profiles. We will be able to use these codes to directly hit the account management APIs for the user.

When provided an environment and email address the script will calculate the internal common subject id, create a JWT with the correct payload for the auth token and sign the JWT with the KMS key used in the environment.

The token expires in 12 hours, allowing a useful but not persistent time window to use it with lower environments.

The IntelliJ HTTP client settings for account management now have dev environments configured, with each baseUrl provided, making using it slightly quicker.

## How to review

1. Code Review
1. `./generate-auth-token.sh dev <email-address>`
1. Use IntelliJ IDEA to open `http/account-management.http` and follow instructions to make a request to change a password

### DIFF
--- a/scripts/export-ics.sh
+++ b/scripts/export-ics.sh
@@ -9,7 +9,11 @@ fi
 
 export AWS_REGION=eu-west-2
 
-sector="identity.$2.account.gov.uk"
+if [[ $2 == authdev* ]]; then
+  sector="identity.$2.sandpit.account.gov.uk"
+else
+  sector="identity.$2.account.gov.uk"
+fi
 
 echo -e "Exporting internalCommonSubjectId for Email = $1 Environment = $2 Sector = ${sector}"
 


http/account-management.http
```diff
@@ -1,12 +1,12 @@
 # Please read guidance on confluence before use
 # (https://govukverify.atlassian.net/wiki/spaces/LO/pages/4606132332/Manually+Testing+Account+Management+API).
-# Populate these values before use. Only do manual testing in lower environments (sandpit, authdev1, authdev2).
-@base-url =
-@bearer-token =
+# Populate the authToken value below use (using scripts/generate-auth-token.sh)
+
+@authToken = unset
```
alhcomer, line 6: Should there be a @baseUrl field here like with @authToken also set to 'unset'? (just to serve as a prompt to do it) Also a mention within the documentation of where to get the base url might be useful

mattvot: @baseurl comes from the env file that sits next to this file and doesn't need stating here.

Good shout on the docs, I can update the page to mention where the URL comes from

dbes-gds (approved):

PR #5736, opened 2025-01-13
ATO-1117: add user info to ipv callback handler

GHSwallow:
This was previously merged here https://github.com/govuk-one-login/authentication-api/pull/5491, but it caused some errors. I've investigated it here https://github.com/govuk-one-login/authentication-api/pull/5716 and here https://github.com/govuk-one-login/authentication-api/pull/5730 and have discovered the issue, and importantly _why_ it happens. It is not a problem with our code - is it the new structure we have, combined with users trying to access parts of the journey they shouldn't be. The second to last commit fixes the issue, and the final commit removes the investigation logs.

## What
Adds authUserInfo to the IPVCallbackHandler. Read access has been added here https://github.com/govuk-one-login/authentication-api/pull/5475.

I first want to verify that authUserInfo is both defined, and that the values are as expected. This PR therefore just adds some logging to compare the values used in this handler and the corresponding values in the orchSession and the authUserInfo table entry.

## How to review
1. Code Review

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/IPVCallbackHandlerIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/IPVCallbackHandlerIntegrationTest.java
@@ -95,12 +95,16 @@ class IPVCallbackHandlerIntegrationTest extends ApiGatewayHandlerIntegrationTest
     private static final String EMAIL = "joe.bloggs@digital.cabinet-office.gov.uk";
     private static final String REDIRECT_URI = "http://localhost/redirect";
     private static final String TEST_EMAIL_ADDRESS = "test@test.com";
-    private static final Subject INTERNAL_SUBJECT = new Subject();
     private static final String SESSION_ID = "some-session-id";
     public static final String CLIENT_NAME = "test-client-name";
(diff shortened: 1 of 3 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/IPVCallbackHandlerIntegrationTest.java
```diff
@@ -555,6 +568,29 @@ private void setupOrchSession(String internalCommonSubjectId) {
...
+                                        "vtm",
+                                        OIDC_BASE_URL + "/trustmark",
+                                        "email",
+                                        TEST_EMAIL_ADDRESS,
+                                        "phone_number",
+                                        TEST_PHONE_NUMBER,
+                                        "salt",
+                                        base64EncodedSalt,
+                                        "local_account_id",
+                                        TEST_SUBJECT.getValue())));
```
Ryan-Andrews99, line 590: Think the `vot` and `vtm` claims aren't present on Auth User info (think they're on the IPV one)

Ryan-Andrews99 (commented):
One comment about test data, should be a super quick fix

PR #5746, opened 2025-01-14
BAU: fix integer max age in request object

Ryan-Andrews99:
### Wider context of change:

As part of the max_age feature, RPs can provide the `max_age` claim in the request object. Our previous code assumed that this would be a string integer (ex: "123") , however it is plausible this could be a integer claim too, and we should support both. The OIDC spec includes an integer max_age in the [example for request objects](https://openid.net/specs/openid-connect-core-1_0.html#:~:text=The%20following%20is%20a%20non%2Dnormative%20example%20of%20the%20Claims%20in%20a%20Request%20Object%20before%20base64url%2Dencoding%20and%20signing%3A), however is not specific on [max_age being a number.](https://openid.net/specs/openid-connect-core-1_0.html#:~:text=OPTIONAL.%20Maximum%20Authentication,prompt%3Dlogin.)

Previously the code would throw if the request object included a number for max_age, this corrects that and supports both. 

### What’s changed:

- Numerical max_age claims in request objects are now supported alongside string representations
- Some refactoring of the requestObject transform method to persist the max_age claim in the top level auth request. This makes dealing with the max_age claim easier in subsequent code.   
- Simplification of some the handler code due to above refactoring

### Manual testing

Deploy to dev
- Run through a max_age reauth journey with number max_age formatting on the stub (deploying this branch: https://github.com/govuk-one-login/relying-party-stub/tree/refs/heads/ATO-000-testing-max-age-as-number-not-string)
- Revert stub changes and test string formatted max_age claim 

### Checklist

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
@@ -1087,7 +1087,7 @@ class AuthorisationIntegrationTest extends ApiGatewayHandlerIntegrationTest {
     }
 
     @Test
-    void shouldUpdateSharedSessionWhenMaxAge() throws Exception {
+    void shouldUpdateSharedSessionWhenMaxAgeExpired() throws Exception {
         registerClient(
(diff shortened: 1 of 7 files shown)


CarlyG55 (approved):
This is really nice, thanks 🙂

PR #5759, opened 2025-01-15
ATO-1134: destroy orch session feature flag cleanup

GHSwallow:
### Wider context of change
This is part of the work to destroy orch session on sign out. There was some special complexity in the ProcessingIdentityHandler, as this was on Auth accounts but needed cross account access to delete from the Orch owned OrchSesssionItem table.

### What’s changed
Deletes the now redundant feature flag

### Manual testing
no need

### Checklist


template.yaml
```diff
@@ -60,13 +60,6 @@ Conditions:
       ],
     ]
   EnableFetchJwks:
```
Ryan-Andrews99, line 62: Looks like this might be accidentally enabling `FetchJwks` in prod

GHSwallow: omg good spot!

Ryan-Andrews99 (dismissed):

Ryan-Andrews99 (approved):

PR #5760, opened 2025-01-15
ATO-1332: Enable feature flag for getting browserSessionId from orch session up to production

cearl1:
### Wider context of change

As part of the migration from auth to orch, we would like to remove the getter for browserSessionId from the shared session. We've already added a feature flag to enable using the orch session to store the browserSessionId, and have tested it works in dev and build.

### What’s changed

This PR enables the feature flag mentioned above in staging, integration, and production.


GHSwallow (approved):
Nice! Merge tomorrow morning, as it's gone 4pm today

PR #5763, opened 2025-01-16
AUT-3895: Correct and improve IPV token call.

andrew-moores:
## What

The request for a token to access the result of a re-verification should contain the original redirect_uri provided in the authorise request.  We were providing the redirect_uri that belongs to orch rather than the auth one.

This was causing the IPV token endpoint to not respond to requests. The ReverificationResult lambda did not have any timeout configuration on the token request and was hanging.  The frontend does have a timeout so we were seeing a timeout error in the frontend.  The Nimbus library httpRequest method defaults to not having a timeout for connect or read so an over-ride has been added to timeout connections and reads if IPV does not respond.

## How to review
1. Code Review

## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/ReverificationResultService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/ReverificationResultService.java
@@ -103,6 +103,12 @@ public class ReverificationResultService {
             count++;
             try {
                 var httpRequest = tokenRequest.toHTTPRequest();
+
+                LOG.info("Sending IPV token request to {}", httpRequest.getURI());
+
+                httpRequest.setConnectTimeout(1000);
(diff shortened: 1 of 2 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/ReverificationResultService.java
```diff
@@ -103,6 +103,12 @@ public HTTPResponse toHTTPResponse() {
             count++;
             try {
                 var httpRequest = tokenRequest.toHTTPRequest();
+
+                LOG.info("Sending IPV token request to {}", httpRequest.getURI());
```
BeckaL, line 107: I'm not super bothered either way, is this a useful log to leave in?

BeckaL (approved):
Minor comment about logging but approving in case you want to leave it in 👍 otherwise happy to do another quick tick if you want to take it out

PR #5771, opened 2025-01-17
AUT-3895: Tidying up environment files.

andrew-moores:
## What

We have two terraform files that provide environment specific values. In addition, some of the environment variables are shared with orchestration and some are just for auth.  This commit does a minor re-organisation of the variables so they are in a section where their use is clearer and grouped together by feature.

Introduced a tfvar for the issuer claim used provided to IPV.  This is highlighted as non-standard by having the _IPV suffix.  IPV currently expect a string rather than the standard URI.  EVCS expect a URI so the name of the variable we use for this can be generic for auth as we would use the URI as our issuer claim in all request claim sets.

## How to review
1. Code Review
2. Deploy to dev using GHA and test

## Checklist


ci/terraform/oidc/build-overrides.tfvars
```diff
@@ -1,25 +1,30 @@
...
+doc_app_authorisation_callback_uri = "https://oidc.build.account.gov.uk/doc-app-callback"
+doc_app_encryption_key_id          = "dcmaw-cri-stubs-build-INqHBvMYWmNodklvbpTCgf1DS10Fv5ic4_8LdoBNjAw"
+doc_app_cri_data_endpoint          = "credentials/issue"
+doc_app_jwks_endpoint              = "https://dcmaw-cri.build.stubs.account.gov.uk/.well-known/jwks.json"
+#ipv_authorisation_client_id
+#ipv_authorisation_callback_uri
+internal_sector_uri = "https://identity.build.account.gov.uk"
+spot_enabled        = false
+#test_clients_enabled
+ipv_no_session_response_enabled             = true
```
dbes-gds, line 19: this was not switched on in build before

PR #5776, opened 2025-01-20
AUT-3924: cross browser

ethanmills:
## What

The second half of the backend changes for cross browser mitigation in the mfa reset journey.

This change creates a new endpoint to retrieve the information stored at the start of the journey in the case of cross browser issue. It also emits an audit event

## How to review
Commit-wise

## Checklist

## Related PRs

---


mattvot (changes requested):
Finished review and other than the audit event name, looks good to me. Will approve once changed.

The audit event should be AUTH_REVERIFY_AUTHORISATION_ERROR_RECEIVED, as it is the one getting reviewed for the event catalogue and is described as
> When Authentication receives an OAuth error from IPV. This also occurs if the cross-browser issue is encountered.

mattvot (approved):

PR #5798, opened 2025-01-23
Aut 3993/feature flag writing to account modifiers

BeckaL:
## What

Adds a new optional field to the json request to the reset password handler, that allows the frontend to control whether or ot we allow mfa reset after password reset. If this field is omitted, the behaviour will default to current (not allowing this).

Currently, when a user resets their password, we don't want them to immediately also be able to reset their MFA for security reasons. However, when we switch on the new mfa reset journey which goes via IPV, there is sufficient security to not need this block. Therefore, when this feature is switched on, we no longer need to add an account modifiers block for the user.

We've made a decision here not to introduce a separate feature flag on the backend that reflects whether the new support mfa journey is on or not. Instead the behaviour is controlled via the frontend feature flag which will be used to populate this field, meaning this new behaviour will all be switched on at once.

## How to review

Code review (testing covered off separately)

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/ResetPasswordIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/ResetPasswordIntegrationTest.java
@@ -56,6 +56,15 @@ public class ResetPasswordIntegrationTest extends ApiGatewayHandlerIntegrationTe
                             }
                             """,
                     PASSWORD);
+    private static final String RESET_PASSWORD_REQUEST_WITH_ALLOW_MFA_RESET =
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/ResetPasswordIntegrationTest.java
andrew-moores, line 1: Commit message:
Improve the integration test so that:
* We can catch deserialisation issues
* It's clearer what the request to this endpoint actually looks like
* We can test optional fields

andrew-moores (approved):

PR #5799, opened 2025-01-23
ATO-1143: Send encryption key id in JWE headers of IPV auth request

cearl1:
### Wider context of change

In a previous PR, we added a feature flag to consume and cache encryption JWKs from an IPV JWKS endpoint.

### What’s changed

To let IPV know which key we encrypted with, we need to include the key ID in the headers of the auth request.

### Checklist


ipv-api/src/main/java/uk/gov/di/authentication/ipv/services/IPVAuthorisationService.java
```diff
@@ -265,10 +269,14 @@ private RSAPublicKey getPublicKeyFromSSM() {
         }
     }
 
-    private RSAPublicKey getPublicKeyFromJwksEndpoint() {
+    private JWK getJwkFromJwksEndpoint() {
+        LOG.info("Getting IPV Auth Encryption JWK via JWKS endpoint");
```
Ryan-Andrews99, line 273: Can we tweak this log line to be something less auth-y sounding, think that name is a hangover of auth/orch as one team?

cearl1: Sure, i've removed the Auth bit so it just reads `IPV Encryption JWK/IPV Encryption Key`

PR #5815, opened 2025-01-28
ATO-1362: Remove use of session id getter from AuthorisationHandler

cearl1:
### Wider context of change

### What’s changed

This PR removes usage of `Session.getSessionId()` from `AuthorisationHandler`. To achieve this, we get the session ID from the cookie first, instead of getting the session directly from the cookie. We can then pass this sessionId variable around and use it where `Session.getSessionId()` was used.

The tests for this handler were also slightly inaccurate. In most cases, they were mocking `generateSession()` with the same session that you already had at the start of the workflow. Instead, I've created a new variable for the new session (the same as the old session except with the new ID), and updated all the assertions to use this.

Finally, I've extracted a test helper class TestIdGeneratorHelper, which adds a shortcut to static mocking the IdGenerator class. You can use it like so: `var result = runWithIds(() -> methodWhichGeneratesIds(), List.of("id1", "id2"));`.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -416,9 +416,10 @@ public class AuthorisationHandler
                 user,
                 auditEventExtensions.toArray(AuditService.MetadataPair[]::new));
 
-        Optional<Session> session = sessionService.getSessionFromSessionCookie(input.getHeaders());
+        Optional<String> sessionId =
+                CookieHelper.getSessionIdFromRequestHeaders(input.getHeaders());
         Optional<OrchSessionItem> orchSessionOptional =
-                orchSessionService.getSessionFromSessionCookie(input.getHeaders());
+                sessionId.flatMap(orchSessionService::getSession);
 
         ClientSession clientSession =
                 clientSessionService.generateClientSession(
@@ -431,7 +432,7 @@ public class AuthorisationHandler
                 authRequest.toParameters(), Optional.of(client))) {
 
             return handleDocAppJourney(
-                    session,
+                    sessionId,
                     orchSessionOptional,
                     clientSession,
                     authRequest,
@@ -446,16 +447,15 @@ public class AuthorisationHandler
         Optional<String> browserSessionIdFromCookie =
                 CookieHelper.parseBrowserSessionCookie(input.getHeaders());
 
-        Optional<Session> sessionWithValidBrowserSessionId = session;
         boolean newAuthenticationRequired = false;
         if (browserSessionIdFromSession.isPresent()
                 && !Objects.equals(browserSessionIdFromSession, browserSessionIdFromCookie)) {
-            sessionWithValidBrowserSessionId = Optional.empty();
+            sessionId = Optional.empty();
             newAuthenticationRequired = true;
         }
 
         return handleAuthJourney(
-                sessionWithValidBrowserSessionId,
+                sessionId,
                 orchSessionOptional,
                 clientSession,
                 authRequest,
@@ -504,7 +504,7 @@ public class AuthorisationHandler
     }
 
     private APIGatewayProxyResponseEvent handleDocAppJourney(
-            Optional<Session> existingSession,
+            Optional<String> existingSessionId,
             Optional<OrchSessionItem> orchSessionOptional,
             ClientSession clientSession,
             AuthenticationRequest authenticationRequest,
@@ -515,15 +515,16 @@ public class AuthorisationHandler
         Session session;
         var newSessionId = IdGenerator.generate();
         var newBrowserSessionId = IdGenerator.generate();
-
-        if (existingSession.isEmpty()) {
+        var existingSession = existingSessionId.flatMap(sessionService::getSession);
+        if (existingSessionId.isEmpty() || existingSession.isEmpty()) {
             session = sessionService.generateSession(newSessionId);
             updateAttachedSessionIdToLogs(newSessionId);
             LOG.info("Created new session with ID {}", newSessionId);
         } else {
-            session = existingSession.get();
-            var previousSessionId = session.getSessionId();
-            sessionService.updateWithNewSessionId(session, newSessionId);
+            var previousSessionId = existingSessionId.get();
+            session =
+                    sessionService.updateWithNewSessionId(
+                            existingSession.get(), previousSessionId, newSessionId);
             updateAttachedSessionIdToLogs(newSessionId);
             LOG.info("Updated session ID from {} to {}", previousSessionId, newSessionId);
         }
@@ -575,13 +576,13 @@ public class AuthorisationHandler
 
         var authorisationRequest = authRequestBuilder.build();
 
-        docAppAuthorisationService.storeState(session.getSessionId(), state);
+        docAppAuthorisationService.storeState(newSessionId, state);
         noSessionOrchestrationService.storeClientSessionIdAgainstState(clientSessionId, state);
 
         auditService.submitAuditEvent(
                 DocAppAuditableEvent.DOC_APP_AUTHORISATION_REQUESTED,
                 client.getClientID(),
-                user.withSessionId(session.getSessionId())
+                user.withSessionId(newSessionId)
                         .withUserId(clientSession.getDocAppSubjectId().getValue()));
 
         URI authorisationRequestUri = authorisationRequest.toURI();
@@ -594,7 +595,7 @@ public class AuthorisationHandler
 
         List<String> cookies =
                 handleCookies(
-                        session,
+                        newSessionId,
                         orchSession,
                         authenticationRequest,
                         persistentSessionId,
@@ -608,7 +609,7 @@ public class AuthorisationHandler
     }
 
     private APIGatewayProxyResponseEvent handleAuthJourney(
-            Optional<Session> existingSession,
+            Optional<String> existingSessionId,
             Optional<OrchSessionItem> existingOrchSessionOptional,
             ClientSession clientSession,
             AuthenticationRequest authenticationRequest,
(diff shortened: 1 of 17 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/SessionService.java
```diff
@@ -79,6 +79,19 @@ private void storeOrUpdateSession(Session session, String oldSessionId) {
         }
     }
 
+    public Session updateWithNewSessionId(
+            Session session, String oldSessionId, String newSessionId) {
```
GHSwallow, line 83: Could you update `updateWithNewSessionId(Session session, String newSessionId)` to use this method, and move this to the bottom of the `updateWithNewSessionId` methods?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
```diff
@@ -77,6 +77,14 @@ public static Optional<HttpCookie> getHttpCookieFromHeaders(
...
 
+    public static Optional<String> getSessionIdFromSessionCookie(Map<String, String> headers) {
+        return parseSessionCookieStatic(headers).map(SessionCookieIds::getSessionId);
+    }
+
+    public static Optional<SessionCookieIds> parseSessionCookieStatic(Map<String, String> headers) {
+        return new CookieHelper().parseSessionCookie(headers);
+    }
+
     public Optional<SessionCookieIds> parseSessionCookie(Map<String, String> headers) {
```
GHSwallow, line 88: I'm not totally sure why `parseSessionCookie` isn't `static`. I think could do a small refactor commit here to make it static, if it makes sense to.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -632,11 +632,10 @@ private APIGatewayProxyResponseEvent handleAuthJourney(
 
         Session session;
         OrchSessionItem orchSession;
-        Optional<String> existingSessionId = existingSession.map(Session::getSessionId);
-
+        var newSessionId = IdGenerator.generate();
```
GHSwallow, line 635: I love this! We always want a new one, so it's generated straight away and we know the session (whether created or cycled) will have this new id

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -679,26 +676,26 @@ && maxAgeExpired(
                         newSessionId);
 
             } else {
-                var sessionToBeUpdated = existingSession.get();
-                var previousSessionId = sessionToBeUpdated.getSessionId();
-                sessionService.updateWithNewSessionId(sessionToBeUpdated);
-                session = sessionToBeUpdated;
+                var previousSession = existingSession.get();
+                var previousSessionId = existingSessionId.orElseThrow();
```
GHSwallow, line 680: I don't love having an `orElseThrow` inline here. We know this should be defined by here. I'd rather add  `existingSessionId.isPresent()` in the `if` above as you did in the DocApp journey so this can just be `existingSessionId.get()`.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
cearl1: Ah cool ok, I left it as is to avoid changing too much, but yeah I'll update it to be static

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
cearl1: Doh, looks like I added an `isEmpty()` check on the `if` further up but forgot to update this `orElseThrow()` line. I'll update this!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/SessionService.java
cearl1: I think i was trying to avoid doing this because it would add another `session.getSessionId()`. It would reduce a lot of duplication though so I'll do this and remove the `getSessionId()` call later.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/CookieHelper.java
```diff
@@ -77,6 +77,14 @@ public static Optional<HttpCookie> getHttpCookieFromHeaders(
         return parseStringToHttpCookie(cookie);
     }
 
+    public static Optional<String> getSessionIdFromSessionCookie(Map<String, String> headers) {
+        return parseSessionCookieStatic(headers).map(SessionCookieIds::getSessionId);
+    }
```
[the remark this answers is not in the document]
cearl1: I've addressed the non-static method call below, but I agree the name of this method is a bit strange. I changed it to `getSessionIdFromRequestHeaders` as you suggested

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -2830,10 +2836,22 @@ void shouldUpdateOrchSessionCorrectlyForRequestObjectRequest(
...
 
     private APIGatewayProxyResponseEvent makeHandlerRequest(APIGatewayProxyRequestEvent event) {
-        var response = handler.handleRequest(event, context);
+        var response =
+                runWithIds(
+                        () -> handler.handleRequest(event, context),
+                        List.of(
+                                NEW_SESSION_ID,
+                                NEW_BROWSER_SESSION_ID,
+                                NEW_SESSION_ID_FOR_PREV_SESSION));
```
GHSwallow, line 2854: This pattern feels quite vulnerable to failure if we, eg, modify the order in the handler, or add in a new IDGenerator. But, I can't actually think of a way around this, and this wasn't being tested well before, so I do think this is a great improvement on our testing coverage!

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -306,15 +312,23 @@ public void setUp() {
...
         when(clientSession.getDocAppSubjectId()).thenReturn(new Subject("test-subject-id"));
         when(clientService.getClient(anyString()))
                 .thenReturn(Optional.of(generateClientRegistry()));
+        when(sessionService.updateWithNewSessionId(any(Session.class), anyString(), anyString()))
+                .then(
+                        invocation -> {
+                            Session sessionToUpdate = invocation.getArgument(0);
+                            sessionToUpdate.setSessionId(invocation.getArgument(2));
+                            return sessionToUpdate;
+                        });
```
GHSwallow, line 331: It makes sense to mock the service here, but I think the code coverage on this method could be better. There's one test in SessionServiceTest that uses it, but it doesn't really cover everything it does.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -677,7 +691,7 @@ void shouldRedirectToLoginWithPromptParamWhenSetToLoginAndExistingSessionIsPrese
                     .submitAuditEvent(
                             OidcAuditableEvent.AUTHORISATION_INITIATED,
                             CLIENT_ID.getValue(),
-                            BASE_AUDIT_USER.withSessionId(session.getSessionId()),
+                            BASE_AUDIT_USER.withSessionId(NEW_SESSION_ID),
```
GHSwallow, line 694: This pattern of not using the test variables in assertions is a pattern I've noticed too, really glad to see it being changed here 🚀

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -2830,10 +2836,22 @@ void shouldUpdateOrchSessionCorrectlyForRequestObjectRequest(
                 verify(orchSessionService).deleteSession(orchSession.getSessionId());
             }
         }
+
+        private APIGatewayProxyResponseEvent makeHandlerRequestForSession(
```
GHSwallow, line 2840: I think this name can be better. If I've understood it correctly, this is making a handler request with a sessionId in the cookie. Personally, I think `ForSession` doesn't encapsulate that fact.

GHSwallow (commented):
Lookin' good! Only a few nits really. I'm otherwise happy to approve this:)

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
cearl1: I think is fine for this handler because there is a consistent order of `IdGenerator.generate()` calls. We know the first 2 are going to be the sessionId and bsid, then if a session already exists, it will call generate again for the previous session. For the case where we have `IdGenerator.generate()` calls 2 branches in logic, e.g:
```java
if (condition){
  var id = IdGenerator.generate()
} else if (otherCondition){
  var id2 = IdGenerator.generate()
}
```
then this approach wont work.

An alternative would be to just generate ids with increasing numbers (e.g `ID_1, ID_2, ID_3`), but that would affect the readability of this test a bit.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
cearl1: Theres a few methods in SessionService that are a bit light on coverage. Mostly cases where exceptions are thrown or empty optionals are returned. We could improve coverage for this class in another PR?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
cearl1: Yep I agree, I've renamed it to match the BSID request method `makeRequestWithSessionIdInCookie()`

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
cearl1: I've added `TestIdGenerator.runWithIncrementalIds(method, prefix)` to the test helper and added a test for it, so we can use it in the future if we want.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
GHSwallow: Agree 👍

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
GHSwallow: brilliant!

PR #5823, opened 2025-01-29
BAU: Add and run tflint

whi-tw:
## What

Add tflint config, plus fix all existing issues

## How to review

Code review

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -29,6 +29,31 @@ repos:
     hooks:
       - id: terraform_fmt
 
+  - repo: local
+    hooks:
+      - id: tflint
+        name: Init tflint (hack)
+        files: ^ci/terraform/
+        pass_filenames: false
+        types_or:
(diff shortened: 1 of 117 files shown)


.pre-commit-config.yaml
```diff
@@ -3,15 +3,6 @@ exclude: |
       .*/test/.*/.*\.approved.json
   )$
 repos:
-  - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v5.0.0
-    hooks:
-      - id: trailing-whitespace
```
mattvot, line 9: I can see we have other formatters in the precommit config.

Are these dropped because having the language specific formatters do their stuff is preferable to these general purpose rules?

whi-tw: no, this would be beacause i'm a fool

whi-tw: fixed in a fixup!

PR #5824, opened 2025-01-29
ATO-1364: Remove use of session id getter from AuthCodeHandler

cearl1:
### Wider context of change

### What’s changed

This PR removes usage of `Session.getSessionId()` from `AuthCodeHandler`. The session id was already available in the request headers, so we did not need to call the getter.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
@@ -50,6 +50,7 @@ import java.util.Optional;
 import static java.util.Objects.isNull;
 import static uk.gov.di.orchestration.shared.conditions.DocAppUserHelper.isDocCheckingAppUserWithSubjectId;
 import static uk.gov.di.orchestration.shared.domain.RequestHeaders.CLIENT_SESSION_ID_HEADER;
+import static uk.gov.di.orchestration.shared.domain.RequestHeaders.SESSION_ID_HEADER;
 import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyErrorResponse;
 import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
 import static uk.gov.di.orchestration.shared.helpers.InstrumentationHelper.addAnnotation;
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -154,15 +155,18 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                    orchSessionService
-                            .getSessionFromRequestHeaders(input.getHeaders())
-                            .orElse(null);
+            sessionId =
+                    getHeaderValueFromHeaders(
+                            input.getHeaders(),
+                            SESSION_ID_HEADER,
+                            configurationService.getHeadersCaseInsensitive());
+            session = sessionService.getSession(sessionId).orElse(null);
+            orchSession = orchSessionService.getSession(sessionId).orElse(null);
```
GHSwallow, line 169: If sessionId is null, I think getSession(null) throw? Looks like `getSessionFromRequestHeaders` can handle a null sessionId. It throws later on in `validateSessions`.

cearl1: Unless redis doesn't like getting values for null keys (i cant see any exceptions being thrown from it), then if the key is null, then the `serializedSession` would be null in `SessionService.getSession()`, which would be returned as an empty optional.

I just wrote a quick test in `SessionService` seeing what would happen if you call `SessionService.getSession(null)`. It returns an empty optional, so should be handled the same way as before.

GHSwallow: The orchSession is on dynamoDb, which I've come across failing due to an empty key being provided

cearl1: Ahh good catch, yep i'm getting exceptions when i try to get a null id. I'll add a check before getting session and orchSession, that would be more reflective of what it was doing before.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -154,15 +155,21 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+            sessionId =
+                    getHeaderValueFromHeaders(
+                            input.getHeaders(),
+                            SESSION_ID_HEADER,
+                            configurationService.getHeadersCaseInsensitive());
+            if (sessionId == null) {
+                return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1000);
+            }
+            session = sessionService.getSession(sessionId).orElse(null);
+            orchSession = orchSessionService.getSession(sessionId).orElse(null);
```
GHSwallow, line 172: imo this shouldn't have an `orElse(null)`, and in `validateSessions`, it should check if either of these are empty. But, I realise that I'm giving you a lot of refactoring comments unrelated to sessionId migration 😅

Approving this as is

GHSwallow (approved):

PR #5831, opened 2025-01-30
ATO-1389: add infrastructure for AuthUserInfo DynamoDb table

GHSwallow:
### Wider context of change
This table will replace the table of the same name that exists on Auth accounts. This table will store the UserInfo that Auth sents to Orch. Auth don't need to know about the table. In this migration, we also fix an issue with the current table, which is only indexed by internalCommonSubjectId. If a user has multiple client sessions, the authUserInfo entry gets overwritten. Hence, we have clientSessionId as a sort key as well. We keep ICSID as the partition key so we can still easily delete user data on request.

### What’s changed
Add infrastructure for AuthUserInfo DynamoDb table 

### Manual testing
Deployed to Dev. Built successfully, and saw that ICSID was the partitian key and ClientSessionId the sort key

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -481,6 +481,66 @@ Resources:
 
   #endregion
 
+  #region AuthUserInfo DynamoDB Table
+
+  AuthUserInfoEncryptionKey:
+    Type: AWS::KMS::Key
+    Properties:
+      Description: KMS encryption key for AuthUserInfo DynamoDB table
+      EnableKeyRotation: true
+      KeyPolicy:
+        Version: 2012-10-17
+        Statement:
+          - Sid: AllowIamManagement
+            Effect: Allow
+            Principal:
+              AWS: !Sub arn:aws:iam::${AWS::AccountId}:root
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -481,6 +481,66 @@ Resources:
...
+          KeyType: HASH
+        - AttributeName: ClientSessionId
+          KeyType: RANGE
+      BillingMode: PAY_PER_REQUEST
+      SSESpecification:
+        SSEEnabled: true
+        KMSMasterKeyId: !GetAtt AuthUserInfoEncryptionKey.Arn
+        SSEType: KMS
+      TimeToLiveSpecification:
+        AttributeName: ttl
```
Ryan-Andrews99, line 535: Just a note that this differs from the current ttl attribute (called `TimeToExist`) - Not really anything to resolve here just something to remember when creating the new DynamoDbBean class

Ryan-Andrews99 (approved):

PR #5838, opened 2025-01-31
ATO-1366: Remove use of session id getter from LogoutService

cearl1:
### Wider context of change

### What’s changed

This PR removes usage of Session.getSessionId() from LogoutService. I've introduced a new data class `DestroySessionsRequest`, which is used in the `LogoutService.destroySessions()` method, and requires setting the session ID directly on creation. This is used by all LogoutService methods that call `destroySessions` .

LogoutRequest was also the last place that `LogLineHelper.attachSessionIdToLogs(Session session)` was used, so I've updated it to use the version of the method where you provide the sessionId directly, and removed the old method signature and associated test.

### Checklist

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -29,6 +29,7 @@ import uk.gov.di.orchestration.shared.api.OrchFrontend;
 import uk.gov.di.orchestration.shared.entity.AccountIntervention;
 import uk.gov.di.orchestration.shared.entity.AuthUserInfoClaims;
 import uk.gov.di.orchestration.shared.entity.ClientRegistry;
+import uk.gov.di.orchestration.shared.entity.DestroySessionsRequest;
 import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
 import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
 import uk.gov.di.orchestration.shared.exceptions.NoSessionException;
@@ -325,7 +326,11 @@ public class IPVCallbackHandler
                 if (configurationService.isAccountInterventionServiceActionEnabled()
                         && (intervention.getBlocked() || intervention.getSuspended())) {
                     return logoutService.handleAccountInterventionLogout(
-                            session, input, clientId, intervention);
+                            new DestroySessionsRequest(sessionId, session),
+                            session.getInternalCommonSubjectIdentifier(),
+                            input,
+                            clientId,
+                            intervention);
                 }
 
                 return ipvCallbackHelper.generateAuthenticationErrorResponse(
@@ -431,7 +436,11 @@ public class IPVCallbackHandler
                 if (configurationService.isAccountInterventionServiceActionEnabled()
                         && (intervention.getBlocked() || intervention.getSuspended())) {
                     return logoutService.handleAccountInterventionLogout(
-                            session, input, clientId, intervention);
+                            new DestroySessionsRequest(sessionId, session),
+                            session.getInternalCommonSubjectIdentifier(),
+                            input,
+                            clientId,
+                            intervention);
                 }
                 var returnCode = userIdentityUserInfo.getClaim(RETURN_CODE.getValue());
                 if (returnCodePresentInIPVResponse(returnCode)) {
--- a/ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandlerTest.java
+++ b/ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandlerTest.java
@@ -49,6 +49,7 @@ import uk.gov.di.orchestration.shared.entity.AccountInterventionState;
 import uk.gov.di.orchestration.shared.entity.ClientRegistry;
 import uk.gov.di.orchestration.shared.entity.ClientSession;
 import uk.gov.di.orchestration.shared.entity.CredentialTrustLevel;
+import uk.gov.di.orchestration.shared.entity.DestroySessionsRequest;
 import uk.gov.di.orchestration.shared.entity.IdentityClaims;
 import uk.gov.di.orchestration.shared.entity.NoSessionEntity;
 import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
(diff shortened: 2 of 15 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
cearl1, line 1: There's been no update to tests for this handler because the area I changed wasn't actually tested. I noticed a lot of this handler wasn't tested - I'll raise another PR to improve coverage for this handler soon.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/entity/LogoutRequestTest.java
```diff
@@ -416,13 +417,12 @@ void shouldCorrectlyParseLogoutRequestWhenLogoutUriInRequestDoesNotMatchClientRe
...
-        return new uk.gov.di.orchestration.shared.entity.Session(SESSION_ID)
-                .addClientSession(CLIENT_SESSION_ID);
+    private Session generateSession() {
+        return new Session(SESSION_ID).addClientSession(CLIENT_SESSION_ID);
     }
 
-    private void generateSessionFromCookie(uk.gov.di.orchestration.shared.entity.Session session) {
-        when(sessionService.getSessionFromSessionCookie(anyMap())).thenReturn(Optional.of(session));
+    private void generateSessionFromCookie(Session session) {
+        when(sessionService.getSession(anyString())).thenReturn(Optional.of(session));
```
cearl1, line 425: This had the full package path for Session where it was used for some reason. I'm guessing this was when we had another class called Session used here? Either way i've removed the full package path and added an import for Session :smile:

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
cearl1: Turns out this handler was pretty thoroughly tested, the coverage was split between the unit test and integration test. I've added some tests to the unit test to cover the cases where you are forced to logout by AIS.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/LogoutRequest.java
```diff
@@ -41,29 +43,35 @@ public class LogoutRequest {
...
-        Optional<String> journeyId = extractClientSessionIdFromCookieHeaders(input.getHeaders());
 
         auditUser =
                 TxmaAuditUser.user()
                         .withIpAddress(extractIpAddress(input))
                         .withPersistentSessionId(
                                 extractPersistentIdFromCookieHeader(input.getHeaders()))
-                        .withSessionId(sessionId.orElse(null))
-                        .withGovukSigninJourneyId(journeyId.orElse(null))
+                        .withSessionId(session.isPresent() ? sessionId.orElse(null) : null)
```
Ryan-Andrews99, line 73: Is the ternary necessary here, can't see what it does that isn't covered by `sessionId.orElse(null)`?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/DestroySessionsRequest.java
```diff
@@ -45,4 +49,18 @@ public boolean equals(Object o) {
...
+                + "sessionId='"
+                + sessionId
+                + '\''
+                + ", clientSessions="
+                + clientSessions
+                + ", emailAddress='"
+                + emailAddress
+                + '\''
+                + '}';
+    }
```
Ryan-Andrews99, line 61: Is this needed for test assertions? I think this could probably be in the commit where we create this class

ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandlerTest.java
Ryan-Andrews99, line 1: Thanks for adding this!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/DestroySessionsRequest.java
```diff
@@ -0,0 +1,48 @@
...
+        return sessionId;
+    }
+
+    public List<String> getClientSessions() {
+        return clientSessions;
+    }
+
+    public String getEmailAddress() {
+        return emailAddress;
+    }
```
Ryan-Andrews99, line 32: Is this method used anywhere? Just checking we need email address on this (or is this potential to migrate to be used when we remove `session.getEmailAddress()`)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/LogoutRequest.java
cearl1: So this was needed because before my changes, if the session ID was present, but no session existed with that session ID, then this session ID would be null (because it was getting the session directly from the request headers). To match this functionality I added this weird tenary operation here.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/DestroySessionsRequest.java
cearl1: Yeah this was mostly to help with fixing tests, I'll include this in the commit for creating the class

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/DestroySessionsRequest.java
cearl1: Yeah this is used in `LogoutRequest.destroySessions()`. If a client session exists in dynamo, it will send a back channel logout message, which takes an email address as a parameter.

Before these 3 fields were all optional parameters to the `destroySessions` method, so I grouped them into this `DestroySessionsRequest` to make it easier to check if it's present or not.

Ryan-Andrews99: Ah makes sense - forgot about backchannel logouts 😄

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/LogoutRequest.java
Ryan-Andrews99: Ah I see what you mean, I wonder if this is any nicer: 

```
session.ifPresent((s) -> auditUser.withSessionId(sessionId.orElse(null)));
```

cearl1: Yeah it could definitely do with being clearer. I've moved it into the `if (sessionId.isPresent() && session.isPresent())` section now.

Ryan-Andrews99 (approved):

PR #5847, opened 2025-02-03
AUT-4048: Allow VPC execute endpoint id of new Auth Accounts 

pskushwaha1:
## What

AUT-4048: Allow VPC execute endpoint id of new Auth Accounts to invoke Auth ext API 

We are moving the Orch stub to new account and the callback function needs to invoke the Auth external IP for user Journey 

## How to review

Code Review
1. Deploy to  Dev using pipeline 
2. check if the resource policy of AUTH Ext API is getting update to Allow VPC endpoint


whi-tw (approved):
LGTM

PR #5865, opened 2025-02-05
AUT-3884: Add consumer contract tests to frontend-api AIS service

ben-fernandes-sw:
## What

Adds contract tests to the `AccountInterventionsService` in the `frontend-api` module.

Similar Orchestration tests already existed in `uk.gov.di.authentication.oidc.contract.AccountInterventionServiceTest`, so I began by copying those. The Orch version of the service has internal logic that we don't replicate (we just forward through the response from AIS), so the contract tests could be slimmed down a bit.

This PR only addresses the `frontend-api` consumer side. The AIS provider side should already be working, given that Orch is already using it in their tests.

I also tweaked the response model files:
* The `@Required` decorator didn't do anything
    * For the primitive types, they naturally reject non-null values
    * For the Object types, they require the custom constructor check
* Removed fields from the `Intervention` model
    * We don't use them all, so we don't need to define them all
* Fixed the `appliedAt` type
    * We get a `number` back from AIS, not a `string`

## How to review

1. Code Review
1. Deploy to sandpit with `./deploy-sandpit.sh -a`
    * [ ] Ensure that requests still make it back to the `frontend-api` after the types of the models being changed

## Related PRs

https://github.com/govuk-one-login/authentication-frontend/pull/2551

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/contract/AccountInterventionsServiceTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/contract/AccountInterventionsServiceTest.java
@@ -1,0 +1,324 @@
+package uk.gov.di.authentication.frontendapi.contract;
+
+import au.com.dius.pact.consumer.MockServer;
+import au.com.dius.pact.consumer.dsl.DslPart;
+import au.com.dius.pact.consumer.dsl.PactDslWithProvider;
+import au.com.dius.pact.consumer.junit.MockServerConfig;
+import au.com.dius.pact.consumer.junit5.PactConsumerTest;
+import au.com.dius.pact.consumer.junit5.PactTestFor;
+import au.com.dius.pact.core.model.PactSpecVersion;
+import au.com.dius.pact.core.model.RequestResponsePact;
+import au.com.dius.pact.core.model.annotations.Pact;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.frontendapi.entity.AccountInterventionsInboundResponse;
+import uk.gov.di.authentication.frontendapi.services.AccountInterventionsService;
+import uk.gov.di.authentication.shared.exceptions.UnsuccessfulAccountInterventionsResponseException;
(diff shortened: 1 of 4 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/AccountInterventionsResponse.java
```diff
@@ -1,11 +1,17 @@
 package uk.gov.di.authentication.frontendapi.entity;
 
 import com.google.gson.annotations.Expose;
-import org.apache.logging.log4j.core.config.plugins.validation.constraints.Required;
```
BeckaL, line 4: Was the issue here actually that this is the wrong import? Should be 
`import uk.gov.di.authentication.shared.validation.Required;`. Looks like this mistake proliferated a bit in our AIS code

BeckaL: (I would be sad if we had to start eroding nice un-boilerplate-y records with constructors etc but haven't tested this out fully)

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/contract/AccountInterventionsServiceTest.java
```diff
@@ -0,0 +1,320 @@
...
+import uk.gov.di.authentication.frontendapi.entity.AccountInterventionsInboundResponse;
+import uk.gov.di.authentication.frontendapi.services.AccountInterventionsService;
+import uk.gov.di.authentication.shared.exceptions.UnsuccessfulAccountInterventionsResponseException;
+import uk.gov.di.authentication.shared.services.ConfigurationService;
+
+import java.net.URI;
+import java.net.http.HttpClient;
+
+import static au.com.dius.pact.consumer.dsl.LambdaDsl.newJsonBody;
+import static org.junit.jupiter.api.Assertions.*;
```
BeckaL, line 23: We should [avoid wildcard imports](https://gds-way.digital.cabinet-office.gov.uk/manuals/programming-languages/java.html#imports) as standard

dbes-gds (approved):

PR #5872, opened 2025-02-06
ATO-646: Add client id authcode validation to token endpoint

cearl1:
### Wider context of change

Previously, it was possible for a client to use another client's valid auth code. We want to add validation for this to prevent this from happening.

### What’s changed

If the client id in the auth code does not match the client id in the request parameters, then a 400 `invalid_client` response will be generated. 

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
@@ -221,8 +221,8 @@ public class TokenHandler
         AuthCodeExchangeData authCodeExchangeData = authCodeExchangeDataMaybe.get();
         if (!Objects.equals(authCodeExchangeData.getClientId(), clientRegistry.getClientID())) {
             LOG.warn("Client ID from auth code does not match client ID from request body");
-            // ATO-646: Make this return 400 error when we are confident that
-            // it won't cause issues with existing clients
+            return generateApiGatewayProxyResponse(
+                    400, OAuth2Error.INVALID_GRANT.toJSONObject().toJSONString());
         }
         updateAttachedLogFieldToLogs(CLIENT_SESSION_ID, authCodeExchangeData.getClientSessionId());
         updateAttachedLogFieldToLogs(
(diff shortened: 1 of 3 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -221,8 +221,11 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
         AuthCodeExchangeData authCodeExchangeData = authCodeExchangeDataMaybe.get();
         if (!Objects.equals(authCodeExchangeData.getClientId(), clientRegistry.getClientID())) {
             LOG.warn("Client ID from auth code does not match client ID from request body");
-            // ATO-646: Make this return 400 error when we are confident that
-            // it won't cause issues with existing clients
+            return generateApiGatewayProxyResponse(
+                    400,
+                    new ErrorObject(OAuth2Error.INVALID_CLIENT_CODE, "Invalid auth code for client")
```
Ryan-Andrews99, line 226: I wonder if it makes more sense to for this message to be less transparent, what if we returned the same `INVALID_GRANT` error as above? This way we're not inadvertently revealing whether an auth code is valid. What do you think?

Ryan-Andrews99 (commented):
Left a sort of 🤔  comment, but outside of that think this looks good 🚀

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
cearl1: Yeah makes sense! I wasn't sure how transparent to be about this. Looking at the javadoc comment for `INVALID_GRANT` it seems perfect for this kind of situation:

`The provided authorisation grant (e.g. authorisation code, resource owner credentials) or refresh token is invalid, expired, revoked, does not match the redirection URI used in the authorization request, or was issued to another client.`

Ryan-Andrews99 (dismissed):

Ryan-Andrews99 (approved):

PR #5933, opened 2025-02-12
ATO-1430: Remove use of session id getter in tests

cearl1:
### Wider context of change

Part of work to migrate sessionId to AuthSessionItem

### What’s changed

Removes usages of the session id getter from tests. In a lot of cases, the shared session was being mocked, so I've replaced the mocked session with a real session.

### Checklist

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandlerTest.java
@@ -121,6 +121,7 @@ class VerifyCodeHandlerTest {
     private static final long LOCKOUT_DURATION = 799;
     private static final URI REDIRECT_URI = URI.create("http://localhost/redirect");
     private static final int MAX_RETRIES = 6;
+    private static final String SESSION_ID_FOR_TEST_CLIENT = "session-id-for-test-client";
     private final Context context = mock(Context.class);
     private final SessionService sessionService = mock(SessionService.class);
     private final CodeStorageService codeStorageService = mock(CodeStorageService.class);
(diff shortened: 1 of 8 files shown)


shared/src/test/java/uk/gov/di/authentication/shared/services/SessionServiceTest.java
```diff
@@ -43,15 +42,19 @@ void shouldPersistSessionToRedisWithExpiry() throws Json.JsonException {
 
     @Test
     void shouldRetrieveSessionUsingRequestHeaders() throws Json.JsonException {
+        var serialisedSession = generateSearlizedSession();
         when(redis.keyExists("session-id")).thenReturn(true);
-        when(redis.getValue("session-id")).thenReturn(generateSearlizedSession());
+        when(redis.getValue("session-id")).thenReturn(serialisedSession);
```
isaac-GDS, line 47: I think "session-id" can be pulled out into a const when we create the session and then referenced

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandlerTest.java
```diff
@@ -195,7 +195,7 @@ void tearDown() {
                                         CLIENT_ID,
                                         TEST_CLIENT_CODE,
                                         SESSION_ID,
-                                        testSession.getSessionId()))));
+                                        "test-client-session-id"))));
```
isaac-GDS, line 198: Not to do with your change but unsure why we have `session` and `testSession` ? If there is a need for both maybe re-name for a better description. This doesn't have to be done in this PR or by you but more a note so it doesn't get lost.

shared/src/test/java/uk/gov/di/authentication/shared/services/SessionServiceTest.java
cearl1: Good shout, I've also changed the value to "test-session-id" because one of the tests was checking if the request param being lowercase would make the session service not retrieve the session id.

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/VerifyCodeHandlerTest.java
cearl1: Yep i agree the naming is confusing, especially how client-session is a different thing. I've renamed the test session to `sessionForTestClient` and added a todo comment to question whether we need both sessions

isaac-GDS (approved):

isaac-GDS (approved):

PR #5942, opened 2025-02-13
ATO-1450: Filter keys by algorithm in JwkCache

cearl1:
### Wider context of change

Previously, the JwkCache did not consider the algorithm of the keys. We use the cache to get an pubic encryption JWK and expect it to be using the algorithm `RSA-OAEP-256`, so if we get an encryption key with a different algorithm, the code would likely break. 

### What’s changed

This PR adds an `algs` field to the JwkCacheEntry, which is used to fetch keys with the right algorithms. 

It's worth noting that this just ensures that if a key with a different algorithm is added, then the code wont break. This PR does not add support for encryption keys using a different algorithm to `RSA-OAEP-256` or `RS256`.

This PR also temporarily looks for encryption keys with the `RS256` alg and the `RSA-OAEP-256` alg. Currently IPV are sending us encryption keys with the `RS256` alg, so once they fix this and send us encryption keys with the correct `RSA-OAEP-256` alg, we can remove the extra test for this, and think about whether we still need a set of `algs` in the cache. 

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/JwkCacheEntry.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/JwkCacheEntry.java
@@ -14,9 +14,9 @@ import java.util.List;
 
 public class JwkCacheEntry {
     private static final Logger LOG = LogManager.getLogger(JwkCacheEntry.class);
-    private final KeyUse keyUse;
     private final URL jwksUrl;
     private final int expirationInSeconds;
+    private final KeyUse keyUse;
     private JWK latestKey;
     private Date expireTime;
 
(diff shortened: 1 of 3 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/JwkCacheEntry.java
```diff
@@ -14,26 +14,29 @@
...
         this.expireTime = NowHelper.nowPlus(this.expirationInSeconds, ChronoUnit.SECONDS);
         this.keyUse = keyUse;
+        this.alg = alg;
         this.latestKey = getKeyFromUrl();
     }
 
-    public static JwkCacheEntry forKeyUse(KeyUse keyUse, URL url, int expirationInSeconds) {
-        return new JwkCacheEntry(url, expirationInSeconds, keyUse);
+    public static JwkCacheEntry forEncryptionKeys(URL url, int expirationInSeconds) {
+        return forKeyUse(url, expirationInSeconds, KeyUse.ENCRYPTION, "RS256");
```
cearl1, line 34: Although this encryption key is a RSA-OAEP-256 key, the `alg` value is `RS256`, which is a signing algorithm. I've added a comment in the test but worth pointing it out here.

isaac-GDS (approved):

PR #5954, opened 2025-02-14
ATO-1418: Add permissions to access new dynamo table

cearl1:
### Wider context of change

Part of the work to migrate to storing client sessions in an orch dynamo table

### What’s changed

This PR adds the relevant permissions to lambdas so they can access the new ClientSession dynamo table. Note that the lambdas aren't using the new table yet this is just for adding permissions so they can use it.

### Checklist

- [x] Impact on orch and auth mutual dependencies has been checked
- [x] Successfully deployed to authdev or not required 
   - [x] Deployment to orch dev was successful. 
   - [x] Deployment to sandpit was successful.
   - [x] Deployment to authdev1 was successful

### DIFF
--- a/ci/terraform/oidc/dynamo-policies.tf
+++ b/ci/terraform/oidc/dynamo-policies.tf
@@ -754,6 +754,22 @@ moved {
   to   = data.aws_iam_policy_document.dynamo_orch_session_encryption_key_cross_account_decrypt_policy_document[0]
 }
 
+data "aws_iam_policy_document" "dynamo_orch_session_cross_account_read_and_delete_access_policy_document" {
+  count = var.is_orch_stubbed ? 0 : 1
+  statement {
+    sid    = "AllowOrchSessionCrossAccountReadAndDeleteAccess"
+    effect = "Allow"
+    actions = [
+      "dynamodb:DescribeTable",
+      "dynamodb:Get*",
+      "dynamodb:DeleteItem",
+    ]
+    resources = [
+      "arn:aws:dynamodb:eu-west-2:${var.orch_account_id}:table/${var.orch_environment}-Orch-Session",
+    ]
+  }
+}
+
 data "aws_iam_policy_document" "dynamo_orch_session_cross_account_read_access_policy_document" {
   count = var.is_orch_stubbed ? 0 : 1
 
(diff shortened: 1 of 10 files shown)


ci/terraform/oidc/dynamo-policies.tf
```diff
@@ -792,6 +792,43 @@ moved {
...
+    ]
+    resources = [
+      "arn:aws:dynamodb:eu-west-2:${var.orch_account_id}:table/${var.orch_environment}-Client-Session",
+    ]
+  }
+}
+moved {
+  from = data.aws_iam_policy_document.dynamo_orch_client_session_cross_account_delete_access_policy_document
+  to   = data.aws_iam_policy_document.dynamo_orch_client_session_cross_account_delete_access_policy_document[0]
+}
```
Ryan-Andrews99, line 830: Are these moved blocks needed if the resource didn't previously exist? Based on [this](https://developer.hashicorp.com/terraform/language/moved#moved) looks like it's used to update how a resource is referred to

Ryan-Andrews99 (commented):
Have left a comment about the moved blocks and also think we might need to deploy this to sandpit/orch dev to see if we're hitting any policy limits with our lambdas. It should be a quick check, it'll fail to deploy if so

ci/terraform/oidc/dynamo-policies.tf
cearl1: Oh maybe, If I'm honest I just replicated what was there for the orch session policies.

Looking at the commit which added this moved block for the orch session policies (https://github.com/govuk-one-login/authentication-api/commit/c408e6250c8299bce0ea373d3e046eb17f6706d7) it might be because of an issue with how `var.is_orch_stubbed` is used?

Ryan-Andrews99: Ah yeah they were introduced as part of a fix for Auth's dev environment where we didn't want to deploy this cross account access policy as Orchestration was stubbed. But the move blocks were specifically added because this policy was already live and in prod, so they shouldn't be needed here as these policies don't exist yet

Ryan-Andrews99 (approved):

PR #5961, opened 2025-02-17
ATO-1456: add email address to AuthSessionItem

Louisasa:
### Wider context of change

This is part of the session migration work, specifically the email migration. User is fetched from the email address, meaning we need to have the email address to check the user, to then check the email address.

### What’s changed

Added email address to AuthSessionItem

### Manual testing

TBD

### Checklist

- [x] Changes have been made to contract tests or not required. **not required**
- [x] Changes have been made to the simulator or not required. **not required**
- [x] Changes have been made to stubs or not required. **not required**
- [x] Successfully deployed to authdev or not required. **not required**
- [x] Successfully run Authentication acceptance tests against sandpit or not required. **not required**

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
@@ -132,6 +132,7 @@ public class CheckUserExistsHandler extends BaseFrontendHandler<CheckUserExistsR
             var userProfile = authenticationService.getUserProfileByEmailMaybe(emailAddress);
             var userExists = userProfile.isPresent();
             userContext.getSession().setEmailAddress(emailAddress);
+            userContext.getAuthSession().setEmailAddress(emailAddress);
 
             var session = userContext.getSession();
             var sessionId = userContext.getAuthSession().getSessionId();
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
@@ -178,6 +178,7 @@ public class SignUpHandler extends BaseFrontendHandler<SignupRequest>
             authSessionService.updateSession(
                     authSessionItem
                             .withAccountState(AuthSessionItem.AccountState.NEW)
+                            .withEmailAddress(request.getEmail())
                             .withInternalCommonSubjectId(internalCommonSubjectId.getValue()));
             LOG.info("Successfully processed request");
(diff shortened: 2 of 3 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
```diff
@@ -134,6 +134,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
             var userProfile = authenticationService.getUserProfileByEmailMaybe(emailAddress);
             var userExists = userProfile.isPresent();
             userContext.getSession().setEmailAddress(emailAddress);
+            userContext.getAuthSession().withEmailAddress(emailAddress);
```
GHSwallow, line 137: Ah I see what you were asking before. I think here, because we're just setting one property, I'd prefer using the `setEmailAddress`. `with_` is typically used just for building, ie chaining adding new values onto it, or creating a new one

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/SignUpHandler.java
```diff
@@ -168,6 +168,7 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
             authSessionService.updateSession(
                     authSessionItem
                             .withAccountState(AuthSessionItem.AccountState.NEW)
+                            .withEmailAddress(request.getEmail())
```
GHSwallow, line 171: here, `with_` is definitely the right call

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
```diff
@@ -130,4 +131,13 @@ public AuthSessionItem withUpliftRequired(boolean upliftRequired) {
         this.upliftRequired = upliftRequired;
         return this;
     }
+
```
GHSwallow, line 134: You'll need to add a `@DynamoDbAttribute(EMAIL)` as a decorator here

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
Louisasa: awkwardly where it's used in SignUpHandler I think it makes sense it's `with` there, do you still think it's worth changing to `set` here?

GHSwallow: yea I agree, SignUpHandler should have `with`, but I think here should have `set` yea

GHSwallow (approved):

PR #5965, opened 2025-02-17
Aut 4062/mfa reset metrics

alhcomer:
## What

Added metrics for MFA reset with IPV:
- Metric in ReverificationResultHandler to count the different responses returned from IPV.
- Metric in IDReverificationStateHandler to count the number of error responses returned from IPV.

Following this PR being merged an additional panel will be added on the ‘Sign In/Sign Up’ [dashboard](https://bhe21058.live.dynatrace.com/#dashboard;gtf=-7d%20to%20now;gf=all;id=1588f7b7-0df7-4730-b726-6b9320bc7cc7) to track these metrics

## How to review

1. Code Review
2. Go through mfa reset journey, see metrics in cloudwatch: 
    - "MfaResetIpvResponse" for regular journey and "MfaResetError" for cross brower journey

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
@@ -13,6 +13,7 @@ import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.CLIENT;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.CLIENT_NAME;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.ENVIRONMENT;
+import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.IPV_RESPONSE;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.IS_TEST;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.MFA_REQUIRED;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.REQUESTED_LEVEL_OF_CONFIDENCE;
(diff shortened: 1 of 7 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
```diff
@@ -123,6 +125,16 @@ public void incrementMfaResetHandoffCount() {
                 Map.of(ENVIRONMENT.getValue(), configurationService.getEnvironment()));
     }
 
+    public void incrementMfaResetIpvResponseCount(String ipvResponse) {
+        incrementCounter(
+                MFA_RESET_IPV_RESPONSE.getValue(),
+                Map.of(
+                        ENVIRONMENT.getValue(),
```
BeckaL, line 133: I might be wrong, but I think this now gets added by default as a result of https://github.com/govuk-one-login/authentication-api/pull/5773/files

shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
```diff
@@ -135,6 +136,12 @@ public void incrementMfaResetIpvResponseCount(String ipvResponse) {
                         ipvResponse));
     }
 
+    public void incrementMfaResetIpvTotalErrorResponseCount() {
```
BeckaL, line 139: I think the naming here could probably be clarified a bit - these aren't standard error responses back from ipv, this is a niche case for users who encounter the cross-browser issue. The equivalent audit event we've called "AUTH_REVERIFY_AUTHORISATION_ERROR_RECEIVED" so maybe we match this naming?

(This applies to all the naming in this commit related to this, not just this one particular function)

shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
alhcomer: Those changes were made in preparation for adding default dimensions but they're not actually used yet. I can't see any reason for this but I'm a little wary of using default dimensions in case their was a reason. What do you think?

shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
alhcomer: That makes sense, I've changed the naming of all mentions to match the audit event

shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
BeckaL: Ah good point, I'd missed that they weren't actually being used by default yet!

BeckaL (approved):

PR #5983, opened 2025-02-19
Ato 1414/clean up client session class

isaac-GDS:
### Wider context of change

Cleaning up Client session class to allow for more clarity in migration

### What’s changed

 - Methods removed
    - Getting and setting the docAppSubjectId
    - getting the clientName (only used in test)
    - getting creationDate (unused)

Removed test as no longer applicable
Removed logic in updating clientSession with DocAppSubjectId in StartHandler as it was never updated and now clientSession doesn't allow updating of docAppSubjectID.
  
### Testing
Running through Tests.
Does it need to be ran through DocApp journey ?


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartHandlerTest.java
```diff
@@ -203,47 +193,6 @@ void shouldReturn200WithStartResponse(String cookieConsentValue, String gaTracki
                         pair("internalSubjectId", AuditService.UNKNOWN));
     }
 
-    @Test
-    void shouldReturn200WhenDocCheckingAppUserIsPresent()
```
cearl1, line 207: With the removal of this test, sonar has spotted that these variables are no longer used: `DOC_APP_SCOPE`, `NONCE`, `CLIENT_ID`,`AUDIENCE`. Might as well remove them in this PR :smile:

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/ClientSessionServiceIntegrationTest.java
cearl1, line 1: Question: isn't this test still relevant while we're in the process of migrating to the dynamo client session store?

isaac-GDS: It is - but since we removed methods in the auth client session we can no longer test this with value. There is much less interaction and these tests also felt like a "just in case" then testing functionality. Hence removing it.

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartHandlerTest.java
isaac-GDS: Good spot. Yes we should.

integration-tests/src/test/java/uk/gov/di/authentication/shared/services/ClientSessionServiceIntegrationTest.java
cearl1: Ah ok yeah fair enough, makes sense.

cearl1 (approved):

PR #6011, opened 2025-02-25
ATO-1371: validate pkce challenge code and method on the `/authorize` endpoint

sw-mattw:
### Wider context of change

We wish to add PKCE support to the `/authorize` endpoint behind a feature flag.

PKCE should be optional to be backward compatible.

If a code challenge is provided, we wish to perform validation on both it and the code challenge method.

Code challenge validation:
- Non-blank if provided
- (We'll perform additional validation on the code verifier - not implemented at present)

Code challenge method validation:
- We wish to only support the "S256" code challenge method as "plain" offers lower security.
- A value must be provided if a code challenge value is provided
  - (Otherwise the method will default to "plain" if not in the request - which we don't want to support (https://datatracker.ietf.org/doc/html/rfc7636#section-4.3))

Validation needs to be performed on the query params (if the request object is null), and otherwise on the request object.

### What’s changed

- Added base validation method to the base validator
- Added call from the query param validator
- Added call from the request object validator
- Added unit tests for both validators
- Added integration tests

### Manual testing

Deployed to Orch dev, tested:
- Query param cases:
  - No code challenge provided, validation successful
  - Code challenge provided, no method, fails with expected validation error
  - Code challenge provided, method set to "plain", fails with expected validation error
  - Code challenge provided, method set to another invalid value (e.g., "abc"), fails with expected validation error
  - Code challenge provided, method set to "S256", validation successful
- Request object cases
  - Same as above, worked as expected

### Checklist

### Related PRs

PKCE FF - https://github.com/govuk-one-login/authentication-api/pull/6010

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
@@ -7,6 +7,8 @@ import com.nimbusds.oauth2.sdk.ResponseType;
 import com.nimbusds.oauth2.sdk.Scope;
 import com.nimbusds.oauth2.sdk.id.ClientID;
 import com.nimbusds.oauth2.sdk.id.State;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallenge;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallengeMethod;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
 import com.nimbusds.openid.connect.sdk.Nonce;
 import com.nimbusds.openid.connect.sdk.OIDCClaimsRequest;
@@ -31,6 +33,7 @@ public class RequestObjectToAuthRequestHelper {
 
     private RequestObjectToAuthRequestHelper() {}
 
+    @SuppressWarnings("deprecation")
     public static AuthenticationRequest transform(AuthenticationRequest authRequest) {
         if (Objects.isNull(authRequest.getRequestObject())) {
             return authRequest;
@@ -82,6 +85,17 @@ public class RequestObjectToAuthRequestHelper {
 
             if (Objects.nonNull(jwtClaimsSet.getClaim("max_age"))) {
                 builder.maxAge(getMaxAge(jwtClaimsSet));
+            }
+
+            var codeChallenge = jwtClaimsSet.getStringClaim("code_challenge");
+
+            if (Objects.nonNull(codeChallenge)) {
+                var parsedCodeChallenge = CodeChallenge.parse(codeChallenge);
+                var parsedCodeChallengeMethod =
+                        CodeChallengeMethod.parse(
+                                jwtClaimsSet.getStringClaim("code_challenge_method"));
+
+                builder.codeChallenge(parsedCodeChallenge, parsedCodeChallengeMethod);
             }
 
             return builder.build();
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/BaseAuthorizeValidator.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/BaseAuthorizeValidator.java
@@ -1,5 +1,8 @@
 package uk.gov.di.authentication.oidc.validators;
 
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallengeMethod;
 import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
 import com.nimbusds.openid.connect.sdk.OIDCClaimsRequest;
 import com.nimbusds.openid.connect.sdk.claims.ClaimsSetRequest;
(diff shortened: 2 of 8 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -161,4 +171,15 @@ private static OIDCClaimsRequest parseClaimsAsString(JWTClaimsSet claimsSet) {
             throw new RuntimeException("Failed to parse OIDC claims", e);
         }
     }
+
+    private static CodeChallengeMethod parseCodeChallengeMethod(JWTClaimsSet claimsSet)
```
sw-mattw, line 175: Feel a bit mixed about this, rather than just setting the code challenge method to `S256`, but decided to leave this to the validator and handle anything it passes through

Easy on changing this so please comment if so

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/QueryParamsAuthorizeValidator.java
```diff
@@ -102,13 +104,27 @@ public Optional<AuthRequestError> validate(AuthenticationRequest authRequest) {
                             state));
         }
 
+        var codeChallengeMethod = authRequest.getCodeChallengeMethod();
+
         // TODO: ATO-1371: Delete the following log once we have verified that
         //  no RPs are using the code challenge method "plain" AND we have
         //  checked which RPs, if any, are using PKCE already.
```
Ryan-Andrews99, line 111: I think we should be able to clear up the logging now 😄

oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -84,6 +87,13 @@ public static AuthenticationRequest transform(AuthenticationRequest authRequest)
                 builder.maxAge(getMaxAge(jwtClaimsSet));
             }
 
+            var codeChallenge = jwtClaimsSet.getStringClaim("code_challenge");
```
Ryan-Andrews99, line 90: Could we move this into the non-null check?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
```diff
@@ -161,4 +171,15 @@ private static OIDCClaimsRequest parseClaimsAsString(JWTClaimsSet claimsSet) {
...
         }
     }
+
+    private static CodeChallengeMethod parseCodeChallengeMethod(JWTClaimsSet claimsSet)
+            throws ParseException {
+        var codeChallengeMethod = claimsSet.getStringClaim("code_challenge_method");
+
+        if (Objects.isNull(codeChallengeMethod)) {
+            return CodeChallengeMethod.getDefault();
+        }
```
Ryan-Andrews99, line 181: I think this getDefault() method returns plain, so not something we want to use here as we're gonna retrieve this value at the token endpoint to check with the `code_verifier`.  What if we change this to just take a string, or inline the work under the non-null check:

```
if (Objects.nonNull(codeChallenge)) {
              var codeChallenge = CodeChallenge.parse( jwtClaimsSet.getStringClaim("code_challenge"));
// Our validation has checked this value is present
              var codeChallengeMethod = CodeChallengeMethod.parse(claimsSet.getStringClaim("code_challenge_method"));
              
              builder.codeChallenge(
                        codeChallenge, codeChallengeMethod);
              
            }

```

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -188,13 +188,24 @@ public Optional<AuthRequestError> validate(AuthenticationRequest authRequest)
...
+
             // TODO: ATO-1371: Delete once verified that no RPs are using code challenge method
             // "plain" AND checked which (if any) RPs are using PKCE already.
-            var codeChallengeMethod = jwtClaimsSet.getClaim("code_challenge_method");
             if (Objects.nonNull(codeChallengeMethod)) {
                 LOG.info("authRequest code challenge method is '{}'", codeChallengeMethod);
             }
 
+            var codeChallenge = jwtClaimsSet.getStringClaim("code_challenge");
+
```
Ryan-Andrews99, line 200: Think we can delete the logging. Also we should move these `getStringClaim` in the non-null check block

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/QueryParamsAuthorizeValidator.java
```diff
@@ -102,13 +104,27 @@ public Optional<AuthRequestError> validate(AuthenticationRequest authRequest) {
                             state));
         }
 
+        var codeChallengeMethod = authRequest.getCodeChallengeMethod();
```
Ryan-Andrews99, line 107: Same here maybe move into non-null block

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/BaseAuthorizeValidator.java
```diff
@@ -93,4 +96,45 @@ protected boolean areClaimsValid(
...
+            logErrorInProdElseWarn(
+                    String.format(
+                            "Invalid value for PKCE code_challenge_method - only '%s' is supported. code_challenge_method value in request: %s",
+                            CodeChallengeMethod.S256.getValue(), codeChallengeMethod));
+            return Optional.of(
+                    new ErrorObject(
+                            OAuth2Error.INVALID_REQUEST_CODE,
+                            "Invalid value for PKCE code_challenge_method parameter."));
+        }
+
```
Ryan-Andrews99, line 137: Most of these error messages can probably omit the term PKCE, eg: `"Invalid value for code_challenge_method parameter.`

Ryan-Andrews99 (commented):
Looks good to me! The comments are really small so should be quick to address. Outside of those maybe add an integration test in `integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java`. Let me know if you need any help with that or testing in dev 😄

oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/RequestObjectToAuthRequestHelper.java
sw-mattw: We use this in the condition and then within the if block - getStringClaim should still return null if the code challenge isn't provided

The alternative is to repeat `jwtClaimsSet.getStringClaim("code_challenge")` in both the condition and if block

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
Ryan-Andrews99: Same response as other comment

Ryan-Andrews99 (approved):

PR #6016, opened 2025-02-25
ATO-1469: write to new identity credentials table

GHSwallow:
### Wider context of change
This is part of the identity credentials migration work.

### What’s changed
In this PR, we write to the new table.

### Manual testing
I've done some logging here to verify clientSessionId is always defined:
https://github.com/govuk-one-login/authentication-api/pull/6013

Testing in dev:
 - I've ran through an identity journey in dev and have confirmed that this works
 - Saw the identity credentials in the new table
### Checklist
- [x] Lambdas have correct permissions for the resources they're accessing. **tbd**
- [x] Changes have been made to contract tests or not required. **Not Needed**
- [x] Changes have been made to the simulator or not required. **Not Needed**
- [x] Changes have been made to stubs or not required. **Not Needed**
- [x] Successfully deployed to authdev or not required. **Not Needed**
- [x] Successfully run Authentication acceptance tests against sandpit or not required. **Not Needed**

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/DynamoIdentityService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/DynamoIdentityService.java
@@ -55,11 +55,12 @@ public class DynamoIdentityService {
     }
 
     public void saveIdentityClaims(
+            String clientSessionId,
             String subjectID,
             Map<String, String> additionalClaims,
             String ipvVot,
             String ipvCoreIdentity) {
-        var identityCredentials =
+        var authIdentityCredentials =
                 new AuthIdentityCredentials()
                         .withSubjectID(subjectID)
                         .withAdditionalClaims(additionalClaims)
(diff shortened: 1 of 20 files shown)


GHSwallow:
31 lines of uncovered code are from the new OrchIdentityCredentials dynamoDb class. Coverage is otherwise 100%, so I'm happy to ignore SonarCloud here

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/DynamoIdentityService.java
```diff
@@ -63,8 +63,9 @@ public Optional<AuthIdentityCredentials> getIdentityCredentials(String subjectID
                 .filter(t -> t.getTimeToExist() > NowHelper.now().toInstant().getEpochSecond());
     }
 
-    public void deleteIdentityCredentials(String subjectID) {
+    public void deleteIdentityCredentials(String clientSessionId, String subjectID) {
         authIdentityCredentialsDynamoService.delete(subjectID);
+        orchIdentityCredentialsDynamoService.delete(clientSessionId);
```
CarlyG55, line 68: Do we want to do this in this PR? How does it behave if a user is mid journey in the deployment?

Ryan-Andrews99: Hmm so the base dynamo service checks if the record is present before calling delete so think it would be safe

cearl1: It looks like `DeleteItem` doesn't throw an error if an item with that key does not exist, so this should be fine?

At https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html
> Unless you specify conditions, the DeleteItem is an idempotent operation; running it multiple times on the same item or attribute does not result in an error response.

It's worth testing though!

cearl1 (approved):

PR #6024, opened 2025-02-26
ATO-1472: Write to and delete from new client session store

cearl1:
### Wider context of change

A previous PR added a new dynamo client session store, as well as a service to interact with the store. We would like to migrate away from using the shared redis store and use this dynamo store instead.

### What’s changed

This PR writes to the dynamo client session store wherever we write to the redis client session store. Since there was only 1 instance of the delete method being used, i've also included deleting from the dynamo client session store in this ticket.

### Manual testing

Deployed to dev and sandpit:
- Did a basic auth journey, saw the client session was created in the new dynamo table. 
- Logged out and saw the client session was deleted from the new dynamo table.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -49,6 +49,7 @@ import uk.gov.di.orchestration.shared.entity.ClientSession;
 import uk.gov.di.orchestration.shared.entity.CredentialTrustLevel;
 import uk.gov.di.orchestration.shared.entity.CustomScopeValue;
 import uk.gov.di.orchestration.shared.entity.ErrorResponse;
+import uk.gov.di.orchestration.shared.entity.OrchClientSessionItem;
 import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
 import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
 import uk.gov.di.orchestration.shared.entity.Session;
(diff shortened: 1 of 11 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -552,6 +567,7 @@ private APIGatewayProxyResponseEvent handleDocAppJourney(
 
         clientSessionService.storeClientSession(
                 clientSessionId, clientSession.setDocAppSubjectId(subjectId));
+        orchClientSessionService.storeClientSession(orchClientSession);
```
Ryan-Andrews99, line 570: Do we not need to set the docAppSubjectId here?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
Ryan-Andrews99, line 1: Can't really see any assertion that we're setting the id token on the client session here. I know this is more something you've inherited but seems like an important thing we should assert against (given I'm sure that not setting it would be bad)

Ryan-Andrews99 (commented):
Really good, a couple of small comments. We're able to do identity journeys in dev now (https://github.com/govuk-one-login/orch-stubs/pull/226) so would be nice to run through one of those before merging

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
cearl1: Oooh yep must have missed that, thanks!

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
cearl1: I updated the integration test (`TokenIntegrationTest`) to assert on the client session id token

Ryan-Andrews99: Nice thanks for adding!

PR #6029, opened 2025-02-27
Aut 3848/add new service retrieve

BeckaL:
# authentication-api PR template picker

> [!TIP]
> Please go to the `Preview` tab and select the appropriate sub-template

## Templates

- [Auth Team](?expand=1&template=auth_template.md)
- [Orchestration Team](?expand=1&template=orch_template.md)

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/GetMfaMethodsHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/GetMfaMethodsHandler.java
@@ -1,0 +1,19 @@
+package uk.gov.di.accountmanagement.lambda;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+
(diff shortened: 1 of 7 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoMfaMethodsService.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public DynamoMfaMethodsService(ConfigurationService configurationService) {
+        this.dynamoService = new DynamoService(configurationService);
+    }
+
+    @Override
+    public List<MfaData> getMfaMethods(String email) {
+        var userProfile = dynamoService.getUserProfileByEmail(email);
+        var userCredentials = dynamoService.getUserCredentialsFromEmail(email);
+        var enabledAuthAppMethod = getPrimaryMFAMethod(userCredentials);
```
andrew-moores, line 21: Do we need to know that primary == enabled here?  Could we just call it primaryMFAMethod?  And is it guaranteed to be an auth app?

andrew-moores (approved):
Couple of minor comments that we can resolve after we start using the new service.

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoMfaMethodsService.java
BeckaL: So it's currently guaranteed to be an auth app, and will always be for a non-migrated user. For migrated users, it could be either. I'd be pro renaming the helper method to "getEnabledMfaMethod" because that's what it actually does, so it was a bit of clarifying naming here.

PR #6057, opened 2025-03-03
Ato 1372/update token endpoint support pkce

isaac-GDS:
### Wider context of change

As part of the PKCE work , at the token endpoint we take the code challenge from the authRequestParams and validated by the code verifier passed through the request

### What’s changed

Added logic in Validating code challenge and verifier
Wrapped in feature flag and code verifier non null check

### Manual testing

Ran through RP stub 

### How to review

 - Look at validation logic
 - Error response
 - enabling PKCE logic

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
@@ -22,6 +22,9 @@ import com.nimbusds.oauth2.sdk.id.Audience;
 import com.nimbusds.oauth2.sdk.id.ClientID;
 import com.nimbusds.oauth2.sdk.id.State;
 import com.nimbusds.oauth2.sdk.id.Subject;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallenge;
+import com.nimbusds.oauth2.sdk.pkce.CodeChallengeMethod;
+import com.nimbusds.oauth2.sdk.pkce.CodeVerifier;
 import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import com.nimbusds.oauth2.sdk.token.RefreshToken;
 import com.nimbusds.oauth2.sdk.token.Tokens;
@@ -53,6 +56,7 @@ import uk.gov.di.orchestration.shared.entity.VectorOfTrust;
 import uk.gov.di.orchestration.shared.helpers.IdGenerator;
 import uk.gov.di.orchestration.shared.helpers.NowHelper;
 import uk.gov.di.orchestration.shared.serialization.Json;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
 import uk.gov.di.orchestration.sharedtest.basetest.ApiGatewayHandlerIntegrationTest;
 import uk.gov.di.orchestration.sharedtest.extensions.OrchClientSessionExtension;
 import uk.gov.di.orchestration.sharedtest.extensions.RpPublicKeyCacheExtension;
@@ -97,7 +101,29 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
     private static final String REFRESH_TOKEN_PREFIX = "REFRESH_TOKEN:";
     private static final String REDIRECT_URI = "http://localhost/redirect";
     private static final Long AUTH_TIME = NowHelper.now().toInstant().getEpochSecond() - 120L;
+    private final CodeVerifier CODE_VERIFIER = new CodeVerifier();
+    private final String CODE_CHALLENGE_STRING = createCodeChallengeFromCodeVerifier(CODE_VERIFIER);
     private static final String CLIENT_SESSION_ID = "a-client-session-id";
+
+    protected static final ConfigurationService configuration =
+            new IntegrationTestConfigurationService(
+                    externalTokenSigner,
+                    storageTokenSigner,
+                    ipvPrivateKeyJwtSigner,
+                    spotQueue,
+                    docAppPrivateKeyJwtSigner,
+                    configurationParameters) {
+
+                @Override
+                public boolean isPkceEnabled() {
+                    return true;
+                }
+
+                @Override
+                public String getTxmaAuditQueueUrl() {
+                    return txmaAuditQueue.getQueueUrl();
+                }
+            };
 
     @RegisterExtension
     public static final RpPublicKeyCacheExtension rpPublicKeyCacheExtension =
@@ -109,7 +135,7 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
 
     @BeforeEach
     void setup() {
-        handler = new TokenHandler(TXMA_ENABLED_CONFIGURATION_SERVICE, redisConnectionService);
+        handler = new TokenHandler(configuration, redisConnectionService);
         txmaAuditQueue.clear();
     }
 
@@ -536,7 +562,8 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
                         Optional.of("Cl.Cm"),
                         Optional.empty(),
                         Optional.of("test-client-1"),
-                        "test-auth-code-2");
+                        "test-auth-code-2",
+                        CODE_VERIFIER.getValue());
         var response =
                 makeTokenRequestWithClientSecretPost(
                         "test-client-1", baseTokenRequest, clientSecret);
@@ -565,13 +592,46 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
                         Optional.of("Cl.Cm"),
                         Optional.empty(),
                         Optional.of("test-client-1"),
-                        "test-auth-code-2");
+                        "test-auth-code-2",
+                        CODE_VERIFIER.getValue());
         var response =
                 makeTokenRequestWithPrivateKeyJWT(
                         "test-client-1", baseTokenRequest, keyPair.getPrivate());
 
         assertThat(response, hasStatus(400));
         assertThat(response, hasBody(OAuth2Error.INVALID_GRANT.toJSONObject().toJSONString()));
+    }
+
+    @Test
+    void shouldCallTokenResourceAndWarnWhenPkceValidationFails() throws Exception {
+        CodeVerifier invalidCodeVerifier = new CodeVerifier();
+        KeyPair keyPair = KeyPairHelper.GENERATE_RSA_KEY_PAIR();
+        Scope scope =
+                new Scope(
+                        OIDCScopeValue.OPENID.getValue(), OIDCScopeValue.OFFLINE_ACCESS.getValue());
+        userStore.signUp(TEST_EMAIL, "password-1", new Subject());
+        registerClientWithPrivateKeyJwtAuthentication(
+                keyPair.getPublic(), scope, SubjectType.PUBLIC);
+        var baseTokenRequest =
+                constructBaseTokenRequest(
+                        scope,
+                        Optional.empty(),
+                        Optional.empty(),
+                        Optional.of(CLIENT_ID),
+                        new AuthorizationCode().toString(),
+                        invalidCodeVerifier.getValue());
+
+        var response = makeTokenRequestWithPrivateKeyJWT(baseTokenRequest, keyPair.getPrivate());
+
+        assertThat(response, hasStatus(400));
+        assertThat(
+                response,
+                hasBody(
+                        new ErrorObject(
+                                        OAuth2Error.INVALID_GRANT_CODE,
+                                        "PKCE code verification failed")
+                                .toJSONObject()
+                                .toJSONString()));
     }
 
     private SignedJWT generateSignedRefreshToken(Scope scope, Subject publicSubject) {
@@ -658,7 +718,8 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
                                 new ClientID(CLIENT_ID),
                                 URI.create("http://localhost/redirect"))
                         .state(state)
-                        .nonce(nonce);
+                        .nonce(nonce)
+                        .customParameter("code_challenge", CODE_CHALLENGE_STRING);
         claimsRequest.ifPresent(builder::claims);
         vtr.ifPresent(v -> builder.customParameter("vtr", v));
 
@@ -707,7 +768,12 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
             Optional<String> clientId)
             throws Json.JsonException {
         return constructBaseTokenRequest(
-                scope, vtr, oidcClaimsRequest, clientId, new AuthorizationCode().toString());
+                scope,
+                vtr,
+                oidcClaimsRequest,
+                clientId,
+                new AuthorizationCode().toString(),
+                CODE_VERIFIER.getValue());
     }
 
     private Map<String, List<String>> constructBaseTokenRequest(
@@ -715,7 +781,8 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
             Optional<String> vtr,
             Optional<OIDCClaimsRequest> oidcClaimsRequest,
             Optional<String> clientId,
-            String code)
+            String code,
+            String codeVerifier)
             throws Json.JsonException {
         List<VectorOfTrust> vtrList = List.of(VectorOfTrust.getDefaults());
         if (vtr.isPresent()) {
@@ -742,6 +809,7 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
         clientId.map(cid -> customParams.put("client_id", Collections.singletonList(cid)));
         customParams.put("code", Collections.singletonList(code));
         customParams.put("redirect_uri", Collections.singletonList(REDIRECT_URI));
+        customParams.put("code_verifier", Collections.singletonList(codeVerifier));
         return customParams;
     }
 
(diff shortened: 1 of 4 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
```diff
@@ -703,6 +715,108 @@ void shouldReturn400IfRedirectUriDoesNotMatchRedirectUriFromAuthRequest()
...
+                        SUCCESSFUL_TOKEN_ISSUED.getValue(),
+                        Map.of(
+                                ENVIRONMENT.getValue(),
+                                configurationService.getEnvironment(),
+                                CLIENT.getValue(),
+                                CLIENT_ID));
+    }
+
+    @Test
+    void shouldReturn400IfCodeVerificationDoesNotExist()
```
CarlyG55, line 769: Is this not doing the same thing as the test above?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
```diff
@@ -690,7 +697,12 @@ void shouldReturn400IfRedirectUriDoesNotMatchRedirectUriFromAuthRequest()
 
         APIGatewayProxyResponseEvent result =
                 generateApiGatewayRequest(
-                        privateKeyJWT, authCode, "http://invalid-redirect-uri", CLIENT_ID, true);
+                        privateKeyJWT,
+                        authCode,
+                        "http://invalid-redirect-uri",
+                        CLIENT_ID,
+                        true,
+                        CODE_VERIFIER_ENCODED_STRING);
```
CarlyG55, line 705: What's this change for?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -245,6 +248,27 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
                     400, e.getErrorObject().toJSONObject().toJSONString());
         }
 
+        if (configurationService.isPkceEnabled()
+                && Objects.nonNull(requestBody.get("code_verifier"))) {
+            var codeChallengeString = authRequest.getCodeChallenge().toString();
+            var codeVerifierString = requestBody.get("code_verifier");
+
+            if (codeVerifierString.isBlank()) {
```
CarlyG55, line 256: What's the difference between this and the non null check above?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -245,6 +248,16 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
                     400, e.getErrorObject().toJSONObject().toJSONString());
         }
 
+        var codeChallengeString = authRequest.getCodeChallenge().toString();
```
Ryan-Andrews99: I have a feeling this `getCodeChallenge().toString()` might throw a null pointer exception given it's perfectly valid for `getCodeChallenge` to return null (ie no code challenge in the request)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -245,6 +248,16 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
...
+        var codeVerifier = new CodeVerifier(requestBody.get("code_verifier"));
```
Ryan-Andrews99: It is a plausible scenario they don't include a verifier so we probably want to handle this get returning null more gracefully

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -321,11 +334,11 @@ private APIGatewayProxyResponseEvent processRefreshTokenRequest(
             tokenStore = objectMapper.readValue(refreshToken.get(), RefreshTokenStore.class);
         } catch (JsonException | NoSuchElementException | IllegalArgumentException e) {
             LOG.warn("Refresh token not found with given key");
-            return generateInvalidGrantCodeApiGatewayProxyResponse();
+            return generateInvalidGrantCodeRefreshTokenApiGatewayProxyResponse();
```
Ryan-Andrews99, line 348: Think it would be nice if we actually kept this a generic function and made the description a parameter, as that's the main bit that changes

integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
```diff
@@ -743,4 +752,16 @@ private Map<String, List<String>> createAuthCodeForClient(
         customParams.put("redirect_uri", Collections.singletonList(REDIRECT_URI));
         return customParams;
     }
+
+    private String createCodeVerifierURLEncodedString() {
+        SecureRandom secureRandom = new SecureRandom();
+        byte[] codeVerifier = new byte[32];
+        secureRandom.nextBytes(codeVerifier);
+        return Base64.getUrlEncoder().withoutPadding().encodeToString(codeVerifier);
+    }
```
Ryan-Andrews99, line 782: Think you can just use the constructor for `new CodeVerifer()` here, it should generate a complaint secure string (and call `getValue` where a string is needed)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -248,7 +248,8 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
                     400, e.getErrorObject().toJSONObject().toJSONString());
         }
 
-        if (configurationService.isPkceEnabled()) {
+        if (configurationService.isPkceEnabled()
+                && Objects.nonNull(requestBody.get("code_verifier"))) {
             var codeChallengeString = authRequest.getCodeChallenge().toString();
             var codeVerifierString = requestBody.get("code_verifier");
```
Ryan-Andrews99, line 254: Same as other comments but think we want to wrap some of this in more null safety/error returning if values are missing or not present. Eg Code verifier but no code challenge or code challenge but no verifier. Happy to talk through this more if you like 🙂

Ryan-Andrews99 (commented):
Looks really close, think it just needs a bit more asserting on exactly if/when things are defined

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Since this is inside `isPkceEnabled` block im assuming it should always return and therefore error on `null`

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Have added validation in a later commit

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Cool - can refactor.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Will add a null check for the `code_challenge` .

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
isaac-GDS: Slight difference in that this test does not pass a code verifier on line 800. Still asserting the same error.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
isaac-GDS: Added a parameter on the method so we are able to pass through a code verifier to test when it is null. Specifically line 800.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: I believe that the first check this if the object exists but this checks if the value is null i.e: `code_verifier: ""` . `shouldReturn400IfCodeVerificationDoesNotExist` should test this logic as it enter the block but errors.

integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
```diff
@@ -95,14 +100,37 @@ public class TokenIntegrationTest extends ApiGatewayHandlerIntegrationTest {
...
+                @Override
+                public boolean isPkceEnabled() {
+                    return true;
+                }
+
+                @Override
+                public String getTxmaAuditQueueUrl() {
+                    return txmaAuditQueue.getQueueUrl();
+                }
+            };
```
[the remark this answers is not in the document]
isaac-GDS: Yes we do this in other integration tests, example: `AuthorisationIntegrationTests`

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Since there are multiple uses of each error response and therefore description have kept the specific error response but added back the generic one. What do you think ?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
isaac-GDS: Ah confusingly the code_challenge is set incorrectly when is should be correct. Will update and hopefully make this test clearer.

integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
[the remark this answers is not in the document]
CarlyG55: Cool, makes sense

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
CarlyG55: Although PKCE is enabled, clients won't always be sending code challenge though - it's optional

CarlyG55:
@isaac-GDS do you mind adding some logging in this PR to check if there are any RPs that are sending us invalid code_verifiers right now? Ie not behind the feature flag? Thanks!

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
```diff
@@ -703,6 +714,159 @@ void shouldReturn400IfRedirectUriDoesNotMatchRedirectUriFromAuthRequest()
...
+                                                                        "Incorrect Code Challenge")
+                                                                .toParameters(),
+                                                        LocalDateTime.now(),
+                                                        List.of(mock(VectorOfTrust.class)),
+                                                        CLIENT_NAME))
+                                        .setAuthTime(AUTH_TIME)
+                                        .setClientId(CLIENT_ID)));
+
+        APIGatewayProxyResponseEvent result =
+                generateApiGatewayRequest(privateKeyJWT, authCode, CLIENT_ID, true);
```
CarlyG55, line 748: Shouldn't this be passing in a code verifier?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Ah gotcha - will update

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
```diff
@@ -703,6 +715,108 @@ void shouldReturn400IfRedirectUriDoesNotMatchRedirectUriFromAuthRequest()
...
+        verify(cloudwatchMetricsService, never())
+                .incrementCounter(
+                        SUCCESSFUL_TOKEN_ISSUED.getValue(),
+                        Map.of(
+                                ENVIRONMENT.getValue(),
+                                configurationService.getEnvironment(),
+                                CLIENT.getValue(),
+                                CLIENT_ID));
+    }
+
```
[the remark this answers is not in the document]
isaac-GDS: Yes, as in `code_challenge:  null || ""` vs code_challenge not being set at all. The authRequestParms just return null for all these cases. Will add a test to make this more obvious.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Since there isnt a way of distinguishing between code_challenger not being sent through or no code challenge. Assuming that if the code_verifier is there we always expect a code_challenge. 

So have kept that same.

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
isaac-GDS: Have added PKCE as the general happy path but will update tests so its just the PKCE work.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -252,6 +256,34 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
...
+                pkceValidated = "is successful";
+            } catch (Exception e) {
+                pkceValidated = "has failed";
+            }
+            LOG.info("PKCE validation {}", pkceValidated);
+        }
+
+        if (configurationService.isPkceEnabled() && Objects.nonNull(codeChallenge)) {
+
+            if (codeVerifierString.isBlank()) {
```
Ryan-Andrews99, line 275: I know there was a conversation about this before so sorry to bring it back up but this line ends throwing a NullPointerException if there is no code verifier attached. I think we don't need this `isBlank()` check. When we call the `new CodeVerifier()` constructor it will validate the length and char set, so should be fine

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
Ryan-Andrews99: Yep happy with this!

integration-tests/src/test/java/uk/gov/di/authentication/api/TokenIntegrationTest.java
Ryan-Andrews99, line 1: Think it would be nice to have one or two PKCE integration tests here

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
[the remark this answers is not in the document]
Ryan-Andrews99: I think we explicitly validate the code_challenge is not an empty string at the authorize endpoint

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
isaac-GDS: Ah yes. Well remembered. The try, catch will be able to return an error from this.

Ryan-Andrews99 (dismissed):

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
isaac-GDS: Have removed.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
Ryan-Andrews99: This has been updated, the code now handles the different PKCE cases

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
```diff
@@ -267,6 +270,26 @@ public APIGatewayProxyResponseEvent tokenRequestHandler(APIGatewayProxyRequestEv
...
+        var codeChallenge = authRequest.getCodeChallenge();
+        var codeVerifierString = requestBody.get("code_verifier");
+
+        try {
+            if (isPKCEValid(
+                    Optional.ofNullable(codeChallenge), Optional.ofNullable(codeVerifierString))) {
+                LOG.info("PKCE validation passed");
+            }
+        } catch (Exception e) {
+            LOG.info("PKCE FAILED {}", e);
```
Ryan-Andrews99, line 283: Sonar is pointing out that this isn't passing the exception into the log message, think you need something like:
```
LOG.info("PKCE FAILED {}", e.getMessage());
```

Ryan-Andrews99 (approved):

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
Ryan-Andrews99: Not sure what this referred to but happy with the changes and tests overall so resolving

PR #6058, opened 2025-03-03
ATO-1419: Verify client session databases are in sync

cearl1:
### Wider context of change

We are currently writing to the dynamo client session store in addition to the redis client session store. We would like to migrate to only using the dynamo store.

### What’s changed

When the client sessions are read from dynamo, this PR adds logging to compare the sessions from the 2 stores. If they do not match, a log message will be generated. If they match, no log messages will be generated

I've also updated the tests so the test output remains accurate. I've done this by making the logging helper throw an exception, run all the relevant tests, and fix them until none were failing, then remove the runtime exception from the logging helper.

(Reviewing commit-by-commit might help break down this chunky PR)

### Manual testing

Deployed to dev and performed a full identity journey. Ran without issues.

### Checklist

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
@@ -37,6 +37,7 @@ import uk.gov.di.orchestration.shared.services.DocAppAuthorisationService;
 import uk.gov.di.orchestration.shared.services.JwksService;
 import uk.gov.di.orchestration.shared.services.KmsConnectionService;
 import uk.gov.di.orchestration.shared.services.NoSessionOrchestrationService;
+import uk.gov.di.orchestration.shared.services.OrchClientSessionService;
 import uk.gov.di.orchestration.shared.services.OrchSessionService;
 import uk.gov.di.orchestration.shared.services.RedirectService;
 import uk.gov.di.orchestration.shared.services.RedisConnectionService;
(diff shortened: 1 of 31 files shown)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
```diff
@@ -222,6 +229,11 @@ public APIGatewayProxyResponseEvent docAppCallbackRequestHandler(
                             .getClientSession(clientSessionId)
                             .orElseThrow(
                                     () -> new DocAppCallbackException("ClientSession not found"));
+            var orchClientSession =
+                    orchClientSessionService.getClientSession(clientSessionId).orElse(null);
```
Ryan-Andrews99, line 233: I know for this we shouldn't throw like the above, but we should probably come back and do this at some point to match the above?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -332,6 +339,11 @@ private ClientSession getClientSession(APIGatewayProxyRequestEvent input)
                 clientSessionService
                         .getClientSessionFromRequestHeaders(input.getHeaders())
                         .orElse(null);
+        var orchClientSession =
+                orchClientSessionService
+                        .getClientSessionFromRequestHeaders(input.getHeaders())
+                        .orElse(null);
```
Ryan-Andrews99, line 345: Took me a minute to realise why it's done this way, think it's because the auth code handler is a machine to machine interaction, so no cookie 🤔

Ryan-Andrews99 (commented):
A really nicely laid our PR, the commits made this slightly daunting PR a really easy review! I've left two comments, the identity progress one is the main thing before approving. But great work!

doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/lambda/DocAppCallbackHandler.java
cearl1: Yep agreed, I'll add a comment to the "swap over" ticket to remember to do this.

ipv-api/src/test/java/uk/gov/di/authentication/ipv/lambda/IdentityProgressFrontendHandlerTest.java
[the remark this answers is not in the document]
cearl1: Ahh i must have forgotten to add this when doing the permissions, nice spot! i'll do that here

Ryan-Andrews99 (approved):

PR #6095, opened 2025-03-06
ATO-1518: Remove doc app code from auth backend

cearl1:
### Wider context of change

As part of the client session migration work, we identified that the auth backend was checking if the doc-app scope was present (using the client session). It seems that this code was leftover from the auth/orch split, as this check is now done at the start of a journey in orch, and the user is directed to a different backend instead, therefore this code for checking doc-app users should be able to be removed from the auth backend. 

### What’s changed

This PR removes the code that checks if an auth request has a `doc-checking-app` scope from the auth backend. It also removes related tests, as well as simplifying the code by removing parameters that are no longer used.

### Manual testing

Tested an auth and identity journey in sandpit

### Checklist

- [x] Changes have been made to contract tests or not required. n/a?
- [x] Successfully deployed to authdev or not required. Done


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -148,22 +147,20 @@ public UserStartInfo buildUserStartInfo(
...
-        var docCheckingAppUser = DocAppUserHelper.isDocCheckingAppUser(userContext);
-        if (Boolean.FALSE.equals(docCheckingAppUser)) {
-            var clientRegistry = userContext.getClient().orElseThrow();
-            identityRequired =
-                    IdentityHelper.identityRequired(
-                            userContext.getClientSession().getAuthRequestParams(),
-                            clientRegistry.isIdentityVerificationSupported(),
-                            identityEnabled);
-        }
+        var docCheckingAppUser = false;
```
isaac-GDS, line 150: Since this is always false are we looking to remove this and it usage in the userStartInfo ?

cearl1: Eventually yeah. The auth frontend uses it at the moment though so i'll need to update that first. I'll make a note of this in the ticket / create a follow-up ticket

Ryan-Andrews99 (approved):

PR #6115, opened 2025-03-10
Bau: move method management feature flag to terraform

BeckaL:
## What

Introduce terraform feature flags for the new mfa method management api endpoints. This replaces logic to look up the environment and disable based on that, and brings us more in line with how we do feature flags elsewhere.

The switch is off in integration and production.

## How to review

1. Code review
1. If you like, you can also deploy to an env toggling the relevant tf var for that environment on / off, and test that api via api gateway. You should immediately get a 400 for either the get or post endpoint (no need to construct a body) when the flag is off, and a relevant response when the flag is on. I've already done this bit of testing so up to you if you'd also like to see.


andrew-moores (approved):
Assume we will address Sonar issue later.

PR #6117, opened 2025-03-10
AUT-2910: Send `password_reset` to TICF CRI API

mattvot:
## What

When a user has attempted to reset their password and when they successfully reset their password it is now recorded in their session. This is then used, in combination with their authenticated state, to determine if the `password_reset=Y` attribute should be added to the payload to the TICF CRI API at the end of the users sign in journey.

This happens if either:
- the user is authenticated and has successfully reset their password
- the user is failed to sign in but attempted to reset their password

At this point the TICF CRI API is not called outside of a successful sign in, so while the code is configured for the attempted password reset, it won't be sent to TICF until we support the CRI API for non-successful sign in journeys.

## How to review

1. Code Review
1. Deploy to a dev environment
1. Sign in as usual
1. Start again and sign in with a password reset
1. Filtering CloudWatch logs in the `*-account-interventions-lambda` with "Invoking TICF CRI with payload" and check the payload for your password reset journey contains `"password_reset":"Y"`, but does not contain it for the sign in journey without the reset

## Checklist

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
+++ b/shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.entity;
 
 import com.google.gson.annotations.Expose;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
+import uk.gov.di.authentication.shared.entity.AuthSessionItem.ResetPasswordState;
 import uk.gov.di.authentication.shared.validation.Required;
 
 import java.util.List;
@@ -19,13 +20,14 @@ public record TICFCRIRequest(
             List<String> vtr,
             String journeyId,
             boolean authenticated,
-            AuthSessionItem.AccountState accountState) {
+            AuthSessionItem.AccountState accountState,
+            ResetPasswordState resetPasswordState) {
         return new TICFCRIRequest(
                 internalPairwiseId,
                 vtr,
                 journeyId,
                 authenticated ? "Y" : "N",
                 accountState == AuthSessionItem.AccountState.NEW ? "Y" : null,
-                null);
+                resetPasswordState.equals(ResetPasswordState.SUCCEEDED) ? "Y" : null);
     }
 }
--- a/shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
+++ b/shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
@@ -22,12 +22,17 @@ public record TICFCRIRequest(
             boolean authenticated,
             AuthSessionItem.AccountState accountState,
             ResetPasswordState resetPasswordState) {
+        boolean passwordResetSuccess = resetPasswordState.equals(ResetPasswordState.SUCCEEDED);
+        boolean reportablePasswordAttempted =
+                (!authenticated && resetPasswordState.equals(ResetPasswordState.ATTEMPTED));
+        boolean passwordReset = passwordResetSuccess || reportablePasswordAttempted;
+
         return new TICFCRIRequest(
                 internalPairwiseId,
                 vtr,
                 journeyId,
                 authenticated ? "Y" : "N",
                 accountState == AuthSessionItem.AccountState.NEW ? "Y" : null,
-                resetPasswordState.equals(ResetPasswordState.SUCCEEDED) ? "Y" : null);
+                passwordReset ? "Y" : null);
     }
 }
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
@@ -26,6 +26,7 @@ public class AuthSessionItem {
 
     public enum ResetPasswordState {
         NONE,
+        ATTEMPTED,
         SUCCEEDED,
     }
 
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
@@ -169,4 +169,26 @@ public class AuthSessionItem {
         this.emailAddress = emailAddress;
         return this;
     }
+
+    /**
+     * Return a string representation of the instance that is safe to record in logs (e.g. does not
+     * contain PII)
+     */
+    public String toLogSafeString() {
+        return "AuthSessionItem{sessionId = '"
+                + sessionId
+                + "', verifiedMfaMethodType = '"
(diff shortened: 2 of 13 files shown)


shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
```diff
@@ -11,21 +12,30 @@ public record TICFCRIRequest(
...
-            AuthSessionItem.AccountState accountState) {
+            AuthSessionItem.AccountState accountState,
+            AuthSessionItem.ResetPasswordState resetPasswordState) {
+        boolean passwordReset = false;
+        if (authenticated
+                && resetPasswordState.equals(AuthSessionItem.ResetPasswordState.SUCCEEDED))
+            passwordReset = true;
+        else if (!authenticated
+                && resetPasswordState.equals(AuthSessionItem.ResetPasswordState.ATTEMPTED))
+            passwordReset = true;
```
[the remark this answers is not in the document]
ethanmills: I also think the test for `authorised` is unnecessary in the first case, either way the value is true

shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
```diff
@@ -11,21 +12,30 @@ public record TICFCRIRequest(
         @Expose @Required List<String> vtr,
         @Expose @Required String govukSigninJourneyId,
         @Expose @Required String authenticated,
-        @Expose String initialRegistration,
-        @Expose String passwordReset) {
+        @Expose @SerializedName("initial_registration") String initialRegistration,
+        @Expose @SerializedName("password_reset") String passwordReset) {
```
ethanmills, line 16: Is this change is necessary, I think we use a serialiser with the correct rules?

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
```diff
@@ -23,10 +24,17 @@ public enum AccountState {
         UNKNOWN
     }
 
+    public enum ResetPasswordState {
+        NOT_APPLICABLE,
```
ethanmills, line 28: Nitpick and happy for this to be left as-is, but perhaps just `NONE`?

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
```diff
@@ -147,4 +169,22 @@ public AuthSessionItem withEmailAddress(String emailAddress) {
         this.emailAddress = emailAddress;
         return this;
     }
+
+    public String toLogSafeString() {
```
ethanmills, line 173: Would you be happy to add a comment to this method making it clear that it can't contain PII please? I initially assumed this means it only contained loggable characters...

ethanmills:
Is there any way to go back to an `ATTEMPTED` state from a `SUCCEEDED` one? I can't think of any, but you know the journeys better than I do

shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
mattvot: I'm probably miss-understanding, but I think so?

22318995a91d416f375329aa0b83a150cbf75c6a makes a correction to `initial_registration` as it's serialised to `initialRegistration` in the `AccountInterventionsHandlerTest.java`. `AccountInterventionsHandler.java` does a gson conversion of the instance, `payload = gson.toJson(ticfRequest);`, so it seems that `AccountInterventionsHandler` was serializing incorrectly.

mattvot:
> Is there any way to go back to an ATTEMPTED state from a SUCCEEDED one? I can't think of any, but you know the journeys better than I do

Yes, in that once you have succeeded and completed the auth journey, if you start a new one (well with the orchstub) it seems to use the same session record again, so it's possible to start the reset password journey again and go back to ATTEMPTED.

shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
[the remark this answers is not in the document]
mattvot: Way simpler. Done

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
mattvot: No preference from me, Done

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
mattvot: Done

shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
ethanmills: Happy to leave this, but it feels weirdly inconsistent to also have `govukSigninJourneyId` when TICF CRI needs `govuk_signin_journey_id`

ethanmills:
> > Is there any way to go back to an ATTEMPTED state from a SUCCEEDED one? I can't think of any, but you know the journeys better than I do
> 
> Yes, in that once you have succeeded and completed the auth journey, if you start a new one (well with the orchstub) it seems to use the same session record again, so it's possible to start the reset password journey again and go back to ATTEMPTED.

I'm not worried about going back to ATTEMPTED in that case, but does this mean that:

mattvot:
> silent login with no password reset

True, there is a way in the StartHandler for me to just reset the ResetPasswordStatus back to `NONE` when a new journey starts

ethanmills:
> > silent login with no password reset
> 
> True, there is a way in the StartHandler for me to just reset the ResetPasswordStatus back to `NONE` when a new journey starts. That way the no `password_reset: Y` would be sent unless the user starts to reset their password even if the session is reused.

Yep, lets do that

shared/src/main/java/uk/gov/di/authentication/entity/TICFCRIRequest.java
mattvot: I've undo this change now

mattvot:
> > > silent login with no password reset
> > 
> > 
> > True, there is a way in the StartHandler for me to just reset the ResetPasswordStatus back to `NONE` when a new journey starts. That way the no `password_reset: Y` would be sent unless the user starts to reset their password even if the session is reused.
> 
> Yep, lets do that

Done, added new commit. Not tested in an env yet. Will do tomorrow

mattvot:
@ethanmills I've successfully tested setting the state back to `NONE` in the `StartHandler` for reused sessions on authdev2 this morning

ethanmills (approved):

PR #6118, opened 2025-03-10
ATO-1471: duplicate roles without old policies

Ryan-Andrews99:
### Wider context of change

We've now stopped writing to the old identity credentials table in the auth account and will be removing the resource. To do this we must remove the policies from each role. However these are not neatly at the end of each `policies_to_attach` array, meaning that removing them will cause the each policy to be removed and re-added to the role, causing downtime. To mitigate this, we're duplicating the existing roles with the policies without the policies we plan to remove. This means the role will not need recreating when we swap the roles over later

### What’s changed:
- Duplicated any role which contained a policy to access the old identity credentials table 
- Removed any old identity credentials table access policies from this role

### Manual testing: 
- Deployed to sandpit and observed new roles created fine
### Checklist


ci/terraform/oidc/identity-progress.tf
```diff
@@ -1,3 +1,23 @@
+// ATO-1471: We're duplicating the role without the old identity credentials table
+// access policies to enable us to safely remove them
+module "identity_progress_role_2" {
```
Louisasa, line 3: does this have to be `role_2`? can it not be `role`?

Louisasa (approved):
one minor comment that's more for my understanding :)

ci/terraform/oidc/identity-progress.tf
Ryan-Andrews99: Ah yeah I didn't really think about that 😄

Louisasa (approved):

PR #6121, opened 2025-03-10
Aut 3485: Add SMS scenario create MFA methods

alhcomer:
## What

Adds an initial success and error scenario in the MFAMethodsCreateHandler:
- P3: Auth app user can add SMS
- E1: Invalid public subject throws error

## How to review

1. Code Review
2. Deploy to dev environment with auth app profile set up
3. Test in API gateway area of the AWS console with the path parameter as your publicSubjectId (found in the user profile table) and the following request body and see 200:

{
"mfaMethod": {
    "priorityIdentifier": "BACKUP",
    "method": {
      "mfaMethodType": "SMS",
      "phoneNumber": "07123123123"
        }
    }
}

AND desired response:

{
  "mfaIdentifier": "a-random-uuid",
  "priorityIdentifier": "BACKUP",
  "methodVerified": true,
  "method": {
  "mfaMethodType": "SMS",
  "phoneNumber": "07123123123"
} }

4. See error for an invalid public subject set as the path parameter

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
@@ -105,7 +105,7 @@ public class MFAMethodsCreateHandler
                             "SerializationService::GSON::fromJson",
                             () ->
                                     objectMapper.readValue(
-                                            input.getBody(), MfaMethodCreateRequest.class));
+                                            input.getBody(), MfaMethodCreateRequest.class, true));
 
         } catch (RuntimeException e) {
             LOG.error("Error during JSON deserialization", e);
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthAppMfaDetail.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthAppMfaDetail.java
@@ -1,8 +1,10 @@
 package uk.gov.di.authentication.shared.entity;
 
 import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
 import uk.gov.di.authentication.shared.validation.Required;
 
 public record AuthAppMfaDetail(
-        @Expose @Required MFAMethodType mfaMethodType, @Expose @Required String credential)
+        @Expose @Required @SerializedName("mfaMethodType") MFAMethodType mfaMethodType,
+        @Expose @Required @SerializedName("credential") String credential)
         implements MfaDetail {}
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/SmsMfaDetail.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/SmsMfaDetail.java
@@ -1,8 +1,10 @@
 package uk.gov.di.authentication.shared.entity;
 
 import com.google.gson.annotations.Expose;
+import com.google.gson.annotations.SerializedName;
 import uk.gov.di.authentication.shared.validation.Required;
 
 public record SmsMfaDetail(
-        @Expose @Required MFAMethodType mfaMethodType, @Expose @Required String phoneNumber)
+        @Expose @Required @SerializedName("mfaMethodType") MFAMethodType mfaMethodType,
+        @Expose @Required @SerializedName("phoneNumber") String phoneNumber)
         implements MfaDetail {}


shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoMfaMethodsService.java
```diff
@@ -87,4 +89,32 @@ private List<MfaMethodData> getMfaMethodsForNonMigratedUser(
...
+
+        if (request.mfaMethod().method() instanceof SmsMfaDetail smsMfaDetail) {
+            dynamoService.addMFAMethodSupportingMultiple(
+                    email,
+                    new SmsMfaData(
+                            smsMfaDetail.phoneNumber(),
+                            true,
+                            true,
+                            request.mfaMethod().priorityIdentifier(),
+                            HARDCODED_SMS_MFA_ID));
```
BeckaL, line 109: Here we don't have to hard-code, we can generate a new UUID. The hard-coding was to fix stuff in read.

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoMfaMethodsService.java
```diff
@@ -87,4 +89,32 @@ private List<MfaMethodData> getMfaMethodsForNonMigratedUser(
             return List.of();
         }
     }
+
+    @Override
+    public MfaMethodData postMfaMethod(String email, MfaMethodCreateRequest request)
```
BeckaL, line 94: I think we should change the language here as this is about the database rather than HTTP layer, so 'add' instead of 'post'. Maybe 'addBackupMfaMethod'?

BeckaL (commented):
A couple of initial comments. As far as I can tell, this PR would also mean we add an auth app method for an SMS user, should we add that into tests as well?

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
```diff
@@ -61,33 +74,38 @@ public APIGatewayProxyResponseEvent mfaMethodsHandler(
             return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
         }
 
+        var maybeUserProfile = dynamoService.getOptionalUserProfileFromPublicSubject(subject);
+        if (maybeUserProfile.isEmpty()) {
+            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
```
BeckaL, line 79: Think this should be a 404 no? This is basically user not found

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoMfaMethodsService.java
BeckaL, line 94: Just pass in the data we need here, not the whole request

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
```diff
@@ -61,25 +74,30 @@ public APIGatewayProxyResponseEvent mfaMethodsHandler(
...
         }
 
+        var maybeUserProfile = dynamoService.getOptionalUserProfileFromPublicSubject(subject);
+        if (maybeUserProfile.isEmpty()) {
+            return generateApiGatewayProxyErrorResponse(404, ErrorResponse.ERROR_1056);
+        }
+        String email = maybeUserProfile.get().getEmail();
+
         try {
             MfaMethodCreateRequest mfaMethodCreateRequest = readMfaMethodCreateRequest(input);
```
andrew-moores, line 84: MfaMethodCreateRequest not needed anymore just a MfaMethodDetail?

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandlerTest.java
```diff
@@ -38,27 +50,58 @@ class MFAMethodsCreateHandlerTest {
...
     @Test
-    void shouldReturn200WhenAndHelloWorld() {
+    void shouldReturn200AndCreateMfaSmsMfaMethod() throws InvalidPriorityIdentifierException {
+        when(dynamoService.getOptionalUserProfileFromPublicSubject(TEST_PUBLIC_SUBJECT))
+                .thenReturn(Optional.of(userProfile));
+        when(userProfile.getEmail()).thenReturn(TEST_EMAIL);
+        when(mfaMethodsService.addBackupMfa(any(), any()))
+                .thenReturn(
+                        new MfaMethodData(
+                                HARDCODED_SMS_MFA_ID,
```
andrew-moores, line 81: Don't need to call this hardcoded.  Convention is TEST_SMS_MFA_ID

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandlerTest.java
```diff
@@ -82,22 +125,30 @@ void shouldReturn200WhenAndHelloWorld() {
...
+                        TEST_PUBLIC_SUBJECT);
 
         var result = handler.handleRequest(event, context);
 
         assertThat(result, hasStatus(400));
     }
 
     @Test
-    void shouldReturn400WhenPathParameterIsIncorrect() {
+    void shouldReturn400WhenPathParameterIsEmpty() {
```
andrew-moores, line 142: This isn't testing a missing path param as "incorrect-subject" is being used as the subject id rather a null?

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandlerTest.java
```diff
@@ -125,22 +199,25 @@ void shouldReturn400WhenJsonIsInvalid() {
...
+            MFAMethodType mfaMethodType,
+            String phoneNumber,
+            String publicSubject) {
         var event = new APIGatewayProxyRequestEvent();
 
-        event.setPathParameters(Map.of("publicSubjectId", "helloPath"));
+        event.setPathParameters(Map.of("publicSubjectId", publicSubject));
         event.setBody(
                 format(
                         """
```
andrew-moores, line 211: Consider moving this up and creating a var to aid readability as the formatting will be better and the setBody method will be a one liner.

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
andrew-moores, line 1: This commit message is a bit difficult to parse.  Last two lines won't be helpful in future, is ti worth finding out why this happened?  Otherwise changet/remove those lines.

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
alhcomer: It is not needed anymore, but I think for the sake of making it easier to understand what is actually being sent to the API gateway I'd prefer to keep this object as is. Happy to talk about it though

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
alhcomer: I had a look into it but couldn't find a reason. Have removed the lines from the message

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandlerTest.java
alhcomer: It was testing that the path parameter was empty, I've changed it to make it clearer now

BeckaL (approved):

PR #6140, opened 2025-03-13
Aut 3487/implement delete mfa method

BeckaL:
## Checklist


BeckaL:
Sonar issue relates to an existing TODO comment that is still TODO (and has a ticket for it)

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
```diff
@@ -724,6 +724,18 @@ public void deleteBatchTestUsers(List<String> emailAddresses) {
         }
     }
 
+    public void deleteMfaMethodByIdentifier(String email, String mfaMethodIdentifier) {
+        var dateTime = NowHelper.toTimestampString(NowHelper.now());
+        dynamoUserCredentialsTable.updateItem(
+                dynamoUserCredentialsTable
+                        .getItem(
+                                Key.builder()
+                                        .partitionValue(email.toLowerCase(Locale.ROOT))
```
andrew-moores, line 733: Any way email could be null at this point?

shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/DynamoMfaMethodsService.java
```diff
@@ -65,8 +69,10 @@ private List<MfaMethodData> getMfaMethodsForMigratedUser(UserCredentials userCre
                 .toList();
     }
 
-    public void deleteMfaMethod(String email, String mfaIdentifier) {
+    public Either<MfaDeleteFailureReason, String> deleteMfaMethod(
+            String email, String mfaIdentifier) {
         dynamoService.deleteMfaMethodByIdentifier(email, mfaIdentifier);
```
andrew-moores, line 95: Do we care if the delete was actually successful, i.e. something was deleted vs nothing needed to be deleted?

integration-tests/src/test/java/uk/gov/di/authentication/services/DynamoMfaMethodsServiceIntegrationTest.java
andrew-moores, line 1: Could the commit message be update to use default instead of primary.  Also add a backup if not present as current backup could be used in the process of deleting the current default.  Also common terminology is to `switch` default and backup.

shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/DynamoMfaMethodsService.java
```diff
@@ -71,25 +71,27 @@ private List<MfaMethodData> getMfaMethodsForMigratedUser(UserCredentials userCre
 
     public Either<MfaDeleteFailureReason, String> deleteMfaMethod(
             String email, String mfaIdentifier) {
+        var userProfile = dynamoService.getUserProfileByEmail(email);
+        if (!userProfile.getMfaMethodsMigrated()) {
```
andrew-moores, line 75: Could userProfile be null?

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsDeleteHandler.java
```diff
@@ -85,6 +86,8 @@ public APIGatewayProxyResponseEvent deleteMFAMethodHandler(
             return generateApiGatewayProxyErrorResponse(404, ErrorResponse.ERROR_1056);
```
andrew-moores, line 91: Should this be a 400 response?  The API should be called with a public subject id and, if it isn't, its something that the client can fix by providing one.  Its not the cases that a specific resource was requested and found not to exist.

shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/DynamoMfaMethodsService.java
BeckaL: So where this is used in the code, no, because we've already done this check in the handler. But was thinking of moving this to in here along with the other 404 case so think this is good reason to do that. Will do in a follow up

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsDeleteHandler.java
BeckaL: ah sorry this is updated later on in a commit - sorry this is some bad git hygiene from me

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
BeckaL: will have a think about in follow up PR

shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/DynamoMfaMethodsService.java
BeckaL: I've added a commit onto the end of this that moves where we do the retrieval and checking of user profile into this method so it's clearer and so that we do all the error handling cases in the same place.

shared/src/main/java/uk/gov/di/authentication/shared/services/DynamoService.java
BeckaL: I think no by this point in the context of how this is called - as we know we have a user profile, and the primary key for the user profile is email so it must be present.

andrew-moores (approved):

PR #6151, opened 2025-03-14
ATO-1533: Get client session from clientSessionStore in TokenHandler

cearl1:
### Wider context of change

We recently added a new client session store to save client sessions to a dynamo table, in addition to redis, and started writing to the new store wherever we were writing to the old one. We also added logging to check whether the 2 sessions were in sync.

We found that if a user hit the `TokenHandler` endpoint multiple times, it would use an outdated view of the `clientSession`. In `AuthenticationCallbackHandler`, the `AuthCodeExchangeData` is saved with the current `clientSession`. The `TokenHandler` uses this `AuthCodeExchangeData` to get the `clientSession`, then updates the `clientSession` with an `idTokenHint` later on. 

### What’s changed

This PR updates the TokenHandler to use the `clientSessionService.getClientSession()`, using the `clientSessionId` already in the `AuthCodeExchangeData`.

Since that was the only usage of the client session in the `AuthCodeExchangeData`, this PR also removes the `clientSession` field from the class.

### Manual testing

Tested auth and identity journey in dev. 
- Tested logging in and signing out
- Tested trying to log in when there was already a session cookie. 

All scenarios worked fine.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
@@ -236,7 +236,19 @@ public class TokenHandler
         updateAttachedLogFieldToLogs(
                 GOVUK_SIGNIN_JOURNEY_ID, authCodeExchangeData.getClientSessionId());
 
-        ClientSession clientSession = authCodeExchangeData.getClientSession();
+        var clientSessionId = authCodeExchangeData.getClientSessionId();
+        var clientSessionOpt = clientSessionService.getClientSession(clientSessionId);
+        if (clientSessionOpt.isEmpty()) {
+            LOG.warn("No client session found for auth code client session id");
+            return generateApiGatewayProxyResponse(
+                    400, OAuth2Error.INVALID_GRANT.toJSONObject().toJSONString());
+        }
+        var clientSession = clientSessionOpt.get();
+        var orchClientSessionOpt = orchClientSessionService.getClientSession(clientSessionId);
+        if (orchClientSessionOpt.isEmpty()) {
+            LOG.warn("No orch client session found for auth code client session id");
+        }
+        logIfClientSessionsAreNotEqual(clientSession, orchClientSessionOpt.orElse(null));
         AuthenticationRequest authRequest;
         try {
             authRequest = AuthenticationRequest.parse(clientSession.getAuthRequestParams());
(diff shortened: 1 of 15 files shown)


Louisasa (approved):

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/TokenHandlerTest.java
```diff
@@ -990,6 +857,42 @@ void shouldReturnClaimsForIdentityJourney() throws JOSEException, TokenAuthInval
...
+                        CLIENT_SESSION_ID,
+                        authRequestParams,
+                        clientSessionCreationTime,
+                        vtr,
+                        CLIENT_NAME);
+        Optional.ofNullable(docAppSubjectId)
+                .ifPresent(subject -> orchClientSession.setDocAppSubjectId(subject.getValue()));
+        when(orchClientSessionService.getClientSession(CLIENT_SESSION_ID))
+                .thenReturn(Optional.of(orchClientSession));
+    }
```
Ryan-Andrews99: This is so much nicer!

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
Ryan-Andrews99, line 1: I'm slightly confused about the scenario described in this commit: 807f8c1df6c4a8296c37c767948392ba53e5904f.  I wouldn't expect the token endpoint to be hit twice as the auth code is single use (and if they had a new auth code they'd have a new/more recently updated client session)

Ryan-Andrews99 (commented):
Have left a couple of comments. Appreciate that I've not massively looked into this so could be missing something. Would be nice to chat about this

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/TokenHandler.java
cearl1: If i recall, the scenario we saw was 2 calls to AuthenticationCallbackHandler that happened really close to eachother, which may not have saved the most up-to-date client session in the auth code exchange data.

cearl1: I'm not sure if this was what we were seeing, but we should really use the dynamo store as the source of truth

PR #6163, opened 2025-03-17
AUT-3485: Add scenarios for MfaMethodsCreateHandler

alhcomer:
## What
Implements the success scenario P2 and error scenarios E2, E5, and E6

- P2: Migrated user with sms mfa only adds an auth app mfa (no OTP presented)
- E2: Migrated user with auth app mfa only attempts to add a second auth app mfa
- E5: Already have 2 MFAs
- E6: Client attempts to add an SMS MFA with the same number as an existing SMS MFA

Error scenario E1 and success scenarios P1 and P3 were implemented in a [previous pull request](https://github.com/govuk-one-login/authentication-api/pull/6121) (albeit for P1 and P3 without OTP handling).

This leaves error scenarios E3 and E4 unimplemented at the time of writing as they are based around the user having provided an invalid OTP when OTP handling has not yet been implemented.

See scenarios [here](https://govukverify.atlassian.net/browse/AUT-3485)

## How to review

1.  Code review
2. Deploy to dev environment with account in required state and make test requests to the mfa-methods-create seeing expected responses.

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
@@ -94,6 +94,9 @@ public class MFAMethodsCreateHandler
                     case INVALID_PRIORITY_IDENTIFIER -> {
                         return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
                     }
+                    case BACKUP_AND_DEFAULT_METHOD_ALREADY_EXIST -> {
+                        return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1068);
+                    }
                 }
             }
 
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
@@ -4,17 +4,18 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import io.vavr.control.Either;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.mfa.MfaMethodCreateRequest;
 import uk.gov.di.authentication.shared.entity.mfa.MfaMethodData;
-import uk.gov.di.authentication.shared.exceptions.InvalidMfaDetailException;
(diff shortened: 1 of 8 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsCreateHandler.java
```diff
@@ -94,6 +94,9 @@ public APIGatewayProxyResponseEvent mfaMethodsHandler(
                     case INVALID_PRIORITY_IDENTIFIER -> {
                         return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
                     }
+                    case AT_MAXIMUM_MFA_COUNT -> {
```
BeckaL, line 97: I'd like to avoid using 'count' terminology here, as I read it as being about mfa attempt counts. Maybe "BACKUP_AND_DEFAULT_METHOD_ALREADY_EXIST"? (and same for elsewhere where we call this count)

account-management-integration-tests/src/test/java/uk/gov/di/accountmanagement/api/MFAMethodsCreateHandlerIntegrationTest.java
```diff
@@ -91,6 +92,56 @@ void shouldReturn200AndMfaMethodDataWhenAuthAppUserAddsSmsMfa() {
...
+
+        String extractedMfaIdentifier =
+                JsonParser.parseString(response.getBody())
+                        .getAsJsonObject()
+                        .get("mfaIdentifier")
+                        .getAsString();
+
+        List<MFAMethod> mfaMethods = userStore.getMfaMethod(TEST_EMAIL);
+
+        assertTrue(
```
BeckaL, line 143: I find this quite hard to read as an assertion.

I think I'd rather grab the sms method and then assert on fields, e.g:

```
        var retrievedSmsMethod = mfaMethods.stream().filter(mfaMethod -> mfaMethod.getMfaMethodType().equals(MFAMethodType.SMS.getValue())).findFirst().get();

assertEquals(expectedFoo, retrievedSmsMethod.foo());
etc...

```

account-management-integration-tests/src/test/java/uk/gov/di/accountmanagement/api/MFAMethodsCreateHandlerIntegrationTest.java
```diff
@@ -71,24 +72,68 @@ void shouldReturn200AndMfaMethodDataWhenAuthAppUserAddsSmsMfa() {
...
+                                constructRequestBody(
+                                        PriorityIdentifier.BACKUP,
+                                        new AuthAppMfaDetail(
+                                                MFAMethodType.AUTH_APP, TEST_CREDENTIAL))),
+                        Collections.emptyMap(),
+                        Collections.emptyMap(),
+                        Map.of("publicSubjectId", TEST_PUBLIC_SUBJECT));
+        assertEquals(200, response.getStatusCode());
+
+        String extractedMfaIdentifier =
```
BeckaL, line 132: It'd be good to assert against a full json response here (e.g. `assertEquals(response.getBody(), someJsonString`. We could potentially do this by grabbing the mfa identifier from the database first, and then chucking that in the json we're expecting. Just helps to have a readable json response when having a look at what the lambda does.

BeckaL (approved):
Comment on a test assertion. Happy for you to fix in a subsequent mini PR just so that we can get this in :-)

PR #6165, opened 2025-03-17
ATO-1117: add test for base64EncodedSalt in spot queue

GHSwallow:
### Wider context of change
Part of the email migration work on IPVCallbackHandler

### What’s changed
The SpotRequest object holds salt as byte[], but in this ticket, I want to refactor it to use base64EncodedSalt instead. This is because the salt on the AuthUserInfo table is stored as a base64 string, and the way we send salt on to SPOT is also as a base64 string. I'm adding this test here to verify the raw string we send to Spot is the same thing now as it is with the new refactor. The one piece of nuance is that the `writeValueAsString` method changes the "=" to "\u003d". That happens currently, and this test proves that it will continue to happen when we pass in a string rather than a byte[].

### Manual testing
n/a

### Checklist

### Related PRs
Created this test after a comment on https://github.com/govuk-one-login/authentication-api/pull/5804. I checked manually that my understanding is correct, but rightly so, it's really important that it's correct, so I'm adding this test to verify the change.


Ryan-Andrews99 (approved):
Thanks for adding this! Sonar check failed but looks like it might just need re-running

PR #6166, opened 2025-03-17
ATO-1116: add AuthUserInfo read policy to logout handler

GHSwallow:
### Wider context of change
Part of the email migration work

### What’s changed
Adds read access to the AuthUserInfo table for the Logout handler.

### Manual testing
n/a

### Checklist


GHSwallow:
I thought we would need to use AuthUserInfo here, as subjectId was being passed through. However, turns out it was just being used to calculate rpPairwiseId, which we can get from the orchClientSessionItem now, so no need for new permissions.

PR #6189, opened 2025-03-21
ATO-1419: Add missing client session sync check

cearl1:
### Background
We have been logging whether the redis client session has been in sync with the dynamo client session. There was one service where this logging check was missing though.

### What
We were getting the client session in NoSessionOrchestrationService, but not logging and comparing with the new dynamo client session. This PR updates the service to compare both client sessions, also updating the tests so the logs are accurate. 

This service is not used in any handlers where we aren't already checking if the client sessions are in sync, so should not be a problem in terms of lambda permissions

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/OrchestrationAuthorizationService.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/OrchestrationAuthorizationService.java
@@ -70,14 +70,14 @@ public class OrchestrationAuthorizationService {
             DynamoClientService dynamoClientService,
             IPVCapacityService ipvCapacityService,
             KmsConnectionService kmsConnectionService,
-            RedisConnectionService redisConnectionService) {
+            RedisConnectionService redisConnectionService,
+            NoSessionOrchestrationService noSessionOrchestrationService) {
         this.configurationService = configurationService;
         this.dynamoClientService = dynamoClientService;
         this.ipvCapacityService = ipvCapacityService;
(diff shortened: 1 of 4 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/OrchestrationAuthorizationService.java
```diff
@@ -70,14 +70,14 @@ public OrchestrationAuthorizationService(
...
+            RedisConnectionService redisConnectionService,
+            NoSessionOrchestrationService noSessionOrchestrationService) {
         this.configurationService = configurationService;
         this.dynamoClientService = dynamoClientService;
         this.ipvCapacityService = ipvCapacityService;
         this.kmsConnectionService = kmsConnectionService;
         this.redisConnectionService = redisConnectionService;
-        this.noSessionOrchestrationService =
-                new NoSessionOrchestrationService(configurationService, redisConnectionService);
+        this.noSessionOrchestrationService = noSessionOrchestrationService;
```
cearl1, line 80: I had to update this constructor to pass in a `NoSessionOrchestrationService` object directly, else the OrchClientSessionService constructor would fail in unit tests (I'm not sure why, but this constructor format matches other unit tests)

GHSwallow: The existing code reuses the redisConnectionService, where this creates a new one. But as you say, there are lots of places where we create a new redisConnectionService it seems.

GHSwallow (approved):

PR #6190, opened 2025-03-21
ATO-1196: add OrchAuthCode table policies

sw-mattw:
### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket. The lambdas writing and reading the auth code need permissions on the DynamoDB table added under #6146.

Handlers requiring write access (calls from `generateAndSaveAuthorisationCode`):
- DocAppCallbackHandler
- IPVCallbackHandler
- AuthCodeHandler
- AuthenticationCallbackHandler

Handlers requiring read access:
- TokenHandler

### What’s changed

- Added two new policies for the OrchAuthCode table - one read, one write.
- Added policies to the lambdas above.

Note that these lambdas do not make use of these permissions yet (will be covered by a subsequent ticket under ATO-937).

### Manual testing

Deployed to orch dev.

Checked both policies have been added. Checked the entities attached to each are as expected.

Tested journeys to trigger each of the lambdas for completeness (although the policies are not used by these yet).

### Checklist

- [ ] Lambdas have correct permissions for the resources they're accessing. **- N/A yet, perms added but not used**

- [ ] Changes have been made to contract tests or **not required**.

- [ ] Changes have been made to the simulator or **not required**.

- [ ] Changes have been made to stubs or **not required**.

- [ ] Successfully deployed to authdev or **not required**.

- [ ] Successfully run Authentication acceptance tests against sandpit or **not required**.

### Related PRs

- #6146

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -1637,6 +1637,7 @@ Resources:
         - !Ref RpPublicKeyCacheTableWriteAccessPolicy
         - !Ref ClientSessionTableReadAccessPolicy
         - !Ref ClientSessionTableWriteAccessPolicy
+        - !Ref OrchAuthCodeTableReadAccessPolicy
         - !Ref OrchAuthCodeTableWriteAccessPolicy
       Tags:
         CheckovRulesToSkip: CKV_AWS_115.CKV_AWS_116.CKV_AWS_173
--- a/template.yaml
+++ b/template.yaml
@@ -1492,6 +1492,7 @@ Resources:
         - !Ref DocAppSigningKmsAccessPolicy
         - !Ref OrchSessionTableReadAccessPolicy
         - !Ref ClientSessionTableReadAccessPolicy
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -1633,6 +1633,7 @@ Resources:
         - !Ref RpPublicKeyCacheTableWriteAccessPolicy
         - !Ref ClientSessionTableReadAccessPolicy
         - !Ref ClientSessionTableWriteAccessPolicy
+        - !Ref OrchAuthCodeTableReadAccessPolicy
       Tags:
```
Ryan-Andrews99, line 1637: Think the token function will need to have write access too, if we're planning to mark each auth code as used once we fetch it

sw-mattw: Added thanks

PR #6191, opened 2025-03-21
ATO-1121: delete the unused isTestClientWithAllowedEmail method

GHSwallow:
### Wider context of change
Part of the email migration work

### What’s changed
Deletes an unused method that was only used in tests. Those tests only verified that method worked as expected.

This was copied directly over from Auth when we initially split. It hasn't been unused or updated since.

### Manual testing
n/a

### Checklist


orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/helpers/TestClientHelperTest.java
cearl1, line 1: Sonar has spotted that the `configurationService` field and the `buildUserContext` method in this class are no longer used, so we could remove them while we're here 🙂

cearl1 (approved):

PR #6198, opened 2025-03-24
Aut 3486/implement update mfa method for default method

BeckaL:
## What

Implements updating an mfa method when the method to update is a default method.

In this case, the only legitimate update is to change phone number or credential value - all other updates are not allowed and will return an appropriate error.

## How to review

1. Code Review -probably easiest to look at the overall diff first, and then commit by commit


integration-tests/src/test/java/uk/gov/di/authentication/services/DynamoServiceIntegrationTest.java
andrew-moores, line 1: 🤔 I wonder if this name change makes more sense in reverse, i.e. the database item called MFAMethodData and the domain model MFAMethod?  Or MFAMethodItem and MFAMethod?

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandler.java
```diff
@@ -53,8 +76,56 @@ public APIGatewayProxyResponseEvent updateMFAMethodsHandler(
...
+            var successfulUpdate = result.get();
+
+            return generateApiGatewayProxyResponse(200, successfulUpdate, true);
+        } catch (Json.JsonException e) {
+            return generateApiGatewayProxyErrorResponse(400, ErrorResponse.ERROR_1001);
+        }
+    }
 
-        return generateApiGatewayProxyResponse(200, "{}");
+    private static APIGatewayProxyResponseEvent handleUpdateMfaFailureReason(
```
andrew-moores, line 121: Is it worth updating the OpenAPI spec with these errors?

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandlerTest.java
```diff
@@ -2,43 +2,188 @@
...
+                """,
+                        MFA_IDENTIFIER, phoneNumber);
+        var expectedResponseParsedToString =
+                JsonParser.parseString(expectedResponse).getAsJsonObject().toString();
+        assertEquals(expectedResponseParsedToString, result.getBody());
+    }
+
+    private static Stream<Arguments> updateFailureReasonsToExpectedResponses() {
+        return Stream.of(
+                Arguments.of(
```
andrew-moores, line 111: Would static imports make this a bit easier to read?

andrew-moores (approved):

integration-tests/src/test/java/uk/gov/di/authentication/services/DynamoServiceIntegrationTest.java
BeckaL: Yeah definitely think we should do a naming clarification, will do in a refactor PR. Liked your idea of calling the database object "Item", not sure what that means for the domain object

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandler.java
BeckaL: Not sure here - the API spec reflects all these statuses, think it's not worth putting the individual cases in but can discuss and do follow up

PR #6200, opened 2025-03-25
ATO-1513 set PreservedReauthCountsForAudit on AuthSessionItem

Louisasa:
### Wider context of change

Moving PreservedReauthCountsForAudit from shared session to AuthSessionItem. Part of the session split.

### What’s changed

Created PreservedReauthCountsForAuditMap on AuthSessionItem and set it in the same places as PreservedReauthCountsForAudit on Session.

### Manual testing

Tested reauth journey on authdev

### Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -436,7 +436,12 @@ public class VerifyMfaCodeHandler extends BaseFrontendHandler<VerifyMfaCodeReque
                 && codeRequest.getMfaMethodType() == MFAMethodType.AUTH_APP
                 && subjectId != null) {
             preserveReauthCountsForAuditIfJourneyIsReauth(
-                    codeRequest.getJourneyType(), subjectId, session, sessionId, maybeRpPairwiseId);
+                    codeRequest.getJourneyType(),
+                    subjectId,
+                    session,
+                    authSession,
+                    sessionId,
+                    maybeRpPairwiseId);
             clearReauthErrorCountsForSuccessfullyAuthenticatedUser(subjectId);
             maybeRpPairwiseId.ifPresentOrElse(
                     this::clearReauthErrorCountsForSuccessfullyAuthenticatedUser,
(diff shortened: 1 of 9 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
```diff
@@ -282,6 +287,28 @@ private void initializeCodeRequestMap() {
         }
     }
 
+    @DynamoDbAttribute(ATTRIBUTE_PRESERVED_REAUTH_COUNTS_FOR_AUDIT_MAP)
+    @DynamoDbConvertedBy(PreservedReauthCountsForAuditMapConverter.class)
+    public Map<CountType, Integer> getPreservedReauthCountsForAuditMap() {
+        return this.preservedReauthCountsForAuditMap;
+    }
+
+    public void setPreservedReauthCountsForAuditMap(
```
Louisasa, line 296: Not convinced this is needed as currently only `with` is used but will double check if DynamoDB needs this

cearl1: I don't think its needed for dynamo, but i've been using this get/set/with pattern to stay consistent with the new orch dynamo classes. There's no harm leaving it there in case someone needs it 🙂

cearl1 (dismissed):

shared/src/main/java/uk/gov/di/authentication/shared/converters/PreservedReauthCountsForAuditMapConverter.java
cearl1, line 1: It would be nice to have some test coverage for this if we havent tested a reauth journey yet. If you set this field on one of the AuthSessionItems in `AuthSessionItemIntegrationTest`, it will make use of this converter

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
Ryan-Andrews99: Ah so there are some docs on this [here](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/ddb-en-client-gs-tableschema.html#ddb-en-client-gs-tableschema-anno-bean). TL;DR any class annotated with `@DynamoDBBean` has to conform to a spec, which includes a public setter and getter

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
Ryan-Andrews99, line 1: Same here would be good to try and add a test for the changes here

cearl1 (approved):

PR #6207, opened 2025-03-25
ATO-1480 validate session auth session

Louisasa:
### Wider context of change

One of the final pieces of the email migration. Validates the email address on the auth session instead of the shared session.

### What’s changed

Validating the email address on the auth session instead of the shared session.

### Manual testing

Logs for the initial email migration.

### Checklist

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
@@ -208,7 +208,7 @@ public class AuthSessionItem {
         this.emailAddress = emailAddress;
         return this;
     }
-    
+
     public boolean validateSession(String emailAddress) {
         return this.emailAddress.equals(emailAddress);
     }
--- a/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
@@ -208,6 +208,10 @@ public class AuthSessionItem {
         this.emailAddress = emailAddress;
         return this;
     }
+    
+    public boolean validateSession(String emailAddress) {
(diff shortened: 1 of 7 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
```diff
@@ -208,6 +208,10 @@ public AuthSessionItem withEmailAddress(String emailAddress) {
         this.emailAddress = emailAddress;
         return this;
     }
+    
+    public boolean validateSession(String emailAddress) {
+        return this.emailAddress.equals(emailAddress);
+    }
```
cearl1, line 214: Is it possible for emailAddress to be null? If so this would throw a NullPointerException. I saw in one of the usages of validateSession it checks if the email is null before calling this method. Maybe we could move that null check inside here?

I know it was like this before your changes, just something i spotted with the existing code 🙂

Louisasa: I don't think the email address should be null by this point @Ryan-Andrews99 when the AuthSession and shared Session emails were mismatched, they wouldn't get to this point to these handlers would they? Or do you think it's worth just covering all cases regardless?

Ryan-Andrews99: Hmm yeah the usages of this method tend to occur after the email has been set on the session. I'd vote to leave it as is given we're just swapping the same behaviour from the shared session

cearl1 (approved):

shared/src/main/java/uk/gov/di/authentication/shared/entity/AuthSessionItem.java
cearl1: Sounds good to me!

PR #6219, opened 2025-03-26
ATO-1435: Use correct type of custom vtr parameter in AuthorisationHandler

cearl1:
### Wider context of change

We recently merged changes to start sending fields from orch to auth frontend, to replace the need for the auth client session. While testing the final part of processing these fields in the auth backend, I noticed that the type of the `vtr_list` parameter was not set correctly.

There is already a `vtr` custom parameter on the auth request, which is in the form of a JSON array (e.g `["Cl.Cm", "P0.Cl.Cm"]`). The new custom field we added to replace this was meant to be in the form of a space separated string list (e.g `"Cl.Cm P0.Cl.Cm"`). However, in the AuthorisationHandler, the logic was incorrect, so was sending the `vtr` parameter as a string, instead of constructing a custom `vtr_list` field.

### What’s changed

The format of the new field has been corrected, and the tests have been updated to reflect what we are expecting.

### Manual testing

Have deployed this to sandpit in conjunction with the other PRs for sending these fields from orch to auth backend, and the fields are what we are expecting.

### Checklist


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/VectorOfTrust.java
```diff
@@ -50,13 +50,7 @@ public boolean containsLevelOfConfidence() {
...
     }
 
-    public static List<VectorOfTrust> parseFromAuthRequestAttribute(List<String> vtr) {
-        if (isNull(vtr) || vtr.isEmpty()) {
-            LOG.info(
-                    "VTR attribute is not present so defaulting to {}",
-                    CredentialTrustLevel.getDefault().getValue());
-            return List.of(new VectorOfTrust(CredentialTrustLevel.getDefault()));
-        }
+    private static JSONArray parseJSONArrayFromAuthRequestAttribute(List<String> vtr) {
```
cearl1, line 53: Diff looks a bit weird, but i've just extracted the code for parsing to a JSON array so the other 2 parse methods can use it. Might be easier to understand looking at the file without the diff

Louisasa (approved):

PR #6236, opened 2025-03-27
ATO-1436: Add new properties onto AuthSessionItem

cearl1:
### Wider context of change

We should now have fields available in the auth backend we can use to replace usages of the clientSession.getAuthParams() method. Some of these usages are outside of the StartHandler however, so we need to add these to the auth session so we can keep track of them.

### What’s changed

This PR adds 3 new fields to the AuthSessionItem: `requestedCredentialStrength`, `requestedLevelOfConfidence` and `clientId`. These fields are also set in the StartHandler.

### Manual testing

Added tests coverage for new fields being set.

Deployed to sandpit, verified that new fields were set on AuthSessionItem in dynamo

### Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -17,10 +17,8 @@ import uk.gov.di.authentication.frontendapi.helpers.ReauthMetadataBuilder;
 import uk.gov.di.authentication.frontendapi.services.StartService;
 import uk.gov.di.authentication.shared.domain.CloudwatchMetrics;
 import uk.gov.di.authentication.shared.entity.ClientRegistry;
-import uk.gov.di.authentication.shared.entity.ClientSession;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
-import uk.gov.di.authentication.shared.entity.LevelOfConfidence;
 import uk.gov.di.authentication.shared.entity.UserProfile;
 import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.ReauthAuthenticationAttemptsHelper;
@@ -39,9 +37,7 @@ import uk.gov.di.authentication.shared.services.SerializationService;
 import uk.gov.di.authentication.shared.services.SessionService;
 
 import java.util.ArrayList;
-import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 
 import static uk.gov.di.authentication.frontendapi.helpers.ReauthMetadataBuilder.getReauthFailureReasonFromCountTypes;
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
```diff
@@ -452,27 +454,45 @@ void shouldAddSessionToDynamoWhenPreviousSessionIsProvidedInRequestBodyButIsNotI
...
+            request += ", \"redirect_uri\": \"" + redirectUriOpt.get() + "\"";
+        }
+        if (levelOfConfidenceOpt.isPresent()) {
+            request +=
+                    ", \"requested_level_of_confidence\": \""
+                            + levelOfConfidenceOpt.get().getValue()
+                            + "\"";
+        }
+
+        return request + "}";
```
Louisasa, line 495: I'd really like a different way of doing this, however I'm coming up short and it is just a test.. but if you can think of a nicer way that'd be excellent

Louisasa (dismissed):

integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
cearl1: Yeah, I'll play around with it for a bit but if I cant come up with a neater way this will have to do. Fortunately its only test code so its not a huge deal, but i agree its a bit ugly

Louisasa (dismissed):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -170,15 +168,20 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-
         boolean isUserAuthenticatedWithValidProfile;
         try {
             var authSession =
                     authSessionService.getUpdatedPreviousSessionOrCreateNew(
                             Optional.ofNullable(startRequest.previousSessionId()),
                             sessionId,
                             startRequest.currentCredentialStrength());
+            authSession.setRequestedCredentialStrength(
+                    retrieveCredentialTrustLevel(startRequest.requestedCredentialStrength()));
```
Ryan-Andrews99, line 179: Random Q - what are these retrieve* methods doing? Looking in the code I can find ones that take a VTR set but I wouldn't expect these to be used here?

Ryan-Andrews99 (commented):
Two small comments - mainly to clarify some stuff 🙂

integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
```diff
@@ -76,6 +78,7 @@ class StartIntegrationTest extends ApiGatewayHandlerIntegrationTest {
                     + "\"requested_credential_strength\": \"%s\", "
                     + "\"scope\": \"%s\""
                     + "}";
+    private static final Gson GSON = new GsonBuilder().create();
```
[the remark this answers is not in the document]
cearl1: Oooh, yeah I forgot about that service when writing this. I'll update it to use that service.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
cearl1: It's mainly so we don't have to call `retrieve*` every time we get the credentialTrustLevel/levelOfConfidence from the auth session. I figured its easier to work with the objects in the auth session instead of the string representations.

Ryan-Andrews99: Oooh I see this converts the string to the enum

Ryan-Andrews99 (approved):

PR #6262, opened 2025-04-01
Aut 3486/implement update for backup

BeckaL:
## What

Implements behaviour for updating a backup method via the new method management put endpoint.

The only valid update for a backup method is to switch it with a default method (thus promoting a backup method to a default method and demoting the existing default method to a backup). All other updates result in failure status code (except for a no-change update, which returns a 204 response and makes no database changes to ensure that repeat identical requests to the PUT endpoint are idempotent).

## Related PRs

https://github.com/govuk-one-login/authentication-api/pull/6198 implemented updates for default methods


BeckaL:
The one sonar issue here is sonar thinking two tests are duplicates, not taking into account that they are parameterized and the data we're passing in is different. So ignoring that here.

andrew-moores (approved):

PR #6279, opened 2025-04-03
Aut 4122/generate mfa identifiers for non migrated users

BeckaL:
Generates an mfa identifier for a non-migrated mfa method on read, to remove the need for a hard-coded mfa identifer.
Ensures that this mfa identifier is migrated alongside its method when a user's mfa methods are migrated.


BeckaL:
one added TODO here left in deliberately (should be updated following results of database analysis)

alhcomer (approved):

PR #6288, opened 2025-04-04
ATO-1571: Add test for auth request missing vtr parameter

cearl1:
### Wider context of change

We have been working on passing new fields from orch to auth frontend to auth backend, with the goal of eventually replacing the need for the auth client session. We have the code for sending the fields from orch to auth frontend, and the code for receiving the fields from auth frontend to auth backend, already merged. 

We tried to merge the PR for passing the field on in auth frontend, but encountered issues with how the vtr field is handled in orch.
 
RPs are not required to send a vtr field to the authorise endpoint. If the vtr field is not provided, the default vtr value is used. The changes we made to pass the vtr field on to the auth frontend did not use the default value, instead using `null`, which in the JWT to auth frontend, was the same as not setting the field. As a result, any RP that sent a request without a vtr would get an error in the auth frontend, complaining that it was missing the required vtr field.

We didn't have any test coverage for this, so this PR is for adding that coverage.

### What’s changed

This PR adds an integration test which asserts that if an auth request is sent to the /authorise endpoint with no vtr field set, then it should use the default vtr value and use this in the JWT that orch send to auth frontend 

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -900,9 +900,8 @@ public class AuthorisationHandler
         orchestrationAuthorizationService.storeState(sessionId, clientSessionId, state);
 
         var vtrList =
-                Optional.ofNullable(authenticationRequest.getCustomParameter(VTR_PARAM))
-                        .map(VectorOfTrust::parseVtrStringListFromAuthRequestAttribute)
-                        .orElse(null);
+                VectorOfTrust.parseVtrStringListFromAuthRequestAttribute(
+                        authenticationRequest.getCustomParameter(VTR_PARAM));
         String reauthSub = null;
         String reauthSid = null;
         if (reauthRequested) {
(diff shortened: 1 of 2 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -902,7 +902,7 @@ private APIGatewayProxyResponseEvent generateAuthRedirect(
         var vtrList =
                 Optional.ofNullable(authenticationRequest.getCustomParameter(VTR_PARAM))
                         .map(VectorOfTrust::parseVtrStringListFromAuthRequestAttribute)
-                        .orElse(null);
+                        .orElse(List.of(CredentialTrustLevel.getDefault().getValue()));
```
Ryan-Andrews99, line 905: I'm a little confused on this being optional in the first place as the method `parseVtrStringListFromAuthRequestAttribute` looks like it handles this?

cearl1: Ahh yeah good point, i can just pass the custom param into the parse method directly. I'll do that.

cearl1:
We only really need the vtr list parameter to check if an identity journey is required in auth (we already send the credential trust level in the `confidence` field). I'll open a new PR to swap this vtr list parameter for an identityRequired parameter.

PR #6297, opened 2025-04-07
Ato 1422: Send identity required flag instead of vtr to auth frontend

cearl1:
### Wider context of change

We would like to eventually remove the auth client session. To do so we need to send fields from orch to the auth backend in order to replace the auth client session usages.

We initially replaced the whole `vtr` list in the client session auth params with a field in the auth backend. However, we had a couple of issues when testing and deploying it. We also realised that we dont really need the whole vtr list, as auth only need it for the MfaHelper and the IdentityHelper.

The MfaHelper only needs the credential trust level (which we already send through to the auth backend in the `confidence` field. The IdentityHelper uses the vtr to calculate whether identity is required. This can be calculated on orch's side, and the flag can be passed through to the auth backend.

### What’s changed

This PR removes the `vtr` field in favor of using the existing `confidence` field (which will be mapped to `requested_credential_strength` in the auth frontend for clarity), and the new `identity_required` flag, which is calculated on orch's side.

### Manual testing

Deployed to sandpit and tested with the auth frontend changes. Auth journey worked fine. 

### Checklist

### Related PRs

Frontend changes (to be merged after this one) TODO


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/StartRequest.java
```diff
@@ -2,21 +2,18 @@
...
         @Expose @SerializedName("authenticated") boolean authenticated,
-        @Expose @SerializedName("current-credential-strength")
-                CredentialTrustLevel currentCredentialStrength,
+        @Expose @SerializedName("current-credential-strength") String currentCredentialStrength,
         @Expose @SerializedName("cookie_consent") String cookieConsent,
         @Expose @SerializedName("_ga") String ga,
-        @Expose @SerializedName("vtr") List<String> vtr,
+        @Expose @SerializedName("requested_credential_strength") String requestedCredentialStrength,
+        @Expose @SerializedName("requested_level_of_confidence") String requestedLevelOfConfidence,
         @Expose @SerializedName("state") String state,
```
Ryan-Andrews99, line 17: I'd be wary of changing the authorize JAR and this in the same PR, Auth's pipeline is slower to prod than ours, so the code will remain expecting fields which we've stopped sending

Ryan-Andrews99 (commented):
Left two comments, main concern is deployment timing. I'd vote to split these changes in two PRs

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -962,8 +964,9 @@ private APIGatewayProxyResponseEvent generateAuthRedirect(
                         .claim("authenticated", orchSession.getAuthenticated())
                         .claim(
                                 "current_credential_strength",
-                                orchSession.getCurrentCredentialStrength())
-                        .claim("vtr", vtrList)
+                                Optional.ofNullable(orchSession.getCurrentCredentialStrength())
+                                        .map(CredentialTrustLevel::getValue)
+                                        .orElse(null))
```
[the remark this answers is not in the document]
cearl1: Yep agreed, I've reverted these changes - theres no harm leaving them as is for the time being

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/entity/StartRequest.java
cearl1: Yep didnt think of that. I've split out another PR for the auth backend changes here: https://github.com/govuk-one-login/authentication-api/pull/6311. 

This PR is just for orch changes now.

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
CarlyG55, line 1: Is there a reason we've dropped the second vtr value in both of these tests?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -899,20 +900,13 @@ private APIGatewayProxyResponseEvent generateAuthRedirect(
...
         String reauthSub = null;
         String reauthSid = null;
         if (reauthRequested) {
             try {
                 SignedJWT reauthIdToken = getReauthIdToken(authenticationRequest);
                 reauthSub = reauthIdToken.getJWTClaimsSet().getSubject();
                 reauthSid = reauthIdToken.getJWTClaimsSet().getStringClaim("sid");
-                if (isNull(authenticationRequest.getCustomParameter(VTR_PARAM))) {
-                    vtrList = extractVoTStringListFromIdTokenHint(reauthIdToken);
-                }
```
CarlyG55, line 915: is there some reauth specific behaviour we're losing here?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -952,7 +946,15 @@ private APIGatewayProxyResponseEvent generateAuthRedirect(
                         .claim("is_one_login_service", client.isOneLoginService())
                         .claim("service_type", client.getServiceType())
                         .claim("govuk_signin_journey_id", clientSessionId)
-                        .claim("confidence", requestedCredentialTrustLevel.getValue())
+                        .claim("confidence", requestedVtr.getCredentialTrustLevel().getValue())
+                        .claim(
+                                "requested_credential_strength",
+                                requestedVtr.getCredentialTrustLevel().getValue())
```
CarlyG55, line 954: have we got a fallback here if there isn't a requested credential trust level?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/VectorOfTrust.java
```diff
@@ -81,25 +81,16 @@ public static List<VectorOfTrust> parseFromAuthRequestAttribute(List<String> vtr
...
-            LOG.info(
-                    "VTR attribute is not present so defaulting to {}",
-                    CredentialTrustLevel.getDefault().getValue());
-            return List.of(CredentialTrustLevel.getDefault().getValue());
-        }
-        JSONArray vtrJsonArray = parseJSONArrayFromAuthRequestAttribute(vtr);
-        List<String> vtrList = vtrJsonArray.stream().map(String::valueOf).toList();
-        LOG.info("VTR list has been processed as vtrStringList list: {}", vtrList);
-        return vtrList;
-    }
```
CarlyG55, line 95: Just want to confirm we're definitely not losing any behaviour in here?

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
cearl1: No not really, i can revert this change so we're testing that we handle multiple vtrs

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
cearl1: Yep so we use the `VectorOfTrust.parseFromAuthRequestAttribute` method to set the `requestedVtr` variable. That method should either default to `Cl.Cm` if no vtr is provided, or error if only a level of confidence is provided (like `P2`)

Before we were using `requestedCredentialTrustLevel`, which was defined in the same way, i've just extracted methods so we can get the requestedVtr, then get the credentialTrustLevel and levelOfConfidence from that.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
cearl1: These checks are done a bit earlier on in this handler. When we initialise the `requestedVtr` variable, we use `getVtrList()`, which delegates to this `extractVotFromIdTokenHint` method.

This was leftover code from when I was trying to send the vtr list as a field to auth backend. In hindsight i'm not sure why i didnt just pass the `requestedVtr` through.....

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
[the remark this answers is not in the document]
cearl1: Ooooh yeah good point, will add some more tests

Ryan-Andrews99 (approved):

PR #6301, opened 2025-04-07
ATO-1576: Client registry api http examples

CarlyG55:
### Wider context of change

Add example HTTP requests to allow testing the client registry API

### Checklist

### DIFF
--- a/http/http-client.private.env.json.template
+++ b/http/http-client.private.env.json.template
@@ -1,0 +1,7 @@
+{
+  "sandpit": {
+    "API-KEY": "value",
+    "clientId": "value",
+    "publicKey": "value"
+  }
+}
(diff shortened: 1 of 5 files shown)


http/http-client.private.env.json.template
```diff
@@ -0,0 +1,7 @@
+{
+  "sandpit": {
+    "API-KEY": "value",
+    "clientId": "value",
+    "publicKey": "value"
```
ethanmills, line 5: Do these need to be private?

ethanmills (approved):

http/http-client.private.env.json.template
CarlyG55: Chatted about this and might follow up with these changes for ease of use, but will leave for now

PR #6311, opened 2025-04-08
ATO-1422: Receive new claims in auth backend

cearl1:
### Wider context of change

We would like to eventually remove the auth client session. To do so we need to send fields from orch to the auth backend in order to replace the auth client session usages.

We initially replaced the whole `vtr` list in the client session auth params with a field in the auth backend. However, we had a couple of issues when testing and deploying it. We also realised that we dont really need the whole vtr list, as auth only need it for the MfaHelper and the IdentityHelper.

The MfaHelper only needs the credential trust level (which we already send through to the auth backend in the `confidence` field). 
The IdentityHelper uses the vtr to calculate whether identity is required. This uses the level of confidence on the lowest requested vtr. We need this level of confidence in some cloudwatch metrics calls, so we can pass this field in too.

### What’s changed

This PR removes the `vtr` field in favor of using the fields `requested_credential_strength` 
 and `requested_level_of_confidence` sent by orch to the auth backend.

### Manual testing

Deployed to sandpit and tested with the auth frontend changes. Auth journey worked fine. 
Also deployed stub and tested in authdev. Auth journey worked fine.

### Checklist

### Related PRs

Orch changes (to be merged before this one) https://github.com/govuk-one-login/authentication-api/pull/6297
Auth frontend changes (to be merged after this one)  https://github.com/govuk-one-login/authentication-frontend/pull/2714

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -20,8 +20,8 @@ import uk.gov.di.authentication.shared.entity.ClientRegistry;
 import uk.gov.di.authentication.shared.entity.ClientSession;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
+import uk.gov.di.authentication.shared.entity.LevelOfConfidence;
 import uk.gov.di.authentication.shared.entity.UserProfile;
-import uk.gov.di.authentication.shared.entity.VectorOfTrust;
 import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.ReauthAuthenticationAttemptsHelper;
 import uk.gov.di.authentication.shared.serialization.Json;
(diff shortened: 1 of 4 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -371,35 +372,44 @@ private static void logIfNewFieldsDoNotMatchClientSessionAuthParameters(
...
+        if (!Objects.equals(
+                startRequest.requestedLevelOfConfidence(), requestedLevelOfConfidence)) {
+            LOG.warn(
+                    "\"requested_level_of_confidence\" field does not match field calculated from auth request params");
+        }
+        if (!Objects.equals(
+                startRequest.requestedCredentialStrength(),
+                requestedVtr.getCredentialTrustLevel().getValue())) {
+            LOG.warn(
+                    "\"requested_credential_strength\" field does not match field calculated from auth request params");
```
Ryan-Andrews99, line 397: Does the request VTR need sorting before we do this comparison?

Ryan-Andrews99 (commented):
Left one comment about sorting the VTR from the client session before comparing

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
cearl1: The `VectorOfTrust.parseFromAuthRequestAttribute()` does the sorting at the end of `parseVtrStringList` (it returns the effective vtr), so I think we should be fine?

cearl1: And the startRequest levelOfConfidence/credentialTrustLevel are set in orch by sorting the vtrList and getting the lowest one.

Ryan-Andrews99: Cool makes sense, was just checking both were sorted

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -371,35 +372,44 @@ private static void logIfNewFieldsDoNotMatchClientSessionAuthParameters(
...
-                Optional.ofNullable(clientSession.getAuthRequestParams().get("vtr"))
-                        .map(vtr -> List.of(VectorOfTrust.parseFromAuthRequestAttribute(vtr)))
-                        .orElse(null);
-        var startRequestVtrList =
-                Optional.ofNullable(startRequest.vtr())
-                        .map(vtr -> List.of(VectorOfTrust.parseVtrStringList(vtr)))
-                        .orElse(null);
-        if (!Objects.equals(startRequestVtrList, authRequestVtrList)) {
-            LOG.warn("\"vtr\" field does match custom parameter in auth request params");
+        var requestedVtr =
```
Ryan-Andrews99, line 381: Thinking about this should we get this value from calling `getEffectiveVectorOfTrust()` which is what the auth side of the codebase use?

Ryan-Andrews99 (approved):

PR #6312, opened 2025-04-08
AUT-4123: Fetch phone number verified stats

mattvot:
## What

The last question to answer for now is "of user profiles with PhoneNumber set, is PhoneNumberVerified always true".

We implement an index here to make the query slightly more performant and use a scan to get the total records with a phone number then filter further to get the count when verified is true.

This also includes several follow on commits that refactor the handler to improve legibility.

## How to review

1. Code Review, commit-by-commit without showing whitespace changes

## Checklist


ci/terraform/shared/dynamodb.tf
```diff
@@ -111,6 +111,11 @@ resource "aws_dynamodb_table" "user_profile_table" {
     type = "N"
   }
 
+  attribute {
```
dbes-gds, line 114: is this so it can be included in an index?

mattvot: Indeed. It was necessary to declare the attribute here in order to use in the index

dbes-gds (approved):

PR #6341, opened 2025-04-14
ATO-1514: remove processingIdentity attempts methods from shared session class

Ryan-Andrews99:
### Wider context of change:
We are in the process of migrating this field from the shared redis session to the Orchestration only DynamoDB session. Now that we're reading the value in the processing identity handler from the new session store, we can remove other usages of methods which set this value. This removes usages of the reset method and any test usages of the increment methods

### What’s changed:
- Removes resetting the value in the authorise handler, given we're not reading it anymore in processing identity handler, we can remove resetting it here. 
- Removes test usages of the increment methods
- Removes the session class methods as they're now unused.

### Manual testing: 
- N/A just removes setting a now unread field

### Checklist


Louisasa (commented):
There's a couple of sonar issues raised but otherwise LGTM

Louisasa (approved):

PR #6344, opened 2025-04-14
AUT-4133: create SSM parameter for frontend session redis cache

pskushwaha1:
## What

AUT-4133: create SSM parameter for frontend session Redis cache output 

## How to review

1. deploy to dev and monitor the deployment , there should be no change in redis front end

### DIFF
--- a/ci/terraform/shared/ssm.tf
+++ b/ci/terraform/shared/ssm.tf
@@ -244,7 +244,7 @@ resource "aws_ssm_parameter" "frontend_redis_password" {
   name   = "${var.environment}-${local.frontend_redis_key}-redis-password"
   type   = "SecureString"
   key_id = aws_kms_alias.frontend_parameter_store_key_alias.id
-  value  = random_password.redis_password.result
+  value  = random_password.frontend_redis_password.result
   lifecycle {
     ignore_changes = [value]
   }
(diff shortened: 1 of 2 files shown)


ci/terraform/shared/ssm.tf
```diff
@@ -174,3 +174,123 @@ resource "aws_iam_role_policy_attachment" "lambda_iam_role_pepper_parameters" {
...
+  value  = "true"
+}
+
+
+
+resource "aws_ssm_parameter" "frontend_redis_password" {
+  name   = "${var.environment}-${local.frontend_redis_key}-redis-password"
+  type   = "SecureString"
+  key_id = aws_kms_alias.frontend_parameter_store_key_alias.id
+  value  = random_password.redis_password.result
```
monhaque, line 247: Which resource is this referencing to?

pskushwaha1: Currently it is referring to https://github.com/govuk-one-login/authentication-api/blob/9b346f253f439bc114eef4adfde7a8445b4d2a3c/ci/terraform/shared/redis.tf#L11 

however this is not the correct password for redis , we need to update this manually with the values which we have store in secret manager

pskushwaha1: I have change this to dedicated frontend radom password

pskushwaha1:
We will be migrating the resource rather creating it along with KMS key from frontend state file into shared state file

PR #6308, opened 2025-04-08
ATO-1205: use orch auth code store in token handler

sw-mattw:
**NOTE:**
- Merge dependent on results of consistency checks (#6287)
- Merge dependent on results of #6316
- Might be easiest to review this PR commit by commit

### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket.

We're now writing to the new DynamoDB table using the new service in the four lambdas: https://github.com/govuk-one-login/authentication-api/pull/6278, https://github.com/govuk-one-login/authentication-api/pull/6250, https://github.com/govuk-one-login/authentication-api/pull/6244, https://github.com/govuk-one-login/authentication-api/pull/6223. We are reading from the DynamoDB table, using the new service, in the TokenFunction lambda (#6287), but not using the retrieved item yet (apart from logging).

Following consistency checks, added in #6287, we now wish to remove the Redis getter call, making DynamoDB the primary. We will stop writing to the Redis store in ATO-1218.

### What’s changed

- Remove consistency check logging
- Remove call to `authorisationCodeService.getExchangeDataForCode`
	- Making `orchAuthCodeService.getExchangeDataForCode` the primary
- Remove references to `authorisationCodeService`
- Remove `getExchangeDataForCode` method from `AuthorisationCodeService` (dead code following changes)
- Update unit test to return 500 when get method throws exception
- Update integration tests to get auth code from the orch auth code generation method

### Manual testing

Deployed to Orch dev, ran through auth only, identity (P2), and doc app journeys successfully.

Reviewed the TokenFunction lambda logs and verified retrieval logs (`Retrieving authorisation code exchange data from orch auth code store`) were present with no errors. Reviewed the DynamoDB table, the associated items all had their `isUsed` attributes set to true as expected. AuthExchangeData looked as expected (email and authTime set to null for a doc app journey).

### Checklist

- [x] Lambdas have correct permissions for the resources they're accessing.
  - No additional permissions added, read/write access already granted under prior PRs and in use by changes under #6287.

- [ ] Impact on orch and auth mutual dependencies has been checked. **- N/A**

- [ ] Changes have been made to contract tests or not required. **- N/A**

- [ ] Changes have been made to the simulator or not required. **- N/A**

- [ ] Changes have been made to stubs or not required. **- N/A**

- [ ] Successfully deployed to authdev or not required. **- N/A**

[...]

PR #6287, opened 2025-04-04
ATO-1198: add orch AuthCode consistency checks

sw-mattw:
~~Note: Wait for #6250 to be merged~~

### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket.

We're now writing to the new DynamoDB table using the new service in the four lambdas: #6278, #6250, #6244, #6223.

We wish to read from the DynamoDB table, using the new service, in the TokenFunction lambda. The DynamoDB table read policy was added to the `TokenFunction` CloudFormation under #6190.

We wish to check the consistency of this data compared to that retrieved from the Redis store. Note that we do not wish to remove the Redis getter call (as this should remain the primary) - this will be done under ATO-1205.

### What’s changed

- Added call to the `getExchangeDataForCode` method on the orch auth code service
- Added consistency checks within logs
- Updated unit tests to check for calls to the new orch auth code service

Note that the integration tests will be updated as part of ATO-1205 - under this ticket's changes we are only catching and logging retrieval errors from the new store and not returning any of these.

### Manual testing

Deployed to Orch dev, ran through an auth only journey successfully. Reviewed the TokenFunction lambda logs and verified consistency logs were present. Reviewed the DynamoDB table, the associated item had its `isUsed` attribute set to true as expected.

### Checklist

- [x] Lambdas have correct permissions for the resources they're accessing.
  - (Read policy (`OrchAuthCodeTableReadAccessPolicy `) previously added to the `TokenFunction` CloudFormation - see  #6190. Reading as expected - see manual testing section above.)

- [ ] Impact on orch and auth mutual dependencies has been checked. **- N/A**

- [ ] Changes have been made to contract tests or **not required**.

- [ ] Changes have been made to the simulator or **not required**.

- [ ] Changes have been made to stubs or **not required**.

- [ ] Successfully deployed to authdev or **not required**.

- [ ] Successfully run Authentication acceptance tests against sandpit or **not required**.

### Related PRs

- Entity and service - #6220
- Update to use the same auth code for both stores to enable consistency checks - #6221
- DynamoDB write policy additions to lambdas - #6190
- Writing in lambdas - #6278, #6250, #6244, #6223

PR #6250, opened 2025-03-31
ATO-1189: set orch AuthCode in IPV Callback lambda

sw-mattw:
### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket.

A new entity and service has been previously added under #6220. Write policy was added to this lambda under #6190.

This work stores the auth code in DynamoDB for the IPV Callback lambda using the new service. Note that we do not wish to remove the Redis setter call - this will be done under ATO-1218.

### What’s changed

- Added call to the `generateAndSaveAuthorisationCode` method on the orch auth code service
- Added test stubs for the new orch auth code service
- Added unit and integration tests for checking we are (or are not) writing to the new DynamoDB table
- Fixed broken unit test - see d92fc3b76b463cdbdd5d2b4f516d2eb3304f8c52

### Manual testing

Deployed to Orch dev, ran through an identity journey (sandpit RP stub, checked "return code" claim, selected "P2" level of confidence, clicked continue, signed in, on the IPV stub form removed the `vot` value, clicked continue, redirected to user info screen).

Observed a request being made to the ipv callback endpoint (for this IPV Callback lambda). This returned 302 and continued on to the main callback, redirecting successfully to the user information screen. The lambda logs showed a call being made to the new orch auth code service.

Reviewed the `dev-Orch-Auth-Code` DynamoDB table - the expected item and attribute values were found: auth code, exchange data, IsUsed=false, and a 5 minute TTL (expected value based on config).

### Checklist

- [x] Lambdas have correct permissions for the resources they're accessing.
  - (Write policy (`OrchAuthCodeTableWriteAccessPolicy`) previously added to the `IPVCallbackFunction` CloudFormation. Writing as expected - see manual testing section above.)

- [ ] Impact on orch and auth mutual dependencies has been checked. **- N/A**

- [ ] Changes have been made to contract tests or **not required**.

- [ ] Changes have been made to the simulator or **not required**.

- [ ] Changes have been made to stubs or **not required**.

- [ ] Successfully deployed to authdev or **not required**.

- [ ] Successfully run Authentication acceptance tests against sandpit or **not required**.

### Related PRs

- Entity and service - #6220
- Update to use the same auth code for both stores to enable consistency checks - #6221
- DynamoDB write policy additions to lambdas - #6190

PR #6244, opened 2025-03-31
ATO-1190: set orch AuthCode in Doc App Callback lambda

sw-mattw:
### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket.

A new entity and service has been previously added under #6220. Write policy was added to this lambda under #6190.

This work stores the auth code in DynamoDB for the Doc App Callback lambda using the new service. Note that we do not wish to remove the Redis setter call - this will be done under ATO-1218.

### What’s changed

- Added call to the `generateAndSaveAuthorisationCode` method on the orch auth code service
- Added test stubs for the new orch auth code service

### Manual testing

Deployed to Orch dev, ran through a valid doc app journey (doc app sandpit, selected valid passport). Observed a request being made to the `doc-app-callback` endpoint (for this Doc App Callback lambda). This returned 302 and continued on to the main callback, redirecting successfully to the user information screen. The lambda logs showed a call being made to the new orch auth code service.

Reviewed the `dev-Orch-Auth-Code` DynamoDB table - the expected item and attribute values were found: auth code, exchange data, IsUsed=false, and a 5 minute TTL (expected value based on config).

### Checklist

- [x] Lambdas have correct permissions for the resources they're accessing.
  - (Write policy (`OrchAuthCodeTableWriteAccessPolicy`) previously added to the `DocAppCallbackFunction` CloudFormation. Writing as expected - see manual testing section above.)

- [ ] Impact on orch and auth mutual dependencies has been checked. **- N/A**

- [ ] Changes have been made to contract tests or **not required**.

- [ ] Changes have been made to the simulator or **not required**.

- [ ] Changes have been made to stubs or **not required**.

- [ ] Successfully deployed to authdev or **not required**.

- [ ] Successfully run Authentication acceptance tests against sandpit or **not required**.

### Related PRs

- Entity and service - #6220
- Update to use the same auth code for both stores to enable consistency checks - #6221
- DynamoDB write policy additions to lambdas - #6190

PR #6354, opened 2025-04-15
ATO-1218: remove Redis auth code setter in auth code handler

sw-mattw:
~**MERGE NOTE: ATO-1205 (#6308) must be merged before this work as this Redis will still be the primary at that point and there is no fallback in place.**~ - merged

### Wider context of change

Currently RP issued auth codes are stored in Redis. We are migrating away from Redis to DynamoDB. The auth code store is being migrated under the parent ticket.

In prior work, we are writing and reading to the new DynamoDB table (using the new orch auth code service) alongside the Redis calls.

Following consistency checks, added in #6287, we now wish to remove the Redis setter call from the handlers, so we're only writing to the DynamoDB.

This PR removes the Redis references in the auth code lambda.

### What’s changed

- Replace call to the `authorisationCodeService` with a call to the `orchAuthCodeService`
- Remove references to `authorisationCodeService`

### Manual testing

Deployed to Orch dev, ran through an identity journey (sandpit RP stub, selected "P2" level of confidence, clicked continue, signed in, clicked continue on the IPV stub form using the default values, redirected to user info screen).

Redirected successfully to the user information screen. The AuthCodeFunction lambda logs showed a call being made to the new orch auth code service.

Reviewed the dev-Orch-Auth-Code DynamoDB table - the expected item and attribute values were found: auth code, exchange data, IsUsed=false, and a 5 minute TTL (expected value based on config).

### Checklist

  - No additional permissions added, write access already granted under prior PRs.

- [ ] Changes have been made to contract tests or not required. **- N/A**

- [ ] Changes have been made to the simulator or not required. **- N/A**

- [ ] Changes have been made to stubs or not required. **- N/A**

- [ ] Successfully deployed to authdev or not required. **- N/A**

- [ ] Successfully run Authentication acceptance tests against sandpit or not required. **- N/A**

### Related PRs

- Entity and service - #6220
- Update to use the same auth code for both stores to enable consistency checks - #6221
- DynamoDB write policy additions to lambdas - #6190
- Writing in lambdas - #6278, #6250, #6244, #6223
- Consistency checks - #6287
- Updating token endpoint to read from DynamoDB - #6308

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
@@ -30,7 +30,6 @@ import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
 import uk.gov.di.orchestration.shared.entity.Session;
 import uk.gov.di.orchestration.shared.entity.VectorOfTrust;
 import uk.gov.di.orchestration.shared.exceptions.ClientNotFoundException;
-import uk.gov.di.orchestration.shared.exceptions.OrchAuthCodeException;
 import uk.gov.di.orchestration.shared.helpers.IpAddressHelper;
 import uk.gov.di.orchestration.shared.helpers.PersistentIdHelper;
 import uk.gov.di.orchestration.shared.serialization.Json.JsonException;
@@ -267,6 +266,7 @@ public class AuthCodeHandler
                             clientSessionId,
                             session,
                             orchSession);
+
             authenticationResponse =
                     orchestrationAuthorizationService.generateSuccessfulAuthResponse(
                             authenticationRequest, authCode, redirectUri, state);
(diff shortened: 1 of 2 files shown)


Ryan-Andrews99 (dismissed):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthCodeHandler.java
```diff
@@ -482,27 +482,10 @@ private AuthorizationCode generateAuthCode(
...
-                    "Failed to generate and save authorisation code to orch auth code DynamoDB store. NOTE: Redis is still the primary at present. Error: {}",
-                    e.getMessage());
-        }
-
-        return authCode;
+        return orchAuthCodeService.generateAndSaveAuthorisationCode(
+                clientID.getValue(),
+                clientSessionId,
+                emailOptional.orElse(null),
+                orchSession.getAuthTime());
```
cearl1, line 483: I've noticed in the other PRs we're still catching the `OrchAuthCodeException` and logging when it happens. Do we need to do that here too? It looks like it will throw the unchecked exception out and i'm not sure if/where it will be handled.

PR #6355, opened 2025-04-15
Aut 4205/migrated user default method signin

BeckaL:
Allow a user with migrated mfa methods (i.e. methods stored under the new structure in user credentials, supporting multiple methods for a user) to sign in with their default method, by returning the correct method in all the places we respond from the frontend api with this information. No change for non-migrated users.

How to review:
1. Code review


andrew-moores (approved):
Deployed to dev and acceptance tests passed.  Code changes look good.

PR #6356, opened 2025-04-15
BAU: Remove bucket from prod/int

andrew-moores:
The new bucket for storing otps during acceptance testing has been created in production and integration but we aren't going to use it there at the moment so removing them.

## What
The new bucket for storing otps during acceptance testing has been created in production and integration but we aren't going to use it there at the moment so removing them.

## How to review
1. Code Review

## Checklist


ci/terraform/account-management/s3.tf
```diff
@@ -10,6 +10,7 @@ resource "aws_s3_bucket" "am-api-acceptance-tests-otp-bucket" {
 }
```
pskushwaha1, line 10: This resource also needs the count index  , else s3 bucket will be not removed for prod

alhcomer (approved):

PR #6378, opened 2025-04-22
ATO-1596: migrate in tests and add comment to not use methods directly

GHSwallow:
### Wider context of change
Part of the rpPairwiseId migration. This was an offshoot of ATO-1117, the email migration. RpPairwiseId got complex enough that it deserved it's own ticket and PRs.

### What’s changed
- Replaced the remaining uses of `getRpPairwiseId` and `getPublicSubjectId` on the orchClientSession with `getCorrectPairwiseIdGivenSubjectType`. These were in tests and a temporary log helper.
- Add a comment to the `getRpPairwiseId` and `getPublicSubjectId` methods to not use them directly. If we still want to access each individual one, this should be done  by passing in the necessary subjectType into `getCorrectPairwiseIdGivenSubjectType` (eg to get exactly rpPairwiseId, can get it via `getCorrectPairwiseIdGivenSubjectType("pairwise")`, as I have done in the first commit on this PR). 

### Manual testing
tbd: depoy the dev. I am 99% confident making the methods private won't affect the interaction with dynamoDb, but I want to be safe, not sorry.

### Checklist
- [x] Lambdas have correct permissions for the resources they're accessing. **No new data accessed**
- [x] Impact on orch and auth mutual dependencies has been checked. **No changed**
- [x] Changes have been made to the simulator or not required.  **N/A**
- [x] Changes have been made to stubs or not required.  **N/A**
- [x] Successfully deployed to authdev or not required.  **N/A**
- [x] Successfully run Authentication acceptance tests against sandpit or not required.  **N/A**

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/OrchClientSessionItem.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/OrchClientSessionItem.java
@@ -136,6 +136,8 @@ public class OrchClientSessionItem {
 
     @DynamoDbAttribute(ATTRIBUTE_RP_PAIRWISE_ID)
     public String getRpPairwiseId() {
+        // This is public because the DynamoDbMapper requires it to be for serialisation.
+        // Do not use it directly - use getCorrectPairwiseIdGivenSubjectType instead.
         return rpPairwiseId;
     }
 
@@ -150,6 +152,8 @@ public class OrchClientSessionItem {
 
     @DynamoDbAttribute(ATTRIBUTE_PUBLIC_SUBJECT_ID)
     public String getPublicSubjectId() {
+        // This is public because the DynamoDbMapper requires it to be for serialisation.
+        // Do not use it directly - use getCorrectPairwiseIdGivenSubjectType instead.
         return publicSubjectId;
     }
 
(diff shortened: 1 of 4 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/OrchClientSessionItem.java
```diff
@@ -135,7 +135,7 @@ public VectorOfTrust getEffectiveVectorOfTrust() {
     }
 
     @DynamoDbAttribute(ATTRIBUTE_RP_PAIRWISE_ID)
-    public String getRpPairwiseId() {
+    private String getRpPairwiseId() {
```
Ryan-Andrews99, line 138: I think this may be the cause of the failing integration tests. Based on the way we interact with Dynamo in BaseDynamoService, I think all the attributes on this class have to be public: https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/ddb-en-client-gs-tableschema.html as we call `TableSchema.fromBean` on the class

GHSwallow: It also has to be named `getRpPairwiseId`, so annoyingly the only thing I can do is add a comment on the method, so I've added that here.

PR #6386, opened 2025-04-23
Bau: Return MFA Method from getMfaMethods, rather than response data object

BeckaL:
## What

We currently have two ways of representing an mfa method in the shared project:

* MFAMethod - which represents the item stored in the database; and
* MfaMethodResponse - which is a representation of the response we return for mfa methods in the method management api.

Originally this second object was introduced as a generic representation of the mfa method that could be used in both the account management api and the frontend api, but:

* we want to decouple these apis as much as possible; and
* there are slight differences in the data required by them (for example, the frontend api only ever returns a partially obscured phone number)

This PR, therefore, is the first step in moving the MfaMethodResponse from shared into the account management api. It focuses on the retrieve mfa methods functionality only, and ensures that in our shared MfaMethodsService, the retrieve method returns a list of MFAMethods, rather than a list of MfaMethodResponses, to make it as reusable as possible.

Note that this is the first step in the overall refactoring - the other methods in this shared service still use the MfaMethodResponse in their return types, so this PR alone does not enable us to move the MfaMethodResponse and associated records out of shared.

## How to review

1. Code Review

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/helpers/MfaMethodResponseConverterHelper.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/helpers/MfaMethodResponseConverterHelper.java
@@ -1,43 +1,30 @@
 package uk.gov.di.accountmanagement.helpers;
 
-import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.shared.entity.Result;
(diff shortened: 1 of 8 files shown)


BeckaL:
Proposing on not fixing the minor difference between test coverage expectation (>80%) and reality here, since this is largely complaining about a comparison method on an object that we only use as a helper for ensuring equality in tests themselves

mattvot (approved):

PR #6387, opened 2025-04-23
ATO-1314: generate new IPV signing key pair in orch account

sw-mattw:
### Wider context of change

At present, orch signs requests to IPV with the private key `IPV_TOKEN_SIGNING_KEY_ALIAS` which is resides in the `gds-di-*` AWS accounts. We wish to move this into the `di-orchestration-*` AWS accounts.

We first wish to add the new key resource to the orch account and publish this to the IPV JWKS endpoint in the same way we do with the active key. We wish to control promotion up to production.

In subsequent PRs, we will make this the active key and remove the key stored in the `gds-di-*` AWS accounts.

### What’s changed

- Added new orch IPV token signing key with same properties as the existing key
- Added new alias for the new signing key
- Added env var to control publishing of the new orch IPV token signing key on the IPV JWKS endpoint (`ORCH_IPV_TOKEN_SIGNING_KEY_PUBLISH_ENABLED`)
- Added new orch IPV token signing key to the JWKS value returned by the IPV JWKS endpoint
  - Conditional on environment

### Manual testing

Deployed to orch dev, navigated to the IPV JWKS endpoint (https://oidc.sandpit.account.gov.uk/.well-known/ipv-jwks.json), observed two keys (active and new inactive), reviewed logs for this lambda and saw both the old active key and the new inactive key were being retrieved, correlated key IDs with those in KMS - as expected.

### Checklist

  - Access policy for the new (inactive) key added to the IPV JWKS lambda. Confirmed permissions worked as expected through manual testing (see above).

- [ ] Changes have been made to contract tests or not required. **- N/A**

- [ ] Changes have been made to the simulator or not required. **- N/A**

- [ ] Changes have been made to stubs or not required. **- N/A**

- [ ] Successfully deployed to authdev or not required. **- N/A**

- [ ] Successfully run Authentication acceptance tests against sandpit or not required. **- N/A**


orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/basetest/IntegrationTest.java
```diff
@@ -336,5 +336,10 @@ public boolean supportMaxAgeEnabled() {
         public boolean isReturnAuthTimeInIdTokenEnabled() {
             return false;
         }
+
+        @Override
+        public boolean isOrchIPVTokenSigningKeyPublishEnabled() {
+            return false;
```
sw-mattw, line 342: Having a think about this.

The existing integration test for the IPV JWKS endpoint checks that the number of keys returned is 1. This PR publishes 2 keys on lower envs temporarily before we switch over to the new key (and remove the old one).

Updating the integration test would require that we register another `TokenSigningExtension` in this file and pass it into the constructor for each initialisation of `TestConfigurationService`. Once the migration is complete we'd need to remove this extension and switch the integration test back to testing that only 1 key is returned.

cearl1 (approved):

PR #6388, opened 2025-04-23
BAU: Allow dev cross account role

mattvot:
## What

We allow cross account roles for build and staging to allow our acceptance tests to run (as they need to touch both accounts while we are in a partial secure pipeline migration).

Here dev support is added so we can run the acceptance tests against it too.

## How to review

1. Code Review

## Checklist

### DIFF
--- a/ci/terraform/shared/iam.tf
+++ b/ci/terraform/shared/iam.tf
@@ -1,8 +1,11 @@
-
-
 # Create a new IAM role for cross account access
+locals {
+  allowed_environments     = ["dev", "build", "staging"]
+  allow_cross_account_role = contains(local.allowed_environments, var.environment)
+}
+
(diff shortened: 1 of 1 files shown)


ci/terraform/shared/iam.tf
```diff
@@ -1,8 +1,10 @@
-
-
 # Create a new IAM role for cross account access
+locals {
+  allow_cross_account_role = var.environment == "dev" || var.environment == "build" || var.environment == "staging"
+}
```
pskushwaha1, line 4: we can have a better list condition 
 locals {
  allowed_environments               = ["dev", "build", "staging"]
  allow_cross_account_role         = contains(local.allowed_environments, var.environment)
}

pskushwaha1 (approved):

PR #6413, opened 2025-04-28
ATO-1128: add cross account read access for authUserInfo table

GHSwallow:
### Wider context of change
Part of email migration work.

### What’s changed
Adds
- cross account encrypt access to the authUserInfo table encryption key
- cross account read access to the authUserInfo table

### Manual testing
N/A

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -570,6 +570,25 @@ Resources:
             Condition:
               ArnLike:
                 kms:EncryptionContext:aws:dynamodb:table/arn: !Sub arn:aws:dynamodb:${AWS::Region}:${AWS::AccountId}:table/*
+          - Sid: AllowAuthUserInfoKeyCrossAccountEncryptActions
+            Effect: Allow
+            Principal:
+              AWS: !Sub
+                - arn:aws:iam::${AuthAccountId}:root
+                - AuthAccountId:
+                    !FindInMap [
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -570,6 +570,20 @@ Resources:
...
+              AWS: !Sub
+                - arn:aws:iam::${AuthAccountId}:root
+                - AuthAccountId:
+                    !FindInMap [
+                      EnvironmentConfiguration,
+                      !Ref Environment,
+                      authAccountId,
+                    ]
+            Action:
+              - kms:Decrypt
```
cearl1, line 585: I'm not sure if its needed, but i remember a client session migration ticket where I needed more than just Decrypt when reading from an encrypted dynamo table. ATO-1491 was the ticket I made to add the missing permissions which has a bit more context.

If this has been tested and all is good, i'm happy to approve, just wary of the key issues.

GHSwallow: ooh ok, yea I'm happy to add those changes 👍

GHSwallow:
I thought we would need authUserInfo in ProcessingIdentity, but alas it seems we don't, so no need for this cross account access. 
Actual migration as below:
https://github.com/govuk-one-login/authentication-api/pull/6427

PR #6421, opened 2025-04-28
ATO-1128: duplicate role and add auth user info access

GHSwallow:
### Wider context of change
Part of email migration. This is the second step in adding cross account access: defining the new role that this lambda will use. We need to define it in a separate PR, as there will be temporary downtime if we overwrite a policy.

### What’s changed
Defines a new policy for adding access to read AuthUserInfo table on orch accounts. In order to do this, I've also created combo policies for other orch tables, so the action access and key access are in one policy, as we were reaching the limit of 20 policies.

### Manual testing
N/A. I will manually test once we swap over to the new one that it deploys ok, and a full identity journey still works.

### Checklist

### Related PRs
PR to add the resource policy onto the AuthUserInfo table, allowing this cross-account access:
https://github.com/govuk-one-login/authentication-api/pull/6413


GHSwallow:
I thought we would need authUserInfo in ProcessingIdentity, but alas it seems we don't, so no need for this cross account access. 
Actual migration as below:
https://github.com/govuk-one-login/authentication-api/pull/6427

PR #6449, opened 2025-05-02
ATO-1435: Improve logging for new claim equality

cearl1:
### Wider context of change

We recently started sending claims through from orch to auth backend, to replace the need for the client session. We are logging whether the new claims are the same as the equivalent fields on the client session. We have noticed a few differences in the logs indicating they are not in sync, but its hard to tell where to start looking.

### What’s changed

This PR refactors the sync logging a bit, outputting whether either field is null. This should help narrow down where to look to fix the sync issues.

### Checklist

Just logging changes, no new functionality added that needs testing.


james-peacock-gds (approved):
LGTM!

PR #6453, opened 2025-05-02
ATO-1616: Emit metric for auth query param size

cearl1:
### Wider context of change

Previously we had an incident where a new field was included in the JAR sent from orch to auth, which caused the query parameters of the redirect to auth to exceed the limit of our WAF. We did increase the query params limit in the WAF, however we will be adding new claims to this JAR and need a way of telling how close we are to the limit.

### What’s changed

This PR adds a new metric called "AuthRedirectQueryParamSize", which is the length of all query parameters in the redirect to auth. Tests have also been updated to assert on this metric being set.

### Manual testing

Deployed to sandpit. Saw new metric in CloudWatch after being redirected to the login page.

### Checklist


james-peacock-gds (approved):
LGTM

PR #6470, opened 2025-05-08
Aut 4159/use ipv jwks endpoint

alhcomer:
## What

Makes the IPV reverification service call the IPV JWKS endpoint to get their public encryption key. This is then used to encrypt the JWT send over in the MFA reset authorize handler.

Also a small refactor (courtesy of @andrew-moores)

## How to review

1. Code Review
    - Take particular care to check that the new logic is only used in staging. Lower environments should still grab the key from 
       the configuration service

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/IPVReverificationService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/IPVReverificationService.java
@@ -23,6 +23,8 @@ import uk.gov.di.authentication.shared.exceptions.MissingEnvVariableException;
 import uk.gov.di.authentication.shared.helpers.IdGenerator;
 import uk.gov.di.authentication.shared.helpers.NowHelper.NowClock;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
+import uk.gov.di.authentication.shared.services.KmsConnectionService;
+import uk.gov.di.authentication.shared.services.RedisConnectionService;
 import uk.gov.di.authentication.shared.services.TokenService;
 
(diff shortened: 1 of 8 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/IPVReverificationService.java
```diff
@@ -39,23 +48,69 @@ public class IPVReverificationService {
     private final JwtService jwtService;
     private final NowClock nowClock;
     private final TokenService tokenService;
+    private final JWKSource<SecurityContext> jwkSource;
+
+    /**
```
dbes-gds, line 53: rogue comments

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -702,4 +704,18 @@ public boolean isUsingStronglyConsistentReads() {
                 .getOrDefault("USE_STRONGLY_CONSISTENT_READS", FEATURE_SWITCH_OFF)
                 .equals(FEATURE_SWITCH_ON);
     }
+
+    public URL getIpvJwksUrl() throws MalformedURLException {
+        return new URL(System.getenv("IPV_JWKS_URL"));
```
andrew-moores, line 709: Could we add a check here to log an error message if IPV_JWKS_URL isn't set?  That would make it easier to see what's gone wrong and triage in prod if its switched on but the config isn't updated.

andrew-moores (approved):

andrew-moores (approved):

dbes-gds (approved):

PR #6495, opened 2025-05-14
BAU: Pin jetty-http2-common

CarlyG55:
This has a high vulnerability, but is a transitive dependency, and there is not a new version of the high level dependency so we need to pin this

### Checklist


james-peacock-gds (approved):
LGTM!

PR #6498, opened 2025-05-14
ATO-1437: Swap to using clientId from orch instead of client session

cearl1:
### Wider context of change

We would like to stop using clientSession and instead use the values of claims we send from orch to auth. By this point, we have claims available in auth backend that can replace the use of the auth request parameters on the clientSession.

We are storing the claims `levelOfConfidence`,`credentialTrustLevel`, and `clientId` in the auth session, as these are used outside of the StartHandler. The rest of the claims (`cookieConsent`, `_ga`, `state`, `scope`, `redirect_uri`) are only used in the StartHandler.

### What’s changed

This PR swaps to using the `clientId` we are storing on the auth session from orch, instead of using the value from the client session auth request params.

### Manual testing

Tested in sandpit and authdev
- Did a couple of auth and identity journeys, seemed to work as expected

### Checklist

### DIFF
--- a/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/AuthSessionExtension.java
+++ b/shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/AuthSessionExtension.java
@@ -82,6 +82,10 @@ public class AuthSessionExtension extends DynamoExtension implements AfterEachCa
         updateSession(getSession(sessionId).orElseThrow().withEmailAddress(email));
     }
 
+    public void addClientIdToSession(String sessionId, String clientId) {
+        updateSession(getSession(sessionId).orElseThrow().withClientId(clientId));
+    }
+
     public void addInternalCommonSubjectIdToSession(
             String sessionId, String internalCommonSubjectIdl) {
         updateSession(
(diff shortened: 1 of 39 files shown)


shared-test/src/main/java/uk/gov/di/authentication/sharedtest/extensions/AuthSessionExtension.java
```diff
@@ -78,6 +79,11 @@ public void addSession(String sessionId) {
         authSessionService.addSession(authSessionService.generateNewAuthSession(sessionId));
     }
 
+    public void addSession(String sessionId, UnaryOperator<AuthSessionItem> updateFn) {
+        var authSession = authSessionService.generateNewAuthSession(sessionId);
+        authSessionService.addSession(updateFn.apply(authSession));
+    }
+
```
CarlyG55, line 86: I'm not totally convinced by this approach - it seems to be a bit 'clever' and maybe a bit hard to read if you're coming in fresh. I will ponder it tomorrow when I have a bit more brain power 😅

CarlyG55: Nice, thanks - I appreciate the thought, but I think it might just be a little easier to understand!

cearl1: Yeah in hindsight it makes sense to stick to the pattern. I'll add another method that updates the client id for an auth session and use that instead

CarlyG55 (approved):

PR #6499, opened 2025-05-14
ATO-1437: Swap to using remaining claims from orch instead of client session

cearl1:
### Wider context of change

We would like to stop using clientSession and instead use the values of claims we send from orch to auth. By this point, we have claims available in auth backend that can replace the use of the auth request parameters on the clientSession.

We are storing the claims `levelOfConfidence`,`credentialTrustLevel`, and `clientId` in the auth session, as these are used outside of the StartHandler. The rest of the claims (`cookieConsent`, `_ga`, `state`, `scope`, `redirect_uri`) are only used in the StartHandler.

### What’s changed

This PR uses the following claims from orch, instead of the auth request parameters in the client session:
- `cookieConsent`
- `_ga`
- `state`
- `scope`
- `redirect_uri`

These claims are only used in the StartHandler, so the scope of these changes are quite small.

### Manual testing

Tested in sandpit and authdev
- Did a couple of auth and identity journeys, seemed to work as expected

### Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -202,9 +202,7 @@ public class StartHandler
             var cookieConsent =
                     startService.getCookieConsentValue(
                             startRequest.cookieConsent(), startRequest.clientId());
-            var gaTrackingId =
-                    startService.getGATrackingId(
-                            userContext.getClientSession().getAuthRequestParams());
+            var gaTrackingId = startRequest.ga();
             var reauthenticateHeader =
                     getHeaderValueFromHeaders(
                             input.getHeaders(),
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
@@ -4,7 +4,7 @@ import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
-import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.id.State;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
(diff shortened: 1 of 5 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
```diff
@@ -271,12 +269,9 @@ private boolean isClientCookieConsentShared(String clientID) throws ClientNotFou
                                                 clientID)));
     }
 
-    private boolean validCookieConsentValueIsPresent(
-            Map<String, List<String>> authRequestParameters) {
-        return authRequestParameters.containsKey(COOKIE_CONSENT)
-                && !authRequestParameters.get(COOKIE_CONSENT).isEmpty()
-                && authRequestParameters.get(COOKIE_CONSENT).get(0) != null
+    private boolean validCookieConsentValueIsPresent(String cookieConsent) {
+        return cookieConsent != null
```
CarlyG55, line 237: Do we need to worry about empty string or anything here?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -202,9 +202,7 @@ public APIGatewayProxyResponseEvent handleRequest(
             var cookieConsent =
                     startService.getCookieConsentValue(
                             startRequest.cookieConsent(), startRequest.clientId());
-            var gaTrackingId =
-                    startService.getGATrackingId(
-                            userContext.getClientSession().getAuthRequestParams());
+            var gaTrackingId = startRequest.ga();
```
CarlyG55, line 213: I assume no worries around the underscore here?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -288,8 +296,10 @@ public APIGatewayProxyResponseEvent handleRequest(
             var errorMessage = "Unable to serialize start response";
             LOG.error(errorMessage, e);
             return generateApiGatewayProxyResponse(400, errorMessage);
-        } catch (ParseException e) {
```
CarlyG55, line 294: Where's the parse exception disappeared from?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/StartService.java
cearl1: I dont think so, if its an empty string it will not exist in the list of valid cookieConsent values (the line below this one), so will return false

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
cearl1: Nope, thats defined in StartRequest
```java
        @Expose @SerializedName("_ga") String ga,
```

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
cearl1: This was thrown from the old `buildClientStartInfo` method, from the call to `AuthenticationRequest.parse()`. For some reason it caught a ParseException, then threw another one with a message.

The new version of the method doesn't try to build an `AuthenticationRequest`, so wont throw a ParseException

CarlyG55 (approved):

PR #6501, opened 2025-05-14
OLH-2621: Check for interventions in authenticate handler 

alex9smith:
## What

Call AIS to check for interventions in the authenticate handler. If the user has a blocked or suspended intervention, return a 403 and an error code.

When a user goes to change eg. their password or MFA settings in the account management frontend they're prompted to re-enter their password. That's checked by calling this API route. If the user has an intervention that should stop them using their account, the account management frontend needs to be able to block that change and log the user out to prevent any further changes or access. 

This is behind a feature flag which is switched off everywhere except the dev environments.

Note this is a synchronous, blocking call. This is intentional - if we can't check for interventions at this stage then we can't allow the user to proceed.

I've also taken the opportunity to give this section of the OpenAPI spec some love. We're going to be writing new behaviour on the consumer side of this API so it'll be useful to have the new error codes documented.

## How to review

1. Code Review

I've done the following in authdev1 to test the functionality:

1. Deploy to dev
1. Generate an access token `./generate-auth-token.sh <environment> <email address>` 
1. Call the `/authenticate` endpoint with your email and password and verify a 204 response:
> `http POST {BASE_URL}/authenticate 'Authorization: Bearer {TOKEN}' 'email={EMAIL_ADDRESS}' 'password={PASSWORD}'`
4. Add a blocked or suspended intervention for your user using the sub returned from the access token script
5. Call the `/authenticate` endpoint again and see the 403 response: 

```bash
authentication-api (OLH-2621-call-ais-on-authenticate) a023b147e $ http POST {BASE_URL}/authenticate 'Authorization: Bearer {TOKEN}' 'email={EMAIL_ADDRESS}' 'password={PASSWORD}'
HTTP/1.1 403 Forbidden
...

{
    "code": 1084,
    "message": "User's account is blocked"
}
```

## Checklist

## Related PRs

#6496


TaherKapasi (approved):

ci/terraform/account-management/variables.tf
```diff
@@ -208,6 +208,18 @@ variable "mfa_method_management_api_enabled" {
   type        = bool
   default     = false
 }
+
+variable "account_intervention_service_uri" {
+  default = "undefined"
```
TaherKapasi, line 213: would default this to `""`

TaherKapasi (dismissed):

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandler.java
```diff
@@ -10,25 +10,18 @@
...
-import uk.gov.di.authentication.shared.services.ConfigurationService;
-import uk.gov.di.authentication.shared.services.DynamoService;
-import uk.gov.di.authentication.shared.services.SerializationService;
+import uk.gov.di.authentication.shared.services.*;
 
 import java.util.Optional;
 
-import static uk.gov.di.accountmanagement.domain.AccountManagementAuditableEvent.AUTH_ACCOUNT_MANAGEMENT_AUTHENTICATE;
-import static uk.gov.di.accountmanagement.domain.AccountManagementAuditableEvent.AUTH_ACCOUNT_MANAGEMENT_AUTHENTICATE_FAILURE;
+import static uk.gov.di.accountmanagement.domain.AccountManagementAuditableEvent.*;
```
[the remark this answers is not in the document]
alex9smith: Ah, IntelliJ defaults to using wildcards when there's more than 4(?) imports from a file. Happy to switch them back if we'd prefer

ci/terraform/account-management/openapi_v2.yaml
TaherKapasi, line 1: Hi @alex9smith - could you bump the api version from version: 1.0.1 -> 1.0.2 please

TaherKapasi (changes requested):

TaherKapasi (approved):

alhcomer (approved):

PR #6519, opened 2025-05-17
AUT-4297: add mobile as enum on orch

GHSwallow:
## What
Added "mobile" to orch's Channel enum. Orch doesn't do much with this value - they just get it from the ClientRegistry and pass it to Auth. The only use of the enum is in `ClientRegistrationRequestHandler`, which checks if that channel is valid.

The default value is still `web`, as a number of clients don't have a channel property, and only a single client will use the new `mobile` channel. 

Currently, no clients use this channel. To complete this work, a client registry change should be made for gov.uk app integration client, client-id `5o5i_n_2oBqvrAiy6f0JW_jzUBE`, and any future client registry entries for gov.uk app. This won't happen until the parent ticket is complete.

## How to review
1. Code Review

## Checklist

- [x] No changes required or changes have been made to stub-orchestration. **See related PRs**

## Related PRs
Change to orch stub: https://github.com/govuk-one-login/authentication-stubs/pull/349

### DIFF
--- a/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
+++ b/oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
@@ -480,8 +480,20 @@ class AuthorisationHandlerTest {
                     expectedClaimSetRequest.toJSONObject(), actualClaimSetRequest.toJSONObject());
         }
 
-        @Test
-        void shouldPassTheChannelClaimToAuth() {
+        private static Stream<Arguments> clientChannelsAndExpectedChannels() {
+            return Stream.of(
+                    arguments(null, Channel.WEB.getValue()),
+                    arguments(Channel.WEB.getValue(), Channel.WEB.getValue()),
+                    arguments(Channel.STRATEGIC_APP.getValue(), Channel.STRATEGIC_APP.getValue()),
+                    arguments(Channel.MOBILE.getValue(), Channel.MOBILE.getValue()));
+        }
+
+        @ParameterizedTest
+        @MethodSource("clientChannelsAndExpectedChannels")
+        void shouldPassTheCorrectChannelClaimToAuth(
+                String clientChannel, String expectedChannelClaim) {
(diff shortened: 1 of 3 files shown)


oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
```diff
@@ -480,8 +480,20 @@ void shouldPassTheCorrectClaimsToAuthForHighLevelTrustJourneys()
...
-        @Test
-        void shouldPassTheChannelClaimToAuth() {
+        private static Stream<Arguments> clientChannelsAndExpectedChannels() {
+            return Stream.of(
+                    arguments(null, Channel.WEB.getValue()),
+                    arguments(Channel.WEB.getValue(), Channel.WEB.getValue()),
+                    arguments(Channel.STRATEGIC_APP.getValue(), Channel.STRATEGIC_APP.getValue()),
+                    arguments(Channel.MOBILE.getValue(), Channel.MOBILE.getValue()));
+        }
+
```
Ryan-Andrews99, line 490: Super minor comment because I'm happy this adds more coverage, but do we need two arguments here? Think this could be simplified to:
```
assertEquals(clientChannel, actualChannelClaim);
```

Ryan-Andrews99 (approved):
One minor comment - feel free to disregard

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandlerTest.java
GHSwallow: We only need the two arguments for the case that the client has a `null` channel, in which case the `actualChannelClaim` is expected to be `WEB`, not `null`. I agree it's little icky when the other 3 cases match, but I can't think of a cleaner way to capture this expected logic. What do you think?

Ryan-Andrews99: Ahhh yeah that makes sense. Sorry completely missed that!

PR #6530, opened 2025-05-19
Aut 4320/check phone number not in use

andrew-moores:
## What

Update the check that a phone number, used for 2FA, is not already being used.  

## How to review

1. Code review
2. Deploy to dev and run frontend and backend acceptance tests.

## Checklist

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
@@ -20,12 +20,13 @@ import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.entity.PendingEmailCheckRequest;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.JourneyType;
-import uk.gov.di.authentication.shared.entity.UserProfile;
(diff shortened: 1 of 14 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/SendOtpNotificationHandler.java
```diff
@@ -246,20 +252,68 @@ public APIGatewayProxyResponseEvent sendOtpRequestHandler(
...
+                    String newPhoneNumber;
+                    try {
+                        newPhoneNumber =
+                                PhoneNumberHelper.formatPhoneNumber(
+                                        sendNotificationRequest.getPhoneNumber());
+                    } catch (Exception e) {
+                        return generateApiGatewayProxyErrorResponse(400, INVALID_PHONE_NUMBER);
                     }
+
+                    if (!userProfile.get().getMfaMethodsMigrated()) {
```
mattvot, line 272: Instead of conditioning here, can we use the MFAMethodService here to get the methods so we don't need to care about migration, which I think would simplify the implementation here.

FWIW I've started adding a helper utility for finding items in the List<MFAMethod> in an open PR, but this is probably not exactly what you need here https://github.com/govuk-one-login/authentication-api/blob/AUT-4248/use-backup-mfa/shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/MFAMethodsService.java#L231C1-L249C6

PR #6536, opened 2025-05-20
ATO-1420: Swap to using orch client session in remaining places

cearl1:
### Wider context of change

We have verified that the orch client session is in sync with the client session previously, and in most places we have been using the orch client session instead of the redis client session for some time now. There were a few prod places where its still using the redis client session though.

### What’s changed

This PR updates the remaining few places to use the orch client session instead of the redis client session. After this PR, the only uses of the redis client session in orch would be in tests, or in the client session service (which will also be removed)

### Manual testing

Deployed to sandpit, tested a few auth and identity journeys. All seemed fine.
Also did a docapp journey to check nothing has broken there

### Checklist


james-peacock-gds (approved):
LGTM!

CarlyG55 (approved):

PR #6538, opened 2025-05-20
ATO-1621: remove ccs auth session

Ryan-Andrews99:
### Wider context of change: 

We're migrating away from using the field CurrentCredential Strength on the new separate sessions we've setup, as we're aiming to make this a solely authentication owned value, which is given to Orchestration. Now we've migrated the main usage of this value to AchievedCredentialStrength, we can now remove getting and setting it.

### What’s changed: 
- Stops sending orchestration the value of Current Credential Strength - because this value was previously null in _some_ cases, the orchestration side of the code is set up to be able to handle this being null - so it's safe to remove this now.
- Stop setting it in VerifyMfaCode handler 
- Stop setting it initially in the start handler based on the Orchestration derived value -  we're removing this value completely so we can remove this.

### Manual testing: 
- Deployed to authdev2 
- Run through a couple of journeys
- Run through an of uplift journey

### Checklist

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
@@ -7,7 +7,6 @@ import software.amazon.awssdk.enhanced.dynamodb.Key;
 import software.amazon.awssdk.enhanced.dynamodb.model.GetItemEnhancedRequest;
 import software.amazon.awssdk.services.dynamodb.DynamoDbClient;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
-import uk.gov.di.authentication.shared.entity.CredentialTrustLevel;
 import uk.gov.di.authentication.shared.exceptions.AuthSessionException;
 import uk.gov.di.authentication.shared.helpers.InputSanitiser;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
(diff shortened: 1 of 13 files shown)


CarlyG55 (commented):
appreciate this is only in draft but it looks good to me!

isaac-GDS (approved):

shared/src/main/java/uk/gov/di/authentication/shared/services/AuthSessionService.java
```diff
@@ -67,9 +66,7 @@ public void addSession(AuthSessionItem authSessionItem) {
     }
 
     public AuthSessionItem getUpdatedPreviousSessionOrCreateNew(
-            Optional<String> previousSessionId,
-            String newSessionId,
-            CredentialTrustLevel currentCredentialStrength) {
```
CarlyG55, line 72: Do we not need to add achieved credential strength here?

Ryan-Andrews99: Nope, because Auth now set achieved credential strength in all cases a user completes a journey -  they don't need to rely on Orchestration doing the working out of this value and sending it back to them

CarlyG55: cool

CarlyG55 (approved):

PR #6547, opened 2025-05-21
ATO-1621: Remove Current Credential Strength Orch session

Ryan-Andrews99:
### Wider context of change: 
Previously we migrated this field to the Orch session from redis. This was done as the existing logic between Authentication and Orchestration required both sides to determine this value, as it was previously not set in all cases. However this left some circular and fairly fragile logic between the components. Given Authentication should be the sole component concerned with this value, we have swapped to a new value called AchievedCredentialStrength which is more reliable and only set by Auth. 

This has meant there is no need currently for Orchestration to keep this value in it's own session store. We have some logging in place 

### What’s changed
- Orchestration no longer set the value of current credential strength on it's session
- Removes it from the Authentication claims 
- Stops sending Current Credential Strength to Auth as it is no longer used
- 

### Manual testing: 

Deployed to sandpit:
- Run through a couple of journeys:
- Ran through a low confidence journey
- Came back to stub, selected medium and was presented with uplift screen

### Checklist

### Related PRs: 
- Orch stub PR which reflects this change: https://github.com/govuk-one-login/authentication-stubs/pull/359

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
@@ -569,21 +569,8 @@ public class AuthenticationCallbackHandler
                                 clientRedirectURI, authCode, null, null, state, null, responseMode);
 
                 sessionService.storeOrUpdateSession(session, sessionId);
-                var currentCredentialStrength =
-                        userInfo.getStringClaim(
-                                AuthUserInfoClaims.CURRENT_CREDENTIAL_STRENGTH.getValue());
-                if (isNull(currentCredentialStrength)
-                        || lowestRequestedCredentialTrustLevel.compareTo(
-                                        CredentialTrustLevel.valueOf(currentCredentialStrength))
-                                > 0) {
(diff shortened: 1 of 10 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthenticationCallbackHandler.java
```diff
@@ -876,7 +876,8 @@ private void handleMaxAgeSession(OrchSessionItem currentOrchSession, TxmaAuditUs
 
     private void logComparisonRequestCredentialTrustAndAchieved(
             UserInfo authUserInfo, CredentialTrustLevel requestedCredentialStrength) {
-
+        // TODO: This logging currently looks fine but in future we should validate that
+        // this value is non-null and >= the requested credential strength
```
CarlyG55, line 880: How much noise is this adding to our logs right now? Want to be a little wary of adding too much

Ryan-Andrews99: Hmm it is being logged for every request but we have a lot of temp logging in this lambda that less useful imo

CarlyG55 (approved):

PR #6556, opened 2025-05-22
ATO-1591: Get client id from auth session instead of from client registry

cearl1:
### Wider context of change

### What’s changed

This PR swaps to using the auth session to get the client id instead of the ClientRegistry. The client ID is already set on the auth session in StartHandler, this is just swapping over the getters. I've also removed all uses of the old client ID getter, but cannot remove it yet because it's the partition key for the dynamo table.

### Manual testing

Tested in sandpit, performed and auth only and identity journey. Both were successful.

Ran acceptance tests too, they all passed

### Checklist

- [x] Successfully run Authentication acceptance tests against sandpit or not required. See above

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckEmailFraudBlockHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckEmailFraudBlockHandlerTest.java
@@ -92,7 +92,6 @@ class CheckEmailFraudBlockHandlerTest {
     @BeforeEach
     void setup() {
         var userProfile = generateUserProfile();
-        when(clientRegistry.getClientID()).thenReturn(CLIENT_ID);
         when(userContext.getClient()).thenReturn(Optional.of(clientRegistry));
         when(userContext.getSession()).thenReturn(session);
         when(userContext.getAuthSession()).thenReturn(authSession);
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandlerTest.java
@@ -144,7 +144,6 @@ class UpdateProfileHandlerTest {
         when(authenticationService.getUserProfileFromEmail(EMAIL))
                 .thenReturn(Optional.of(generateUserProfile()));
         when(clientService.getClient(CLIENT_ID.getValue())).thenReturn(Optional.of(clientRegistry));
-        when(clientRegistry.getClientID()).thenReturn(CLIENT_ID.getValue());
 
         var body =
                 format(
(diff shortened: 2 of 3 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
```diff
@@ -37,6 +38,23 @@ public static Subject getSubject(
         }
     }
 
+    public static Subject getSubject(
```
Louisasa: v minor but this gets removed in a later commit, so could just not be added?

shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
```diff
@@ -60,20 +61,36 @@ void shouldReturnDifferentSubjectIDForMultipleClientsWithDifferentSectors() {
                         PAIRWISE.toString(),
                         "https://test.com",
                         false);
+        var authSession1 =
```
Louisasa, line 64: again minor, I'm conscious there's a lot of duplicated code here, is there any we could pull out or commonise i.e. all `authSession1` and `authSession2` and only use as needed, maybe even pull out `test-client-id` and `test-auth-session`?

shared/src/main/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelper.java
cearl1: I'm a bit confused. The method above this one is removed in a later commit (the one that does not take an AuthSessionItem as a parameter).

I've added another method alongside the existing one so I can update each handler in its own commit, and so i dont break any tests.

shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
cearl1: Yeah to be honest i dont think the auth sessions need to have different session IDs. I did have a helper method that took a sessionId and clientId, but found that it didnt make it that much more readable and it was the same amount of code.

I'll have a play around and see if i can improve this test setup!

cearl1: Ahh nice yeah I ended up doing what you suggested - extracting some constants for client id 1 and 2, and auth sessions for each client id. Now we just use these constants in the tests so they're a lot less busy

shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
```diff
@@ -146,35 +165,43 @@ void shouldReturnPairwiseSubjectIdWhenClientTypeIsPairwise() {
...
-                        keyPair, "test-client-id-1", PAIRWISE.toString(), "https://test.com", true);
+                        keyPair, CLIENT_ID_1, PAIRWISE.toString(), "https://test.com", true);
         var clientRegistry2 =
                 generateClientRegistryPairwise(
-                        keyPair,
-                        "test-client-id-1",
-                        PAIRWISE.toString(),
-                        "https://test.com",
-                        false);
+                        keyPair, CLIENT_ID_2, PAIRWISE.toString(), "https://test.com", false);
```
cearl1, line 171: I noticed this test was using the same client id for different client registry entries, which was a bit unrealistic. I've updated it to use a unique id instead. This doesn't affect the test fortunately

PR #6561, opened 2025-05-23
ATO-1665: update backchannel logout logging

Louisasa:
### Wider context of change

A couple of RPs have been timing out, resulting in no splunk query showing but still being added to the DLQ.

### What’s changed

Added a new timeout variable (default copied from Account Intervention Service) and new Exception.

### Manual testing

I haven't tested it because it's not easy to test. Let me know if you can think of a good way to test (although I don't think it's that big of a change that needs testing)

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/BackChannelLogoutRequestHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/BackChannelLogoutRequestHandler.java
@@ -10,6 +10,7 @@ import com.nimbusds.jwt.JWTClaimsSet;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
+import uk.gov.di.authentication.oidc.exceptions.HttpRequestTimeoutException;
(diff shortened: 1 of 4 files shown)


cearl1 (approved):

cearl1:
Sorry forgot to add a comment when i approved it. I think the new test you added which mocks the post request and throws the timeout exception should be enough test coverage for this!

ci/terraform/oidc/variables.tf
```diff
@@ -686,6 +686,12 @@ variable "cmk_for_back_channel_logout_enabled" {
   description = "Feature flag which toggles whether the back channel logout queue is encrypted using CMK"
 }
 
+variable "back_channel_logout_call_timeout" {
+  default     = 3000
+  type        = number
+  description = "The HTTP Client connection timeout for requests to back channel logout URI (in milliseconds)."
+}
+
```
CarlyG55, line 694: I think we should make this a bit longer - the lambda can run for up to 30 seconds. AIS is one of our own services, so I think we'd expect that to come back quicker than a random RP service. How about 15 seconds?

CarlyG55 (changes requested):
Oh actually just realised that this hasn't been added to the cloudformation template, so we won't actually have a value for the variable - we've just added it to terraform which is the old deployment

CarlyG55 (approved):

PR #6565, opened 2025-05-23
ATO-1670: Set client_name on auth session in StartHandler

cearl1:
### Wider context of change

We are now sending the `client_name` claim from the frontend to the backend. We would like to use this claim instead of using the ClientRegistry.getClientName() method. To do this, we will need to add this claim to the auth session

### What’s changed

The `clientName` field has been added to the auth session. We are also setting this field in the StartHandler, using the claim we get from the frontend.

### Manual testing

Deployed to sandpit. Started a journey and checked the AuthSession dynamo table in AWS. Confirmed that the ClientName field was set there.

Finished a couple of journeys and they all passed successfully.

### Checklist


cearl1:
> Assuming the Orch stub is already sending this value too?

Yep, it's already sending client_name

PR #6600, opened 2025-05-30
ATO-1676: use cookie_consent_shared instead of client registry

Louisasa:
### Wider context of change

### What’s changed

This PR swaps to using the cookie_consent_shared claim sent from orch instead of the ClientRegistry.isCookieConsentShared() method.

### Manual testing

Tested in authdev2, did a couple of journeys and all got the cookie from the frontend.

### Checklist


integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
```diff
@@ -144,7 +144,6 @@ void shouldReturn200AndStartResponse(
                 "clientName":"test-client-name",
                 "scopes":["openid"],
                 "serviceType":"MANDATORY",
-                "cookieConsentShared":false,
```
cearl1, line 147: I think this field should still be here as this is what we send back to the frontend. This might be causing this test to fail.

PR #6616, opened 2025-06-02
ATO-1682: Use isSmokeTest from auth session instead of client registry

cearl1:
### Wider context of change

### What’s changed

This PR swaps to using the isSmokeTest field stored on the auth session instead of the ClientRegistry.isSmokeTest() method.

### Manual testing

Tested in authdev with a few journeys - all worked fine

### Checklist


Ryan-Andrews99 (commented):
Might be good to add some logging to check things are in sync before merging this

cearl1:
Have introduced logging to check whether the new claim in the auth session is in sync with the client registry. There was no cases where they were not in sync overnight, so we can be fairly confident this swap will be safe.

PR #6644, opened 2025-06-05
ATO-1661: Create GlobalLogout lambda

cearl1:
- This just adds the infrastructure (a lambda triggered by a test SQS queue)

### Wider context of change

We would like a lambda that is triggered from an SQS queue. When it receives a message with an internalCommonSubjectId, it should log the user out of all their sessions and client sessions.

### What’s changed

This PR adds the infrastructure needed to achieve the above ((a lambda triggered by a test SQS queue for the time being). We will be provided an SQS queue from TxMA in higher environments later on, but for now we have created the `GlobalLogoutTestQueue` to help test.

### Manual testing

Deployed to orch dev, confirmed that the lambda has the right permissions to read from the queue

### Checklist


template.yaml
```diff
@@ -2083,6 +2083,168 @@ Resources:
             Unit: Count
   #endregion
 
+  #region Global Logout Lambda
+  GlobalLogoutFunction:
+    Type: AWS::Serverless::Function
+    # checkov:skip=CKV_AWS_115: Reserved concurrency will be looked into in ATO-657
+    # checkov:skip=CKV_AWS_116: DLQ is not appropriate for a Lambda invoked by an API
+    # checkov:skip=CKV_AWS_117: Lambda needs open egress so isn't in VPC
```
Ryan-Andrews99, line 2091: I know these are copy-pasted but some of these could probably be tweaked:
```
# checkov:skip=CKV_AWS_116: Lambda is triggered by SQS so DLQ not needed
```
I also wouldn't think this lambda needs open egress so we should probably put it inside the VPC?

template.yaml
```diff
@@ -2083,6 +2083,168 @@ Resources:
...
+              Dimensions:
+                - Name: FunctionName
+                  Value: !Sub ${Environment}-GlobalLogoutFunction
+            Period: 60
+            Stat: Sum
+            Unit: Count
+  GlobalLogoutTestQueue:
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: !Sub ${Environment}-GlobalLogoutTestQueue
```
Ryan-Andrews99, line 2222: Really minor comment, but do you mind putting a space between the resources here, just reads a bit nicer 😄

template.yaml
```diff
@@ -2083,6 +2083,168 @@ Resources:
...
+                ]
+
+          TXMA_AUDIT_QUEUE_URL:
+            Fn::ImportValue: !Sub orchestration-${Environment}-txma-QueueURL
+          REDIS_KEY: session
+      Policies:
+        - !Ref OrchSessionTableReadAndDeleteAccessPolicy
+        - !Ref ClientSessionTableReadAndDeleteAccessPolicy
+        - !Ref RedisParametersAccessPolicy
+        - !Ref GlobalLogoutTestQueueConsumePolicy
```
Ryan-Andrews99, line 2133: Also if we make the test queue conditional this policy attachment should be also

template.yaml
```diff
@@ -2083,6 +2083,168 @@ Resources:
...
+    Type: AWS::SQS::Queue
+    Properties:
+      QueueName: !Sub ${Environment}-GlobalLogoutTestDlq
+      KmsMasterKeyId: !GetAtt GlobalLogoutTestQueueEncryptionKey.Arn
+      KmsDataKeyReusePeriodSeconds: 300
+      MessageRetentionPeriod: 259200
+  GlobalLogoutTestQueueEncryptionKey:
+    Type: AWS::KMS::Key
+    Properties:
+      Description: KMS encryption key for global logout test queue and dlq
```
Ryan-Andrews99, line 2241: Think we could slap a:
```
Condition: IsDevOrBuild
```
as we don't need to deploy this to any higher environments

Ryan-Andrews99 (commented):
Have left a few comments, nothing major!

template.yaml
cearl1: Yep agreed :) have done so!

template.yaml
cearl1: I didnt realise these were comments after the checkov skip bit! I've updated them to be more accurate

template.yaml
cearl1: Done, thanks!

PR #6647, opened 2025-06-06
ATO-1705: Add logging for invalid channel values

cearl1:
### Wider context of change

We would like to start accepting `channel` in the authorise request to orch. Before we add validation for this we need to check if any RPs are already sending `channel` in the request, and if this will cause the validation to fail.

### What’s changed

This PR checks if the `channel` parameter exists on the authorise request. If it does, it will log the value, and also log whether the claim is considered valid (`web` or `generic_app`)

### Manual testing

Updated a test to send `channel` in the auth request. Saw the logs appear in the test log. I think this should be enough?

### Checklist


Ryan-Andrews99 (commented):
Don't think this covers request objects as the RequestObjectToAuthRequestHelper.transform method only explicitly passes through params. You'll need to add something like:
```
   if (Objects.nonNull(jwtClaimsSet.getStringClaim("channel"))) {
                builder.customParameter(
                        "channel", jwtClaimsSet.getStringClaim("channel"));
            }
```
to the transform method

Ryan-Andrews99 (approved):

PR #6655, opened 2025-06-10
ATO-1709: send is identity verification required to frontend

Louisasa:
### Wider context of change

### What’s changed

Sends is_identity_verification_required to frontend.

### Manual testing

TODO: test on dev

### Checklist


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -918,6 +918,11 @@ private APIGatewayProxyResponseEvent generateAuthRedirect(
                         .map(List::stream)
                         .flatMap(Stream::findFirst);
         var levelOfConfidenceOpt = Optional.ofNullable(requestedVtr.getLevelOfConfidence());
+        var identityRequested =
```
cearl1, line 921: Small nit, but could we rename this to `identityRequired`? I'm just thinking if someone looks at the claims they might get a bit confused :smile:

PR #6659, opened 2025-06-10
ATO-1693: Pass channel claim to auth

cearl1:
Paired on this with @james-peacock-gds 

### Wider context of change

We would like to support RPs providing us with a channel claim, that will be used instead of the channel defined in the client registry. We have already checked with logs that there are no clients sending us this claim at the moment, and we are validating that the channel we receive from the RP is a valid one (not present, `web`, or `generic_app`)

### What’s changed

This PR changes what channel we send based on the claim we get from the RP in the authRequest.
- If the channel is present in the auth request, and it is valid, then it used as the channel value in the JAR we sent to auth.
- If the channel is not present, it will fall back to using the client registry
- If there is no channel set in the client registry, then a default of `web` is used (note this was existing functionality)

### Manual testing

Tested in sandpit. Hit /authorize with a channel parameter of `generic_app` and saw a different login page to when we pass in `web`.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/AuthRequestHelper.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/helpers/AuthRequestHelper.java
@@ -1,0 +1,23 @@
+package uk.gov.di.authentication.oidc.helpers;
+
+import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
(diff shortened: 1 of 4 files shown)


james-peacock-gds:
> is it worth adding logging to check the channel value is the same as client.getChannel() first? happy to leave if you think it's minor enough to not do logging first

Thanks @Louisasa . We've previously done logging to check that RPs aren't currently sending channel in requests. We're expecting that there will be cases where the channel value in request overrides what's in registry (that's the purpose of this particular change) so don't think any further checks are needed.

Louisasa (approved):

PR #6670, opened 2025-06-12
ATO-1696/Add channel to audit event

james-peacock-gds:
### Wider context of change

- We are now allowing RPs to specify `channel` in the auth request and if they do so, this overrides the value in the client registry.

### What’s changed

- When RPs specify `channel` in the auth request, include `channel` when submitting the request parsed event to audit 
- Refactored authoriseRequestHandler to improve readability: extracted a method to sendAuthRequestParsedAuditEvent.


cearl1:
This may need https://github.com/govuk-one-login/event-catalogue/pull/432 to be merged first, not sure if its required but lets play it safe and hold off merging this one :smile:

cearl1 (approved):

PR #6673, opened 2025-06-12
ATO-1660: Logout all sessions by internalCommonSubjectId

cearl1:
### Wider context of change

We would like to support the ability to log out all active sessions that a user has, given their internalCommonSubjectId. We have previously created a lambda triggered by an SQS event from a queue, and we are validating the message we get from SQS has the correct fields.

### What’s changed

This PR adds the logic for logging out users. The new GlobalLogoutService searches for all sessions associated with an internalCommonSubjectId, then reuses some of the LogoutService code to perform the actual logout for these sessions. This service is also used in GlobalLogoutHandler

Although there are no test for whether back channel logout messages are sent, the method we use on the LogoutService has full test coverage, so we can be fairly confident this will work.

### Manual testing

Tested in sandpit by logging into 2 sessions, then manually sending a global logout message for my ICSID (found by looking at my dynamo orch session). 
- Observed the logs of the lambda, saw it delete 2 sessions and 2 client sessions
- Saw both sessions get deleted in dynamo
- Tried doing a reauth journey for both sessions, which said the page had expired and prompted me to login again.

Also tested back channel logouts were still sent by updating sandpit to add a back channel logout URI. The requests were sent to the back channel logout request handler when calling global logout.

### Checklist


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
```diff
@@ -106,7 +106,7 @@ private APIGatewayProxyResponseEvent generateLogoutResponse(
                 302, "", Map.of(ResponseHeaders.LOCATION, logoutUri.toString()), null);
     }
 
-    private void destroySessions(DestroySessionsRequest request) {
+    protected void destroySessions(DestroySessionsRequest request) {
```
Louisasa, line 109: I'm trying to think if it's worth having global logout service as part of logout service to avoid making this anything other than private :/ it's probably fine but I don't really like that theoretically another file can access this. Do you have any opinions on this?

integration-tests/src/test/java/uk/gov/di/authentication/api/GlobalLogoutIntegrationTest.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Test
+    void shouldLogoutAllSessionsForIcsid() {
+        withSessions(
+                sessionWithClientSessions("same-icsid", SESSION_ID_1, CLIENT_SESSION_ID_1),
+                sessionWithClientSessions("same-icsid", SESSION_ID_2, CLIENT_SESSION_ID_2),
+                sessionWithClientSessions("different-icsid", SESSION_ID_3, CLIENT_SESSION_ID_3));
+        var inputMessage =
+                createGlobalLogoutMessage("same-icsid", SESSION_ID_1, CLIENT_SESSION_ID_2);
```
Louisasa, line 79: minor but should the session ID and client session ID be mixed like this?

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandlerTest.java
```diff
@@ -17,7 +18,8 @@
 
 public class GlobalLogoutHandlerTest {
     private final Context context = mock(Context.class);
-    private final GlobalLogoutHandler globalLogoutHandler = new GlobalLogoutHandler();
+    private final GlobalLogoutHandler globalLogoutHandler =
```
Louisasa, line 21: would `mock(GlobalLogoutService.class)` not work?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
cearl1: Yeah - i must admit this was the "line of least resistance" approach :smile:

It has been bugging me how we need to instantiate the LogoutService (which involves the logout redirect stuff we dont actually use in global logout). It would be great if we had `LogoutService` and `LogoutWithRedirectService`, where the latter calls the former, but that might be a bit too much work.

I'll see what it looks like if I duplicate a bit of the code here and put it in GlobalLogoutService

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandlerTest.java
cearl1: I decided to remove this unit test and merge it into the integration test class, as it only had 1 test in it

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandler.java
```diff
@@ -58,6 +71,6 @@ private void processMessage(SQSEvent.SQSMessage message) throws Json.JsonExcepti
                 "Received request to global logout user with session {} and client session {}",
                 request.getSessionId(),
                 request.getClientSessionId());
-        // TODO: Add logic for global logout (ATO-1660)
+        globalLogoutService.logoutAllSessions(request.getInternalCommonSubjectId());
```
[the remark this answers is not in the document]
cearl1: I see what you mean yeah. I'll update this PR to add that extra validation instead of raising another one

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandler.java
```diff
@@ -58,6 +71,6 @@ private void processMessage(SQSEvent.SQSMessage message) throws Json.JsonExcepti
                 "Received request to global logout user with session {} and client session {}",
```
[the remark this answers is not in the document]
cearl1: I'll raise a PR to add this and merge it before this one

oidc-api/src/test/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandlerTest.java
cearl1, line 1: I decided to remove this unit test as it was only testing 1 thing, and merge it with the integration test. The downside is it now takes slightly longer to run, but the upside is that all the handler tests are in the same file

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/LogoutService.java
cearl1: For now I've copied the code from LogoutService into a new service. There's a bit of duplication but I've also created a tech-debt ticket in the backlog to consider ways of reducing this in the future

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandler.java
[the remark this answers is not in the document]
cearl1: So null values will be handled by `SerialisationService` (by `RequiredFieldValidator` specifically) if the field has the `@Required` annotation.

Empty strings are being handled by this PR: https://github.com/govuk-one-login/authentication-api/pull/6699

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/GlobalLogoutHandler.java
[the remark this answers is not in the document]
cearl1: This PR addresses this: https://github.com/govuk-one-login/authentication-api/pull/6694

template.yaml
```diff
@@ -2111,12 +2111,22 @@ Resources:
...
+                  !Ref Environment,
+                  authAccountId,
+                ]
+              AuthEnvironment:
+                !FindInMap [
+                  EnvironmentConfiguration,
+                  !Ref Environment,
+                  authEnvironment,
+                ]
+          REDIS_KEY: session
```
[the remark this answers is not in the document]
cearl1: Ooooh yes good point i'll remove it

PR #6679, opened 2025-06-13
ATO-1661: Add DLQ alarm for global logout

cearl1:
### What’s changed

We forgot to add this when creating the lambda. This adds an alarm for if a message arrives on the global logout test DLQ. 

### Checklist


template.yaml
```diff
@@ -2263,6 +2263,29 @@ Resources:
       FunctionName: !GetAtt GlobalLogoutFunction.Arn
       FunctionResponseTypes:
         - ReportBatchItemFailures
+
+  GlobalLogoutDLQAlarm:
```
Louisasa, line 2267: some thoughts:
- is it worth renaming to be clear this alarm is for test dlq only?
- do we need a runbook if this is only going to be for dev and build?

cearl1: So I thought we can keep the alarm named generically if we dont get provided an alarm by TxMA. That way we can conditionally set the queue like we do with the creation of the test queue.

I'm not sure what the runbook should say to be honest, but i've updated the link in the alarm and created a blank confluence page for the time being.

Louisasa: we chatted about this and TxMA will be providing and monitoring alarms so this will now just be for testing

PR #6680, opened 2025-06-13
ATO-1637 Disallow POST authorize requests

james-peacock-gds:
### Wider context of change

Support for POST authorise requests was added in 2023 as it was anticipated that this would be required by HMRC. This is no longer the case so there is no requirement to support POST authorise requests. 

### What’s changed

- update terraform for API gateway and authorize endpoint to only allow authorise requests with GET
- In AuthorisationHandler, only support authorise requests with GET
- update unit test for allowed http methods
- remove redundant unit tests for POST requests

### Manual testing

Deployed terraform changes to sandpit and observed the following:
- POST request from RP stub is not permitted (403 response)
- GET request from RP stub remains successful

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
@@ -113,7 +113,6 @@ import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachOrchSes
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachSessionIdToLogs;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.updateAttachedLogFieldToLogs;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.updateAttachedSessionIdToLogs;
-import static uk.gov.di.orchestration.shared.helpers.RequestBodyHelper.parseRequestBody;
 import static uk.gov.di.orchestration.shared.services.AuditService.MetadataPair.pair;
 
 public class AuthorisationHandler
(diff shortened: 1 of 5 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -265,21 +264,15 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
-                                    String.format(
-                                            "Authentication request sent with invalid HTTP method %s",
-                                            input.getHttpMethod()));
-                            throw new InvalidHttpMethodException(
-                                    String.format(
-                                            "Authentication request does not support %s requests",
-                                            input.getHttpMethod()));
-                        }
-                    };
+            if (!"GET".equals(input.getHttpMethod())) {
```
Louisasa, line 267: absolutely tiny pick, but I really dislike that it's static string equals variable instead of variable instead of static string lol it reads weird to me

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -265,21 +264,15 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
+                LOG.warn(
+                        String.format(
+                                "Authentication request sent with invalid HTTP method %s",
+                                input.getHttpMethod()));
+                throw new InvalidHttpMethodException(
+                        "Authentication request only supports GET requests");
```
Louisasa, line 273: is it worth rewording back to what it was so we know which request was sent in the exception? not sure where the exception is used so might not be necessary

Louisasa (commented):
not certain but I think `AuthorisationFunctionCrossAccountPostPermission` might need to be removed from template.yaml?

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
cearl1: It's useful for avoiding a null pointer when `input.getHttpMethod()` is null, and i think it was being handled by the switch's default statement. Not sure how likely it is for httpMethod to be null though 🤷

james-peacock-gds: I share your instinctive dislike for this, @Louisasa , but it seems to be the standard for defensive programming reasons including what Craig says above. TIL it's called the 'Yoda condition'!

james-peacock-gds:
> not certain but I think `AuthorisationFunctionCrossAccountPostPermission` might need to be removed from template.yaml?

thanks, this is now done.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -270,7 +270,9 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
                                 "Authentication request sent with invalid HTTP method %s",
                                 input.getHttpMethod()));
                 throw new InvalidHttpMethodException(
-                        "Authentication request only supports GET requests");
+                        String.format(
```
Louisasa, line 273: minor minor nit, can these comment changes be a part of the first commit instead of their own commit? happy to give a hand if you're unsure of the best way to do this change :) otherwise LGTM

james-peacock-gds: I was planning to squash all commits into one as this is quite a small PR, is that ok?

Louisasa (approved):

PR #6707, opened 2025-06-19
ATO-1699: Stop using hardcoded DCMAW key ID in dev and build

cearl1:
### Wider context of change

We are currently getting the encryption key for doc app requests by fetching a key with a specific ID from the JWKS URL. This key ID is hardcoded at the moment. We would like to stop using this hardcoded key in favor of getting the first available encryption key on the JWKS endpoint.

### What’s changed

This PR caches the keys on the JWKS endpoint and fetches the first available encryption key, instead of looking for a specific key by ID. 

This is currently locked behind a feature flag, only enabled in dev and build

### Manual testing

TODO

### Checklist


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
```diff
@@ -174,6 +174,10 @@ public URI getDocAppJwksURI() {
         return getURIOrEmpty("DOC_APP_JWKS_URL");
     }
 
+    public URL getDocAppJwksUrl() {
+        return getURLOrThrow("DOC_APP_JWKS_URL");
+    }
+
```
Ryan-Andrews99, line 180: What's the difference between this and the above method `getDocAppJwksURI`? Is the one above used?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/DocAppAuthorisationService.java
```diff
@@ -239,20 +241,22 @@ private EncryptedJWT encryptJWT(SignedJWT signedJWT) {
...
+    private JWK getPublicEncryptionKey() {
         try {
             LOG.info("Getting Doc App Auth Encryption Public Key via JWKS endpoint");
-            var encryptionJWK =
-                    jwksService.retrieveJwkFromURLWithKeyId(
-                            configurationService.getDocAppJwksURI().toURL(),
-                            configurationService.getDocAppEncryptionKeyID());
-            return new RSAKey.Builder((RSAKey) encryptionJWK).build().toRSAPublicKey();
+            JWK encryptionJWK;
+            if (configurationService.isUseAnyKeyFromDocAppJwks()) {
```
Ryan-Andrews99, line 248: Can we add a `//TODO: ATO-1755` to remind us to remove this feature flag once it's turned on in all envs

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
cearl1: So the method above returns a `URI` instead of a `URL`. The JWKS caching code expects a URL so you either need to call `URI.toUrl()` which can throw a `MalformedUrlException`, or we can use this helper method to catch this checked exception and rethrow it as an unchecked exception.

I'm not really sure of the best approach to this. The method above is used in a few places, but everywhere its called, it immediately calls `toURL()`, and the checked exception is handled whenever this is called

I suppose i could change `jwksService.getDocAppJwk()` to take a URL as a parameter, and have the exception handling code that's already in `DocAppAuthorisationService` handle the MalformedUrlException.

Ryan-Andrews99: Nice!

Ryan-Andrews99 (approved):
LGTM!

PR #6710, opened 2025-06-20
AUT-4452: MFAMethodType agnostic CodeStorageService CODE_BLOCKED keys

mattvot:
## What

Currently we have a lock set in redis using the `CODE_BLOCKED_KEY_PREFIX` prefix. This is then appended with a concatenation of JourneyType and MFAMethodType.

Going forward we want to make this lock MFAMethodType agnostic (so locks triggered by SMS or AUTH_APP are the same lock).

This will have the effect blocking entering an MFA OTP on the users DEFAULT MFA method if they managed to lock themselves out by entering too many incorrect OTPs using their BACKUP on the same journey and where DEFAULT and BACKUP are different MFAMethodTypes.

This is done by narrowing the possible options in the CodeRequestType enum, generalising SMS and AUTH_APP mentions to a generic MFA value (as an aside concatenation of the CodeRequestType for the PASSWORD_RESET_MFA journey is flipped to match the others).

This also has the effect of ignoring any existing lockouts for the deprecated prefixes. 

To address this we introduce a temporary measure to ensure existing locks using the deprecated prefixes are still used when determining if a lock needs to be applied.

This can be reverted when all existing locks using the deprecated prefixes have expired.

## How to review

1. Code Review
1. Deploy to dev
1. See acceptance test passes as expected

## Checklist


dbes-gds (approved):
nice 👍

PR #6730, opened 2025-06-24
ATO-1770: Set rpSectorIdentifierHost on auth session and add logging

cearl1:
### Wider context of change

We have been calculating the `rpSectorIdentifierHost` in orch and have started to send the claim from orch to auth backend. By this point auth frontend should be sending this claim to auth backend as `rp_sector_identifier_host`. 

This claim is currently being calculated in auth, but the calculation is the same as the one in orch. We want to set this new claim on the auth session and use that instead of calculating the `rpSectorIdentifierHost` a second time. This is mainly to remove usages of the auth client registry, so we can rely on orch's client registry instead.

### What’s changed

This PR sets the rpSectorIdentifierHost on the auth session, and also adds logging to check whether the calculation using the client registry is the same as the one using the pre-calculated auth session field. 

I've also added logging to check if the subject IDs calculated from both fields above are equal too. 

### Manual testing

Tested by deploying FE changes and this PR to sandpit. 
- Did a full auth journey and saw the logs confirming that both the new claim on the auth session and the field in the client registry are equal.
- Also saw the logs that confirmed the subject id calculated from both of these fields was the same
Will test in dev once FE change has been merged.

### Checklist

### Related PRs

Need this PR to be merged first: https://github.com/govuk-one-login/authentication-frontend/pull/2830

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/MfaResetAuthorizeHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/MfaResetAuthorizeHandlerTest.java
@@ -103,7 +103,8 @@ class MfaResetAuthorizeHandlerTest {
             new AuthSessionItem()
                     .withSessionId(SESSION_ID)
                     .withInternalCommonSubjectId(INTERNAL_COMMON_SUBJECT_ID)
-                    .withClientId(CLIENT_ID);
+                    .withClientId(CLIENT_ID)
+                    .withRpSectorIdentifierHost("gov.uk");
     private static MfaResetAuthorizeHandler handler;
     private static UserProfile userProfile = new UserProfile();
 
@@ -116,7 +117,7 @@ class MfaResetAuthorizeHandlerTest {
         when(userContext.getUserProfile()).thenReturn(Optional.of(userProfile));
         when(userContext.getTxmaAuditEncoded()).thenReturn(ENCODED_DEVICE_DETAILS);
 
-        when(clientRegistry.getSectorIdentifierUri()).thenReturn("htttps://gov.uk");
+        when(clientRegistry.getSectorIdentifierUri()).thenReturn("https://gov.uk");
 
         when(userContext.getClient()).thenReturn(Optional.of(clientRegistry));
         when(authSessionService.getSessionFromRequestHeaders(anyMap()))
--- a/shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
+++ b/shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
@@ -9,11 +9,8 @@ import uk.gov.di.authentication.shared.entity.AuthSessionItem;
 import uk.gov.di.authentication.shared.entity.ClientRegistry;
 import uk.gov.di.authentication.shared.entity.UserProfile;
 import uk.gov.di.authentication.shared.services.AuthenticationService;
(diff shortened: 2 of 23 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/MfaResetAuthorizeHandlerTest.java
```diff
@@ -103,7 +103,8 @@ class MfaResetAuthorizeHandlerTest {
             new AuthSessionItem()
                     .withSessionId(SESSION_ID)
                     .withInternalCommonSubjectId(INTERNAL_COMMON_SUBJECT_ID)
-                    .withClientId(CLIENT_ID);
+                    .withClientId(CLIENT_ID)
+                    .withRpSectorIdentifierHost("gov.uk");
```
Louisasa, line 107: so very minor but I noticed in that file the sector identifier uri returned is `"htttps://gov.uk"` would you mind fixing that?

cearl1: I think this is correct. In the `ClientSubjectHelper` it compares the rpSectorIdentifierHost on the auth session with the one calculated from the client registry sector identifier URI. 

In `ClientSubjectHelper.getSectorIdentifierForClient` It returns the host from the sector identifier URI.
```java
return client.getSectorIdentifierUri() != null
        ? returnHost(client.getSectorIdentifierUri())
        : returnHost(client.getRedirectUrls().stream().findFirst().orElseThrow());
```

shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
```diff
@@ -126,10 +146,18 @@ void shouldReturnSameSubjectIDForMultipleClientsWithSameSector() {
...
+                        PUBLIC.toString(),
+                        "https://" + CLIENT_SECTOR_HOST_1,
+                        false);
         ClientRegistry clientRegistry2 =
                 generateClientRegistryPairwise(
-                        keyPair, CLIENT_ID_2, PUBLIC.toString(), "https://test.com", false);
+                        keyPair,
+                        CLIENT_ID_2,
+                        PUBLIC.toString(),
+                        "https://" + CLIENT_SECTOR_HOST_1,
```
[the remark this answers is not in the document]
cearl1: Yep that will make it a lot less cluttered. I'll make those changes, thanks!

shared/src/test/java/uk/gov/di/authentication/shared/helpers/ClientSubjectHelperTest.java
```diff
@@ -54,41 +51,27 @@ void setUp() {
...
 
     @Test
     void shouldReturnDifferentSubjectIDForMultipleClientsWithDifferentSectors() {
-        ClientRegistry clientRegistry1 =
-                generateClientRegistryPairwise(
-                        keyPair, CLIENT_ID_1, PAIRWISE.toString(), "https://test.com", false);
-        ClientRegistry clientRegistry2 =
-                generateClientRegistryPairwise(
-                        keyPair, CLIENT_ID_2, PAIRWISE.toString(), "https://not-test.com", false);
+        var clientRegistry1 = createClient(CLIENT_ID_1, PAIRWISE.toString(), "test.com", false);
```
Louisasa, line 58: sorry for the confusion, I reckon keep `test.com` as a variable like `CLIENT_SECTOR_HOST` as it is used a lot and not have `not.test.com` as a variable as it's only used once? otherwise I think your refactor here is great :)

PR #6745, opened 2025-06-26
ATO-1779: add ratelimit to client registry

Louisasa:
### Wider context of change

We want to add rateLimit to the client registry to be able to add rate limiting.

### What’s changed

Rate limit has been added to the client registry. It's been made an `Integer` so it allows null as not all client registries will have rate limiting. I've made all instances as `Integer` unlike `pkceEnforced` which is only `Boolean` for some and `boolean` for the rest as I wasn't sure we'd want `0` as the default if the rate limit is null.

### Manual testing

Ran journeys on sandpit and all good

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ClientRegistry.java
@@ -47,6 +47,7 @@ public class ClientRegistry {
     private String channel;
     private boolean maxAgeEnabled = false;
     private boolean pkceEnforced = false;
+    private Integer rateLimit;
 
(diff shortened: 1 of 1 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/DynamoClientService.java
```diff
@@ -81,7 +81,8 @@ public void addClient(
             List<String> clientLoCs,
             String channel,
             boolean maxAgeEnabled,
-            boolean pkceEnforced) {
+            boolean pkceEnforced,
+            Integer rateLimit) {
```
Ryan-Andrews99, line 85: Hmm thinking about this, do we need this here if we're always gonna set the value to null as we're not planning to add this to SSE?

PR #6746, opened 2025-06-26
Aut 4372/add extra dimensions to existing metric

mattvot:
## What

Add MFAMethodType, JourneyType and PriorityIdentifier to the AuthenticationSuccess metric when an MFA method has been used. This will give us additional context to how MFAs are being used for authentication.

Note that the VerifyMfaCodeHandler is used for SMS code verification only on REGISTRATION and ACCOUNT_MANAGEMENT journeys, hence their priority values default to DEFAULT.

## How to review

1. Code Review

## Checklist


andrew-moores (approved):
I assume it makes no difference that the mfa-method in AudtiEvent extensions is lower case: default rather then DEFAULT for example?

mattvot:
> I assume it makes no difference that the mfa-method in AudtiEvent extensions is lower case: default rather then DEFAULT for example?

No, not really AFAIK. They are datasets for different purposes and it doesn't functionally matter for metric dimension values

dbes-gds (approved):
tested in authdev1

PR #6752, opened 2025-06-27
Refactoring: ApiGateway response generation

mrwilson:
- **BAU: Add new ApiResponse.ok method which wraps exception handling to generate 400 + invalid_request errors**
- **BAU: Demonstrate `ApiResponse.ok` in `TrustMarkHandler`**


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/ApiResponse.java
```diff
@@ -0,0 +1,37 @@
...
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.AccessTokenResponse;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.oauth2.sdk.OAuth2Error;
+
+import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
+
+public class ApiResponse {
+
+    public static <T> APIGatewayProxyResponseEvent ok(T supplier) {
```
cearl1, line 12: Small nit but could we rename this parameter to `payload`? When I first looked at it I thought this was a `Supplier<T>` instead of just `T`. 

From the commit history it looks like this used to be the case, so probably just a case of missing a rename :smile:

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/helpers/ApiResponse.java
```diff
@@ -0,0 +1,37 @@
...
+            } else {
+                return generateApiGatewayProxyResponse(200, supplier);
+            }
+        } catch (Exception e) {
+            return generateApiGatewayProxyResponse(
+                    400, OAuth2Error.INVALID_REQUEST.toJSONObject().toJSONString());
+        }
+    }
+
+    public static <T> APIGatewayProxyResponseEvent badRequest(T supplier) {
```
cearl1, line 25: Same as above with the supplier parameter

cearl1 (approved):
Looks good to me! 2 minor nits but nothing super important

mrwilson:
> Looks good to me! 2 minor nits but nothing super important

Yep, I'll change that in the follow up, thanks!

PR #6768, opened 2025-07-01
AUT-4381: Add mfa-method to AUTH_CODE_MAX_RETRIES_REACHED

ben-fernandes-sw:
## What

The `mfa-method` extension attribute has been added to the AUTH_CODE_MAX_RETRIES_REACHED audit event.

There are two places where we generate this event:

**VerifyCodeHandler** in the `processBlockedCodeSession` method
This work was already done. The extension is attached if a `maybeRequestedSmsMfaMethod` is present, which should be the case when this handler is invoked.

**VerifyMfaCodeHandler** in the `metadataPairsForEvent` method
We don't have the ID of the method passed in, so we can't query the method for which priority it is. Therefore, we use a workaround to determine the priority (see the bottom of VerifyMfaCodeHandler). I've extracted the `getPriorityIdentifier` method that was pre-existing on `main` to re-use it.

## How to review

1. Code Review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
@@ -62,6 +62,7 @@ import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent
 import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent.AUTH_INVALID_CODE_SENT;
 import static uk.gov.di.authentication.frontendapi.helpers.ReauthMetadataBuilder.getReauthFailureReasonFromCountTypes;
 import static uk.gov.di.authentication.shared.domain.AuditableEvent.AUDIT_EVENT_EXTENSIONS_ATTEMPT_NO_FAILED_AT;
+import static uk.gov.di.authentication.shared.domain.AuditableEvent.AUDIT_EVENT_EXTENSIONS_JOURNEY_TYPE;
 import static uk.gov.di.authentication.shared.domain.AuditableEvent.AUDIT_EVENT_EXTENSIONS_MFA_METHOD;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.ENVIRONMENT;
 import static uk.gov.di.authentication.shared.domain.CloudwatchMetricDimensions.FAILURE_REASON;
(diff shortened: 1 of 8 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -650,6 +661,18 @@ private AuditService.MetadataPair[] metadataPairsForEvent(
...
+        if (journeyType == JourneyType.REGISTRATION) {
+            return Optional.of(PriorityIdentifier.DEFAULT);
+        } else {
+            return Optional.of(
+                    mfaMethodsService.isAuthAppDefaultMfaMethod(email)
+                            ? PriorityIdentifier.DEFAULT
+                            : PriorityIdentifier.BACKUP);
+        }
+    }
+
```
ChristianSoftwire, line 675: Would it be worth leaving a comment about why this logic works as I can imagine it being hard to follow in the future. I know comments aren't ideal but I can see this being a source of confusion.

shared/src/main/java/uk/gov/di/authentication/shared/services/mfa/MFAMethodsService.java
```diff
@@ -609,6 +609,23 @@ public Optional<MfaMigrationFailureReason> migrateMfaCredentialsForUser(
...
+                                                        .getValue()
+                                                        .equals(mfaMethod.getMfaMethodType())
+                                                && PriorityIdentifier.DEFAULT
+                                                        .toString()
+                                                        .equals(mfaMethod.getPriority()))
+                        .findFirst()
+                        .orElse(null);
+        return Objects.nonNull(authMfaMethodWithDefaultPriority);
+    }
+
```
ChristianSoftwire, line 628: Might be good to add a test for this method (but I understand if we might want to move this to after release given the deadline)

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/VerifyMfaCodeHandler.java
```diff
@@ -650,6 +661,18 @@ private AuditService.MetadataPair[] metadataPairsForEvent(
...
+            String email, JourneyType journeyType) {
+        if (journeyType == JourneyType.REGISTRATION) {
+            return Optional.of(PriorityIdentifier.DEFAULT);
+        } else {
+            return Optional.of(
+                    mfaMethodsService.isAuthAppDefaultMfaMethod(email)
+                            ? PriorityIdentifier.DEFAULT
+                            : PriorityIdentifier.BACKUP);
+        }
+    }
```
ChristianSoftwire, line 674: Given the else in this method, I don't think the return type needs to be Optional anymore as a `PriorityIdentifier` will always be returned

ChristianSoftwire (commented):
LGTM just some comments to think about

PR #6775, opened 2025-07-02
ATO-1750: Read state from dynamo and compare with redis

cearl1:
### Wider context of change

We are now saving the state value to a dynamo table stored in orch, as well as storing the state value in redis. In order to swap to just using dynamo to store the state value, we need to make sure that the values are equal whenever we read the state from redis.

### What’s changed

This PR adds logging to the places where we read the state from redis, and compares the state value from redis with the state value from dynamo. Permissions were added where necessary.

### Manual testing

Tested in sandpit. Performed:
- an auth journey
- a docapp journey
- an identity journey

All journeys completed successfully. Also observed the logs for all 3 callback handlers and saw the log message `Is state from redis equal to state from dynamo? true` in all of them

### Checklist

- [x] Lambdas have correct permissions for the resources they're accessing. Added as part of the ticket


james-peacock-gds (approved):
LGTM!

PR #6782, opened 2025-07-03
ATO-1768 Refactor Client Store Extension

james-peacock-gds:
- add a builder class to make it easier to register test clients within integration tests
- update integration tests to use this builder
- tidy up some of the helper methods in integration tests, rationalising the helper methods and reducing duplicated steps


orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/extensions/ClientStoreExtension.java
```diff
@@ -400,4 +408,188 @@ private void createClientRegistryTable(String tableName) {
...
+        private String clientSecret = null;
+        private String tokenAuthMethod = ClientAuthenticationMethod.PRIVATE_KEY_JWT.getValue();
+        private String idTokenSigningAlgorithm = null;
+        private List<String> clientLoCs = emptyList();
+        private String channel = Channel.WEB.getValue();
+        private boolean maxAgeEnabled = false;
+        private boolean pkceEnforced = false;
+        private String landingPageUrl = null;
+
+        public void build() {
```
cearl1, line 135: I find it a bit strange having a `build()` method on a builder that returns nothing. Maybe we can rename this method to `registerClient()` or `saveToDynamo()`? 

I'm thinking of how it would read when seeing it in a test. Something like
```java
clientStore
    .createClient()
    .withClientId("test-client-id")
    .saveToDynamo()
```
Would make a bit more sense in my opinion. Let me know your thoughts though!

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
cearl1, line 1: I think in the future it would be helpful if we split out the tests into nested tests in their own commit - as the combined diff is a bit hard to follow.

cearl1 (commented):
A few small things. I noticed Sonar has spotted some unused fields too so would be good to get rid of them while we're in the area.

Overall though really like these changes! Reduces a lot of duplication across tests and the builder pattern is nice if you only care about specific fields, and you can default the rest of them

orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/extensions/ClientStoreExtension.java
james-peacock-gds: I agree, this is clearer - I've made these changes.

integration-tests/src/test/java/uk/gov/di/authentication/api/AuthorisationIntegrationTest.java
james-peacock-gds: I've removed the nesting changes - I'm not sure they added much value and not really in scope here anyway.

cearl1 (approved):
LGTM!

PR #6794, opened 2025-07-07
ATO-1744: write to new docapp credential table

isaac-GDS:
### Wider context of change

Migrating the doc App credential table from the auth account to the orch account

### What’s changed

Added new service and method to add to the new table
Added permissions in docAppcallback lambda

### Manual testing

deployed and ran through doc app journey in dev

### DIFF
--- a/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DynamoDocAppCriService.java
+++ b/doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DynamoDocAppCriService.java
@@ -1,0 +1,32 @@
+package uk.gov.di.authentication.app.services;
+
+import uk.gov.di.authentication.app.entity.DocAppCredential;
+import uk.gov.di.orchestration.shared.helpers.NowHelper;
+import uk.gov.di.orchestration.shared.services.BaseDynamoService;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


doc-checking-app-api/src/main/java/uk/gov/di/authentication/app/services/DynamoDocAppCriService.java
```diff
@@ -0,0 +1,38 @@
...
+import uk.gov.di.authentication.app.entity.DocAppCredential;
+import uk.gov.di.orchestration.shared.helpers.NowHelper;
+import uk.gov.di.orchestration.shared.services.BaseDynamoService;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+
+import java.time.temporal.ChronoUnit;
+import java.util.List;
+import java.util.Optional;
+
+public class DynamoDocAppCriService extends BaseDynamoService<DocAppCredential> {
```
cearl1, line 12: Minor nit, but do we need to include Dynamo in the name? We could name it like `OrchDocAppCriService` to make it distinct from the auth dynamo service (`DynamoDocAppService`)

isaac-GDS: I would say it should include `dynamo` in the name as this service is only for interacting with the dynamo table. Am open to other alternative names given that in mind.

isaac-GDS: I see - since at the moment its just for Dynamo I would say keep the service name with `dynamo` but if we extend the service we can always change the name.

PR #6800, opened 2025-07-08
ATO-1795: Add session invalidation handling ipv callback

Ryan-Andrews99:
### Wider context of change

IPV will be returning a special oAuth error to Orchestration to convey some when a user's session should be invalidated. We should respond to this error by logging a user out and sending them to a generic `You've been signed out page`

### What’s changed:
- Tweaks our validation to add a new case for this session_invalidated case
- Adds handling of this which destroys a user's session, audits the logout, and redirects them to the generic "You've been Signed Out" page
- 
### Manual testing: 
- Deployed to dev 
- Deploy this branch to ipv stub dev: https://github.com/govuk-one-login/orch-stubs/pull/322
- Run through an identity journey
- When you reach the stub, configure it to return the `session_invalidated` oAuth error
- See you are redirected to a generic "You've been signed out" page 
- Run through another authorize journey to check that you reach the sign in page with no silent login

Also did:
- Run through another ID journey without an oAuth error 
- All is as expected 

### Checklist

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -293,7 +293,7 @@ public class IPVCallbackHandler
 
                 return ipvCallbackHelper.generateAuthenticationErrorResponse(
                         authRequest,
-                        new ErrorObject(ACCESS_DENIED_CODE, errorObject.get().getDescription()),
+                        new ErrorObject(ACCESS_DENIED_CODE, errorObject.get().errorDescription()),
                         false,
                         clientSessionId,
                         sessionId);
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -274,6 +274,7 @@ public class IPVCallbackHandler
                             persistentId);
 
             if (errorObject.isPresent()) {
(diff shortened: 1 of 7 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -287,16 +290,25 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                            input,
+                            clientId);
+                }
+
                 return ipvCallbackHelper.generateAuthenticationErrorResponse(
                         authRequest,
-                        new ErrorObject(ACCESS_DENIED_CODE, errorObject.get().getDescription()),
+                        new ErrorObject(
+                                ACCESS_DENIED_CODE, validationResult.getFailure().description()),
                         false,
```
Ryan-Andrews99, line 300: Note here that we were previously returning an error object from this validation function, but **only using the error_description** here to populate what error we redirect back to the RP with.  The error_code is hardcoded to access_denied. 

Outside of the refactoring - we should probably consider if the error description back to the RP makes sense here. For example the below:

```
if (headers == null || headers.isEmpty()) {
            LOG.warn("No Query parameters in IPV Authorisation response");
            return Optional.of(
                    new ErrorObject(
                            OAuth2Error.INVALID_REQUEST_CODE, "No query parameters present"));
        }
```

Is what we returned if out IPV callback gets no query params - we currently propagate this error description (`"No query parameters present"`) back in our error response - which seems odd?

isaac-GDS:
Think we should update the contract tests with IPV as we expect a specific error: `session_invalidated` . Happy for this to be in another PR/ticket but would be good to track.

PR #6808, opened 2025-07-09
ATO 1745: Swap reads to new doc app cred table

isaac-GDS:
### Wider context of change

Migrating the doc App credential table from Auth to Orch

### What’s changed

Swapping the read in the userInfo service over to the new doc app table.

### Manual testing

Testing in dev
Checked the writing to the new table,


james-peacock-gds (approved):
LGTM

PR #6894, opened 2025-07-25
ATO-1857: Apply CSID mismatch checks in all cases where possible

Ryan-Andrews99:
### Wider context of change: 

We have uncovered that our handling of concurrent client sessions is currently not the best. This PR aims to resolve the _"happy path"_ case in which we don't get a cross browser error from IPV, but we *do* have a different client session  id in the cookie from the one that initiated the IPV authorisation. We can correlate this using the state parameter.

### What’s changed:
- Tweaks the previously added method in `NoSessionOrchestrationService` to ensure we apply this check in all cases we are able to - ie the state value is present - which is the minimum amount of information we need to handle this. 
- This change will me almost all requests to this callback endpoint will go through this validation as opposed to just cross browser ones which we added previously.
- This also has the effect of now redirecting a user to ipv-session-expiry page if there is no state param present in the callback response from IPV - as we need the state to be able to determine _if_ they are in the mismatch case
- If we detect a mismatch case - we return the cross browser `access_denied` response
- This uses the feature flag from the previous PR and is still only enabled in **dev**, **build** and **staging**

Again we should consider some refactoring of our validation here

### Manual testing:
Slightly convoluted - the RP stub we use for testing is not best setup to handle these multi client journeys but I've got around this by running one copy of the stub locally, configured to point at dev and the other using the deployed stub:

- Get two copies of the stub - this required some fiddling of the config for the stub to add a new localhost redirect uri in dev
- Open a browser window with each copy of the stub in separate tabs
- In one tab start an identity journey - get to IPV stub page
- In the other tab initiate another journey - I chose an auth-only - see this through to at least past /authorize to get new cookies
- Now on tab 1 - press continue to send a _successful_ IPV response to our ipv callback endpoint
- See you hit IPV callback and are redirected,  and validate a couple of things:
    - You are on the correct domain for the original journey (either localhost or deployed domain)
    - You see the following error and error_description:
    - **Error: access_denied Error description: Access denied for security reasons, a new authentication request may be successful**

Also ran through a normal identity journey and all was as expected.
### Checklist

### Related PRs:
- Previous PR which added the bulk of this work and the feature flag: https://github.com/govuk-one-login/authentication-api/pull/6889

### DIFF
--- a/orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/NoSessionOrchestrationServiceTest.java
+++ b/orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/NoSessionOrchestrationServiceTest.java
@@ -1,5 +1,6 @@
 package uk.gov.di.orchestration.shared.services;
 
+import com.nimbusds.oauth2.sdk.AuthorizationCode;
 import com.nimbusds.oauth2.sdk.OAuth2Error;
 import com.nimbusds.oauth2.sdk.ParseException;
 import com.nimbusds.oauth2.sdk.ResponseType;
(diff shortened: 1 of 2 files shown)


orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/NoSessionOrchestrationServiceTest.java
```diff
@@ -296,16 +332,15 @@ void itShouldReturnEmptyIfNotAccessDeniedError() throws NoSessionException {
         }
 
         @Test
-        void itShouldReturnEmptyIfAccessDeniedErrorWithNoState() throws NoSessionException {
+        void itShouldReturnEmptyIfNotAccessDeniedError() throws NoSessionException {
```
CarlyG55, line 335: Not really sure what value we get out of this test - given we're not checking the error from IPV here anymore

Ryan-Andrews99: Ooh good point - will rewrite this one!

Ryan-Andrews99: Ended up removing it because I think we're covering all the conditions already

CarlyG55 (approved):

PR #6901, opened 2025-07-28
ATO-1868: Refactor and rename orch KeyPairHelper

cearl1:
### Wider context of change

This helper class had a few problems: 
- It has a public static method for generating key pairs, which is named like a static constant using  `UPPER_SNAKE_CASE` instead of `camelCase`, which is inconsistent with other public static methods 
- The method for generating key pairs caches the first key it generates, then reuses this key if multiple calls are made to this method. This functionality wasnt actually used anywhere, and it means that for some tests like `AuthorisationIntegrationTest`, its using the same key for the RP signing key and the internal auth encryption key - which is unrealistic. It’s also hard to tell if the right key is being used!
- There are a lot of cases where an identical method to generate the key pair is defined locally in tests.

### What’s changed

This PR removes the key caching functionality of this helper class, so that every call to generateRsaKeyPair() is a unique key pair. As a result of this, the helper class was stateless, so I've also renamed it to `KeyPairUtils` and moved it into the `utils` package.  

Finally, i've searched for duplicate method definitions of `generateRsaKeyPair` in the orch test code and replaced them with using the `KeyPairUtils` static method.

### Checklist

### DIFF
--- a/ipv-api/src/test/java/uk/gov/di/authentication/ipv/services/IPVAuthorisationServiceTest.java
+++ b/ipv-api/src/test/java/uk/gov/di/authentication/ipv/services/IPVAuthorisationServiceTest.java
@@ -48,9 +48,6 @@ import uk.gov.di.orchestration.sharedtest.helper.TestClockHelper;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
-import java.security.KeyPair;
-import java.security.KeyPairGenerator;
-import java.security.NoSuchAlgorithmException;
 import java.security.PrivateKey;
 import java.security.interfaces.RSAPublicKey;
 import java.text.ParseException;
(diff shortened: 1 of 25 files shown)


ipv-api/src/test/java/uk/gov/di/authentication/ipv/services/IPVAuthorisationServiceTest.java
```diff
@@ -466,15 +464,4 @@ private SignedJWT decryptJWT(EncryptedJWT encryptedJWT) throws JOSEException {
...
-    private KeyPair generateRsaKeyPair() {
-        KeyPairGenerator kpg;
-        try {
-            kpg = KeyPairGenerator.getInstance("RSA");
-        } catch (NoSuchAlgorithmException e) {
-            throw new RuntimeException(e);
-        }
-        kpg.initialize(2048);
-        return kpg.generateKeyPair();
-    }
```
Ryan-Andrews99, line 479: Nice!

Ryan-Andrews99 (approved):
Looks good to me!

PR #6905, opened 2025-07-28
ATO-1777: Setup PagerDuty alarms in orch

cearl1:
### Wider context of change

We currently have pagerduty alarms configured in auth, but not in orch.  We should move the pagerduty alarms that are orch-related into orch too.

### What’s changed

This PR creates 2 new cloudwatch alarms:
- One will alarm if more than 20 errors appear in the IPV callback handler and SPOT response handler over 10 mins. 
- One will alarm if more than 20 errors appear in the DocApp callback handler over 10 mins.

Both these alarms are configured to publish the alarm event on a new SNS topic: `${Environment}-pagerduty-alerts `. There is also a subscription to this topic configured in the template file, which uses the PagerDuty URL defined in a secret. 

These alarms will only be created if the environment is production and there is a `pagerDutySecretId` defined in the environment configuration for production. This secret should have a key of `integrationURL` and a value of the PagerDuty integration URL

These alarms will also send a slack notification to the channel configured in the `pagerdury-slack-channel-id` SSM parameter. This is already set in production.

### Manual testing

Have tested this in dev by observing alerts appearing in the PagerDuty dashboard. Once the alarms were triggered, the alert appeared in the dashboard.

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -6431,6 +6431,7 @@ Resources:
               Resource:
                 - !Sub "arn:aws:ssm:eu-west-2:${AWS::AccountId}:parameter/${Environment}-slack-channel-id"
                 - !Sub "arn:aws:ssm:eu-west-2:${AWS::AccountId}:parameter/${Environment}-slack-hook-url"
+                - !Sub "arn:aws:ssm:eu-west-2:${AWS::AccountId}:parameter/pagerduty-slack-channel-id"
       Events:
         SNS:
           Type: SNS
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -7514,9 +7516,11 @@ Resources:
       ActionsEnabled: true
       AlarmActions:
         - !Ref PagerDutyEvents
+        - !Ref OrchProdSlackEvents
```
Louisasa, line 7519: just to check, are we saying we want a pager alarm, pager duty slack message, and orch prod slack message? Do we want a message to 2nd still?

cearl1: Ahh yep, i think this will work but it's a bit confusing as the logic isn't quite right.

We'll send a message to pagerduty to trigger an alarm there, and we will send a slack message to the channel defined in `pagerduty-slack-channel-id` (2nd line). 

I've updated the slackNotifications.js file to check if the alarm description contains `PagerDuty`, and if it does, and it's in production, then it will use the channel defined above. At the moment, this logic will overwrite the prod alerts logic (which will work, but i might change it to make it clearer!)

Louisasa: sorry the slack notifications javascript throws me every time

Louisasa (dismissed):

Louisasa (dismissed):

PR #6953, opened 2025-08-07
BAU: Log current quota consumption

mrwilson:
This logs out the current consumption of the limit for RPs with rate limiting enabled

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/SlidingWindowAlgorithm.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/SlidingWindowAlgorithm.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.oidc.entity;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.message.ObjectMessage;
 import uk.gov.di.authentication.oidc.services.ClientRateLimitDataService;
 import uk.gov.di.orchestration.shared.helpers.NowHelper;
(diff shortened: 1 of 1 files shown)


CarlyG55 (dismissed):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/entity/SlidingWindowAlgorithm.java
```diff
@@ -70,6 +70,12 @@ public boolean hasRateLimitExceeded(ClientRateLimitConfig rateLimitConfig) {
         var previousCountInWindow =
                 previousCount
                         * ((double) secondsFromPreviousPeriodInWindow / PERIOD_LENGTH_IN_SECONDS);
+
+        if (rateLimit > 0) {
+            var consumption = (previousCountInWindow + currentCount) / rateLimit;
+            LOG.info("Client = {} :: Consumption = {}", clientId, consumption);
```
Joe-Edwards-GDS, line 76: Is there any structured logging in Orch at the moment? Only because it's much easier to write queries when the parameters are structured!

mrwilson: True! We could reasonably use `ObjectMessage` here wrapping a Map. Let me tweak.

mrwilson: The next change going in will make it `"message":{"client":"test-client-id","consumption":0.9}` in the log json

PR #6991, opened 2025-08-15
AUT-4569: Emit notification metrics

dbes-gds:
## What

Adds two new methods which collectively emit four new metrics when either emails or sms are sent via Notify, and also when client errors happen when sending.

emitMetricForNotification

- SmsNotificationSent
- EmailNotificationSent

emitMetricForNotificationError

- SmsNotificationError
- EmailNotificationError

At the moment Notify metrics are only created on the back of delivery receipts received from Notify (and the networks).  As there is no guarantee that we’ll get a delivery receipt the metrics are not a true reflection of the number of messages actually sent.

## How to review

- [x] Deployment of this PR will not break active user journeys

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
@@ -14,6 +14,7 @@ import uk.gov.di.accountmanagement.entity.NotificationType;
 import uk.gov.di.accountmanagement.entity.NotifyRequest;
 import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.serialization.Json.JsonException;
+import uk.gov.di.authentication.shared.services.CloudwatchMetricsService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 import uk.gov.di.authentication.shared.services.NotificationService;
 import uk.gov.di.authentication.shared.services.SerializationService;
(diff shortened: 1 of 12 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
```diff
@@ -238,4 +251,63 @@ public DimensionSet getDimensions(Map<String, String> dimensions) {
...
+            String destination,
+            Boolean isTestDestination,
+            Application application) {
+        String metricName;
+        var dimensions =
+                getNotificationBaseMetricDimensions(
+                        notificationType, isTestDestination, application);
+
+        if (notificationType.isForPhoneNumber()) {
+            dimensions.put(COUNTRY.getValue(), maybeGetCountry(destination).orElse("INVALID"));
```
Joe-Edwards-GDS, line 291: Have you checked how this shows up in CloudWatch? I have a slight concern is that this is quite a lot of dimensions and it's slightly fiddly to aggregate them, but it's useful to have all the data broken down. 🙂

All the dimensions are bounded, so not critical, so long as we can easily get a query which answers: 'how many [international/domestic] SMSs am I sending, and what % are successful'.

Joe-Edwards-GDS (approved):
Looks sensible to me, thanks. A couple of things to consider but nothing necessary to change.

shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
dbes-gds: will take a look, resolving to merge

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
```diff
@@ -226,8 +232,19 @@ private void sendEmailNotification(
...
+                                ONE_LOGIN_HOME);
                     } catch (NotificationClientException e) {
                         LOG.error(ERROR_SENDING_WITH_NOTIFY, e.getMessage());
+                        cloudwatchMetricsService.emitMetricForNotificationError(
+                                notifyRequest.getNotificationType(),
+                                destination,
+                                false,
+                                ONE_LOGIN_HOME,
+                                e);
                     } catch (RuntimeException e) {
```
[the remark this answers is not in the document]
dbes-gds: will take a look, resolving to merge

PR #7008, opened 2025-08-18
AUT-4539: Use kms key alias in managed policies

monhaque:
## What

Use kms alias in managed policies, instead of directly referencing a key id


monhaque:
Tested in dev account and not working. Closing this PR.

PR #7045, opened 2025-08-26
AUT-4197: Set email check retry policy

alhcomer:
## What

- Sets the retry policy to what is outlined in AUT-4197
- Sets permissions to allow the email check account to send SQS messages to the pending email check DLQ. These permissions are the allowance of ingress of SQS messages from the email check account to the pending email check DLQ, and the allowance of the email check account to use the pending_email_check_queue_encryption_key to encrypt those messages.

## How to review

1. Code Review
2. Deploy to dev with the associated PR. Send a message to the lambda in the AWS test area with the email 'stubemailassessmentfailure@example.com'. See that this SQS message appears in the pending email check DLQ.

## Related PRs

https://github.com/govuk-one-login/authentication-check-experian/pull/514

### DIFF
--- a/ci/terraform/shared/sqs-policy.tf
+++ b/ci/terraform/shared/sqs-policy.tf
@@ -40,3 +40,28 @@ resource "aws_sqs_queue_policy" "pending_email_check_queue_subscription" {
 
   policy = data.aws_iam_policy_document.pending_email_check_queue_subscription_policy_document.json
 }
+
+data "aws_iam_policy_document" "pending_email_check_dlq_policy_document" {
+  statement {
+    effect = "Allow"
+    sid    = "AllowCrossAccountSendMessageToDLQ"
+
+    principals {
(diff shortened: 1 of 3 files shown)


ci/terraform/shared/sqs-policy.tf
```diff
@@ -40,3 +40,28 @@ resource "aws_sqs_queue_policy" "pending_email_check_queue_subscription" {
...
+    effect = "Allow"
+    sid    = "AllowCrossAccountSendMessageToDLQ"
+
+    principals {
+      type        = "AWS"
+      identifiers = ["arn:aws:iam::${var.auth_check_account_id}:root"]
+    }
+
+    actions = [
+      "sqs:SendMessage"
```
mattvot, line 55: We may need to allow encryption too to write to the queue (and maybe similar on the experian repo side too

PR #7048, opened 2025-08-27
AuUT-4633 Add daily sms limit reached alarms

andrew-moores:
## What

When the international SMS quota is exceeded we start to get 429 responses from Notify.

This change adds two alarms for exceeding the international quota and the domestic quota.  In production the alarms with trigger PagerDuty.

This is part one of two changes.  The next change will be an early warning alarm when the number of messages being sent looks like the daily quota will be exceeded.

## How to review

1. Deploy to a dev environment.
2. Check the metrics and alarms are created.
3. Use the aws cli to emit updates to the metric and check the alarm is activated.
4. Optionally create a subscription on the SNS topic to confirm the alarm is emitting a notification.

I have already tested this in authdev1 and confirmed the alarms are activated and received notifications from SNS.

## Checklist

- [x] Deployment of this PR will not break active user journeys

### DIFF
--- a/ci/terraform/oidc/alerts.tf
+++ b/ci/terraform/oidc/alerts.tf
@@ -68,6 +68,46 @@ resource "aws_cloudwatch_metric_alarm" "spot_request_sqs_cloudwatch_p1_alarm" {
   alarm_actions     = [var.environment == "production" ? data.aws_sns_topic.pagerduty_p1_alerts[0].arn : local.slack_event_sns_topic_arn]
 }
 
+resource "aws_cloudwatch_metric_alarm" "domestic_sms_limit_exceeded_alarm" {
+  alarm_name          = "${var.environment}-${var.environment == "production" ? "P1-" : ""}domestic-sms-limit-exceeded-alarm"
+  comparison_operator = "GreaterThanOrEqualToThreshold"
+  evaluation_periods  = "1"
+  metric_name         = "SmsLimitExceeded"
+  namespace           = "Authentication"
+  period              = "300"
+  statistic           = "Sum"
+  threshold           = "2"
+  treat_missing_data  = "notBreaching"
+
+  dimensions = {
+    Environment        = var.environment
+    SmsDestinationType = "DOMESTIC"
+  }
+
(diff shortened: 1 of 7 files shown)


mattvot (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/NotificationHandler.java
```diff
@@ -194,12 +194,27 @@ private void sendNotifyMessage(NotifyRequest request) {
                     "Error sending with Notify using NotificationType: {}, reference: {}",
                     request.getNotificationType(),
                     reference);
+
+            if (e.getHttpResult() == 429 && isPhoneNotification(request.getNotificationType())) {
```
dbes-gds, line 198: Is there a way to differentiate between the daily and the per minute limits?  The per minute limit could trigger a 429 on either local or int numbers

ci/terraform/oidc/alerts.tf
```diff
@@ -68,6 +68,46 @@ resource "aws_cloudwatch_metric_alarm" "spot_request_sqs_cloudwatch_p1_alarm" {
...
+  statistic           = "Sum"
+  threshold           = "2"
+  treat_missing_data  = "notBreaching"
+
+  dimensions = {
+    Environment        = var.environment
+    SmsDestinationType = "INTERNATIONAL"
+  }
+
+  alarm_description = "${var.environment == "production" ? "CRITICAL: " : ""}International SMS daily limit exceeded (${var.environment}). 2+ 429 responses from Notify. ACCOUNT: ${local.aws_account_alias}"
```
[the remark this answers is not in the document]
andrew-moores: I'll add this in next change.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/NotificationHandler.java
andrew-moores: As per slack thread we don't think this will be an issue so will push to prod and keep under observation.

PR #7049, opened 2025-08-27
ATO-1941: Send slack alerts to orch prod alerts channel

cearl1:
### Wider context of change

We would like to make the orch-nonprod-alerts slack channel less noisy. We already have a prod-alerts slack channel for build notifications. 

### What’s changed

This PR creates a new version of the `SlackNotificationsFunction` specifically for production alerts. This involves creating a new encrypted SNS topic. All the new Prod specific infrastructure is only deployed in the prod environment. This new version (ProdSlackNotificationsFunction) is only used in the back channel logout DLQ alarm at the moment.

It's best to review this commit-by-commit, as the penultimate commit applies our JS linting rules to the whole `slackNotifications.js` file and makes the diff look awful :smile:

I've also added the new SSM params needed for the prod alarms to work

### Manual testing

Tested in dev by temporarily adding the dev environment to the condition for the new infrastructure, and manually triggering the back channel logout DLQ alarm. The slack message got sent to the prod alerts channel as expected.

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -4853,7 +4853,7 @@ Resources:
       AlarmActions:
         - !If
           - IsProduction
-          - !Ref ProdSlackEvents
+          - !Ref OrchProdSlackEvents
           - !Ref SlackEvents
       Namespace: AWS/SQS
       MetricName: ApproximateNumberOfMessagesVisible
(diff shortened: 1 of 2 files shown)


template.yaml
```diff
@@ -6195,11 +6260,47 @@ Resources:
...
+      Policies:
+        - Statement:
+            - Sid: AllowGetParameters
+              Effect: "Allow"
+              Action:
+                - "ssm:GetParameter"
+                - "ssm:GetParameters"
+              Resource:
+                - !Sub "arn:aws:ssm:eu-west-2:${AWS::AccountId}:parameter/${Environment}-slack-channel-id"
+                - !Sub "arn:aws:ssm:eu-west-2:${AWS::AccountId}:parameter/${Environment}-slack-hook-url"
```
isaac-GDS, line 6275: Know we do this for non prod as well but curious on why these values are stored in parameter store and not values set in the template.

template.yaml
```diff
@@ -6153,6 +6153,52 @@ Resources:
               StringEquals:
                 AWS:SourceAccount: !Ref AWS::AccountId
 
+  ProdSlackEvents:
```
Louisasa, line 6159: in an effort to not be confusing for the future, do you think this is worth renaming something to make it clear it's not going to 2nd line? maybe OrchProdSlackX?

isaac-GDS (dismissed):

template.yaml
cearl1: Yeah, although that may look a bit different to the existing slack infra,  we could rename the existing slack infra to `OrchNonProdSlackX` too?

Louisasa (approved):

PR #7058, opened 2025-08-28
ATO-1942: Adjust alarm sensitivity to reduce noise

james-peacock-gds:
Context: error rate alarms are currently firing when errors exceed 10% of invocations in a 60 second period. This means that for low-volume services, you only need a couple of errors to cause the alarm, which means lots of false positives.

- adjust error rate alarms so that as well as exceeding the allowed rate of 10% of invocations in a 60 second period, there must be more than 3 errors in that period.

Tested in dev environment using `aws cloudwatch put-metric-data` to check that alarms are triggered under certain conditions. Confirmed that following the change, 4 or more errors in 10 invocations would trigger the alarm, as would 11 in 100, but 3 in 10 would not.

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -35,6 +35,10 @@ Parameters:
   LambdaDeploymentPreference:
     Type: String
     Description: Specifies the configuration to enable gradual Lambda deployments
+  MinErrorThreshold:
+    Type: Number
+    Description: Minimum number of errors before error rate alarms can be triggered
+    Default: 4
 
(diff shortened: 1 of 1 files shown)


james-peacock-gds:
Following changes to move the error threshold into a variable, I've retested in dev to confirm that following the change, 4 or more errors in 10 invocations will trigger the alarm, as will 11 in 100, but 3 in 10 will not.

PR #7093, opened 2025-09-03
ATO-1941: Send slack alerts to orch prod channel

cearl1:
Reopening PR as prod deployment failed because of clashing topic names. I've changed the existing slack events topic to only get created when its a non-prod environment. This means only 1 topic will be created in the prod env

### Wider context of change

We would like to make the orch-nonprod-alerts slack channel less noisy. We already have a prod-alerts slack channel for build notifications. 

### What’s changed

This PR creates a new version of the `SlackNotificationsFunction` specifically for production alerts. This involves creating a new encrypted SNS topic. All the new Prod specific infrastructure is only deployed in the prod environment. This new version (`OrchProdSlackNotificationsFunction`) is only used in the back channel logout DLQ alarm at the moment.

It's best to review this commit-by-commit, as one of the commits applies our JS linting rules to the whole `slackNotifications.js` file and makes the diff look awful :smile:

I've also added the new SSM params needed for the prod alarms to work

### Manual testing

Tested in dev by temporarily adding the dev environment to the condition for the new infrastructure, and manually triggering the back channel logout DLQ alarm. The slack message got sent to the prod alerts channel as expected.

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -68,7 +68,7 @@ Conditions:
     ]
   IsDevOrBuild:
     !Or [!Equals [dev, !Ref Environment], !Equals [build, !Ref Environment]]
-  IsNotProduction: !Not [!Equals [!Ref Environment, production]]
+  IsNotProduction: !Not [!Condition IsProduction]
   IsNotIntegration: !Not [!Equals [!Ref Environment, integration]]
   IsNotDevOrProd:
     !Not [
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -6093,7 +6096,7 @@ Resources:
     Type: AWS::SNS::Topic
     Properties:
       DisplayName: !Sub "Slack Alerts for ${Environment}"
-      TopicName: !Sub ${Environment}-slack-events
+      TopicName: !Sub nonprod-slack-events
```
Louisasa, line 6099: would it make sense to change the prod slack events to orch-prod as that's been the name elsewhere? I'm not sure if we need to be see which env the slack events are but it might be good to keep them separated?
as another note, I'm perturbed that this gets created for prod and not used, is it worth putting under a conditional or would that break things?

cearl1: Ok I've got the deployment working by only creating the slack events topic when its not a production environment. This should fix the prod deployment issues as only 1 production slack events topic will be created

Louisasa (approved):

PR #7242, opened 2025-10-02
WIP: Cucumber API tests

Joe-Edwards-GDS:
## What

Spike to show what API tests running in GHA could look like


.github/workflows/pre-merge-checks.yml
```diff
@@ -443,3 +443,96 @@ jobs:
...
+
+  run-api-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    defaults:
+      run:
+        working-directory: ./api-tests
+    needs:
+      - build
+    services:
```
Joe-Edwards-GDS, line 455: Copied from integration tests above - could also use the docker-compose, but this _might_ be faster (as it uses native GHA containers)

api-tests/cucumber.js
Joe-Edwards-GDS, line 1: Needs tidying up - copied from IPV

PR #7248, opened 2025-10-03
ATO-1995: update rate limit from ManualUpdateClientRegistryHandler

Louisasa:
### Wider context of change

For TSD to be able to update the client rate limit, we need a new lambda to manually update the client registry.

### What’s changed

Lambda logic and validation added.

### Manual testing

Tested on authdev3
- no `rate_limit` or `rate_limit = null` does not break and does not turn off rate limit
- only whole numbers are accepted
- `rate_limit = "''"` will turn off rate limit

### Checklist

### DIFF
--- a/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ManualUpdateClientRegistryHandler.java
+++ b/client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ManualUpdateClientRegistryHandler.java
@@ -2,13 +2,118 @@ package uk.gov.di.authentication.clientregistry.lambda;
 
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.clientregistry.services.ManualUpdateClientRegistryValidationService;
+import uk.gov.di.orchestration.shared.entity.ManualUpdateClientRegistryRequest;
+import uk.gov.di.orchestration.shared.serialization.Json;
+import uk.gov.di.orchestration.shared.services.AuditService;
+import uk.gov.di.orchestration.shared.services.ClientService;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+import uk.gov.di.orchestration.shared.services.DynamoClientService;
+import uk.gov.di.orchestration.shared.services.SerializationService;
 
-public class ManualUpdateClientRegistryHandler implements RequestHandler<String, Void> {
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
 
-    public ManualUpdateClientRegistryHandler() {}
+import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachTraceId;
+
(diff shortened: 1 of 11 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientService.java
```diff
@@ -42,7 +43,10 @@ void addClient(
 
     ClientID generateClientID();
 
-    ClientRegistry updateClient(String clientId, UpdateClientConfigRequest updateRequest);
+    ClientRegistry updateClientFromSSE(String clientId, UpdateClientConfigRequest updateRequest);
```
Louisasa, line 46: not convinced by this naming or the AWS naming, but I wanted to make it clear why there were two, especially as this is only used in SSE.

Maybe the AWS one should be renamed `manualUpdateClient`? idk thoughts?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/entity/ManualUpdateClientRegistryRequest.java
Ryan-Andrews99, line 1: Minor one but it would be great if this was a record instead of a class, doesn't make a practical difference but does cut down massively on the boilerplate needed. An example of this is here: https://github.com/govuk-one-login/authentication-api/commit/543e42ae3725ccfd61deef8532542bfadc14b090

client-registry-api/src/main/java/uk/gov/di/authentication/clientregistry/lambda/ManualUpdateClientRegistryHandler.java
```diff
@@ -2,13 +2,114 @@
...
+                                "Failed validation. ErrorCode: %s. ErrorDescription: %s",
+                                errorResponse.get().getCode(),
+                                errorResponse.get().getDescription()));
+            }
+
+            clientService.updateClientFromAWS(clientId, manualUpdateClientRegistryRequest);
+
+            LOG.info("Client updated");
+            response.put(RESULT_KEY, RESULT_SUCCESS);
+            response.put(MESSAGE_KEY, "Successfully updated client");
```
Ryan-Andrews99, line 99: Thinking it might be good to echo back some of the input here - so maybe something like:
`Successfully updated client: "<SOME_CLIENT_ID>" with rate limit "<RATE_LIMIT>"`  - what do you think?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientService.java
Ryan-Andrews99: Hmm I see your point - I think it's good to name these methods separately.  I like the name `manualUpdateClient` for the one we use in this lambda. Don't love the name `updateClientFromSSE` but think that's because I don't like the name `SSE` 😆

Ryan-Andrews99 (commented):
Left a few comments -  nothing major just some thoughts and one place we can cut down on boilerplate - otherwise LGTM!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientService.java
Louisasa: I'll give it a little more thought, but I think SSE has gotta be in the name for it to be clear :(

PR #7264, opened 2025-10-07
ATO-2024 Multi-session recovery

Joe-Edwards-GDS:
### Wider context of change

https://govukverify.atlassian.net/browse/ATO-2024

The V2 App introduces a new scenario where the app login session clashes with the original client session. In some cases (where the browser is the same) we may still be able to recover.

### What’s changed

If

- The client session in the state does not match the client session in the cookie
- AND the client session in the state is an active client session
- AND the client session in the state is linked to the active orch session

Then we continue processing and redirect to the RP using that client session.

### Manual testing

TBD

### Checklist

- [ ] Successfully deployed to authdev

### Related PRs

https://github.com/govuk-one-login/authentication-api/pull/6889

### DIFF
--- a/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
+++ b/ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
@@ -216,30 +216,6 @@ public class IPVCallbackHandler
                     crossBrowserOrchestrationService.generateEntityForMismatchInClientSessionId(
                             input.getQueryStringParameters(), clientSessionIdFromCookie);
 
-            if (mismatchedEntity.isPresent()) {
-                if (orchSession
-                        .getClientSessions()
-                        .contains(mismatchedEntity.get().getClientSessionId())) {
-                    LOG.info("Recovering client session from state");
-                } else {
-                    var authRequestFromStateDerivedRP =
-                            AuthenticationRequest.parse(
-                                    mismatchedEntity
-                                            .get()
-                                            .getClientSession()
-                                            .getAuthRequestParams());
-                    attachLogFieldToLogs(
-                            CLIENT_ID, authRequestFromStateDerivedRP.getClientID().getValue());
-
-                    return ipvCallbackHelper.generateAuthenticationErrorResponse(
-                            authRequestFromStateDerivedRP,
-                            mismatchedEntity.get().getErrorObject(),
-                            false,
-                            mismatchedEntity.get().getClientSessionId(),
-                            AuditService.UNKNOWN);
-                }
-            }
-
             var clientSessionId =
                     mismatchedEntity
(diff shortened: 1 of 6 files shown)


ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -198,37 +199,54 @@ public APIGatewayProxyResponseEvent handleRequest(
...
-                return ipvCallbackHelper.generateAuthenticationErrorResponse(
-                        authRequestFromStateDerivedRP,
-                        mismatchedEntity.get().getErrorObject(),
-                        false,
-                        mismatchedEntity.get().getClientSessionId(),
-                        AuditService.UNKNOWN);
+                if (orchSession
+                        .getClientSessions()
+                        .contains(mismatchedEntity.get().getClientSessionId())) {
+                    LOG.info("Recovering client session from state");
```
Joe-Edwards-GDS, line 221: This might also be where we need to update the `gs` cookie and possibly update some other state to make the SPOT spinner work.

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
```diff
@@ -492,8 +500,23 @@ public APIGatewayProxyResponseEvent handleRequest(
                 return RedirectService.redirectToFrontendErrorPageWithErrorLog(
                         frontend.errorURI(), new Error("Failed to create redirectURI"));
             }
-            return generateApiGatewayProxyResponse(
-                    302, "", Map.of(ResponseHeaders.LOCATION, redirectURI.toString()), null);
+
+            // If we've recovered from a mismatched client session then we need to update the
+            // session cookie
```
Joe-Edwards-GDS, line 505: I'm not sure if we need to do this on any of the error scenarios as well - I'm hoping it's ok, because they tend to be dead end pages anyway 🤞

ipv-api/src/main/java/uk/gov/di/authentication/ipv/lambda/IPVCallbackHandler.java
Joe-Edwards-GDS, line 1: A reflection here is that we could simplify this whole process a lot if we:
- Removed `clientSessionId` from the session cookie
- Validated directly that the `clientSessionId` in the `state` is present on the (orch) session (and doing cross-browser recovery logic if it isn't)

Then the happy/mismatch paths all look the same: use the client session from the state, verify it's correctly bound to the session, and return an error if it isn't.

I'm tempted to do the second point now, but it's tangled up in other refactoring and I don't want to confuse things...

pskushwaha1:
Orch code are no longer maintained in  this repo code are moved to https://github.com/govuk-one-login/orchestration-api

PR #7265, opened 2025-10-07
Ato 2010/ensure scopes dont have commas

cearl1:
### Wider context of change

We would like to upgrade nimbus, but in order to do so, we need to remove support scopes which are a comma-separated list. To check if any RPs are currently sending us comma-separated scopes, we should add some log lines. 
### What’s changed

This PR adds logging in AuthorisationHandler if the scope parameter contains commas. It will also log the client ID if this is the case.

I've also removed any uses of comma-separated scopes in tests.

### Manual testing

Will deploy to dev and test

### Checklist

- [ ]


james-peacock-gds (approved):
LGTM!

PR #7783, opened 2026-02-04
ATO-2292: force empty client locs when sse registering

Ryan-Andrews99:
### Wider context of change:

We previously removed setting the client LOCs value in SSE, because it was doing  it in a slightly wrong way and ran the risk of overwriting values which we manually configured for clients. So we decided to remove this and allow or runtime defaulting using the `IsIdentityVerificationEnabled` flag to determine the permitted Levels of Confidence, which would also allow us to persist manually updated values. 

### What’s changed:
- Uses the DynamoEnhancedClientExtension's beforeWrite hook to override the value of ClientLoCs on creation and updating the DB. This is **only** applied in the SSE create and update handlers.
- Includes an exclusion for clients using manually configured values (`P1` or `P3`)

### Manual testing

TODO: Deployed to dev:
- registered a new client - saw empty LOCs in dynamo ✅ 
- updated an existing client configured with `[P0]` - saw empty LOCs in dynamo ✅
- updated an existing client configured with `[P0]` to turn on identity verification - saw empty LOCs in dynamo and made a P2 request successfully ✅
- Updated a client manually configured with `P3` and saw the the client locs unaffected ✅

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/dynamodb/SelfServiceClientRegistryExtension.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/dynamodb/SelfServiceClientRegistryExtension.java
@@ -1,0 +1,40 @@
+package uk.gov.di.orchestration.shared.dynamodb;
+
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbEnhancedClientExtension;
+import software.amazon.awssdk.enhanced.dynamodb.DynamoDbExtensionContext;
+import software.amazon.awssdk.enhanced.dynamodb.extensions.WriteModification;
+import software.amazon.awssdk.services.dynamodb.model.AttributeValue;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+import static uk.gov.di.orchestration.shared.entity.LevelOfConfidence.HIGH_LEVEL;
+import static uk.gov.di.orchestration.shared.entity.LevelOfConfidence.LOW_LEVEL;
+
+public class SelfServiceClientRegistryExtension implements DynamoDbEnhancedClientExtension {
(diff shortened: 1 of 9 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/dynamodb/SelfServiceClientRegistryExtension.java
```diff
@@ -0,0 +1,40 @@
...
+    @Override
+    public WriteModification beforeWrite(DynamoDbExtensionContext.BeforeWrite context) {
+        var clientItem = context.items();
+
+        if (clientItem == null) {
+            return WriteModification.builder().build();
+        }
+
+        if (!clientItem.containsKey("ClientLoCs")) {
+            return WriteModification.builder().build();
```
cearl1: I'm trying to think of the situation where this could be hit - or is it just a safeguard for accessing the ClientLoCs below?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/dynamodb/SelfServiceClientRegistryExtension.java
```diff
@@ -0,0 +1,40 @@
...
+        }
+
+        if (clientItem.get("ClientLoCs").l().stream()
+                .anyMatch(loc -> manuallyConfiguredLocValues.contains(loc.s()))) {
```
cearl1: If we wanted to ensure this is only called for new clients, i suppose we could add a new flag to the client registry like `isNewClient` and set it to true, then in this extension we could force it to false once we've identified it is a new client

Just an idea though! Not sure what that will look like in practice, or how happy we are with adding a new field to an already massive object :smile:

PR #8149, opened 2026-04-09
AUT-5323: Build AMAPI-to-ADAPI passkey proxy endpoints

ben-fernandes-sw:
## What

Implements the Account Data API proxying endpoints within the Account Management API. These are quite dumb endpoints that simply forward through the request/response to the ADAPI.

Requires a bit more CloudFormation config to set the request URI.

I've also added WireMock to the account-management-integration-tests for the first time, to mock the ADAPI that's called. These tests allow us to check that the body was returned with no interference by the AMAPI.

## How to review

1. Code Review
1. Deploy to a dev environment using the most appropriate [GitHub dev deployment workflow](https://github.com/govuk-one-login/authentication-api/actions), or using the deployment scripts in this repo for the old account (e.g., `./deploy-authdevs.sh -c -b --all`).
1. Test calling the new endpoints to make sure they perform the actions you'd expect on the ADAPI.

### DIFF
--- a/account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandlerTest.java
+++ b/account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandlerTest.java
@@ -3,35 +3,88 @@ package uk.gov.di.accountmanagement.lambda;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
+import uk.gov.di.authentication.shared.entity.ErrorResponse;
+import uk.gov.di.authentication.shared.exceptions.UnsuccessfulAccountDataApiResponseException;
+import uk.gov.di.authentication.shared.services.AccountDataApiService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
+
+import java.net.http.HttpResponse;
+import java.util.Map;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
 import static uk.gov.di.accountmanagement.helpers.CommonTestVariables.VALID_HEADERS;
 import static uk.gov.di.authentication.sharedtest.helper.CommonTestVariables.IP_ADDRESS;
+import static uk.gov.di.authentication.sharedtest.helper.CommonTestVariables.PUBLIC_SUBJECT_ID;
 import static uk.gov.di.authentication.sharedtest.helper.RequestEventHelper.contextWithSourceIp;
+import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasBody;
+import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasJsonBody;
 import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasStatus;
 
 public class PasskeysDeleteProxyHandlerTest {
     private final Context context = mock(Context.class);
     private final ConfigurationService configurationService = mock(ConfigurationService.class);
+    private final AccountDataApiService accountDataApiService = mock(AccountDataApiService.class);
 
     private PasskeysDeleteProxyHandler handler;
 
     @BeforeEach
     void setUp() {
-        handler = new PasskeysDeleteProxyHandler(configurationService);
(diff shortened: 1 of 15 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
```diff
@@ -917,6 +917,11 @@ public String getAccountDataURI() {
         return System.getenv("ACCOUNT_DATA_API_URI");
     }
 
+    public long getAccountDataApiCallTimeout() {
+        return Long.parseLong(
+                System.getenv().getOrDefault("ACCOUNT_DATA_API_CALL_TIMEOUT", "6000"));
```
ben-fernandes-sw, line 922: This is currently set to 6s. However, I've seen some 500s while testing, I think because of cold starts from ADAPI. Not sure how to fix that. Might be retrying in the proxy handler. Might be lengthening this timeout.

ChristianSoftwire: I don't hate the idea of retries but would probably be abit of extra work here. Maybe we can see how this performs after merge and then increase the timeout if necessary?

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandlerTest.java
```diff
@@ -3,35 +3,88 @@
...
+                    .thenReturn(mockHttpResponse);
+
+            // Act
+            var result = handler.handleRequest(passkeysDeleteProxyRequest(), context);
+
+            // Assert
+            assertThat(result, hasStatus(200));
+            assertThat(result, hasBody("{\"deleted\": true}"));
+            verify(accountDataApiService).deletePasskey(PUBLIC_SUBJECT_ID, PASSKEY_IDENTIFIER);
+        }
```
ChristianSoftwire, line 60: Nitpick - I would prefer if we had a standard across the tests of Given, When, Then VS Arrange, Act, Assert. Maybe one for dev chat but it seems we are mixing the 2 currently

account-management-integration-tests/src/test/java/uk/gov/di/accountmanagement/api/PasskeysRetrieveProxyHandlerIntegrationTest.java
```diff
@@ -0,0 +1,122 @@
...
+                makeRequest(
+                        Optional.empty(),
+                        Collections.emptyMap(),
+                        Collections.emptyMap(),
+                        Map.of("publicSubjectId", publicSubjectId));
+
+        // Assert
+        assertThat(response, hasStatus(200));
+        assertThat(response.getBody(), equalTo(adapiResponse));
+    }
```
ChristianSoftwire, line 90: super readable tests, very nice!

shared/src/test/java/uk/gov/di/authentication/shared/services/AccountDataApiServiceTest.java
```diff
@@ -0,0 +1,236 @@
...
+                service.retrievePasskeys("testPublicSubjectId");
+
+                verify(httpClient).send(httpRequestCaptor.capture(), any());
+                assertThat(
+                        httpRequestCaptor.getValue().uri(),
+                        equalTo(
+                                URI.create(
+                                        "https://example.com/accounts/testPublicSubjectId/authenticators/passkeys")));
+                assertThat(httpRequestCaptor.getValue().method(), equalTo("GET"));
+            }
```
ChristianSoftwire, line 74: Can we add the given, when, then or arrange, act, assert to these tests aswell please?

ChristianSoftwire (dismissed):
Left some minor nitpicky comments but definitely non blocking! Nice

shared/src/test/java/uk/gov/di/authentication/shared/services/AccountDataApiServiceTest.java
ben-fernandes-sw: Oh yeah, why'd I miss it out there 🤦

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandlerTest.java
ben-fernandes-sw: I'll stick with AAC for this PR. I've put it to a poll here:
https://gds.slack.com/archives/C092B9SMB5Y/p1775813377438919
You're right, we should be consistent. There are also plenty of places where we don't add comments in at all.

shared/src/main/java/uk/gov/di/authentication/shared/services/ConfigurationService.java
ben-fernandes-sw: I've posted here
https://gds.slack.com/archives/C0946SZ9LUS/p1775814302929029
to see if we have an idea of expected traffic, and whether we need to worry about this at all.

I'll resolve this comment and we can come back to it if it's a problem.

ChristianSoftwire (approved):

PR #8164, opened 2026-04-13
ATO-2258: Add hosted zone template for orchestration

Ryan-Andrews99:
### Wider context of change
We're planning on migrating the management of the oidc domain to our own AWS account, so this requires us to create a hosted zone for the domain. We're doing this in Cloudformation and this PR creates a template for that.

### What’s changed:
- Adds a hosted zone template containing a few feature:
  - Ability to conditionally deploy an ACM cert
  - Ability to conditionally deploy dev subdomains
  - Ability to add an suffix to differentiate the domain (e.g `oidc-orch.account.gov.uk`)

### Manual testing:

This one isn't deployed anywhere yet so not really much to test yet, but the commit message explains a bit of why I've set out things the way I have in the template 

### Checklist

### DIFF
--- a/ci/stack-orchestration/manual-stacks/domains/template.yaml
+++ b/ci/stack-orchestration/manual-stacks/domains/template.yaml
@@ -1,0 +1,302 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Transform:
+  - AWS::Serverless-2016-10-31
+
+Parameters:
+  Environment:
+    Type: String
+    Description: The name of the environment to deploy to
+    AllowedValues:
+      - dev
+      - build
+      - staging
+      - integration
+      - production
+  EndpointSuffix:
+    Description: >
+      (Optional) domain suffix i.e. -orch. Example output FQDN - oidc-orch.build.account.gov.uk
+    Type: String
+    Default: ""
+    AllowedPattern: "[a-z0-9-]*"
+    ConstraintDescription: >
+      must be a valid domain name, consists of lowercase letters, numbers and hyphens OR blank
+  DeployHostedZone:
+    Description: >
+      Deploy the Route53 hosted zone
+    Type: String
+    Default: "Yes"
+    AllowedValues:
+      - "Yes"
+      - "No"
+  DeployCertificate:
+    Description: >
+      Deploy domain certificate
+    Type: String
+    Default: "No"
+    AllowedValues:
+      - "Yes"
+      - "No"
+  DeploySubEnvDomains:
+    Description: >
+      (Optional) Deploy the sub-environment subdomains
+    Type: String
+    Default: "No"
+    AllowedValues:
+      - "Yes"
+      - "No"
+
+Conditions:
+  DeployDevSubDomains:
+    !And [
+      !Equals [!Ref Environment, dev],
+      !Equals [!Ref DeploySubEnvDomains, "Yes"],
+    ]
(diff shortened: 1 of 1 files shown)


ci/stack-orchestration/manual-stacks/domains/template.yaml
```diff
@@ -0,0 +1,302 @@
...
+      domainSuffix: account.gov.uk
+
+Resources:
+  OidcHostedZone:
+    Type: AWS::Route53::HostedZone
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
+    Properties:
+      Name: !Sub
+        - oidc${EndpointSuffix}.${DomainSuffix}
```
isaac-GDS, line 76: Is it always `-orch` for the endpointSuffix ?

ci/stack-orchestration/manual-stacks/domains/template.yaml
```diff
@@ -0,0 +1,302 @@
...
+        - DomainName: !Sub
+            - oidc${EndpointSuffix}.${DomainSuffix}
+            - DomainSuffix:
+                !FindInMap [
+                  EnvironmentConfiguration,
+                  !Ref Environment,
+                  domainSuffix,
+                ]
+          HostedZoneId: !Ref OidcHostedZone
+      SubjectAlternativeNames: !If
```
isaac-GDS, line 132: Cool - does this mean all the subdomains are valid from this one cert - you dont have to create a cert for each subdomain ?

isaac-GDS:
All looks good - so is the main reason for hosted zone per subdomain , circular dependency ?

ci/stack-orchestration/manual-stacks/domains/template.yaml
Ryan-Andrews99: Nope we can provide it as a template parameter but it'll likely be something like that

ci/stack-orchestration/manual-stacks/domains/template.yaml
Ryan-Andrews99: Yep exactly, we can issue one cert that covers the sub domains

ci/stack-orchestration/manual-stacks/domains/template.yaml
isaac-GDS: This is a personal preference but how do you feel with having the dash in the construction of the name and the EndpointSuffix only being the word with no dash. E.g `oidc-${EndpointSuffix}.${DomainSuffix}`

ci/stack-orchestration/manual-stacks/domains/template.yaml
isaac-GDS: Nice - didnt know that !

isaac-GDS:
Cool - dont think there is really any implications of creating a new hosted zone instead of records ? Cost increase by an insignificant amount but we have better and easier access control to our subdomains.

ci/stack-orchestration/manual-stacks/domains/template.yaml
Ryan-Andrews99: Ah I didn't do a great job explaining this, we'll at some point need to move from this being `oidc-orch` to just `oidc`, hence why the - needs to be included

isaac-GDS: Ah i see, make sense !

isaac-GDS (approved):

PR #8202, opened 2026-04-17
ATO-2508: Use default credentials provider, not Environment variable based

Ryan-Andrews99:
### Wider context of change

We previously enabled snapstart on the token handler[1] which started throwing errors due to the usage of EnvironmentVariableCredentialsProvider.

When using Snapstart[2], the environment variables based credentials are not set in the lambda environment, as these would have likely expired due to the cached execution environment.

When explicitly setting EnvironmentVariableCredentialsProvider here, we overide the default credential provider chain[3] lambda uses, which means lambda only tries to use the Environment variables to invoke the SDK client, causing exceptions to be thrown because the env vars are not set when Snapstart is enabled.

The fix here is to use the Default Credential provider class, as there is no real benefit to us overriding the default here. This means that once snapstart is re-enabled, we will no longer throw an exception here.

[1] - https://github.com/govuk-one-login/authentication-api/pull/8168
[2] - https://docs.aws.amazon.com/lambda/latest/dg/snapstart-activate.html#snapstart-credentials
[3] - https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html

### What’s changed
- Removes any usages of EnvironmentVariableCredentialsProvider with DefaultCredentialsProvider in client signature validation service. 
- Also removes it from a service in the an SNS service in account management

### Manual testing:
- This is really quite unlikely to break anything as the DefaultCredentialsProvider is more expansive than the EnvironmentVariableCredentialsProvider however I will test that our JWKS fetching is unaffected by this:

- Deploy to dev and configure stub for JWKS (deployed these [stub](https://github.com/govuk-one-login/relying-party-stub/pull/793) changes alongside this)
- Run through a journey and confirm the fetch jwks handler is invoked and a full journey is unaffected
- Saw the fetch JWKS handler was invoked fine 
- Saw the public key cache was populated fin
- Also deployed to auth dev and checked using the bulk account deletion handler and was fine

- I think I scanned through the codebase and saw that we're not using EnvironmentVariableCredentialsProvider anywhere outside of code that is called in integration tests, but would be good for someone to pull the branch down and confirm this!

### Checklist

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
@@ -1,5 +1,6 @@
 package uk.gov.di.accountmanagement.services;
 
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.regions.Region;
@@ -21,9 +22,10 @@ public class AwsSnsClient {
         var builder =
                 SnsClient.builder()
                         .region(Region.of(region))
-                        .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
+                        .credentialsProvider(DefaultCredentialsProvider.builder().build());
         if (endpointUri != null && !endpointUri.isEmpty()) {
-            builder.endpointOverride(URI.create(endpointUri));
+            builder.endpointOverride(URI.create(endpointUri))
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
         }
         this.snsClient = builder.build();
         this.topicArn = topicArn;
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
@@ -2,6 +2,7 @@ package uk.gov.di.accountmanagement.services;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.auth.credentials.DefaultCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.EnvironmentVariableCredentialsProvider;
 import software.amazon.awssdk.core.exception.SdkClientException;
 import software.amazon.awssdk.regions.Region;
@@ -24,9 +25,10 @@ public class AwsSnsClient {
         var builder =
                 SnsClient.builder()
                         .region(Region.of(region))
-                        .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
+                        .credentialsProvider(DefaultCredentialsProvider.builder().build());
         if (endpointUri != null && !endpointUri.isEmpty()) {
-            builder.endpointOverride(URI.create(endpointUri));
+            builder.endpointOverride(URI.create(endpointUri))
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
         }
         this.snsClient = builder.build();
         this.topicArn = topicArn;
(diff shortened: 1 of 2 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
```diff
@@ -21,9 +22,10 @@ public AwsSnsClient(String region, String topicArn, String endpointUri) {
         var builder =
                 SnsClient.builder()
                         .region(Region.of(region))
-                        .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
+                        .credentialsProvider(DefaultCredentialsProvider.builder().build());
         if (endpointUri != null && !endpointUri.isEmpty()) {
-            builder.endpointOverride(URI.create(endpointUri));
+            builder.endpointOverride(URI.create(endpointUri))
+                    .credentialsProvider(EnvironmentVariableCredentialsProvider.create());
```
Ryan-Andrews99, line 28: This is fine to do inside this if as the presence of endpointUri means we're running against localstack and should therefore use the dummy env var credentials

CarlyG55: Do we assume that this was the original reason env variable credentials provider was used in the first place?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
```diff
@@ -59,7 +59,7 @@ public ClientSignatureValidationService(ConfigurationService configurationServic
         this.lambdaClient =
                 LambdaClient.builder()
                         .region(Region.of(configurationService.getAwsRegion()))
-                        .credentialsProvider(EnvironmentVariableCredentialsProvider.create())
+                        .credentialsProvider(DefaultCredentialsProvider.builder().build())
```
CarlyG55, line 66: are there any notes in the commit message as to why this was set in the first place?

Ryan-Andrews99: So here is the [commit](https://github.com/govuk-one-login/authentication-api/pull/4939/changes/370a49c8bf9ba86a64055b7c736d8d5523f17c23) which introduces this. It is correct in saying that using an EnvironmentVariable credentials provider is recommended as something to [reduce the SDK startup time](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/lambda-optimize-starttime.html#:~:text=Use%20the%20EnvironmentVariableCredentialProvider). However there is also a lovely warning right underneath it:

_If you intend to use Lambda SnapStart for Java at some point, you should rely on the default credentials provider chain to lookup credentials. If you specify the EnvironmentVariableCredentialsProvider, the initial credentials lookup works, but when SnapStart is activated, the Java runtime sets container credentials environment variables. On activation, the environment variables used by the EnvironmentVariableCredentialsProvider—access key environment variables—are not available to the Java SDK._

I think turning on SnapStart massively outweighs any time we're saving from using the EnvironmentVariableCredentialsProvider.

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
Ryan-Andrews99: Hmm potentially, this [commit](https://github.com/govuk-one-login/authentication-api/commit/997b2a4f9e5351d1b9834a31d4cf253c1deb1a4e) includes the changes but doesn't give any reasoning as to why it was introduced 🤔

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
CarlyG55: 🫠 thanks for looking into it

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AwsSnsClient.java
CarlyG55: might be worth dropping a message to Dan? it's relatively recent so he might remember

CarlyG55 (approved):

PR #8209, opened 2026-04-20
Aut 5321/account data api authorizer

BeckaL:
**What**

Adds a new lambda for the account data api, which sits in front of all its endpoints as a custom authorizer (see aws documentation [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-use-lambda-authorizer.html)

It ensures that an access token is sent as an authorization header on all requests. It verifies the signature of the access token with the key served by the internal api's jwks endpoint.

If the token cannot be validated for whatever reason, the lambda throws an error and does not return a policy. This is translated by api gateway into a 401 unauthorized response. If the token is valid, the lambda responds with an access policy which will then be attached to the request to the api endpoints.

This is similar to [what we do in the account management api](https://github.com/govuk-one-login/authentication-api/blob/main/account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/AuthoriseAccessTokenHandlerTest.java#L3), although I have simplified the policy generation as we just need to return some json that represents the access policy (see aws docs [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html)).

A subsequent PR will add validation to all the api endpoints that the access policy contains the subject id in the request path - ie the authorizer has granted access to the specific resource being requested.

A general philosophy of this work has been not to add any additional dependency on shared code (either test or implementation) since at some point this api will be split out from this repo. So in some cases, code has been copied across and simplified.

**Related PRs**

Depends on https://github.com/govuk-one-login/authentication-api/pull/8213

[Subsequent PR](https://github.com/govuk-one-login/authentication-api/pull/8214) to add verification of the subject id

### DIFF
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
@@ -1,0 +1,57 @@
+package uk.gov.di.authentication.accountdata.lambda;
+
+import com.amazonaws.services.lambda.runtime.Context;
+import com.amazonaws.services.lambda.runtime.RequestHandler;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayCustomAuthorizerEvent;
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.jwt.util.DateUtils;
+import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.token.AccessToken;
+import com.nimbusds.oauth2.sdk.token.AccessTokenType;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.accountdata.entity.UnauthorizedException;
+import uk.gov.di.authentication.shared.entity.Result;
+import uk.gov.di.authentication.shared.helpers.NowHelper;
+
+import java.util.Date;
+
+public class AuthorizeHandler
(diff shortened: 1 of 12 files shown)


account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandlerTest.java
```diff
@@ -0,0 +1,277 @@
...
+            throws MalformedURLException {
+        var configurationService = mock(ConfigurationService.class);
+        when(configurationService.getAccountDataJwksUrl())
+                .thenThrow(new MalformedURLException("uh oh"));
+
+        assertThrows(
+                AuthorizeException.class,
+                () -> new AuthorizeHandler(configurationService),
+                "Expected to throw exception");
+    }
```
csophocleous, line 271: Non blocking - would be nice if these were in nested test classes (especially as we might add more validation to the claims in the future)

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -0,0 +1,148 @@
...
+    @Override
+    public Map<String, Object> handleRequest(
+            APIGatewayCustomAuthorizerEvent apiGatewayCustomAuthorizerEvent, Context context) {
+        var token = apiGatewayCustomAuthorizerEvent.getAuthorizationToken();
+        try {
+            var accessToken = AccessToken.parse(token, AccessTokenType.BEARER);
+            var signedAccessToken = SignedJWT.parse(accessToken.getValue());
+            var claimsSet = signedAccessToken.getJWTClaimsSet();
+
+            var maybeValidatedClaims =
```
csophocleous: Why is this not `validatedClaimsResult`?

ChristianSoftwire (commented):
Looks good! Just a couple minor naming/organisational comments

ChristianSoftwire (approved):
LGTM

PR #8213, opened 2026-04-20
AUT-5340: Include ADAPI access token in request to account data when retrieving passkeys

ChristianSoftwire:
## What

When we make a request to the `CheckUserExistsHandler` we are making a request to check if the user has a passkey. As we are adding an authoriser to the ADAPI, we need to send the access token in the request

## How to review

1. Code review
2. Deploy to dev env to test (contingent on authoriser being available)

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/passkeys/PasskeysService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/passkeys/PasskeysService.java
@@ -1,12 +1,16 @@
 package uk.gov.di.authentication.frontendapi.services.passkeys;
 
 import com.google.gson.JsonParseException;
+import com.nimbusds.oauth2.sdk.token.BearerAccessToken;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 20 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/passkeys/PasskeysService.java
```diff
@@ -89,4 +121,27 @@ private Result<PasskeyRetrieveError, PasskeysRetrieveResponse> parseResponse(
...
+                        internalCommonSubjectId,
+                        AccountDataScope.PASSKEY_RETRIEVE,
+                        sessionId,
+                        nowClock.now(),
+                        nowClock.nowPlus(ADAPI_ACCESS_TOKEN_LIFETIME, ChronoUnit.MINUTES),
+                        configurationService.getAuthToAccountDataApiAudience(),
+                        configurationService.getAuthIssuerClaim(),
+                        configurationService.getAMCClientId(),
+                        configurationService.getAuthToAccountDataSigningKey())
+                .fold(
```
BeckaL: Think you could probably just do .mapFailure here as you're not transforming the success value, but no biggie

BeckaL (dismissed):
Tiny non blocking comment

BeckaL (approved):

PR #8214, opened 2026-04-20
Aut 5321/subject id validation

BeckaL:
## What

We have a custom authorizer for the account data api which sits in front of all requests to the account data api and attaches an access token to authorized requests based on the token sent in the request. 

This means we now can add validation into each of the account data api handlers that the public subject id in the api path matches the one sent through in the token and attached to the request via the authorizer and api gateway.

This PR achieves that by retrieving the principal id from the request context (which will be populated by api gateway after the request has gone through the authorizer), and checking this against the public subject id in the api request path. If these match, we allow the request, but if not we return a 401 status.

## Related PRs

Depends on https://github.com/govuk-one-login/authentication-api/pull/8209

### DIFF
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/SubjectIdAuthorizerHelper.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/SubjectIdAuthorizerHelper.java
@@ -1,0 +1,36 @@
+package uk.gov.di.authentication.accountdata.helpers;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+
+public class SubjectIdAuthorizerHelper {
+    private static final Logger LOG = LogManager.getLogger(SubjectIdAuthorizerHelper.class);
+
+    private SubjectIdAuthorizerHelper() {
+        /* This utility class should not be instantiated */
+    }
+
+    private static final String PRINCIPAL_ID_FIELD = "principalId";
+
+    public static boolean isSubjectIdAuthorized(
+            String publicSubjectId,
+            APIGatewayProxyRequestEvent.ProxyRequestContext requestContext) {
+        if (requestContext.getAuthorizer() == null
+                || !requestContext.getAuthorizer().containsKey(PRINCIPAL_ID_FIELD)) {
+            LOG.warn("Authorizer missing from request context or principalId not present");
+            return false;
+        }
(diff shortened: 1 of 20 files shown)


account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/SubjectIdAuthorizerHelper.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    private static final String PRINCIPAL_ID_FIELD = "principalId";
+
+    public static boolean isSubjectIdAuthorized(
+            String publicSubjectId,
+            APIGatewayProxyRequestEvent.ProxyRequestContext requestContext) {
+        if (requestContext.getAuthorizer() == null
+                || !requestContext.getAuthorizer().containsKey(PRINCIPAL_ID_FIELD)) {
+            LOG.warn("Authorizer missing from request context or principalId not present");
```
csophocleous, line 21: Should this be a LOG.error()? Might make it easier to find that log if something does go wrong?

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/PasskeysCreateHandler.java
```diff
@@ -102,6 +107,16 @@ private Result<PasskeysCreateFailureReason, PasskeysCreateContext> parseRequest(
         return Result.success(new PasskeysCreateContext(publicSubjectId, passkeysCreateRequest));
     }
 
+    private Result<PasskeysCreateFailureReason, PasskeysCreateContext> validateAuthorizedSubjectId(
+            PasskeysCreateContext context, APIGatewayProxyRequestEvent input) {
+        if (isSubjectIdAuthorized(context.publicSubjectId(), input.getRequestContext())) {
+            return Result.success(context);
+        } else {
+            LOG.warn("SubjectId in path parameter does not match Authorizer principalId");
```
csophocleous, line 115: Ditto to the other comment, I feel like this should be a LOG.error

ChristianSoftwire (approved):

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/SubjectIdAuthorizerHelper.java
BeckaL: Thoughts here are that it's not an error on our side (is an expected case for when the authorizer is null) but indicates a problem on the client's side. So warning more appropriate here. But we're definitely not consistent with our log levels so this might not be true everywhere.

PR #8215, opened 2026-04-21
AUT-5322: Send ADAPI bearer token from proxy

alhcomer:
## What

- `PasskeysRetrieveProxyHandler` and `PasskeysDeleteProxyHandler` receives `X-ADAPI-AccessToken` header with token, extracting the value.
- The proxy handler then passes this value on to ADAPI in bearer authentication `Authorization: Bearer <token>` format

## How to review

1. Code Review
2. Test in dev alongside built authorizer

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
@@ -53,10 +53,23 @@ public class PasskeysDeleteProxyHandler
         LOG.info("PasskeysDeleteProxyHandler invoked");
 
         var publicSubjectId = input.getPathParameters().getOrDefault("publicSubjectId", "");
+        if (publicSubjectId.isEmpty()) {
+            LOG.warn("No publicSubjectId in path parameters, request will likely fail");
+        }
+
         var passkeyIdentifier = input.getPathParameters().getOrDefault("passkeyIdentifier", "");
+        if (passkeyIdentifier.isEmpty()) {
+            LOG.warn("No passkeyIdentifier in path parameters, request will likely fail");
+        }
+
+        var token = input.getHeaders().getOrDefault("X-ADAPI-AccessToken", "");
+        if (token.isEmpty()) {
+            LOG.warn("No X-ADAPI-AccessToken in headers, request will likely fail");
+        }
 
         try {
-            var response = accountDataApiService.deletePasskey(publicSubjectId, passkeyIdentifier);
+            var response =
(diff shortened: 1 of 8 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
```diff
@@ -54,9 +54,11 @@ public APIGatewayProxyResponseEvent passkeyDeleteProxyHandler(
 
         var publicSubjectId = input.getPathParameters().getOrDefault("publicSubjectId", "");
         var passkeyIdentifier = input.getPathParameters().getOrDefault("passkeyIdentifier", "");
+        var token = input.getHeaders().getOrDefault("X-ADAPI-AccessToken", "");
```
BeckaL, line 57: I wonder here whether we should be defaulting or e.g. returning a 401 (which is what should happen when this request is forwarded to api gateway). Kind of think we should be early returning with unsuccessful status code rather than defaulting as it's another layer to debug if this isn't being sent. Let's log when it's empty too

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysRetrieveProxyHandler.java
```diff
@@ -53,9 +53,10 @@ public APIGatewayProxyResponseEvent passkeyRetrieveProxyHandler(
         LOG.info("PasskeysRetrieveProxyHandler invoked");
 
         var publicSubjectId = input.getPathParameters().getOrDefault("publicSubjectId", "");
+        var token = input.getHeaders().getOrDefault("X-ADAPI-AccessToken", "");
```
BeckaL, line 56: same here

account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/PasskeysRetrieveProxyHandlerTest.java
```diff
@@ -25,13 +26,16 @@
 import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasJsonBody;
 import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasStatus;
 
-public class PasskeysRetrieveProxyHandlerTest {
+class PasskeysRetrieveProxyHandlerTest {
```
BeckaL, line 29: If you're making change, squash this commit into the previous as the code changes for these tests are in there (don't mind both handlers being changed in one test) - preference for atomic commits

BeckaL (commented):
Generally looks sensible to me!

BeckaL (dismissed):

BeckaL (approved):

PR #8221, opened 2026-04-22
BAU: Orchestration local running

Joe-Edwards-GDS:
### Wider context of change

Setup to run Orchestration API and associated stubs locally without an AWS environment.

This is very similar to the Authentication and IPV Core local running setup, using docker containers and harnesses to replace API Gateway and other AWS infrastructure.

### What’s changed

- Add `docker-compose.yml` which runs containers for Orchestration, RP Stub and Auth/IPV/AIS stubs (as well as the dynamodb, redis and localstack containers from the base configuration).
- Add `.env.local` with configuration for use in the local setup
- Add Javalin application which emulates the API Gateway interface and invokes lambda handlers for the corresponding routes

### Manual testing

- Run `docker compose up` and try to run an auth-only journey.
- Set up a debugger in IntelliJ and try attaching to the Java process to use breakpoints and step-through debugging.

### Related PRs

- RP Stub: https://github.com/govuk-one-login/relying-party-stub/pull/794
- Orch Stubs: https://github.com/govuk-one-login/orch-stubs/pull/543


orchestration-local-running/src/main/java/uk/gov/di/orchestration/local/LocalOrchestrationApi.java
```diff
@@ -0,0 +1,69 @@
...
+import uk.gov.di.authentication.oidc.lambda.TrustMarkHandler;
+import uk.gov.di.authentication.oidc.lambda.UserInfoHandler;
+import uk.gov.di.authentication.oidc.lambda.WellknownHandler;
+
+import static uk.gov.di.orchestration.local.handlers.ApiGatewayLambdaHandler.handlerFor;
+
+public class LocalOrchestrationApi {
+    private static final int DEFAULT_PORT = 4400;
+
+    // These path mappings must match those in the infrastructure-as-code configuration
```
Joe-Edwards-GDS, line 29: This is the biggest trade-off - if merged, we will need to keep this in sync with the cloudformation.

(FWIW it can also be useful as API documentation - much easier to decipher than cloudformation IMO!)

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/api/OidcAPI.java
```diff
@@ -9,9 +9,11 @@
 public class OidcAPI {
 
     private final URI oidcApiBaseUrl;
+    private final URI oidcFrontendBaseUrl;
 
     public OidcAPI(ConfigurationService configurationService) {
         oidcApiBaseUrl = configurationService.getOidcApiBaseURL();
+        oidcFrontendBaseUrl = configurationService.getOidcFrontendBaseURL();
```
Joe-Edwards-GDS, line 16: Open to better names here... Essentially the problem is that we need to distinguish the frontchannel URLs (i.e. ones loaded by the browser) from backchannel URLs (i.e. ones made server to server).

When running locally:
- Frontchannel needs to use `localhost` (your browser can't resolve docker domains)
- Backchannel needs to use `host.docker.internal` (localhost resolves to the container, not the host machine)

Alternatively we could require devs to set up a custom domain in `hosts` which will work for both, but it's nicer if it _just works_!

Joe-Edwards-GDS: Naming here matches our internal components (auth, IPV, etc.) which need different URLs in production as well (frontend vs. API)

orchestration-local-running/src/main/java/uk/gov/di/orchestration/local/handlers/ApiGatewayLambdaHandler.java
Joe-Edwards-GDS, line 1: (A lot of this mapping code and the initialisers is duplicated from the Auth equivalent)

orchestration-local-running/.env.local
```diff
@@ -0,0 +1,75 @@
...
+
+# SPOT
+SPOT_REQUEST_QUEUE_URL=http://sqs.eu-west-2.localhost:45678/000000000000/local-spot-request-queue
+SPOT_RESPONSE_QUEUE_URL=http://sqs.eu-west-2.localhost:45678/000000000000/local-spot-response-queue
+
+# Features
+IDENTITY_ENABLED=true
+SEND_STORAGE_TOKEN_TO_IPV_ENABLED=true
+USE_SYNC_WAIT_FOR_SPOT=true
+NEW_SPOT_REQUEST_QUEUE_WRITING=true
```
Joe-Edwards-GDS, line 75: Any other feature flags/config that should be included?

orchestration-local-running/src/main/java/uk/gov/di/orchestration/local/LocalOrchestrationApi.java
cearl1: Maybe we can update the PR template to remind us to keep this up to date?

orchestration-local-running/src/main/java/uk/gov/di/orchestration/local/App.java
```diff
@@ -0,0 +1,122 @@
...
+        //
+        // Most tables are initialised automatically in the 'warm up' step
+        var dynamoInitialiser = new DynamoDbInitialiser();
+        dynamoInitialiser.addRecords(
+                "local-client-registry",
+                ClientRegistry.class,
+                List.of(
+                        // Client for the local RP stub
+                        new ClientRegistry()
+                                .withClientID("local-rp-stub")
```
cearl1, line 68: Do we want to mention in the README that if you want to do an identity journey, you'll need to update the ClientLoCs to include P2?

Joe-Edwards-GDS: Interesting - I seem to be able to run identity journeys without doing that :suspect:. I think if you have `identityVerificationSupported = true` then it defaults to `[P0, P2]`?

However, we should make sure this config is as realistic as possible - so I'll update to specify. Is there anything else you think is obviously missing?

cearl1: Ahh yep, i think what happened is i requested P1 too which would have made the request fail. I think this is the only field we'll need to consider though!

Joe-Edwards-GDS: Added them all for good measure :)

cearl1 (approved):

PR #8223, opened 2026-04-23
Bau/improve test assertions

BeckaL:
## What

Switches all tests for audit events in the account management api to assert on the full event.

This is stage one in a refactor for how we construct audit events in the account management api, which is convoluted and has some minor issues - as demonstrated here in the TODOs.

We were previously asserting on partial bits of the audit event, which hid a few issues:

1. We were adding duplicates of the same data (not an issue since the downstream event would just contain one of these, but a symptom of overly convoluted construction
2. In some cases, we are adding contradictory data - ie two bits of metadata with the same key but different values, which leads to one of these being discarded in the actual submitted audit event, which is a problem.

This PR does not fix these issues, but makes them clear in tests, and will serve a base from which to refactor the production code safely.

The TODOs are left in deliberately, and will be fixed in a subsequent PR, so the sonar issues highlighting this can be ignored.

## How to review

1. Code review


ChristianSoftwire (approved):
LGTM - much easier to read

PR #8227, opened 2026-04-23
BAU: Attempt to fix some flakey tests

BeckaL:
In auth, we have some tests that are frequently flakey in CI. These fail on a before hook to create an sns queue, but in auth these are no longer used except in one handler, so we should stop creating these queues unecessarily for every integration test.

See, for example, a screenshot from a test report from a failed integration test in CI:
Screenshot 2026-04-23 at 14 43 14

Hopefully this will fix some of the flakiness. It should definitely avoid creating a lot of unnecessary resources for integration tests and works as a general tidy.


BeckaL:
Minor sonar issues flagged here are not to do with code introduced or touched in my PR, so ignoring (existing minor things)

alhcomer (dismissed):

alhcomer (approved):

PR #8237, opened 2026-04-24
ATO-2527: UPRN fix

Joe-Edwards-GDS:
### Wider context of change

We are currently serialising UPRNs incorrectly in the UserInfo response - they should be preserved in the format they are returned from IPV, but due to the need to deserialise + serialise (for compatibility with Nimbusds) we end up mangling the format as it goes via Double.

E.g.: `10022812929` => `1.0022812929E10`

N.B. Although UPRN is often formatted with leading zeroes (i.e. as a string), this is not in the spec: https://github.com/co-cddo/open-standards/issues/68#issuecomment-491245902

### What’s changed

Update the GSON strategy to use `LONG_TO_DOUBLE`, so it tries to use a long first (which doesn't lose any precision).

An ideal fix might eliminate this round-trip serialisation entire, but that's hard to achieve with our use of the Nimbusds library.

### Manual testing

Run locally and passed the userinfo claim back from the IPV stub.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/orchestration/oidc/lambda/UserInfoIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/orchestration/oidc/lambda/UserInfoIntegrationTest.java
@@ -16,8 +16,6 @@ import com.nimbusds.openid.connect.sdk.OIDCScopeValue;
 import com.nimbusds.openid.connect.sdk.UserInfoErrorResponse;
 import com.nimbusds.openid.connect.sdk.claims.ClaimsSetRequest;
 import com.nimbusds.openid.connect.sdk.claims.UserInfo;
-import net.minidev.json.JSONArray;
-import net.minidev.json.JSONObject;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.RegisterExtension;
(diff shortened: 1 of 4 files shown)


integration-tests/src/test/java/uk/gov/di/orchestration/oidc/lambda/UserInfoIntegrationTest.java
```diff
@@ -221,17 +223,16 @@ void shouldReturn200WhenIdentityIsEnabledAndIdentityClaimsArePresent() throws Pa
...
         var drivingPermitClaim =
-                userInfoResponse.getJSONArrayClaim(ValidClaims.DRIVING_PERMIT.getValue());
+                (JSONArray) userInfoResponse.getClaim(ValidClaims.DRIVING_PERMIT.getValue());
         var returnCodeClaim =
-                userInfoResponse.getJSONArrayClaim(ValidClaims.RETURN_CODE.getValue());
-
-        assertThat(addressClaim.toJSONString(), equalTo(ADDRESS_CLAIM));
-        assertThat(passportClaim.toJSONString(), equalTo(PASSPORT_CLAIM));
-        assertThat(drivingPermitClaim.toJSONString(), equalTo(DRIVING_PERMIT));
-        assertThat(returnCodeClaim.toJSONString(), equalTo(RETURN_CODE));
```
Ryan-Andrews99, line 234: I managed to get a reliable assertion against the claim values doing something like this:

```
 var mapper = new ObjectMapper();
        assertEquals(mapper.readTree(addressClaim.toJSONString()), mapper.readTree(ADDRESS_CLAIM));
        assertEquals(mapper.readTree(passportClaim.toJSONString()), mapper.readTree(PASSPORT_CLAIM));
        assertEquals(mapper.readTree(drivingPermitClaim.toJSONString()), mapper.readTree(DRIVING_PERMIT));
        assertEquals(mapper.readTree(returnCodeClaim.toJSONString()), mapper.readTree(RETURN_CODE));
```

I reverted the serialisation service changes and the test failed, so this does catch some differences where GSON is mangling things

Ryan-Andrews99 (approved):
Looks good to me, I don't think there is anywhere else we're serialising numbers so think this should be fine.  I've added an optional comment about asserting against the exact claim values

integration-tests/src/test/java/uk/gov/di/orchestration/oidc/lambda/UserInfoIntegrationTest.java
Joe-Edwards-GDS: Ah, using Jackson to check GSON 😂

It's a bit fiddly, so given that we check it more precisely in the UT I'm tempted to leave this one...

PR #8242, opened 2026-04-27
AUT-5324: Allow list AMC VPCE on internal API

alhcomer:
## What

When curling the internal API from AMC hit a resource policy deny due to the AMC VPCE not being allow-listed. This should resolve that.

## How to review

1. Code Review

### DIFF
--- a/ci/cloudformation/auth/api/auth-internal-api.yaml
+++ b/ci/cloudformation/auth/api/auth-internal-api.yaml
@@ -50,6 +50,32 @@ Resources:
                           DefaultValue: "",
                         ]
                       - !Ref AWS::NoValue
+                    - !If
+                      - AddAmcApiVpcEndpointId
+                      - !FindInMap [
+                          EnvironmentConfiguration,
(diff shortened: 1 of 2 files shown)


ci/cloudformation/auth/api/auth-internal-api.yaml
```diff
@@ -50,6 +50,32 @@ Resources:
...
+                    - !If
+                      - AddAmcApiVpcEndpointId
+                      - !FindInMap [
+                          EnvironmentConfiguration,
+                          !Ref Environment,
+                          amcApiVpcEndpointId,
+                          DefaultValue: "",
+                        ]
+                      - !Ref AWS::NoValue
+            - !If
```
pskushwaha1, line 62: We have a deny statement above which cover the All endpoint deny if they are not coming from allowed VpC endpoint , why are we adding explicitly deny for AMC ?

pskushwaha1 (approved):

PR #8253, opened 2026-04-29
Bau/assert on full audit event

BeckaL:
## What

A test refactor - Stop using argument captors to assert on individual fields in the submitted audit event's context, instead assert on the entire audit context.

This is done for:
* readability / simplicity
* to ensure that we don't mask issues where we add conflicting metadata (multiple pairs with the same key but different values) to the audit context (which does not show up in tests where we just assert on individual items)
* to provide a solid foundation for refactoring

I'm planning a subsequent change to stop allowing metadata items to be set on the audit context (at the moment it can be sent through for the audit event in multiple ways), so this is a preparatory step to ensure we have a good base of tests to make this change from.


alhcomer (approved):
lgtm

PR #8263, opened 2026-04-30
BAU: Retry calling the ADAPI once if a timeout was encountered

ben-fernandes-sw:
## What

There is a potential issue where, on the first call of an ADAPI lambda, there is no SnapStart point to restore from, so it'll take a while to warm up. Should be no more than 10s.

This will try calling the lambda once (with a timeout of 6s), and if there's a timeout try again (waiting for another 6s). That should give the lambda enough time to warm up.

For the majority of requests though, the lambda will already be warm, or will be able to restore using SnapStart, so should return in 1s/2s.

## How to review

1. Code Review
1. Deploy Account Management API to a dev environment and validate that the retrieve/delete proxy requests still work

## Initial testing
I deployed to authdev2 and tested a request from AMAPI via the API Gateway test console. I successfully got a 401 from the ADAPI Authorizer and confirmed this by checking its logs in CloudWatch.

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/services/AccountDataApiService.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/services/AccountDataApiService.java
@@ -1,5 +1,7 @@
 package uk.gov.di.authentication.shared.services;
 
+import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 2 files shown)


ChristianSoftwire (approved):
LGTM

PR #8275, opened 2026-05-06
Aut 5358/delete passkey

BeckaL:
## What

Implement delete passkey endpoint. The barebones lambda already exists, but does not have any of the delete logic in.

This implements this along with appropriate error handling. It also updates the open api spec to detail the 400 response from this endpoint.

## How to review

### DIFF
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/services/DynamoPasskeyService.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/services/DynamoPasskeyService.java
@@ -4,6 +4,7 @@ import uk.gov.di.authentication.accountdata.constants.AccountDataConstants;
 import uk.gov.di.authentication.accountdata.entity.Authenticator;
 import uk.gov.di.authentication.accountdata.entity.passkey.Passkey;
 import uk.gov.di.authentication.accountdata.entity.passkey.failurereasons.PasskeysCreateFailureReason;
+import uk.gov.di.authentication.accountdata.entity.passkey.failurereasons.PasskeysDeleteFailureReason;
 import uk.gov.di.authentication.accountdata.entity.passkey.failurereasons.PasskeysUpdateFailureReason;
 import uk.gov.di.authentication.shared.entity.Result;
 import uk.gov.di.authentication.shared.services.BaseDynamoService;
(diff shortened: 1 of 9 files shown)


account-data-api/src/main/java/uk/gov/di/authentication/accountdata/services/DynamoPasskeyService.java
```diff
@@ -52,8 +53,14 @@ public Result<PasskeysUpdateFailureReason, Passkey> updatePasskey(
...
-        delete(publicSubjectId, sortKey);
+    public Result<PasskeysDeleteFailureReason, Void> deletePasskey(
+            String publicSubjectId, String passkeyId) {
+        return getPasskeyForUserWithPasskeyId(publicSubjectId, passkeyId)
+                .map(
+                        pk -> {
+                            delete(publicSubjectId, buildSortKey(passkeyId));
+                            return Result.<PasskeysDeleteFailureReason, Void>success(null);
+                        })
+                .orElseGet(() -> Result.failure(PasskeysDeleteFailureReason.PASSKEY_NOT_FOUND));
```
csophocleous, line 64: I think the `delete` method does the get check already so I think this would be reading from the db twice

account-data-api/src/test/java/uk/gov/di/authentication/accountdata/services/PasskeysServiceTest.java
```diff
@@ -249,25 +275,55 @@ void shouldReturnFailedToGetPasskeysIfExceptionThrown() {
...
+
+                // Then
+                assertEquals(passkeyDeleteResult, result);
+            }
+        }
+
+        @Nested
+        class Failure {
+            @Test
+            void shouldReturnFailedToDetePasskeysIfExceptionThrown() {
```
csophocleous: Typo in test name

account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/PasskeysDeleteHandlerTest.java
```diff
@@ -97,25 +102,46 @@ void shouldReturn500WhenDatabaseOperationFails() {
...
+                            Map.of("publicSubjectId", PUBLIC_SUBJECT_ID),
+                            ErrorResponse.MISSING_PASSKEY_ID),
+                    Arguments.of(
+                            Map.of("publicSubjectId", PUBLIC_SUBJECT_ID, "passkeyId", ""),
+                            ErrorResponse.MISSING_PASSKEY_ID));
+        }
+
+        @ParameterizedTest
+        @MethodSource("pathParamsWithMissingFields")
+        void shouldReturn400WhenPublicSubjectIdNotPresent(
```
csophocleous: Could we rename this test to be something like `shouldReturn400WhenPathParamsHaveMissingFields`

ChristianSoftwire (commented):
Overall looks good. Just some comments

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/services/DynamoPasskeyService.java
BeckaL: Ah good spot - switched to the `delete(T item)` method which doesn't do the read first and ifPresent check

ChristianSoftwire (approved):
LGTM

PR #8290, opened 2026-05-11
AUT-4939: Possibly return hasActivePasskey=null

ben-fernandes-sw:
## What

Previously, we were always returning `hasActivePasskey=true` or `=false`. However, in the case where the `PasskeyService` request failed, we lied and still sent back a `false`.

This changes the `hasActivePasskey` check to only return a boolean response if passkeys are enabled, and if we successfully make the request to get the user's current state. Otherwise, we'll always return `null`, to represent the "unknown" state.

The response object will either include the `true`/`false` value if it's available, or `null` if it's not.

## How to review

1. Code Review
2. Test

## Related PRs

https://github.com/govuk-one-login/authentication-frontend/pull/3378

### DIFF
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
@@ -197,7 +197,7 @@ class CheckUserExistsHandlerTest {
                     "mfaMethodType":"SMS",
                     "phoneNumberLastThree":"%s",
                     "lockoutInformation":[],
-                    "hasActivePasskey":false,
+                    "hasActivePasskey":null,
                     "needsForcedMFAResetAfterMFACheck":false}
                     """,
                             EMAIL_ADDRESS, phoneNumber.substring(phoneNumber.length() - 3));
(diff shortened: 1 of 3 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
```diff
@@ -197,7 +197,7 @@ void shouldReturn200WithRelevantMfaMethod() {
                     "mfaMethodType":"SMS",
                     "phoneNumberLastThree":"%s",
                     "lockoutInformation":[],
-                    "hasActivePasskey":false,
+                    "hasActivePasskey":null,
```
BeckaL, line 200: Minor point - I'd rather the default for tests that aren't specifically testing the passkeys response was that there either was or wasn't an active passkey, as that better reflects the average response here, and we just had the single test case of it being null when erroring as you've added below. Can mock the standard passkeys response up in the before block if we want to avoid crowding out all the tests

ben-fernandes-sw: The logic in the code atm is basically
* If user exists
  * If `PasskeyService` responds successfully
    * Return a `true`/`false`

So really, in the `WhenUserExists` `@BeforeEach`, `hasActivePasskey` should always return `false`, unless:
* we're simulating a `PasskeyService` failure, or
* we're simulating the actual response being returned correctly

BeckaL: Ah from chat - the key is that we now return hasActivePasskey = null when the feature flag is off. Given feature flag is off in most envs at the moment, these nulls do reflect the default case at the moment

BeckaL (approved):

PR #8295, opened 2026-05-12
AUT-5319: Return ADAPI access token to Orch in UserInfo claims

alhcomer:
## What

- Orchestration request the ADAPI access token by including
  the user info claim account_data_api_access_token sent in
  the JAR to the frontends authorize endpoint (the entry point
  to auth from orch)
- This claim signifies that Orch require an ADAPI access token
  for Home to use when making requests to ADAPI
- The user info claims are then:
  - Sent to the AuthorizationAuthCodeHandler and put in the
    auth code store
  - Put in the AccessTokenStore in the auth external api's
    TokenHandler
  - Retrieved in the UserInfoHandler from the AccessTokenStore
- If `account_data_api_access_token` is present, we populate the
  claim with an ADAPI access token for a Home client

## How to review

## Checklist


auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
```diff
@@ -122,12 +129,14 @@ private void addClaimsFromToken(
                             configurationService.getHomeClientId(),
                             configurationService.getAuthToAccountDataSigningKey());
 
-            if (result.isSuccess()) {
```
BeckaL: Given this is a vital step in all sign in journeys, I'm nervous about going straight to failing here in case we have e.g. config issues in prod. Could we do this in two steps - firstly just adding warning logging and then reviewing in each envs to make sure we're not going to suddenly cause an outage on all journeys

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccessTokenStoreService.java
```diff
@@ -31,15 +31,15 @@
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.LogFieldName.CLIENT_ID;
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachLogFieldToLogs;
 
-public class AccessTokenService {
+public class AccessTokenStoreService {
```
BeckaL: I don't think we should do this - this isn't our code but orch's. Possibly get rid of this commit entirely?

auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
BeckaL: Actually sorry I missed that this only happens when orch request the token is sent - do we know if this is happening at all at the moment? I just want to make sure we don't break any journeys

oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AccessTokenStoreService.java
alhcomer: That was some careless intellij renaming on my part - have fixed now

auth-external-api/src/main/java/uk/gov/di/authentication/external/services/UserInfoService.java
alhcomer: The claim is put in for all for all `am` scoped RP requests, not on a per-environment basis (see: https://github.com/govuk-one-login/authentication-api/blob/c119971ad21f53fb8dbadced1d5fd28638f39b9c/oidc-api/src/main/java/uk/gov/di/authentication/oidc/services/AuthenticationAuthorizationService.java#L339-L342)

Maybe we pause the pipeline at staging to test in the RP stub

BeckaL (approved):

PR #8305, opened 2026-05-14
ATO-2525: alternate domain params

Ryan-Andrews99:
### Wider context of change:

As part of our migration, we would like to deploy our infrastructure on an alternate domain to facilitate testing. To do this we may need to be able alternate the parameters we pass to our 

### What’s changed:
- Adds a parameter to our deploy-dev script for the hosted zone to alternate the parameters we use for the live or alternate domain
- Does the same for cloudfront for the distribution and TLS cert

### Manual testing:

- N/A will test when using to deploy

### Checklist

### DIFF
--- a/ci/stack-orchestration/deploy-cloudfront.sh
+++ b/ci/stack-orchestration/deploy-cloudfront.sh
@@ -37,7 +37,7 @@ function usage {
 
   Options:
     -e, --environment        The environment you wish to deploy to i.e dev, build, staging, integration, or production
-    -c, --certificate        Creates certificate in us-east-1 region for CloudFront Distribution.
+    -c, --certificate         Creates certificate in us-east-1 region for CloudFront Distribution.
     -d, --distribution       Creates the CloudFront distribution
     -m, --monitoring         Deploys CloudFront Extended Monitoring stack in us-east-1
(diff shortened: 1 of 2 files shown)


ci/stack-orchestration/configuration/dev/hosted-zone/alternative-domain-parameters.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  {
+    "ParameterKey": "DeploySubEnvDomains",
+    "ParameterValue": "No"
+  },
+  {
+    "ParameterKey": "EndpointSuffix",
+    "ParameterValue": "-orch"
+  }
+]
```
Ryan-Andrews99: This isn't deployed to dev, more included as an example

cearl1: Should we be adding an example for the alternative cloudfront certificate parameters too?

Ryan-Andrews99: Hmm thinking about it, it might be better to not have parameters here where they don't reflect what is actually deployed. I'll remove this

cearl1 (approved):

PR #8335, opened 2026-05-15
AUT-5396: ADAPI authorizer validation

alhcomer:
## What

- Extends the account-data-api authorizer to validate all JWT claims: 
   - issuer 
   - audience 
   - not-before time 
   - client_id (AMC + Home allow list) 
   - scope-to-HTTP-method mapping (AMC client restricted to create/retrieve only) 

## How to review

2. Test in dev

## Checklist

### DIFF
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
@@ -3,6 +3,7 @@ package uk.gov.di.authentication.accountdata.lambda;
 import com.amazonaws.services.lambda.runtime.Context;
 import com.amazonaws.services.lambda.runtime.RequestHandler;
 import com.amazonaws.services.lambda.runtime.events.APIGatewayCustomAuthorizerEvent;
+import com.amazonaws.services.lambda.runtime.events.IamPolicyResponseV1;
 import com.nimbusds.jose.JOSEException;
 import com.nimbusds.jose.JWSAlgorithm;
 import com.nimbusds.jose.JWSVerifier;
@@ -25,10 +26,9 @@ import uk.gov.di.authentication.shared.services.ConfigurationService;
 import java.net.MalformedURLException;
 import java.util.Date;
 import java.util.List;
-import java.util.Map;
 
 public class AuthorizeHandler
-        implements RequestHandler<APIGatewayCustomAuthorizerEvent, Map<String, Object>> {
+        implements RequestHandler<APIGatewayCustomAuthorizerEvent, IamPolicyResponseV1> {
     private static final Logger LOG = LogManager.getLogger(AuthorizeHandler.class);
     private RemoteJwksService jwksService;
 
@@ -50,7 +50,7 @@ public class AuthorizeHandler
     }
 
     @Override
-    public Map<String, Object> handleRequest(
+    public IamPolicyResponseV1 handleRequest(
             APIGatewayCustomAuthorizerEvent apiGatewayCustomAuthorizerEvent, Context context) {
         var token = apiGatewayCustomAuthorizerEvent.getAuthorizationToken();
         try {
@@ -131,19 +131,19 @@ public class AuthorizeHandler
         return Result.success(claimsSet);
     }
 
-    private Map<String, Object> getAllowExecuteApiPolicyForSubject(
+    private IamPolicyResponseV1 getAllowExecuteApiPolicyForSubject(
             String subject, String methodArn) {
-        var executeApiStatement =
-                Map.ofEntries(
-                        Map.entry("Action", "execute-api:Invoke"),
-                        Map.entry("Effect", "Allow"),
-                        Map.entry("Resource", methodArn));
-        return Map.ofEntries(
-                Map.entry("principalId", subject),
-                Map.entry(
-                        "policyDocument",
-                        Map.ofEntries(
-                                Map.entry("Version", "2012-10-17"),
-                                Map.entry("Statement", List.of(executeApiStatement)))));
+        var statement = IamPolicyResponseV1.allowStatement(methodArn);
+
+        var policyDocument =
+                IamPolicyResponseV1.PolicyDocument.builder()
+                        .withStatement(List.of(statement))
+                        .withVersion(IamPolicyResponseV1.VERSION_2012_10_17)
+                        .build();
+
+        return IamPolicyResponseV1.builder()
+                .withPolicyDocument(policyDocument)
+                .withPrincipalId(subject)
+                .build();
     }
 }
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
@@ -155,6 +155,14 @@ public class AuthorizeHandler
             LOG.warn("Access Token audience is invalid");
             return Result.failure(new UnauthorizedException());
         }
+        if (claimsSet.getNotBeforeTime() == null) {
+            LOG.warn("Access Token is missing nbf claim");
+            return Result.failure(new UnauthorizedException());
+        }
+        if (DateUtils.isAfter(claimsSet.getNotBeforeTime(), NowHelper.now(), 0)) {
+            LOG.warn("Access Token is not yet valid (nbf: {})", claimsSet.getNotBeforeTime());
+            return Result.failure(new UnauthorizedException());
+        }
         return Result.success(claimsSet);
     }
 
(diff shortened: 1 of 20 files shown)


account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -154,6 +154,11 @@ private Result<UnauthorizedException, JWTClaimsSet> validateClaimsSet(JWTClaimsS
             LOG.warn("Access Token audience is invalid");
             return Result.failure(new UnauthorizedException());
         }
+        if (claimsSet.getNotBeforeTime() != null
```
BeckaL: Do we want the token to be pass validation if the not before time is null?

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -159,6 +159,13 @@ private Result<UnauthorizedException, JWTClaimsSet> validateClaimsSet(JWTClaimsS
             LOG.warn("Access Token is not yet valid (nbf: {})", claimsSet.getNotBeforeTime());
             return Result.failure(new UnauthorizedException());
         }
+        var expectedClientId = configurationService.getAMCClientId();
```
BeckaL: I find it a bit confusing here that we have an 'expected client id' and then a 'home client id', either of which we accept. suggest something like:
```
var allowedClientIds = List.of(configurationService.getAmcClientId, configurationService.getHomeClientId())
if (allowedClientIds.contains(clientId)...)
```

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -146,4 +153,39 @@ private IamPolicyResponseV1 getAllowExecuteApiPolicyForSubject(
...
+
+    private String extractHttpMethod(String methodArn) {
+        String[] parts = methodArn.split("/");
+        if (parts.length < 3) {
+            throw new UnauthorizedException();
+        }
+        return parts[2];
+    }
+
+    private Result<UnauthorizedException, JWTClaimsSet> validateScope(
```
BeckaL: So I think we don't want to do this here - here we should perhaps just validate the scope is part of an allow list, but in the individual endpoints we want to validate the scope matches the endpoint requested. This means passing the scope through in the created policy document

BeckaL (commented):
Looking good! Nice find on the policy document class. A few comments - lmk if you want to discuss. We should also update documentation for the account data api when this goes in (doesn't have to be added to this PR, can be in a follow up). Hoping to get [this](https://github.com/govuk-one-login/authentication-api/pull/8300) PR in v soon, so the token generation will need to be updated

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -77,73 +83,110 @@ public Map<String, Object> handleRequest(
...
 
-    private Result<UnauthorizedException, Void> verifySignature(SignedJWT signedJWT) {
-        var failure = Result.<UnauthorizedException, Void>failure(new UnauthorizedException());
+    private Optional<JWTClaimsSet> validateToken(
+            SignedJWT signedAccessToken, JWTClaimsSet claimsSet, String httpMethod) {
+        if (!verifySignature(signedAccessToken)) {
+            return Optional.empty();
+        }
+
+        return Optional.of(claimsSet)
```
BeckaL: So this logic works, but I find the use of filtering options here to validate a bit unidiomatic (as opposed to using results or doing some quite boring early returns with logging) - maybe we can discuss

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
BeckaL: Ah sorry I see this is updated to something different later on

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
alhcomer: I agree, I just really wanted a nice contained claims check block. I was hoping you might have some ideas on how to get things looking nice with Result or something :)

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
alhcomer: My thinking was that scope checking is authorizer logic and all of it can be contained in a single lambda handler, and that this leaves the endpoint lambdas containing only functional passkeys logic. Would be helpful to discuss

BeckaL: Yeah let's chat:

Thoughts:

* we might add more endpoints in to the api, so the current logic around http methods will be flimsy at that point
* we already split logic for the authorizer between api endpoints and this handler (logic for validating the principal id). Fundamentally there are some things we can validate in here, and some things (to do with request paths) that sit better in the api endpoints and I think this is OK

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
alhcomer: added in another check to make sure it's there

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -63,14 +65,20 @@ public IamPolicyResponseV1 handleRequest(
                             .flatMap(success -> verifySignature(signedAccessToken))
                             .flatMap(success -> validateClaimsSet(claimsSet));
 
-            if (validatedClaimsResult.isFailure()) {
-                throw validatedClaimsResult.getFailure();
+            var methodArn = apiGatewayCustomAuthorizerEvent.getMethodArn();
```
BeckaL, line 68: Minor: This methodArn interrupts the flow of this method a bit since it's only used on line 81 - would be easier to read if it was defined closer to where it was used and this bit was only about the validations

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -144,6 +152,19 @@ private IamPolicyResponseV1 getAllowExecuteApiPolicyForSubject(
         return IamPolicyResponseV1.builder()
                 .withPolicyDocument(policyDocument)
                 .withPrincipalId(subject)
+                .withContext(Map.of("scope", scope))
                 .build();
     }
+
+    private Result<UnauthorizedException, JWTClaimsSet> validateScope(JWTClaimsSet claimsSet) {
```
BeckaL, line 159: Minor: I think now that we're just doing some basic validation here, this should go inside the validate claims method

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -63,14 +65,20 @@ public IamPolicyResponseV1 handleRequest(
...
+                            .flatMap(this::validateScope)
+                            .map(JWTClaimsSet::getSubject);
+
+            if (result.isFailure()) {
+                throw result.getFailure();
             }
 
-            var subject = validatedClaimsResult.getSuccess().getSubject();
-            var methodArn = apiGatewayCustomAuthorizerEvent.getMethodArn();
+            var scope = (String) claimsSet.getClaim("scope");
```
BeckaL, line 79: We should extract this string to a constant. Better still if the validation returns the validated scope rather than having to retrieve this separately but that a bit more of a rewrite

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -150,6 +150,11 @@ private Result<UnauthorizedException, JWTClaimsSet> validateClaimsSet(JWTClaimsS
             LOG.warn("Access Token issuer is invalid");
             return Result.failure(new UnauthorizedException());
         }
+        var expectedAudience = configurationService.getAuthToAccountDataApiAudience();
+        if (!claimsSet.getAudience().contains(expectedAudience)) {
```
BeckaL, line 154: Why contains and not equals here?

BeckaL (approved):
Minor things - I've tested and it works, so I'm happy if you'd like to ship and address these in a small follow up (or discuss and not address). Also happy to re-review if you'd like to push any changes

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
alhcomer: 'getAudience()' returns a 'List<String>'

BeckaL: Ah sorry was looking at the configuration service method not getAudience 🤦

PR #8337, opened 2026-05-15
BAU: Upgrade Gradle to 9.6.1 and turn on Configuration Cache

alhcomer:
## What

- Upgrade Gradle wrapper from 8.14.5 to 9.6.1
- Fix all deprecated API usages removed in Gradle 9
- Bump plugins for compatibility (docker-compose 0.17.21, Pact 4.7.0)
- Consolidate CVE dependency constraints into shared data structures for easier maintenance
- Remove dead Terraform Gradle tasks (no longer used by any CI workflow)
- Enable the [Gradle configuration cache](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:intro) with full CI reuse

### CI performance improvement

Compared against [PR #8631](https://github.com/govuk-one-login/authentication-api/pull/8631) (a typical source-only PR on Gradle 8.14.5, no configuration cache):

| Job | [Baseline](https://github.com/govuk-one-login/authentication-api/actions/runs/29755941838) | [This PR (CC reuse)](https://github.com/govuk-one-login/authentication-api/actions/runs/29804531977/attempts/2) | Improvement |
|-----|---|---|---|
| build | 70s | 27s | -61% |
| run-unit-tests | 256s | 30s | -88% |
| run-integration-tests-api | 223s | 97s | -56% |
| run-integration-tests-utils | 248s | 107s | -57% |
| run-integration-tests-orch | 208s | 105s | -50% |
| run-integration-tests-services | 144s | 107s | -26% |
| run-account-management-and-DR | 200s | 97s | -52% |
| run-account-data-api | 116s | 108s | -7% |
| **Wall-clock** | **~7m50s** | **~4m41s** | **-40%** |
| **Cumulative compute** | **27m32s** | **14m15s** | **-48%** |

### How configuration cache reuse works across PRs

The configuration cache is keyed on build file hashes (`build.gradle`, `gradle.properties`, `libs.versions.toml`, `settings.gradle`). Once this PR merges:

1. The first CI run on main stores CC entries for each job
2. Any subsequent PR that only changes source code (`.java`, tests, configs, templates) shares the same cache key, so it gets CC reuse on its first run with no warm-up needed
3. CC entries are only invalidated when someone changes build files, e.g. when applying dependency constraints
4. After a build file change, the first run stores new entries automatically and all following runs reuse them

This means the ~40% wall-clock reduction applies to the majority of day-to-day PRs without any manual intervention.

### Configuration cache setup

- `GRADLE_ENCRYPTION_KEY` repository secret (already added)
- Per-job `actions/cache` step for `.gradle/configuration-cache/` keyed on build file hashes
- `cache-read-only: false` on all jobs so each persists its own complete Gradle Home cache

## How to review

2. Check GHA job stack traces to check only intended behaviour has changed e.g. `Reusing configuration cache.`
3. It works if CI passses


.github/workflows/pre-merge-checks.yml
```diff
@@ -53,6 +54,9 @@ jobs:
         uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v4
         with:
           gradle-version: wrapper
+          develocity-injection-enabled: false
+          add-job-summary: never
+          dependency-graph: disabled
```
Joe-Edwards-GDS: Are any of these non-default values? Can we add a comment to explain why we've changed them?

.github/workflows/pre-merge-checks.yml
```diff
@@ -53,6 +54,9 @@ jobs:
         uses: gradle/actions/setup-gradle@50e97c2cd7a37755bbfafc9c5b7cafaece252f6e # v4
```
Joe-Edwards-GDS, line 54: It looks like the tag comment is out of sync with the tag itself - probably because it's not got the full version number and dependabot doesn't understand it :(

This is actually v6.1.0, which introduces new caching behaviour - https://blog.gradle.org/choice-clarity-future-caching-gradle-actions

@alex9smith do you know if there's a programme position on this? This version changes the default caching behaviour to use a closed-source implementation/licence. It's free to use for public repos (but only a 'free preview' for private repos).

We can avoid it by using `cache-provider: basic` if necessary...

.github/workflows/pre-merge-checks.yml
```diff
@@ -106,6 +113,12 @@ jobs:
             ${{ runner.os }}-build-${{ github.event.pull_request.base.sha }}
             ${{ runner.os }}-build-
 
+      - name: Cache Gradle configuration cache
+        if: needs.check-changed-files.outputs.java_changed == 'true'
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: .gradle/configuration-cache/
+          key: ${{ runner.os }}-cc-style-checks-${{ hashFiles('build.gradle', 'gradle.properties', 'gradle/libs.versions.toml', 'settings.gradle') }}
```
Joe-Edwards-GDS: How does this differ from the built-in cache in the `setup-gradle` action?

The docs suggest that it will cache everything in `.gradle` (i.e. including the configuration-cache): https://docs.gradle.org/current/userguide/github-actions.html#github_actions_caching
> setup-gradle automatically saves and restores the Gradle User Home (~/.gradle), which contains the Gradle distribution, resolved dependencies, and the local build cache. A cache entry is written at job completion when the job is a push to the default branch; other jobs restore from existing entries.

If you're seeing a significant speedup then it must be working, I'm just a bit surprised we need to do this manually.

Joe-Edwards-GDS: Is this key sufficient? Do we also need to key on `build.gradle` within every subproject (i.e. `**/build.gradle`) as well?

gradle/wrapper/gradle-wrapper.properties
Joe-Edwards-GDS, line 1: Should the wrapper script have changed as well? Did you follow the upgrade instructions (https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper)

.github/workflows/pre-merge-checks.yml
alhcomer: Gradle _did_ have support for it, but removed it per this [blog post](https://blog.gradle.org/github-actions-for-gradle-v6#:~:text=The%20only%20functional%20change%20in%20v6%20is%20the%20removal%20of%20the%20experimental%20Configuration%20Cache%20support.%20It%20was%20not%20production%2Dready%20and%20worked%20only%20for%20a%20limited%20set%20of%20projects.%20We%E2%80%99re%20reworking%20this%20feature%20so%20that%20it%20works%20for%20most%20projects%20and%20more%20users%20can%20benefit%20(more%20on%20this%20below).) and the [changelog](https://github.com/gradle/actions/releases#:~:text=Existing%2C%20rudimentary%2C%20configuration%2Dcache%20support%20has%20been%20removed%2C%20pending%20a%20fully%20functional%20implementation%20in%20gradle%2Dactions%2Dcaching.) due to not being ready. There hasn't been a mention of the configuration cache since then in the changelog

alhcomer: You're right, it should key on the subprojects as well. Thanks Joe

.github/workflows/pre-merge-checks.yml
alhcomer: That tag comment was confusing... will fix 

As for the programme position, if we don't want a closed-source implementation, we need to turn it off. ['enhanced' is the default ](https://github.com/gradle/actions/blob/50e97c2cd7a37755bbfafc9c5b7cafaece252f6e/setup-gradle/action.yml#L18)

gradle/wrapper/gradle-wrapper.properties
alhcomer: This has been added

.github/workflows/pre-merge-checks.yml
alhcomer: Changed it to hash on '**/build.gradle' instead of just the root build.gradle

.github/workflows/pre-merge-checks.yml
alhcomer: I've actually removed these, they were a hangover from experimenting for a problem where the solution was adding the encryption key

.github/workflows/pre-merge-checks.yml
Joe-Edwards-GDS: I'll delegate to @alex9smith on whether it's allowed.

In any case, would be nice to fix the tag comments if we can!

.github/workflows/pre-merge-checks.yml
Joe-Edwards-GDS: Cool, thanks - hopefully we can switch back to the built-in at some point, but if it's saving minutes per build then worth doing now :)

Joe-Edwards-GDS (dismissed):
Looks like a good improvement to build time - nice one! 😄

I think it's worth checking with Alex on the cache licence before merging, and maybe getting eyes from another dev on the build.gradle refactoring (if nothing else, just so they know it's coming!) but nothing major from me.

dbes-gds (approved):

.github/workflows/pre-merge-checks.yml
CarlyG55: @alhcomer would we be able to work this out before merging? 🙂 if you want it in before the end of the week, could ask Andrew Moores for a lead dev opinion?

Joe-Edwards-GDS: I got a view from @danielmason-gds that it was probably fine for now - the updated licence is much clearer about data ownership etc.

It's not completely ideal (and we might have to change if/when they attempt to monetise) but no hard blockers at the moment.

CarlyG55 (approved):

PR #8346, opened 2026-05-18
ATO-2537: add feature flag to stop publishing old external token key …

Louisasa:
…(while still supporting the old key for reauth using the stored public key)

### Wider context of change

To decommission the old keys, we first want to stop publishing them to the jwks endpoint.

### What’s changed

Adds a feature flag to publish the old keys to the jwks endpoint. Note: I did add a funky thing where if using the stored public key isn't enabled, we do publish the key anyway. I'm not sure if that's the way round we want but wanted to ensure reauth doesn't get accidentally broken.

### Manual testing

- Stored old public key enabled and publishing enabled - keys show at endpoint
- Stored old public key enabled and publishing disabled - keys don't show at endpoint
- Stored old public key disabled and publishing enabled - keys show at endpoint
- Stored old public key disabled and publishing disabled - keys show at endpoint
Also checked a reauth journey still works
Also also tested rollback plan works

### Checklist

- [x] Added new endpoints to local running (`LocalOrchestrationApi.java`) or not required. **N/A**

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
@@ -64,11 +64,14 @@ public class JwksHandler
 
             List<JWK> signingKeys = new ArrayList<>();
 
-            signingKeys.add(jwksService.getPublicTokenJwkWithOpaqueId());
             signingKeys.add(jwksService.getPublicDocAppSigningJwkWithOpaqueId());
 
-            if (configurationService.isRsaSigningAvailable()) {
-                signingKeys.add(jwksService.getPublicTokenRsaJwkWithOpaqueId());
+            if (configurationService.isPublishOldExternalTokenSigningKeysEnabled()) {
+                signingKeys.add(jwksService.getPublicTokenJwkWithOpaqueId());
+
+                if (configurationService.isRsaSigningAvailable()) {
+                    signingKeys.add(jwksService.getPublicTokenRsaJwkWithOpaqueId());
+                }
             }
 
             if (configurationService.isPublishNextExternalTokenSigningKeysEnabledV2()) {
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenValidationService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenValidationService.java
@@ -132,8 +132,8 @@ public class TokenValidationService {
     }
 
     private boolean validateWithOldECPublicKey(SignedJWT jwt) throws JOSEException {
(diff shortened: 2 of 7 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
```diff
@@ -64,11 +64,15 @@ public APIGatewayProxyResponseEvent jwksRequestHandler(
 
             List<JWK> signingKeys = new ArrayList<>();
 
-            signingKeys.add(jwksService.getPublicTokenJwkWithOpaqueId());
             signingKeys.add(jwksService.getPublicDocAppSigningJwkWithOpaqueId());
 
-            if (configurationService.isRsaSigningAvailable()) {
-                signingKeys.add(jwksService.getPublicTokenRsaJwkWithOpaqueId());
+            if (configurationService.isPublishOldExternalTokenSigningKeysEnabled()
+                    || !configurationService.isUseStoredOldIdTokenPublicKeysEnabled()) {
```
cearl1: I find this logic a bit confusing as the new feature flag for publishing keys can be false but we will still publish keys. Could we rely on us setting the environment up for useStoredOldIdTokenPublicKeys so we don't need this fallback logic? Or maybe have a log message to say we're falling back to needing an old public key on the endpoint?

cearl1: We could move this logic into the getter for `isPublishOldExternalTokenSigningKeysEnabled()`, although thats sort of moving the problem elsewhere haha

Louisasa: Yeah I wasn't sure if we should fall back to publishing them or using the stored old public keys, I'll chat to Carly and ask

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenValidationService.java
```diff
@@ -145,6 +145,7 @@ private boolean validateWithOldECPublicKey(SignedJWT jwt) throws JOSEException {
             return optionalPublicKey.isPresent()
                     && jwt.verify(new ECDSAVerifier(optionalPublicKey.get()));
         } else {
```
cearl1, line 147: Could we move the `isPublishOldExternalTokenSigningKeysEnabled()` check onto this `else if` bit? Then have another else that just returns false with a log message saying "No key found to validate" or something

cearl1: I'm probably overthinking this, the publish key feature flag wont stick around for long

cearl1 (dismissed):

cearl1 (approved):

PR #8353, opened 2026-05-19
AUT-5359: Send confirmation email to user when they successfully delete their passkey

ChristianSoftwire:
## What

- When a user deletes a passkey on their account, we want to send an email that they deleted it. This email will vary depending on if there are still passkeys left on their account after the deletion.
- In order to find the number of passkeys on a users account we are using the get passkeys endpoint from the account-data-api.
- However, if we retrieve the passkeys after a deletion, there is a risk that the get request will fail and we won't send the email but will have deleted the passkey. Therefore, I have flipped the order of the requests. So we will first get the users passkeys, store the count of passkeys in a context record, then after deletion we will use the count to decide what email to send to the user. This means that if the retrieve fails, we don't delete or send an email to the user.

## How to review

1. Code review
2. Code deploy and test you receive the email (will require a passkey on your account) - can be tested through the console


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
```diff
@@ -67,16 +109,107 @@ public APIGatewayProxyResponseEvent passkeyDeleteProxyHandler(
...
+                    "Email notification not sent because delete passkey response was {}",
+                    context.deletePasskeyResponse.statusCode());
+            return Result.success(context);
+        }
+
+        var userProfile =
+                dynamoService.getOptionalUserProfileFromPublicSubject(context.publicSubjectId);
+        if (userProfile.isEmpty()) {
+            LOG.warn("No user profile found for public subject ID '{}'", context.publicSubjectId);
+            return Result.success(context);
```
csophocleous: Not sure if this is the right call, but I have chosen to not return a 500 in the case where the userProfile is not found. Instead, we just skip sending the email entirely and continue through the pipeline and return the response from ADAPI

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/NotificationHandler.java
```diff
@@ -48,6 +48,9 @@ public class NotificationHandler implements RequestHandler<SQSEvent, Void> {
     private static final Logger LOG = LogManager.getLogger(NotificationHandler.class);
     public static final String VALIDATION_CODE_PERSONALISATION = "validation-code";
     public static final String EMAIL_ADDRESS_PERSONALISATION = "email-address";
+    public static final String HAS_PASSKEYS_REMAINING_PERSONALISATION = "has_passkeys_remaining";
+    public static final String DOES_NOT_HAVE_PASSKEYS_REMAINING =
```
BeckaL: Minor: should this have _PERSONALISATION suffix for consistency?

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
```diff
@@ -67,16 +109,107 @@ public APIGatewayProxyResponseEvent passkeyDeleteProxyHandler(
...
+        sqsClient.send(serializationService.writeValueAsString((notifyRequest)));
+        LOG.info(
+                "Notify request sent with notification type '{}' and publicSubjectId '{}'",
+                notificationType,
+                context.publicSubjectId);
+
+        return Result.success(context);
+    }
+
+    private record PasskeysDeleteProxyContext(
```
BeckaL: I find this object mixes concerns a bit, containing both the input and some outcomes. I'd rather have these separate even if it makes the chaining a bit less nice.

BeckaL: I've got an example branch [here](https://github.com/govuk-one-login/authentication-api/commit/170c045ad01a94b4669217e40f45805d59e621cd) where I've rewritten some of the logic in this handler - see what you think.

Key bits are:
* Create a request context object that doesn't mix any concerns and can be used where necessary
* Move the lookup of the passkey count into the send notification function - in my head it's an internal concern of this sending logic and can be hidden away in there, so that reading the core logic of the function is simpler. Also doing the lookup here means we don't have to remember to -1 from the count.
* The method that validates the request doesn't actually return any errors, so this renamed to reflect this, and the signature is changed from a Result<Foo, Bar> to Bar.
* Not integral to the changes I'm suggesting here, which is mainly about removing the PasskeysDeleteProxyContext and simplifying stuff a bit, but I'm not sure if we should fail the request if the notification sending fails - after all, we have still deleted the passkey, so not sure what we should do in this case.

The logic in the main handler is a bit more ploddy, but I think it's more straightforward to understand and less surprising in terms of what's going on.

account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/PasskeysDeleteProxyHandler.java
BeckaL: Hmm yeah I think we should clarify the behaviour here with someone - and then make it consistent for any failure to send the notification (count retrieval failure, failure from notify)

csophocleous: Checked with UCD and product and the decision is to fail the entire delete passkey request if we can't send the comms

BeckaL (dismissed):
LGTM!

BeckaL (approved):

PR #8360, opened 2026-05-19
Bau: assert on full audit events in send notification handler tests

BeckaL:
Provides a more solid base for refactoring.


ChristianSoftwire (approved):
LGTM - thanks for doing that

PR #8362, opened 2026-05-19
ATO-2559: Add feature flag for defaulting tokenAuthMethod

cearl1:
### Wider context of change

We previously allowed for tokenAuthMethod to be null, and treated `private_key_jwt` as the default in the `PrivateKeyJwtClientAuthValidator`. To make it a bit clearer it would be nice to move this defaulting into the client registry getter.

### What’s changed

This PR adds a feature flag (enabled in dev, build, and staging) to default the tokenAuthMethod to `private_key_jwt` if it is null. This is done in a utility class and the method is used in all places where we access the tokenAuthMethod. I also added some tests for the utility class.

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
@@ -470,6 +470,10 @@ public class ConfigurationService implements BaseLambdaConfiguration, AuditPubli
         return Long.parseLong(System.getenv().getOrDefault("SYNC_WAIT_FOR_SPOT_TIMEOUT", "5000"));
     }
 
+    public boolean isUseDefaultTokenAuthMethod() {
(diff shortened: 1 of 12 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
```diff
@@ -470,6 +470,10 @@ public long getSyncWaitForSpotTimeout() {
         return Long.parseLong(System.getenv().getOrDefault("SYNC_WAIT_FOR_SPOT_TIMEOUT", "5000"));
```
Louisasa, line 470: bit weird but the first commit mentions conflicts with template.yaml lol? I assume a bad rebase

cearl1: Ahh yep that was conflicting with your changes 😄 I'll remove that comment from the commit

Louisasa (dismissed):

Louisasa (approved):

PR #8366, opened 2026-05-20
BAU: Support multiple scopes in ADAPI token

ben-fernandes-sw:
## What

In some cases, we need to generate ADAPI tokens with multiple scopes (e.g. a token might need the ability to read and write passkeys). Previously, our token generation helper method only allowed you to set one - now it allows multiple. The scopes are passed around as a `List` until the point when they're written to the token body as a space-separated `String`.

This commits can be roughly grouped like this:
* 8374ac071a6098a0544bbf812fc4e83e3c95c93a
  * Supporting generation with multiple scopes
* 701b71d1435133109a39c46ce7a5f26c19fb7643, 9203d7dcf7c672ea14e5c492e7f60af053e1b71f, 611902dd72f17dc6213cdc25dc688adb7585b4fb
  * Updating places where we generate tokens to make sure they're generated with the correct scopes
* b1abcc63a3317eba1a92880987adb1992faa20de, d1b69bb6a9d9c2a6e04cd238773b3713d7fced42
  * Validating the scopes correctly in the ADAPI
* e077a34b70a643a853620fb9182ca53db519a69d
  * Updating the ADAPI token generator script

## How to review

1. Code Review
2. Test end-to-end journeys with AMC/Home once it reaches Staging

### DIFF
--- a/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/ScopeAuthorizerHelper.java
+++ b/account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/ScopeAuthorizerHelper.java
@@ -5,6 +5,8 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.authentication.shared.entity.AccountDataScope;
 
+import java.util.Arrays;
+
 public class ScopeAuthorizerHelper {
     private static final Logger LOG = LogManager.getLogger(ScopeAuthorizerHelper.class);
     private static final String SCOPE_FIELD = "scope";
@@ -20,7 +22,7 @@ public class ScopeAuthorizerHelper {
             return false;
         }
         var scope = requestContext.getAuthorizer().get(SCOPE_FIELD).toString();
-        var matches = expectedScope.getValue().equals(scope);
+        var matches = Arrays.asList(scope.split(" ")).contains(expectedScope.getValue());
         if (!matches) {
             LOG.warn("Scope {} does not match expected scope {}", scope, expectedScope.getValue());
         }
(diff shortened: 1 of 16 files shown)


account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/ScopeAuthorizerHelper.java
```diff
@@ -20,7 +20,7 @@ public static boolean isScopeAuthorized(
             return false;
         }
         var scope = requestContext.getAuthorizer().get(SCOPE_FIELD).toString();
-        var matches = expectedScope.getValue().equals(scope);
+        var matches = java.util.Arrays.asList(scope.split(" ")).contains(expectedScope.getValue());
```
alhcomer: could we remove the java.util.Arrays import chain here?

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandler.java
```diff
@@ -192,4 +197,20 @@ private IamPolicyResponseV1 getAllowExecuteApiPolicyForSubject(
...
     }
+
+    private Result<UnauthorizedException, Void> validateScope(String scopeValue) {
+        if (scopeValue == null || scopeValue.isBlank()) {
+            LOG.warn("Invalid or missing scope: {}", scopeValue);
+            return Result.failure(new UnauthorizedException());
+        }
+        var allValid =
+                java.util.Arrays.stream(scopeValue.split(" "))
+                        .allMatch(s -> AccountDataScope.fromValue(s).isPresent());
```
alhcomer: do we want to keep validating beyond just checking that the scope is present here now? If an unrecognised scope is sent alongside otherwise valid scopes, should we be erroring there, or just ignore it? Considering we're now validating more than a single scope and we do full validation in the handlers, could it be acceptably more flexible for future adapi development to ignore? Kind of like accepting a request that has extra fields that are not yet used

Not necessarily advocating for either way

ben-fernandes-sw: Have removed

account-data-api/src/main/java/uk/gov/di/authentication/accountdata/helpers/ScopeAuthorizerHelper.java
ben-fernandes-sw: Have replaced with an import at the top of the file

alhcomer (approved):

PR #8371, opened 2026-05-21
ATO-2575: remove new v2 key feature flags

Louisasa:
### Wider context of change

As part of the key rotation, the new V2 keys have been successfully rotated and no longer need to be behind a feature flag.

### What’s changed

Remove UseV2Key, PublishV2Key, and ProvisionV2Key feature flags, as well as tidy up fetching the old public key for non-reauth journeys.

### Manual testing

Tested in dev

### Checklist

- [x] Added new endpoints to local running (`LocalOrchestrationApi.java`) or not required. **N/A**

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
@@ -74,12 +74,10 @@ public class JwksHandler
                 }
             }
 
-            if (configurationService.isPublishNextExternalTokenSigningKeysEnabledV2()) {
-                signingKeys.add(jwksService.getNextPublicTokenJwkWithOpaqueIdV2());
+            signingKeys.add(jwksService.getNextPublicTokenJwkWithOpaqueIdV2());
 
-                if (configurationService.isRsaSigningAvailable()) {
-                    signingKeys.add(jwksService.getNextPublicTokenRsaJwkWithOpaqueIdV2());
-                }
+            if (configurationService.isRsaSigningAvailable()) {
+                signingKeys.add(jwksService.getNextPublicTokenRsaJwkWithOpaqueIdV2());
             }
 
             JWKSet jwkSet = new JWKSet(signingKeys);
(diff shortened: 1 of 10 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/TokenValidationService.java
```diff
@@ -63,27 +63,17 @@ public boolean isTokenSignatureValid(String tokenValue) {
             if (JWSAlgorithm.RS256 == jwt.getHeader().getAlgorithm()
                     && configuration.isRsaSigningAvailable()) {
                 if (configuration.isPublishNextExternalTokenSigningKeysEnabledV2()) {
-                    var oldPublicKey = jwksService.getPublicTokenRsaJwkWithOpaqueId();
```
Louisasa, line 66: I made this change because I suddenly panicked about removing the old keys from the JWKs endpoint would break fetching them, but this actually gets it straight from KMS so wouldn't break, but it isn't needed anymore. Happy to remove from this PR tho as it isn't entirely related

arthurs-gds: Yeah, I lean toward thinking that's better done as BAU or with a fresh ticket.

Louisasa: I'm kind of tempted to leave it in though so it doesn't get missed. It's not entirely unrelated but I can update the commit to at least be BAU?

arthurs-gds: Yeah, I think at least split that change out into a separate commit--without your comment I would have been really confused about how that change connected to everything else.

arthurs-gds (approved):

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/JwksHandler.java
```diff
@@ -74,12 +74,10 @@ public APIGatewayProxyResponseEvent jwksRequestHandler(
...
             }
 
-            if (configurationService.isPublishNextExternalTokenSigningKeysEnabledV2()) {
-                signingKeys.add(jwksService.getNextPublicTokenJwkWithOpaqueIdV2());
+            signingKeys.add(jwksService.getNextPublicTokenJwkWithOpaqueIdV2());
 
-                if (configurationService.isRsaSigningAvailable()) {
-                    signingKeys.add(jwksService.getNextPublicTokenRsaJwkWithOpaqueIdV2());
-                }
+            if (configurationService.isRsaSigningAvailable()) {
```
Ryan-Andrews99, line 79: Not a blocker, but I'd like to remove this feature flag at some point as I'm sure it's enabled in all envs!

Ryan-Andrews99 (approved):
Looks good to me, added one comment on removing a redundant feature flag - but that's a later problem!

PR #8393, opened 2026-05-29
AUT-5370: Fix TICF partially created account state

sw-mattw:
## What

- Added an `isPartiallyCreatedAccount` attribute to `AuthSessionItem` to track users resuming incomplete registration journeys.
- Updated `LoginHandler` to flag sessions as partially created when a user logs in but their `mfaMethodVerified` status is false.
- Modified `AccountInterventionsHandler` to send `AccountState.NEW` to the TICF CRI for partially created accounts, resolving the issue where they were incorrectly recorded as `EXISTING`.

## How to review

1. Code review.

## Checklist

- [ ] Deployment of this PR will not break active user journeys **- Should be ok: the new `isPartiallyCreatedAccount` session attribute defaults to false and does not invalidate existing session data**
- [ ] No changes required or changes have been made to stub-orchestration. **- N/A**
- [ ] A UCD review has been performed. **- N/A**


BeckaL (approved):

sw-mattw:
Ran through journeys for:
- Standard sign in
- Partial account creation, then resuming to set up MFA
- Standard account creation (full end to end, rather than partial creation)

PR #8404, opened 2026-06-01
Bau: Stop setting metadata on audit context in Mfa Methods Put Handler

BeckaL:
## What

Switches all events emitted in the mfa methods put handler over to pass metadata items in as varargs rather than setting them on the audit context.

This is a step towards removing the metadata items entirely from the audit context - currently it is confusing as they can be set in either place, but the audit context is supposed to contain only data that is common to all events emitted by a handler (and can just be created once per handler) rather than any event-specific metadata. 

## How to review

1. Code Review commit by commit - I've done this refactor very incrementally so it should be easy to see what's going on when going through each commit individually.

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandler.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandler.java
@@ -15,6 +15,7 @@ import uk.gov.di.accountmanagement.helpers.PrincipalValidationHelper;
 import uk.gov.di.accountmanagement.services.AwsSqsClient;
 import uk.gov.di.accountmanagement.services.CodeStorageService;
 import uk.gov.di.accountmanagement.services.MfaMethodsMigrationService;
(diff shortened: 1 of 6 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/lambda/MFAMethodsPutHandler.java
```diff
@@ -626,6 +640,10 @@ private AuditContext buildAuditContext(
                         : AuditService.UNKNOWN;
         var context = baseContext.withPhoneNumber(phoneNumber);
 
+        if (auditEvent.equals(AUTH_MFA_METHOD_SWITCH_COMPLETED)) {
```
BeckaL, line 643: A note to help anyone reviewing follow this refactor: The idea here is to go leave the bulk of this method intact, and as we switch over each event add them to the condition here so that we short-circuit any of the following logic to add metadata pairs to the audit context. This condition, and the majority of the code in this method, is completely removed in a subsequent commit in this PR.

alhcomer (approved):

PR #8410, opened 2026-06-02
Aut 5462: Add AMC Authorisation Initiated Event

BeckaL:
## What

Emit a new audit event when we're about to hand off to amc for passkeys. This is emitted just before we return to the frontend with the amc redirect url. It won't be emitted in the case of any failures to construct the redirect url (signing failures, jwks retrieval errors etc).

## How to review

1. Deploy to a dev env, go through a sign in journey where you're prompted to set up a passkey, choose to set up a passkey and then see the audit events emitted in sqs (you can also use the new script in this repo)

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AMCAuthorizeHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AMCAuthorizeHandler.java
@@ -25,8 +25,11 @@ import uk.gov.di.authentication.frontendapi.errormapper.AMCFailureHttpMapper;
 import uk.gov.di.authentication.frontendapi.services.AMCService;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
+import uk.gov.di.authentication.shared.entity.JourneyType;
 import uk.gov.di.authentication.shared.entity.Result;
+import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
+import uk.gov.di.authentication.shared.helpers.PersistentIdHelper;
 import uk.gov.di.authentication.shared.lambda.BaseFrontendHandler;
 import uk.gov.di.authentication.shared.serialization.Json;
 import uk.gov.di.authentication.shared.services.AccessTokenConstructorService;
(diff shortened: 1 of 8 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AMCAuthorizeHandler.java
```diff
@@ -190,8 +190,16 @@ private void emitAuthorizationRequestedAuditEvent(
                         AuditService
                                 .UNKNOWN, // the schema does not include phone number for this event
                         PersistentIdHelper.extractPersistentIdFromHeaders(input.getHeaders()));
+
+        var amcScopeForAuditEvent =
+                switch (request.amcJourneyType()) {
+                    case PASSKEY_CREATE -> "passkey-create";
+                    case SFAD -> "sfad"; // note we'll probably want to review this when we get to
+                        // implementing sfad fully
+                };
```
ben-fernandes-sw, line 199: Would it be crazy to bake this into the enum, rather than having to do a switch/case here?

integration-tests/src/test/java/uk/gov/di/authentication/api/AMCAuthorizeHandlerIntegrationTest.java
```diff
@@ -50,7 +53,12 @@
...
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static uk.gov.di.authentication.frontendapi.domain.FrontendAuditableEvent.AUTH_AMC_AUTHORISATION_REQUESTED;
+import static uk.gov.di.authentication.sharedtest.helper.AuditAssertionsHelper.assertNoTxmaAuditEventsReceived;
+import static uk.gov.di.authentication.sharedtest.helper.AuditAssertionsHelper.assertTxmaAuditEventsReceived;
 import static uk.gov.di.authentication.sharedtest.matchers.APIGatewayProxyResponseEventMatcher.hasStatus;
+import static uk.gov.di.authentication.testsupport.AuditTestConstants.EXTENSIONS_AMC_SCOPE;
+import static uk.gov.di.authentication.testsupport.AuditTestConstants.EXTENSIONS_JOURNEY_TYPE;
 
 @ExtendWith(SystemStubsExtension.class)
 class AMCAuthorizeHandlerIntegrationTest extends ApiGatewayHandlerIntegrationTest {
```
ben-fernandes-sw, line 64: How about having a test in here that validates the actual JSON output of the event against what we expect.

BeckaL: So we don't tend to do this because the full event contains some values that are hard to determine in advance (timestamps etc). The way I've done it in here is an existing pattern, although I kind of don't love it. Is it OK if we leave it and on the course of my continued adventures in audit events I'll have a think if there's a better way?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/AMCAuthorizeHandler.java
BeckaL: It's quite metadata pair specific, I don't know whether to pollute the enum with metadata-specific knowledge. Can we leave this for now, and as part of implementing the other audit events I'll work out where to move if this makes sense to be shared

ben-fernandes-sw (approved):

PR #8414, opened 2026-06-03
BAU: config service for account data api

BeckaL:
## What

Adds a ConfigurationService for the account data api, and removes any dependency on the shared configuration service in the account data api.

This is useful as:
* a step towards fully breaking the dependency of the account data api on shared, in order to eventually move it to its own repo
* a way of simplifying the configuration service and making it clear which environment variables are used in the account data api.

## How to review

1. Code Review


BeckaL:
Proposing to ignore the code coverage complaint here - mostly complaining about the non-unit tested BaseDynamoService, but this is in fact tested via the integration tests for the handlers / the passkeys service, so I'm dismissing as a false positive

dbes-gds (approved):

PR #8422, opened 2026-06-05
Aut 5465: Add passkey authentication generated event

BeckaL:
Adds the passkey authentication generated event as the last thing we do in the start passkey assertion handler before returning a response.

Notes:

* we have not yet set the user verification on the passkey request. The event is currently populated from this, so this will be set to an empty string. Future work will set this.

## How to review

1. Code review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
@@ -63,7 +63,7 @@ public class AccountDataCredentialRepository implements CredentialRepository {
 
     Set<AuthenticatorTransport> transportStringsToSet(List<String> passkeyTransports) {
         return passkeyTransports.stream()
-                .map(AuthenticatorTransport::valueOf)
+                .map(s -> AuthenticatorTransport.valueOf(s.toUpperCase()))
                 .collect(Collectors.toSet());
     }
 
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
@@ -2,6 +2,7 @@ package uk.gov.di.authentication.frontendapi.services.webauthn;
 
 import com.yubico.webauthn.CredentialRepository;
 import com.yubico.webauthn.RegisteredCredential;
+import com.yubico.webauthn.data.AuthenticatorTransport;
 import com.yubico.webauthn.data.ByteArray;
 import com.yubico.webauthn.data.PublicKeyCredentialDescriptor;
 import com.yubico.webauthn.data.PublicKeyCredentialType;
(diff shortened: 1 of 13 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartPasskeyAssertionHandlerTest.java
```diff
@@ -104,6 +130,12 @@ void shouldReturn200ForValidRequest() throws Exception {
                                     session ->
                                             session.getPasskeyAssertionRequest()
                                                     .equals(expectedJson)));
+
+            verify(auditService)
+                    .submitAuditEvent(
```
ben-fernandes-sw, line 191: Do you think it'd be useful to test the counter-case too? e.g. When an error was encountered, `submitAuditEvent` was not called with `AUTH_PASSKEY_AUTHENTICATION_GENERATED`.

ben-fernandes-sw (dismissed):

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartPasskeyAssertionHandlerTest.java
ben-fernandes-sw: Will add in follow-up PR

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
```diff
@@ -59,6 +61,12 @@ private Optional<PublicKeyCredentialDescriptor> toCredentialDescriptor(
         }
     }
 
+    Set<AuthenticatorTransport> transportStringsToSet(List<String> passkeyTransports) {
+        return passkeyTransports.stream()
+                .map(s -> AuthenticatorTransport.valueOf(s.toUpperCase()))
```
BeckaL, line 66: At the moment this will throw an exception if the authenticator transport is not valid. We should perhaps handle that more gracefully - I'll do in a follow up if we decide what the behaviour should be

ben-fernandes-sw: These transport values are fixed in the spec
https://www.w3.org/TR/webauthn/#enumdef-authenticatortransport
so I think it's probably fine to throw if it's not valid.

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepositoryTest.java
```diff
@@ -236,7 +240,7 @@ private static PasskeysRetrieveResponse.PasskeyResponse aPasskeyResponse(String
                 "some-aaguid",
                 true,
                 5,
-                List.of("BLE"),
+                List.of("BLE", "internal"),
```
ben-fernandes-sw, line 243: Is the casing correct here?

BeckaL: So the idea here was to allow the transports to be transformed regardless of case, so this was intentional

ben-fernandes-sw (approved):

PR #8423, opened 2026-06-05
BAU: Explicitly unset env vars in test using system stub

BeckaL:
Means real env vars can't get in the way here


alhcomer (approved):

Joe-Edwards-GDS (approved):
Looks like a more robust test to me (and should fix the issue) - they're probably not super high-value tests, but will let you and the auth team decide whether to keep long-term. 🙂

PR #8430, opened 2026-06-08
BAU: Replace Optional<String> with String for txmaAuditEncoded in AuditContext

alhcomer:
## What

- Remove Optional wrapping from txmaAuditEncoded in AuditContext
  and all call sites
- Use AuditService.UNKNOWN as for missing values, consistent with
  all other audit context fields
- Move warning logs from AuditService event emission to point of
  header extraction in TxmaAuditHelper and AuditHelper
- Rename methods to getTxmaAuditEncodedHeaderOrUnknown to signal
  that null is never returned
- Deduplicate txma audit header extraction in account management, use existing shared functionality instead

## How to review

1. Code Review

### DIFF
--- a/shared/src/main/java/uk/gov/di/authentication/shared/helpers/TxmaAuditHelper.java
+++ b/shared/src/main/java/uk/gov/di/authentication/shared/helpers/TxmaAuditHelper.java
@@ -1,18 +1,32 @@
 package uk.gov.di.authentication.shared.helpers;
 
 import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyRequestEvent;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.shared.services.AuditService;
 
 import java.util.Optional;
 
 import static uk.gov.di.authentication.shared.helpers.RequestHeaderHelper.getOptionalHeaderValueFromHeaders;
 
 public class TxmaAuditHelper {
+    private static final Logger LOG = LogManager.getLogger(TxmaAuditHelper.class);
     public static final String TXMA_AUDIT_ENCODED_HEADER = "txma-audit-encoded";
 
     private TxmaAuditHelper() {}
 
-    public static Optional<String> getTxmaAuditEncodedHeader(APIGatewayProxyRequestEvent request) {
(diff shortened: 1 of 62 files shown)


shared/src/main/java/uk/gov/di/authentication/shared/helpers/TxmaAuditHelper.java
```diff
@@ -1,18 +1,32 @@
...
+                        request.getHeaders(), TXMA_AUDIT_ENCODED_HEADER, false);
+        if (header.isEmpty()) {
+            LOG.warn("Encoded device information for audit event is not present.");
+            return AuditService.UNKNOWN;
+        }
+        if (header.get().isEmpty()) {
+            LOG.warn("Encoded device information for audit event present but empty.");
+            return AuditService.UNKNOWN;
+        }
+        return header.get();
```
csophocleous, line 30: Not really part of this but just curious why this isn't used in account-management?

shared/src/main/java/uk/gov/di/authentication/shared/services/AuditService.java
```diff
@@ -41,22 +38,14 @@ public AuditService(ConfigurationService configurationService) {
...
-                        txmaAuditEvent.addRestricted("device_information", Map.of("encoded", s));
-                    } else {
-                        LOG.warn("Encoded device information for audit event present but empty.");
-                    }
-                },
-                () -> LOG.warn("Encoded device information for audit event is not present."));
+        if (!txmaAuditEncoded.isEmpty()) {
+            txmaAuditEvent.addRestricted("device_information", Map.of("encoded", txmaAuditEncoded));
+        }
     }
```
csophocleous, line 49: nice cleanup

shared/src/main/java/uk/gov/di/authentication/shared/helpers/TxmaAuditHelper.java
alhcomer: oooh that's a good spot, it really should use it I think. will add a new commit to move things over to it

ChristianSoftwire (dismissed):
LGTM - I left a small question!

ChristianSoftwire (approved):
Lovely bit of cleanup!

PR #8431, opened 2026-06-09
BAU: Rename auth session verified attributes to be clearer

ChristianSoftwire:
## What

Rename the hasVerified<Authenticator> attributes to hasVerifiedWith<Authenticator> to make it clearer that the attributes mean the user has verified with a particular authenticator (rather than the user has a type of authenticator).

## How to review

1. Code review


ben-fernandes-sw (approved):

sw-mattw (changes requested):
Blocking current proposed changes due to use of hasVerifiedPassword and hasVerifiedMfa in existing live sessions

PR #8432, opened 2026-06-09
ATO-2607: Create Custom Cloudfront distribution stack

Ryan-Andrews99:
### Wider context of change

We would like to be able to point our Cloudfront at the API Gateway execute endpoint directly as a temporary step during our migration. This requires a couple of things, such as the ability to provide a custom API Gateway execute endpoint and use of an origin path, and the AllViewerExceptHostHeader managed origin request policy.

### What’s changed
- Imports the Cloudfront distribution template
- Makes the necessary changes for our use case
-  Updates our deployment script to be able to deploy this version of the stack instead

### Manual testing
- I deployed this to build as part of ATO-2491

### Checklist


ci/stack-orchestration/manual-stacks/cloudfront-distribution/template.yaml
```diff
@@ -192,6 +202,8 @@ Conditions:
           "none",
         ],
       ]
+  UseApiGatewayOrigin: !Not [!Equals [!Ref ApiGatewayOriginDomain, "none"]]
+  UseOriginPath: !Not [!Equals [!Ref OriginPath, "none"]]
```
cearl1, line 206: Is this condition being used at the moment? or are we going to be using it in the future?

Ryan-Andrews99: Ah good spot - I noticed this when deploying build- I've pushed a fix so we're using that condition properly now!

cearl1 (approved):

PR #8437, opened 2026-06-09
BAU: Add new verified authenticator attributes for clarity

ChristianSoftwire:
## What

Currently, we have `HasVerified<Authenticator>` attributes in the AuthSessionItem, but this naming isn't indicative of what this attribute is. We set these attributes when a user has authenticated with one of the authenticators (eg if they authenticate with a passkey we set `HasVerifiedPasskey` to true). A clearer name would be `HasVerifiedWith<Authenticator>`. 

This PR adds these new attributes and writes to them in the same way we were writing to the `HasVerified<Authenticator>` attributes. After approximately an hour after merging, we know that all of these attributes should align in all sessions (as session expiration is set to an hour). We can then remove the old attributes and use the new ones in our validation.

## How to review

1. Code review 
2. Deploy to a dev env, and check that at the end of a successful journey with a given authenticator, in the `orch-auth-code-lambda` you can see the `Checking if auth code is blocked` and the hasVerified<Authenticator> and hasVerifiedWith<Authenticator> attributes align


BeckaL (approved):
Looks good to me, checked out the code and couldn't see any places where we're setting the old values without setting the new ones alongside them 👍

PR #8439, opened 2026-06-09
ATO-2632: Rejig feature flags

Ryan-Andrews99:
### Wider context of change

During our API migration, we're going to have several intermediate states which we will need to be able to switch on via feature flags. Currently our template is not organised to support this. This solves that by breaking each distinct step which involves this template into separate feature flags we can turn on. There is some built in dependency between these, which represent an ordering of steps.  

### What’s changed:
- Adds several new feature flags to represent our intermediate states during the migration 
- All of these resources were behind the `UseCloudfront` feature flag, which only applied to dev
- They are now broken down into different feature flags, which allow us to turn parts on step-by-step

### Manual testing
- Deployed to dev - observe no change because all the new flags are switched on - representing our end state
- No changes to our DNS records observed 

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -125,6 +125,7 @@ Conditions:
   ProvisionNewApiClientRegistryApiKey:
     !And [!Condition ProvisionNewApiGateway, !Condition IsNotProduction]
   UseCloudfront: !Equals [dev, !Ref Environment]
+  DeployAlternateDomainRecord: false
   CreateApiGwCustomDomain: !Equals [dev, !Ref Environment]
   #endregion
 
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -123,6 +120,34 @@ Conditions:
       !Equals [production, !Ref Environment],
     ]
 
+  #region API Migration Feature Flags
+  # Please keep separate from the other ones
+  ProvisionNewApiGateway: !Equals [dev, !Ref Environment]
+  ProvisionNewApiClientRegistryApiKey:
+    !And [!Condition ProvisionNewApiGateway, !Condition IsNotProduction]
+  UseCloudfront: !Equals [dev, !Ref Environment]
```
cearl1, line 128: Are we keeping this feature flag or was it supposed to be replaced with one of the new flags you introduced?

Ryan-Andrews99: Ooh so we do want to keep this one - the git diff isn't great for highlighting this but it's used to attach the WAF from the Cloudfront stack to the API Gateway

cearl1 (approved):

PR #8445, opened 2026-06-11
BAU: Try to optimise storage token jwk latency

Ryan-Andrews99:
### Wider context of change

This lambda is called fairly infrequently and therefore has poor latency with snapstart enabled. To try and fix this, we can try what AWS refer to as "Invoke priming"[1]. Given our lambda handler has relatively few side-effects and is argumentless, we can call the handler in the beforeCheckpoint hook. This ensures that the classes in our hot code path are pre JIT compiled at the time the snapshot is created. This should hopefully ensure we have less latency on this endpoint

[1] - https://aws.amazon.com/blogs/compute/optimizing-cold-start-performance-of-aws-lambda-using-advanced-priming-strategies-with-snapstart/

### What’s changed
- Adds a new dependency `crac` to add the interface for our runtime hooks
- Implements the `beforeCheckpoint` and calls our handler.

### Manual testing
Deployed to dev and test by hitting the `/.well-known/storage-token-jwk.json` endpoint

Our latency on a new restore on this lambda in dev before:  ~4.4 seconds
Our latency one a new restore  after these changes: 470ms

These both included an initial KMS call, showing that the key was not cached on the first request

Inspected the restore logs:
- Before: `Restore Duration: 1269.88 ms`
- After `Restore Duration: 1084.98 ms`

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/StorageTokenJwkHandler.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/StorageTokenJwkHandler.java
@@ -9,6 +9,8 @@ import com.nimbusds.jose.jwk.JWKSet;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.ThreadContext;
+import org.crac.Core;
+import org.crac.Resource;
 import uk.gov.di.orchestration.shared.services.ConfigurationService;
 import uk.gov.di.orchestration.shared.services.JwksService;
 import uk.gov.di.orchestration.shared.services.KmsConnectionService;
@@ -24,7 +26,8 @@ import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachLogFiel
 import static uk.gov.di.orchestration.shared.helpers.LogLineHelper.attachTraceId;
 
 public class StorageTokenJwkHandler
-        implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent> {
+        implements RequestHandler<APIGatewayProxyRequestEvent, APIGatewayProxyResponseEvent>,
+                Resource {
 
     private final JwksService jwksService;
     private static final Logger LOG = LogManager.getLogger(StorageTokenJwkHandler.class);
@@ -41,7 +44,17 @@ public class StorageTokenJwkHandler
 
     public StorageTokenJwkHandler() {
         this(ConfigurationService.getInstance());
+        Core.getGlobalContext().register(this);
     }
+
+    @Override
+    public void beforeCheckpoint(org.crac.Context<? extends Resource> context) throws Exception {
+        LOG.info("Executing before checkpoint");
+        this.storageTokenJwkRequestHandler();
+    }
+
+    @Override
+    public void afterRestore(org.crac.Context<? extends Resource> context) throws Exception {}
 
     @Override
     public APIGatewayProxyResponseEvent handleRequest(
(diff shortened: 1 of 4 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/StorageTokenJwkHandler.java
```diff
@@ -24,25 +26,40 @@
...
+                Resource {
 
     private final JwksService jwksService;
+    private final ConfigurationService configurationService;
     private static final Logger LOG = LogManager.getLogger(StorageTokenJwkHandler.class);
 
-    public StorageTokenJwkHandler(JwksService jwksService) {
+    public StorageTokenJwkHandler(
+            JwksService jwksService, ConfigurationService configurationService) {
+        this.configurationService = ConfigurationService.getInstance();
```
Joe-Edwards-GDS: This is a bit odd - we're ignoring the configurationService passed into the constructor? Is this a relic of a previous iteration or is it needed for a more subtle reason...

Joe-Edwards-GDS (commented):
Nice discovery!

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/StorageTokenJwkHandler.java
Ryan-Andrews99: Ooh deffo a relic - will remove this!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/JwksService.java
```diff
@@ -155,4 +155,8 @@ private PublicKey createPublicKey(GetPublicKeyResponse publicKeyResponse) {
             throw new RuntimeException();
         }
     }
+
+    public void emptyCache() {
+        this.KEY_CACHE = new HashMap<>();
```
Joe-Edwards-GDS: (could probably just be `KEY_CACHE.clear()` but this works too!)

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/JwksService.java
```diff
@@ -36,7 +36,7 @@
 public class JwksService {
     private final ConfigurationService configurationService;
     private final KmsConnectionService kmsConnectionService;
-    private static final Map<String, JWK> KEY_CACHE = new HashMap<>();
+    private Map<String, JWK> KEY_CACHE = new HashMap<>();
```
Joe-Edwards-GDS: Making this non-static does have a (small) caching impact on any lambda that instantiates more than one `JwksService` - e.g. the `IPVCallbackHandler`.

Arguably a more modular solution is to keep it as an instance property but make sure we properly dependency inject all our services, but that's probably more work...

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/JwksService.java
Ryan-Andrews99: Calling `.clear()` is much nicer, it means I can keep the key cache static

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/JwksService.java
Ryan-Andrews99: I've removed these changes, calling `.clear()` on the cache means we don't need to re-assign here

Joe-Edwards-GDS (approved):

PR #8454, opened 2026-06-12
BAU: provision acceptance test params as IaC

alhcomer:
## What

SSM parameters introduced to be deployed on a per-module/pipeline basis, as opposed to the current approach of manually provisioning.

When this is merged the pipeline will have to be reserved so that we can delete the existing, manually provisioned SSM params to then be re-created.

## How to review

1. Code Review
2. Delete all cucumber tags in parameter store, trigger the utils pipeline in dev, see the tags re-created. Trigger auth api, FE, and AM pipeline runs and observe tests work as expected

### DIFF
--- a/ci/cloudformation/account-management/parameters/acceptance-test-tags.yaml
+++ b/ci/cloudformation/account-management/parameters/acceptance-test-tags.yaml
@@ -1,0 +1,37 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Conditions:
+  HasAcceptanceTests:
+    Fn::Or:
+      - Fn::Equals: [!Ref Environment, dev]
+      - Fn::Equals: [!Ref Environment, build]
+      - Fn::Equals: [!Ref Environment, staging]
+
+# /acceptance-tests/{env}/AM_CUCUMBER_FILTER_TAGS
+Mappings:
+  AcceptanceTestTags:
+    dev:
+      tags: >-
+        @AccountManagementAPI and not (@under-development or
+        @AcceptInternationalNumbers)
+    build:
+      tags: >-
+        @AccountManagementAPI and not (@under-development
+        or @AcceptInternationalNumbers)
+    staging:
+      tags: >-
+        @AccountManagementAPI and not (@under-development
+        or @AcceptInternationalNumbers or @new-mfa-reset-with-ipv)
+
+Resources:
+  AcceptanceTestAMCucumberFilterTags:
+    Type: AWS::SSM::Parameter
+    Condition: HasAcceptanceTests
+    Properties:
+      Name: !Sub /acceptance-tests/${Environment}/AM_CUCUMBER_FILTER_TAGS
+      Type: String
+      Value: !FindInMap
+        - AcceptanceTestTags
+        - !Ref Environment
+        - tags
+        - DefaultValue: "unused"
--- a/ci/cloudformation/auth/parameters/acceptance-test-tags.yaml
+++ b/ci/cloudformation/auth/parameters/acceptance-test-tags.yaml
@@ -1,0 +1,40 @@
+AWSTemplateFormatVersion: "2010-09-09"
+
+Conditions:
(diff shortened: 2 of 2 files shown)


ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
```diff
@@ -0,0 +1,91 @@
...
+        @UI and not (@under-development or @AccountInterventions
+        or @new-mfa-reset-with-ipv or @old-mfa-without-ipv
+        or @AcceptInternationalNumbers or @InternationalSmsSendingDisabled
+        or @SignInWithoutPasskeys or @PasskeyUsageEnabled
+        or @EnvironmentWithAMCStubDeployed)
+      api: >-
+        @UI and not (@under-development or @AccountInterventions or @Reauth
+        or @old-mfa-without-ipv or @new-mfa-reset-with-ipv
+        or @AcceptInternationalNumbers or @InternationalSmsSendingDisabled
+        or @SignInWithoutPasskeys or @PasskeyUsageEnabled
```
BeckaL: Didn't we remove the SignInWithoutPasskeys tag? Might be worth going through and getting the params for each env to double check these are all up to date

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
```diff
@@ -0,0 +1,91 @@
...
+  HasAcceptanceTests:
+    Fn::Or:
+      - Fn::Equals: [!Ref Environment, dev]
+      - Fn::Equals: [!Ref Environment, build]
+      - Fn::Equals: [!Ref Environment, staging]
+
+Mappings:
+  AcceptanceTestTags:
+    dev:
+      ui: >-
```
BeckaL: We shouldn't do the frontend tags in this repo, should be in the frontend repo

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
```diff
@@ -0,0 +1,91 @@
...
+        @UI and not (@AccountInterventions or @Reauth or @old-mfa-without-ipv
+        or @AccountRecovery or @InternationalSmsSendingDisabled
+        or @PasskeyUsageEnabled) and not (@InternationalNumbersIndefiniteLockout
+        and @under-development)
+      api: >-
```
BeckaL: api here is perhaps a bit confusing. Maybe make it more explicit (internal-and-external-api)

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
```diff
@@ -0,0 +1,91 @@
...
+        @UI and not (@AccountInterventions or @Reauth or @old-mfa-without-ipv
+        or @AccountRecovery or @InternationalSmsSendingDisabled
+        or @PasskeyUsageEnabled) and not (@InternationalNumbersIndefiniteLockout
+        and @under-development)
+      am: >-
```
BeckaL: Oh and I don't think we should run this as part of the utils pipelines - should run in the relevant pipelines

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
alhcomer: I've copied all the tests directly from what's in the envs now

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
alhcomer: it's all just `tags` now they're in separate CF templates

ci/cloudformation/utils/parameters/acceptance-test-tags.yaml
alhcomer: I've separated out by module, other PR [here](https://github.com/govuk-one-login/authentication-frontend/pull/3461) for FE's tags

BeckaL (approved):

PR #8469, opened 2026-06-15
BAU: Use standard key factory, don't initialise BouncyCastleProvider 

Ryan-Andrews99:
### Wider context of change

We have been using BouncyCastle here to convert a KMS key response into a public key object. To do this, we initialise a `new BouncyCastleProvider()` and use it to convert the key response. This is incredibly expensive and given the way we initialise it combined with our use of Snapstart, means we have to make this call every time there is a Snapstart restore. From testing in dev, this can add up to 2 seconds to our latency on a fresh restore. We can instead use the `KeyFactory` provided in the Java standard library. This is much more lightweight, and should shave a good amount of time off our latency when we hit a restore

### What’s changed

- Use KeyFactory instead of `JcaPEMKeyConverter` and removes the initialisation of `BouncyCastleProvider` from this service

### Manual testing

- Deployed to dev:
- Ran through auth, identity and doc app journeys  - all fine 
- Hit our jwks endpoints - all fine 
- Saw latency improvements on our IPV and Auth JWKS endpoint
### Checklist


cearl1 (commented):
Can we get rid of `bouncycastle` from the `build.gradle` file now?

cearl1 (approved):
Ah yep, can see the other uses now, thats fine!

PR #8485, opened 2026-06-17
ATO-2648: Add script to update the TTL of NS records

Ryan-Andrews99:
### Wider context of change:

As part of migrating our hosted zone, we would like to lower the TTL of our NS records in both the old and new hosted zone. This is so that, ahead of the migration, we can ensure that our clients are re-fetching our records frequently and that if we go live with a bad record set, the records are not cached for the default TTL (2 days)

### What’s changed

Adds a script to update the TTL of our NS records in a hosted zone for a given hosted zone ID. Has some sensible limits on the TTL: can't be more than the default of two days 172800,  or less than 5 mins (300) - the latter is just a arbitrary decision as the real minimum is 60 seconds

### Manual testing

- Ran against dev: 
- Lowered our TTL to one day 
- Raised it back to the default

### Checklist

### DIFF
--- a/ci/stack-orchestration/update-ns-records-ttl.sh
+++ b/ci/stack-orchestration/update-ns-records-ttl.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+NEW_TTL_VALUE=300
+
+function usage() {
+  cat << USAGE
+  Script to lower the ttl of NS records in a specified hosted zone
+
+  Usage:
+    $0 [-z|--zone-id] <ZONE-ID> [-t|--ttl] <TTL>
+    ex: ./update-ns-records-ttl.sh -z Z148QEXAMPLE8V -ttl 900 -e dev
+
+  Options:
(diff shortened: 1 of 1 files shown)


ci/stack-orchestration/update-ns-records-ttl.sh
```diff
@@ -0,0 +1,155 @@
...
+
+NEW_TTL_VALUE=300
+
+function usage() {
+  cat << USAGE
+  Script to lower the ttl of NS records in a specified hosted zone
+
+  Usage:
+    $0 [-z|--zone-id] <ZONE-ID> [-t|--ttl] <TTL>
+    ex: ./lower-ns-records-ttl.sh -z Z148QEXAMPLE8V -ttl 900 -e dev
```
cearl1: Minor thing but this example is using the wrong script name :smile:

Ryan-Andrews99: Ah this is my fault for renaming stuff 😄

cearl1 (approved):

PR #8491, opened 2026-06-19
BAU: Detect unsigned requests

Joe-Edwards-GDS:
### Wider context of change

Today, unsigned requests cause a `ClassCastException` when the code assumes a `SignedJWT`.

### What’s changed

- Consolidate fatal request validation exceptions into `InvalidAuthorizeRequestException`s
- Add a check for the JWT type and throw an `InvalidAuthorizeRequestException` if unsigned.

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidAuthorizeRequestException.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidAuthorizeRequestException.java
@@ -2,11 +2,12 @@ package uk.gov.di.authentication.oidc.exceptions;
 
 import com.nimbusds.oauth2.sdk.ErrorObject;
 
-public class InvalidAuthenticationRequestException extends Exception {
+public class InvalidAuthorizeRequestException extends Exception {
 
     private final ErrorObject error;
 
-    public InvalidAuthenticationRequestException(ErrorObject error) {
+    public InvalidAuthorizeRequestException(ErrorObject error) {
+        super(error != null ? error.getDescription() : null);
         this.error = error;
     }
 
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
@@ -80,7 +80,8 @@ public class RequestObjectAuthorizeValidator extends BaseAuthorizeValidator {
         } else {
             logErrorInProdElseWarn("Request object JWT was unsigned");
             throw new InvalidAuthorizeRequestException(
-                    OAuth2Error.INVALID_REQUEST_OBJECT.setDescription(
+                    new ErrorObject(
+                            OAuth2Error.INVALID_REQUEST_OBJECT_CODE,
                             "Request object JWT must be signed"));
         }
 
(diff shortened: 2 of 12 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -68,17 +67,25 @@ public RequestObjectAuthorizeValidator(ConfigurationService configurationService
...
-        var signedJWT = (SignedJWT) authRequest.getRequestObject();
-        clientSignatureValidationService.validate(signedJWT, client);
+        if (authRequest.getRequestObject() instanceof SignedJWT signedJWT) {
+            clientSignatureValidationService.validate(signedJWT, client);
+        } else {
+            logErrorInProdElseWarn("Request object JWT was unsigned");
+            throw new InvalidAuthorizeRequestException(
+                    OAuth2Error.INVALID_REQUEST_OBJECT.setDescription(
+                            "Request object JWT must be signed"));
+        }
```
Joe-Edwards-GDS, line 86: This is the new logic - previously it would just throw a `ClassCastException`.

We could also quite easily process plain JWTs (so long as `jarValidationRequired = false`), but that's a recipe for confusion with RPs who might misconfigure unsigned JWTs thinking they were signed.

oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidAuthorizeRequestException.java
Joe-Edwards-GDS, line 1: As discussed, we have a bit of inconsistency at the moment with checked/unchecked exceptions being thrown in various places and bubbling all the way up to AWS!

We could probably rationalise a bit and display a proper error pages when we can't redirect, but definitely out of scope here.

FWIW IPV had a model with two different exceptions which were caught by the handler:
- 'recoverable' validation errors, where we can redirect back to the RP
- 'unrecoverable' validation errors, where we cannot (i.e. if the client id/redirect URL is not valid)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/validators/RequestObjectAuthorizeValidator.java
```diff
@@ -68,17 +67,25 @@ public RequestObjectAuthorizeValidator(ConfigurationService configurationService
...
 
-        var signedJWT = (SignedJWT) authRequest.getRequestObject();
-        clientSignatureValidationService.validate(signedJWT, client);
+        if (authRequest.getRequestObject() instanceof SignedJWT signedJWT) {
+            clientSignatureValidationService.validate(signedJWT, client);
+        } else {
+            logErrorInProdElseWarn("Request object JWT was unsigned");
+            throw new InvalidAuthorizeRequestException(
+                    OAuth2Error.INVALID_REQUEST_OBJECT.setDescription(
+                            "Request object JWT must be signed"));
```
cearl1, line 85: I know its not related to your changes, but i noticed that we've got 2 ways of defining `ErrorObject`s with constant error codes and custom descriptions. 

We have the way we're doing it here, taking an error constant and returning a new error with an updated description:
```java
OAuth2Error.INVALID_REQUEST.setDescription("Invalid redirect URI")
```

and further below we create an error object directly with an error code:
```java
new ErrorObject(
        OAuth2Error.INVALID_REQUEST_CODE,
        "Request is missing state parameter"),
```

The only difference I see is that the first one sets the HTTPStatusCode to `HTTPResponse.SC_BAD_REQUEST`, which is `400`. We don't actually use this variable in the code though

Might be worth staying consistent with how we do this? Personally i was a bit confused at first when i saw the setDescription call on a constant (i didnt expect the implementation to return a new error object :smile: )

Joe-Edwards-GDS: Yeah, it would be a lot clearer if it was `withDescription` - more standard for immutable classes!

There could be an argument for reworking the handler to use the status code in future, but that's more complicated and I'm happy to match the existing pattern.

cearl1 (approved):

PR #8499, opened 2026-06-22
Aut 5466/more verification failed events

BeckaL:
Emit verification failure audit events in the case of any failure within the finish passkey assertion.

These events don't have all the data the existing failure event has (which is emitted in the case of the yubico library returning an unsuccessful assertion result). Various fields are nulled which means they won't appear on the audit event. This is because, depending on when the failure occurs, we won't have all the data required to populate the full event.

## Related PRs

In [this previous PR](https://github.com/govuk-one-login/authentication-api/pull/8489) we started emitting the failure event in one of the failure cases. The current PR implements the event for all failure cases.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
@@ -114,7 +114,7 @@ public class PasskeyAssertionService {
                         assertionResult.getSignatureCount(),
                         assertionResult.isBackedUp(),
                         passkeyCredentialDeviceTypeFrom(assertionResult),
-                        "UnknownError");
+                        "Passkey assertion result was not successful");
         var event =
                 AuthPasskeyVerificationFailed.create(
                         auditContext,
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
@@ -63,7 +63,8 @@ public class PasskeyAssertionService {
         try {
             credential = jsonParser.parsePublicKeyCredential(publicKeyCredentialJson);
         } catch (Exception e) {
-            emitAuthPasskeyVerificationFailedEvent(auditContext);
+            var failureReason = "Public key credential in request failed to parse";
+            emitAuthPasskeyVerificationFailedEvent(auditContext, failureReason, Optional.empty());
             return Result.failure(FinishPasskeyAssertionFailureReason.PARSING_PKC_ERROR);
         }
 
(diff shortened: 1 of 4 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
```diff
@@ -72,6 +72,7 @@ public Result<FinishPasskeyAssertionFailureReason, AssertionResult> finishAssert
         try {
             credential = jsonParser.parsePublicKeyCredential(publicKeyCredentialJson);
         } catch (Exception e) {
+            emitAuthPasskeyVerificationFailedEvent(auditContext);
             return Result.failure(FinishPasskeyAssertionFailureReason.PARSING_PKC_ERROR);
         }
```
csophocleous, line 77: Nothing to do with your PR specifically - Looking at this, I would expect there to be a log both in this catch and the one above. If parsing the request or the pkc causes a failure, it won't be immediately obvious from the logs.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
```diff
@@ -101,16 +105,22 @@ public Result<FinishPasskeyAssertionFailureReason, AssertionResult> finishAssert
...
-        var passkeyDetail =
-                PasskeyDetail.verificationCouldNotProceed(
-                        "Public key credential in request failed to parse");
+    private void emitAuthPasskeyVerificationFailedEvent(
+            AuditContext auditContext,
+            String failureReason,
+            Optional<
+                            PublicKeyCredential<
+                                    AuthenticatorAssertionResponse,
+                                    ClientAssertionExtensionOutputs>>
```
csophocleous, line 114: Not sure if there is much you can do other than change the param passed in but I'm not a fan of the formatters way of doing this 😆

ChristianSoftwire (approved):

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
BeckaL: agreed, hate it :-D

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/PasskeyAssertionService.java
BeckaL: So this is actually covered off, as we log the error when we generate the error response: https://github.com/govuk-one-login/authentication-api/blob/de7c50ae0d223e53f8aedd4b70256fca2121a9e7/shared/src/main/java/uk/gov/di/authentication/shared/helpers/ApiGatewayResponseHelper.java#L64-L68

PR #8502, opened 2026-06-23
ATO-2668: Setup `sis-api` module and create `SISJwksFunction`

cearl1:
### Wider context of change

We would like a new package for components that interact with SIS. We'd also like to setup the infra for a Jwks lambda to serve JWKS keys.

### What’s changed

This PR creates a new gradle package in the project `sis-api`. The config for this was mostly copied from `ipv-api` but with a few differences
- I've removed any dependencies we're not using at the moment (as well as removing the contract testing config)
- I've renamed the package for the lambda, and the group for the build.gradle file to use `orchestration` instead of `authentication` to be clear it's owned by orch.
- I've added a temporary lambda that just outputs a success message, so we can test the new project is packaged and uploaded in the deploy pipeline successfully

### Manual testing

Deployed to dev, saw the new lambda in the AWS console and manually triggered it. Saw the test output was a success.

### Checklist


template.yaml
```diff
@@ -3641,6 +3641,34 @@ Resources:
             Unit: Count
   #endregion
 
+  #region SIS Jwks Lambda
+
+  SISJwksFunction:
```
Louisasa, line 3646: your commit message might need updating as this isn't terraform 😅

.github/workflows/deploy-orch.yml
```diff
@@ -139,7 +139,7 @@ jobs:
 
       - name: SAM build
         run: |
-          ./gradlew --no-daemon --parallel :oidc:buildZip :ipv-api:buildZip :doc-checking-app-api:buildZip :client-registry-api:buildZip
+          ./gradlew --no-daemon --parallel :oidc:buildZip :ipv-api:buildZip :doc-checking-app-api:buildZip :client-registry-api:buildZip :sis-api:buildZip
```
Louisasa, line 142: I'm not certain but I think we should also add `:sis-api:buildZip` to `deploy-api-auth-sp.yml` and `deploy-auth-api-combined-dev-sp.yml` as `ipv-api:buildZip` is there and if they rely on ipv-api, they will also rely on sis-api

I also think there's some places it should be added as an ignore (`deploy-api-account-management-sp.yml` for example)

And also some places we'll want the sis-api tests to run too

I'd search for `ipv-api` in the repo and double check some places

template.yaml
cearl1: doh, will update, thanks! :smile:

.github/workflows/deploy-orch.yml
cearl1: I couldn't find any references to the ipv-api tests, i believe the orch integration tests run by looking for looking for tests in packages starting with `uk.gov.di.orchestration`, and the unit tests just exclude certain projects. I've added a test to verify that the tests are running though

Louisasa: ah so ipv-api is technically `uk.gov.di.authentication` I did add `Run Unit Tests` in `deploy-orch` which I had to specify which tests to run because I couldn't use `uk.gov.di.orchestration`

Louisasa: I think `deploy-api-modules.yml` and `deploy-api-modules-dev.yml` are still missing. I think this maybe also needs to be added to `integration-tests/build.gradle` for when we have integration tests (I'm conscious it could be missed if we leave until we make an integration test).

I would also say maybe the `docs/orchestration/README` needs updating. And we'll need to add this to the local running, idk if that should be part of this PR or not tho.

sis-api/build.gradle
```diff
@@ -0,0 +1,58 @@
+plugins {
+    id "java"
+    id "jacoco"
+}
+
+group = "uk.gov.di.orchestration.sisapi"
+version = "unspecified"
+
+File terraformDir = new File(project.rootDir, "ci/terraform/oidc")
```
Ryan-Andrews99: Think we can bin this terraformDir bit

.github/workflows/deploy-orch.yml
Ryan-Andrews99: I think the reliance on ipv-api in auth is for processing identity handler. I don't think we'd need to build this module and upload it as part of Auth's Github actions, especially given its new and shouldn't have any reliance on auth stuff hopefully

sis-api/build.gradle
cearl1: done, thanks!

.github/workflows/deploy-orch.yml
cearl1: Have updated our Run unit tests action to run the sis tests, as well as anywhere else IPV-API is used (apart from auth modules as mentioned above)

Louisasa: oooh good point, yes lets leave the auth parts then :)

Louisasa (approved):

PR #8503, opened 2026-06-23
Aut 5529/suppress passkey setup for new accounts

BeckaL:
Introduces a new field on the check user exists response, which indicates whether the backend knows of any reason the passkey registration prompt should be suppressed. 

For now, the only rule it applies is whether the account is less than 2 hours old or not - if so, it will return a response indicating the passkey registration prompt should be skipped by the frontend. This is so that users who are signing into the OL app to prove their identity (after creating their account) are not prompted to set up a passkey during an ID proving journey.

In future, we will add more rules into this logic as appropriate.

This can be merged before the relevant change on the frontend to start responding to this new field is in place - currently without that this change is a no-op in terms of journeys. The frontend does not perform strict validation on the shape of response so an extra field is just ignored.

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelper.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelper.java
@@ -1,0 +1,44 @@
+package uk.gov.di.authentication.frontendapi.helpers;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.authentication.shared.entity.UserProfile;
+
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeParseException;
+import java.util.Objects;
+
+public class PasskeyRegistrationPromptHelper {
+    private static final Logger LOG = LogManager.getLogger(PasskeyRegistrationPromptHelper.class);
+
+    private PasskeyRegistrationPromptHelper() {
+        /* This utility class should not be instantiated */
+    }
+
+    public static boolean shouldSuppressPasskeyRegistrationPrompt(UserProfile userProfile) {
+        return accountIsLessThanTwoHoursOld(userProfile);
+    }
+
+    private static boolean accountIsLessThanTwoHoursOld(UserProfile userProfile) {
+        if (Objects.isNull(userProfile.getCreated()) || userProfile.getCreated().isEmpty()) {
+            LOG.warn(
+                    "created at date on user profile is null or empty, not suppressing create passkey prompt");
+            return false;
+        }
+        try {
+            var createdDate = LocalDateTime.parse(userProfile.getCreated());
+            var nowMinusTwoHours = LocalDateTime.now(ZoneId.of("UTC")).minusHours(2);
+            var shouldSuppressBasedOnAccountCreation = createdDate.isAfter(nowMinusTwoHours);
+            if (shouldSuppressBasedOnAccountCreation) {
+                LOG.info(
+                        "suppressing passkey registration prompt as account is less than two hours old");
+            }
+            return shouldSuppressBasedOnAccountCreation;
+        } catch (DateTimeParseException e) {
+            LOG.warn(
+                    "created at date on user profile could not be parsed as local date time, not suppressing create passkey prompt");
+            return false;
+        }
+    }
+}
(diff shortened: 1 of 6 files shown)


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelper.java
```diff
@@ -0,0 +1,46 @@
...
+import java.util.Objects;
+
+public class PasskeyRegistrationPromptHelper {
+    private static final Logger LOG = LogManager.getLogger(PasskeyRegistrationPromptHelper.class);
+
+    private PasskeyRegistrationPromptHelper() {
+        /* This utility class should not be instantiated */
+    }
+
+    private static final DateTimeFormatter DATE_FORMATTER = DateTimeFormatter.ISO_LOCAL_DATE_TIME;
```
alhcomer: nitpick: Don't think we need to declare this explicitly as the one-arg `LocalDateTime.parse` uses `ISO_LOCAL_DATE_TIME` as default:

```
public static LocalDateTime parse(CharSequence text) {
    return parse(text, DateTimeFormatter.ISO_LOCAL_DATE_TIME);
}
```

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelperTest.java
```diff
@@ -0,0 +1,61 @@
...
+
+class PasskeyRegistrationPromptHelperTest {
+    private static final String YESTERDAY = LocalDateTime.now().minusDays(1).toString();
+    private static final String TWO_HOURS_ONE_MINUTE_AGO =
+            LocalDateTime.now().minusMinutes(121).toString();
+    private static final String ONE_HOUR_59_AGO = LocalDateTime.now().minusMinutes(119).toString();
+    private static final String THIRTY_MINUTES_AGO =
+            LocalDateTime.now().minusMinutes(30).toString();
+
+    private static Stream<Arguments> foo() {
```
alhcomer: could this be renamed to be more descriptive?

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandlerTest.java
```diff
@@ -662,6 +665,30 @@ void shouldReturnResultOfHasActivePasskeysWhenFeatureFlagIsOn(boolean hasActiveP
             assertEquals(hasActivePasskey, checkUserExistsResponse.hasActivePasskey());
         }
 
+        private static Stream<Arguments> foo() {
```
alhcomer: and this one?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
```diff
@@ -180,14 +181,16 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
             var userMfaDetail = UserMfaDetail.noMfa();
             Optional<Boolean> hasActivePasskey = Optional.empty();
             var needsForcedMFAResetAfterMFACheck = false;
+            var shouldSuprressPasskeyRegistrationPrompt = false;
```
csophocleous: typo in this one :)

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelperTest.java
BeckaL: Omg lol what foo makes perfect sense. Whoops

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/CheckUserExistsHandler.java
```diff
@@ -180,14 +181,16 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
 
             AuthSessionItem authSession = userContext.getAuthSession();
 
             if (userExists) {
+                var userProfile = maybeUserProfile.get();
                 auditableEvent = FrontendAuditableEvent.AUTH_CHECK_USER_KNOWN_EMAIL;
                 rpPairwiseId =
                         ClientSubjectHelper.getSubject(
-                                        userProfile.get(),
+                                        maybeUserProfile.get(),
```
csophocleous: Could this just be userProfile?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/helpers/PasskeyRegistrationPromptHelper.java
BeckaL: Good spot

alhcomer (approved):
LGTM :)

PR #8510, opened 2026-06-24
AUT-5503: Remove reference to @old-mfa-reset-with-ipv in acceptance tests for backend

BeckaL:
All the tests tagged with this tag have been removed from the acceptance test repo, as the feature flag has been removed from code so we have no way of toggling back to this old way of resetting mfa. Therefore, reference to this tag can now safely be removed from the tags for each environment

## Related PRs

Depends on https://github.com/govuk-one-login/authentication-acceptance-tests/pull/934

### DIFF
--- a/ci/cloudformation/auth/parameters/acceptance-test-tags.yaml
+++ b/ci/cloudformation/auth/parameters/acceptance-test-tags.yaml
@@ -12,18 +12,18 @@ Mappings:
   AcceptanceTestTags:
     dev:
       tags: >-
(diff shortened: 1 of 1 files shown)


ci/cloudformation/auth/parameters/acceptance-test-tags.yaml
```diff
@@ -12,18 +12,18 @@ Mappings:
   AcceptanceTestTags:
```
BeckaL, line 12: Comment to block merge until acceptance tests have been merged

ben-fernandes-sw (approved):

PR #8511, opened 2026-06-24
ATO-2686: fixups after running in build 

Ryan-Andrews99:
### Wider context of change

After running the migration in build, there are a couple of things that needed minor tweaks, fixes or scripts that needed to be written to ease the process. This add these things.

### What’s changed
- Adds a load of parameter files to represent our intermediate states for the hosted zone and cloudfront distribution stacks
- Updates the scripts to support these intermediate states
- Adds a script to automate the getting/setting of SSM params needed for the migration
- Adds a script to test the origin configuration for step 4.4 of the migration plan
- fixes typos in the any existing scripts

### Manual testing

- N/A these were ran against build

### Checklist

### DIFF
--- a/ci/stack-orchestration/configuration/build/build-oidc-cloudfront/live-with-rollback-parameters.json
+++ b/ci/stack-orchestration/configuration/build/build-oidc-cloudfront/live-with-rollback-parameters.json
@@ -1,0 +1,42 @@
+[
+  {
+    "ParameterKey": "DistributionAlias",
+    "ParameterValue": "oidc.build.account.gov.uk"
+  },
+  {
+    "ParameterKey": "CloudFrontCertArn",
+    "ParameterValue": "arn:aws:acm:us-east-1:767397776536:certificate/d7f6e974-d9e5-4c42-ba24-551f64d14494"
+  },
+  {
+    "ParameterKey": "FraudHeaderEnabled",
+    "ParameterValue": "false"
+  },
+  {
+    "ParameterKey": "StandardLoggingEnabled",
+    "ParameterValue": "true"
+  },
+  {
+    "ParameterKey": "LogDestination",
+    "ParameterValue": "none"
+  },
+  {
+    "ParameterKey": "OriginCloakingHeaderManagedSecretRotationMonthWeekDaySchedule",
+    "ParameterValue": "WED#3"
+  },
+  {
+    "ParameterKey": "OriginCloakingHeader",
+    "ParameterValue": "none"
+  },
+  {
+    "ParameterKey": "PreviousOriginCloakingHeader",
+    "ParameterValue": "none"
+  },
+  {
+    "ParameterKey": "ApiGatewayOriginDomain",
+    "ParameterValue": "w2kildk4ee.execute-api.eu-west-2.amazonaws.com"
+  },
+  {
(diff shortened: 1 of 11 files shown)


ci/stack-orchestration/manual-stacks/domains/template.yaml
```diff
@@ -177,16 +169,6 @@ Resources:
       Type: String
       Value: !Ref OidcHostedZoneRootCertificate
 
-  OidcDomainValidationTxtRecord:
```
CarlyG55, line 180: Am I going blind - was this meant to have the condition on it?

ci/stack-orchestration/configuration/build/build-oidc-cloudfront/live-with-rollback-parameters.json
```diff
@@ -0,0 +1,42 @@
...
+    "ParameterKey": "OriginCloakingHeader",
+    "ParameterValue": "none"
+  },
+  {
+    "ParameterKey": "PreviousOriginCloakingHeader",
+    "ParameterValue": "none"
+  },
+  {
+    "ParameterKey": "ApiGatewayOriginDomain",
+    "ParameterValue": "w2kildk4ee.execute-api.eu-west-2.amazonaws.com"
```
CarlyG55, line 36: possible stupid question - there's no reason why we shouldn't be exposing these?

CarlyG55 (commented):
More generally, just want to make sure we aren't exposing any values here we shouldn't be!

ci/stack-orchestration/manual-stacks/domains/template.yaml
Ryan-Andrews99: Yeah I think it was, but I may have forgotten this initially. Either way it's being deleted because it doesn't work

CarlyG55: No worries - does the new version have/need a condition?

ci/stack-orchestration/configuration/build/build-oidc-cloudfront/live-with-rollback-parameters.json
Ryan-Andrews99: Not that I can think of - the API still has the cloaking WAF on so you can't hit the API directly, even with that domain name. The [docs don't mention anything about it being sensitive](https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html) either.

Ryan-Andrews99: We can also [disable this endpoint](https://docs.aws.amazon.com/apigateway/latest/developerguide/rest-api-disable-default-endpoint.html) once we're done with the migration - if we're worried about it

ci/stack-orchestration/manual-stacks/domains/template.yaml
Ryan-Andrews99: Ah so we're adding this to the main domain template (see this [one](https://github.com/govuk-one-login/domains/pull/182)) so there is no need to add a condition  - we'll just add it as/when it's needed

CarlyG55 (approved):

PR #8518, opened 2026-06-25
Aut 5546: metrics for passkeys

BeckaL:
Increments metrics for various passkey journeys.

These metrics have mostly been introduced shadowing audit events.


shared/src/main/java/uk/gov/di/authentication/shared/services/CloudwatchMetricsService.java
```diff
@@ -159,6 +159,10 @@ public void incrementAuthenticationSuccessWithMfa(
                 mfaMethodPriorityIdentifier);
     }
 
+    public Map.Entry<String, String> environmentDimension() {
+        return Map.entry(ENVIRONMENT.getValue(), configurationService.getEnvironment());
+    }
```
alhcomer: this isn't being used anywhere atm. should it be?

BeckaL: Ah yes sorry experimented with adding this but it made testing much worse because the metrics service is stubbed, should have been removed from the commit, out now

alhcomer (approved):

PR #8519, opened 2026-06-25
ATO-2011: Bump nimbus OIDC package to 11.22.2

cearl1:
### Wider context of change

We have previously analysed the changes from our previous version of `com.nimbusds:oauth2-oidc-sdk` to `11.21` and have ensured that RPs are not sending invalid auth requests.

### What’s changed

Unfortunately, due to `11.7.1` introducing a dependency with a vulnerability, we have had to update to `11.22.2` instead, where the vulnerable dependency was updated to a non-vulnerable version. This PR also replaces all the now-deprecated methods and constructors, and replaces them with their equivalent in `11.22.2`.

### Manual testing

Deployed to dev, ran through an auth only journey, an identity journey, and a docapp journey. All 3 journeys were successful.

### Checklist

### DIFF
--- a/auth-external-api/src/main/java/uk/gov/di/authentication/external/validators/TokenRequestValidator.java
+++ b/auth-external-api/src/main/java/uk/gov/di/authentication/external/validators/TokenRequestValidator.java
@@ -8,6 +8,7 @@ import com.nimbusds.oauth2.sdk.auth.ClientAuthenticationMethod;
 import com.nimbusds.oauth2.sdk.auth.PrivateKeyJWT;
 import com.nimbusds.oauth2.sdk.auth.verifier.ClientAuthenticationVerifier;
 import com.nimbusds.oauth2.sdk.auth.verifier.InvalidClientException;
+import com.nimbusds.oauth2.sdk.auth.verifier.JWTAudienceCheck;
 import com.nimbusds.oauth2.sdk.id.Audience;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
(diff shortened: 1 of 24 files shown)


auth-external-api/src/main/java/uk/gov/di/authentication/external/validators/TokenRequestValidator.java
```diff
@@ -77,7 +78,8 @@ public void validatePrivateKeyJwtClientAuth(
             ClientAuthenticationVerifier<?> signatureVerifier =
                     new ClientAuthenticationVerifier<>(
                             new PrivateKeyJwtAuthPublicKeySelector(publicKeys, KeyType.EC),
-                            expectedAudience);
+                            expectedAudience,
+                            JWTAudienceCheck.LEGACY);
```
Louisasa, line 82: I think LEGACY is fine here but I'm wondering if we should add a log to see if we could update to use STRICT? might be worth a chat to Carly to see if necessary but I think using STRICT if we can would be good

Louisasa (approved):

auth-external-api/src/main/java/uk/gov/di/authentication/external/validators/TokenRequestValidator.java
cearl1: Yeah good shout, i think at the moment it will break the orch > auth token endpoint so i'll leave it as LEGACY for now, but will chat with Carly

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ClientSignatureValidationService.java
```diff
@@ -110,7 +111,8 @@ public void validateTokenClientAssertion(PrivateKeyJWT privateKeyJWT, ClientRegi
                             new PrivateKeyJwtAuthPublicKeySelector(publicKey),
                             Set.of(
                                     new Audience(oidcAPI.tokenURI().toString()),
-                                    new Audience(oidcAPI.getIssuerURI().toString())));
+                                    new Audience(oidcAPI.getIssuerURI().toString())),
+                            JWTAudienceCheck.LEGACY);
```
cearl1, line 115: This needs to stay as `LEGACY` for now as applying `STRICT` would require a lot of RPs to change how they send token requests.

auth-external-api/src/main/java/uk/gov/di/authentication/external/validators/TokenRequestValidator.java
cearl1: So orch are sending a single audience value, but the `ClientAuthenticationVerifier` is being instantiated with 2 audiences here. This is something auth will need to update

PR #8525, opened 2026-06-26
BAU: Refactor IntegrationTestConfigurationService

cearl1:
### Wider context of change

Previously, this static class took several extensions as parameters, stored them locally, and overrode some ConfigurationService methods to use these extensions. However, in all cases where this constructor was used, the test were passing in the protected extensions defined in `IntegrationTest` (as the integration tests inherit from this class). This was kind of pointless, as these protected extensions were available to the `IntegrationTestConfgurationService`, so we were passing these extensions around for no reason.

### What’s changed

This PR removes the parameters for extensions in the static class `IntegrationTestConfigurationService`, instead using the protected extensions defined in the `IntegrationTest` class (which is where `IntegrationTestConfigurationService` resides). 

There were also some constants for common `IntegrationTestConfigurationService` instances, like `TEST_CONFIGURATION_SERVICE` and `TXMA_ENABLED_CONFIGURATION_SERVICE`, which i've used in some places instead of instantiating `IntegrationTestConfigurationService` again.

I've also tidied up some class references, removing any unnecessary class prefixes when the static class was already in the class.

### Checklist


Ryan-Andrews99 (approved):
Net -200 lines 🚀

PR #8529, opened 2026-06-26
BAU: Extract OrchJwtService

cearl1:
### Wider context of change

We have multiple places where we are signing and encrypting a JWT, and we are about to add another (for SIS). The signing and encrypting parts of the code are the same across all the places we are constructing JWTs (apart from the signing and encryption keys). It would be nice if we could have a service that handles the signing and encrypting for us, so we don't need to copy big chunks of code around.

### What’s changed

This PR extracts the signing and encryption of a JWT to a new service called `OrchJwtService`. This service has a method which takes a `JWTClaimSet`, a signing key alias, and an encryption key. It then signs and encrypts the JWT.

I've started to use this in `OrchestrationAuthorizationService`,`DocAppAuthorisationService` and `IPVAuthorisationService`. I have had to change the tests a bit to assert on the claims that we pass into the OrchJwtService, instead of decrypting the JWT we get back from it.

`OrchestrationAuthorizationService` is quite small now so we might be able to move the use of `OrchJwtService` into `AuthenticationAuthorisationService`. I've left it as is for now though

### Manual testing

Deployed to dev, ran through an auth only journey, an identity journey, and a docapp journey. All 3 journeys were successful.

### Checklist

### DIFF
--- a/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidJWEException.java
+++ b/oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidJWEException.java
@@ -1,7 +1,0 @@
-package uk.gov.di.authentication.oidc.exceptions;
-
-public class InvalidJWEException extends RuntimeException {
-    public InvalidJWEException(String message, Throwable cause) {
-        super(message, cause);
-    }
-}
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchJwtService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchJwtService.java
@@ -1,0 +1,138 @@
+package uk.gov.di.orchestration.shared.services;
+
+import com.nimbusds.jose.EncryptionMethod;
+import com.nimbusds.jose.JOSEException;
+import com.nimbusds.jose.JWEAlgorithm;
+import com.nimbusds.jose.JWEHeader;
+import com.nimbusds.jose.JWEObject;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.JWSHeader;
+import com.nimbusds.jose.Payload;
+import com.nimbusds.jose.crypto.RSAEncrypter;
+import com.nimbusds.jose.crypto.impl.ECDSA;
+import com.nimbusds.jose.util.Base64URL;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import software.amazon.awssdk.core.SdkBytes;
+import software.amazon.awssdk.services.kms.model.MessageType;
+import software.amazon.awssdk.services.kms.model.SignRequest;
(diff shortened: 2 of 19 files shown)


oidc-api/src/main/java/uk/gov/di/authentication/oidc/exceptions/InvalidJWEException.java
cearl1, line 1: This was moved to `orchestration-shared` to be used by the new service!

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/OrchJwtServiceTest.java
```diff
@@ -0,0 +1,149 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+import static uk.gov.di.orchestration.sharedtest.utils.KeyPairUtils.generateRsaKeyPair;
+
+public class OrchJwtServiceTest {
+    private static final String SIGNING_KEY_ID = "test-key-id";
+    private static final String SIGNING_KEY_ALIAS = "test-key-alias";
+    // 5000 chars
+    private static final String LONG_CLAIM =
```
mrwilson: This can be replaced with `"1".repeat(5000)`

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchJwtService.java
```diff
@@ -0,0 +1,132 @@
...
+                        configurationService, new KmsConnectionService(configurationService)));
+    }
+
+    public OrchJwtService(KmsConnectionService kmsConnectionService, JwksService jwksService) {
+        this.kmsConnectionService = kmsConnectionService;
+        this.jwksService = jwksService;
+    }
+
+    public EncryptedJWT signAndEncryptJWT(
+            JWTClaimsSet jwtClaimsSet, String signingKeyAlias, RSAPublicKey publicEncryptionKey) {
```
mrwilson, line 54: If the claimsSet has a unique `jti`, might be worth adding to the log context so all loglines in this method get that ID attached?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchJwtService.java
```diff
@@ -0,0 +1,138 @@
...
+        if (jwtClaimsSet.getJWTID() != null) {
+            attachLogFieldToLogs(JWT_ID, jwtClaimsSet.getJWTID());
+        }
+        LOG.info("Generating signed and encrypted JWT");
+        var signingKey = jwksService.getPublicJWKWithKeyId(signingKeyAlias);
```
Ryan-Andrews99, line 59: I _think_ in the Auth equivalent of this they try/catch-ed this bit, incase the alias was bad/invalid  - might be worth doing here?

Ryan-Andrews99 (approved):
Left one optional comment

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/OrchJwtService.java
cearl1: I can't see where this try-catch is, maybe you're thinking of something else?

PR #8563, opened 2026-07-01
ATO-2568: Create S3 bucket to hold JWKS files

cearl1:
### Wider context of change

We would like to move away from creating JWKS responses in lambdas, and instead serve files in an S3 bucket. This should not only reduce response time, but also is a prerequesite to using the identity key rotation stacks, which will rotate keys for us. To start though, we need an S3 bucket to store JWKS json files.

### What’s changed

This PR creates a new S3 bucket for storing JWKS files. It's configured to use the new access logs bucket that has been set up for us by devplatform. 

### Manual testing

Deployed to dev, saw the bucket existed in S3 and saw it was configured with the correct access logs bucket.

### Checklist

### DIFF
--- a/template.yaml
+++ b/template.yaml
@@ -198,6 +198,7 @@ Mappings:
         }]'
       oidcDomainName: "oidc.dev.account.gov.uk"
       oidcAlternateDomainName: "oidc-orch.dev.account.gov.uk"
+      serviceNameForKeyRotationStack: "orch"
     build:
       dynatraceSecretArn: arn:aws:secretsmanager:eu-west-2:216552277552:secret:DynatraceNonProductionVariables
       dynatraceLayerArn: arn:aws:lambda:eu-west-2:216552277552:layer:Dynatrace_OneAgent_1_331_39_20260206-150338_with_collector_java_arm:1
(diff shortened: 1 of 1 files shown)


template.yaml
```diff
@@ -9631,6 +9603,29 @@ Resources:
       EvaluationPeriods: 1
   #endregion
 
+  #region S3 bucket for JWKS files
+  WellKnownJwksBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Sub
+        - govuk-one-login-${ServiceName}-published-keys-${Environment}
+        - ServiceName: "orch"
```
cearl1: I'm explicitly setting the ServiceName here as we will need it in the future to setup the key rotation pipeline, and it's easier to see if we use a variable than if it was part of the bucket name!

Ryan-Andrews99: I'd suggest we put this in the environment mappings bit - even if it's the same per env. That way we can name it a bit clearer so something like:

```
 BucketName: !Sub
        - govuk-one-login-${ServiceName}-published-keys-${Environment}
        -  ServiceName: !FindInMap [EnvironmentConfiguration, !Ref Environment, serviceNameForKeyRotationStack ]
```

cearl1: Oooh yeah that would be a bit cleaner i'll do that!

Ryan-Andrews99 (approved):

PR #8564, opened 2026-07-01
Aut 5532/do not allow setup passkey when 2fa not entered

BeckaL:
We only want to allow passkey setup in 2fa journeys (where password and mfa have been entered). We already do this in the frontend prompting rules, but this adds an additional check into the backend for this before allowing passkey create journeys to proceed to AMC to shore up this logic.

## Related PRs

[Frontend PR which ensures 1fa / uplift journeys cannot be prompted to setup passkeys](https://github.com/govuk-one-login/authentication-frontend/pull/3511) - this needs to be merged first
Acceptance test PR - can be merged first


ChristianSoftwire (approved):
LGTM

PR #8575, opened 2026-07-03
ATO-2724: Create sendRequest method in SISAuthorisationService

cearl1:
### Wider context of change

We would like to redirect users to SIS before sending them to IPV. At the moment we have code to create a request JWT and we are asserting the correct claims are set on this request object. We would now like to use this request object and return an API response to redirect the user to the SIS authorisation URI, including the request object as a query parameter like we do for IPV.

### What’s changed

This PR adds a new method `sendRequest` to the `SISAuthorisationService`. This method creates a new auth request for SIS, using the `constructRequestJWT` method. It then saves the new state generated to the `stateStorageService` and `crossBrowserOrchestrationService`. Finally, it returns an API response redirect the user to the redirect URI generated by the auth request

There are still a few bits missing such as audit events and cloudwatch metrics, but these will be addressed in future PRs

### Checklist

### DIFF
--- a/orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/utils/JwtUtils.java
+++ b/orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/utils/JwtUtils.java
@@ -1,0 +1,52 @@
+package uk.gov.di.orchestration.sharedtest.utils;
+
+import com.nimbusds.jose.EncryptionMethod;
+import com.nimbusds.jose.JWEAlgorithm;
+import com.nimbusds.jose.JWEHeader;
+import com.nimbusds.jose.JWEObject;
+import com.nimbusds.jose.JWSAlgorithm;
+import com.nimbusds.jose.JWSHeader;
+import com.nimbusds.jose.Payload;
+import com.nimbusds.jose.crypto.ECDSASigner;
+import com.nimbusds.jose.crypto.RSAEncrypter;
+import com.nimbusds.jose.jwk.Curve;
+import com.nimbusds.jose.jwk.gen.ECKeyGenerator;
+import com.nimbusds.jose.jwk.gen.RSAKeyGenerator;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.oauth2.sdk.ResponseType;
+
+public class JwtUtils {
+    private JwtUtils() {}
+
+    public static EncryptedJWT createDummyJwt() throws Exception {
+        var jwtClaimsSet =
+                new JWTClaimsSet.Builder()
+                        .claim("redirect_uri", "http://test-redirect.com")
+                        .claim("response_type", ResponseType.CODE.toString())
+                        .claim("client_id", "test-client-id")
+                        .claim("govuk_signin_journey_id", "test-csid")
+                        .issuer("test-client-id")
+                        .build();
+        var ecSigningKey =
+                new ECKeyGenerator(Curve.P_256)
+                        .keyID("key-id")
+                        .algorithm(JWSAlgorithm.ES256)
+                        .generate();
+        var ecdsaSigner = new ECDSASigner(ecSigningKey);
+        var jwsHeader = new JWSHeader(JWSAlgorithm.ES256);
+        var signedJWT = new SignedJWT(jwsHeader, jwtClaimsSet);
+        signedJWT.sign(ecdsaSigner);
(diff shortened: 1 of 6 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
```diff
@@ -283,6 +283,10 @@ public URL getIPVJwksUrl() {
         return getURLOrThrow("IPV_JWKS_URL");
     }
 
+    public URI getSISAuthorisationURI() {
+        return getURIOrEmpty("IPV_AUTHORISATION_URI");
```
Louisasa: don't we know SIS's authorisation URI now and can update this to be SIS's actual URI?

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
```diff
@@ -1,45 +1,118 @@
...
+                        reproveIdentity);
+
+        var sisAuthRequest =
+                new AuthorizationRequest.Builder(
+                                new ResponseType(ResponseType.Value.CODE),
+                                new ClientID(configurationService.getSISAuthorisationClientId()))
+                        .endpointURI(configurationService.getSISAuthorisationURI())
+                        .requestObject(encryptedJWT)
+                        .build();
+        // TODO: Do we want to use a different state storage prefix
```
Louisasa: the reason I thought no was because the state will be different (as that's come from the constructed JWT) however, I'm not 100% sure on how this works so might need double checking

orchestration-shared-test/src/main/java/uk/gov/di/orchestration/sharedtest/utils/JwtUtils.java
```diff
@@ -0,0 +1,52 @@
...
+import com.nimbusds.jose.jwk.gen.RSAKeyGenerator;
+import com.nimbusds.jwt.EncryptedJWT;
+import com.nimbusds.jwt.JWTClaimsSet;
+import com.nimbusds.jwt.SignedJWT;
+import com.nimbusds.oauth2.sdk.ResponseType;
+
+public class JwtUtils {
+    private JwtUtils() {}
+
+    public static EncryptedJWT createDummyJwt() throws Exception {
```
Louisasa, line 23: nice!

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
```diff
@@ -1,45 +1,118 @@
...
+    public APIGatewayProxyResponseEvent sendRequest(
+            AuthenticationRequest authRequest,
+            UserInfo userInfo,
+            String rpClientID,
+            String sessionId,
+            String clientSessionId,
+            Boolean reproveIdentity,
+            List<String> levelsOfConfidence) {
+
+        attachLogFieldToLogs(CLIENT_ID, rpClientID);
```
Louisasa, line 81: I think this should still have an identity check at the beginning like IPV right?

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/ConfigurationService.java
cearl1: I was thinking we could set them once we have it hooked up in AuthCallback, because otherwise we wont know if it works or not.

Either way this env var needs to change so i'll update that!

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
cearl1: Yeah I think i forgot about this line - we're storing by sessionId but because the state storage service uses a dynamo Put call it will just overwrite any existing value, so we should be fine to not use a different prefix

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
cearl1: Ahh yep, thanks!

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
Louisasa: might be worth double checking with Carly that overwriting it is okay? I'm not sure if we'd need to keep it in the db for any reason

Louisasa (approved):

PR #8579, opened 2026-07-06
AUT-5648: remove old phone check SQS policy and SQS url

pskushwaha1:
## What

AUT-5648: remove old phone check SQS policy and SQS url

Not this lambda is not in use its legacy Lambda , removing the policy attachment so we can delete legacy Experian phone check resource   

## How to review

1. Code Review


sw-mattw (approved):
Static review, looks sensible

PR #8582, opened 2026-07-07
ATO-2728: Add response validation for SIS

cearl1:
### Wider context of change

We would like to handle errors in the callback response from SIS. We have a similar thing in `IPVAuthorisationService.validateResponse`. SIS is slightly different as we want to make sure we have the ability to differentiate between the `record_update_requested` error, and other errors. This is because we need to add an extra claim to the IPV request if the user requests to update their details.

### What’s changed

This PR adds a method `validateResponse` to the `SISAuthorisationService`. This method checks the query parameters of the response from SIS, and does some standard checks that we also do in IPV (state exists, state matches what we expect, auth code exists). It also checks if any errors are returned, and differentiates between the `record_update_requested` error by using a flag on the custom error class. 

### Manual testing

Added a bunch of unit tests with 100% coverage

### Checklist


sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
```diff
@@ -182,4 +186,64 @@ private ClaimsSetRequest buildClaimsRequest(
...
+            return Optional.of(
+                    new SISCallbackValidationError(
+                            OAuth2Error.INVALID_REQUEST_CODE, "No query parameters present"));
+        }
+        if (queryParams.containsKey("error")) {
+            if (RECORD_UPDATE_REQUESTED.equals(queryParams.get("error_description"))) {
+                LOG.warn("User requested to update their details");
+                return Optional.of(
+                        new SISCallbackValidationError(queryParams.get("error"), null, true));
+            }
```
Ryan-Andrews99: This check needs to be slightly more specific I think, as the [RFC](https://github.com/govuk-one-login/architecture/pull/1050/changes) says that we should have `error=access_denied` && `error_description=record_update_requested` (or any other `error_description`). Right now, any present error code matches this

cearl1: Ahhh good catch, will update the PR!

Joe-Edwards-GDS: There are 3 OAuth error cases we need to handle
- `access_denied` + `record_update_requested` (route to IPV with update requested)
- `access_denied` (route to IPV normally)
- default (a genuine error - we may still want to route to IPV but it should be logged as an error)

Really the first 2 cases are not errors, so we shouldn't use elevated logging (i.e. info at most), and we should be consistent about where the business logic for it is - either map the error code/description here for _both_ cases, or at the call-site for both.

cearl1: Ahh ok, i assumed we wouldn't get any other errors from the table in the RFC. I'll update the PR again to handle these cases better

Joe-Edwards-GDS: Hopefully we won't - but if we do they are real errors 😄

sis-api/src/main/java/uk/gov/di/orchestration/sis/service/SISAuthorisationService.java
```diff
@@ -182,4 +184,16 @@ private ClaimsSetRequest buildClaimsRequest(
...
                         .withValues(List.of(storageToken.getValue())));
     }
+
+    public Optional<SISCallbackValidationError> validateResponse(
+            Map<String, String> queryParams, String sessionId) {
+        if (queryParams == null || queryParams.isEmpty()) {
+            LOG.warn("No Query parameters in SIS Authorisation response");
+            return Optional.of(
+                    new SISCallbackValidationError(
+                            OAuth2Error.INVALID_REQUEST_CODE, "No query parameters present"));
```
CarlyG55, line 194: Right now we actually show session expired for auth-callback/ipv-callback for this case, and I recently removed a ticket to do specific handling as I thought that was a decent response (realistically, people will only find it through their browser history). Won't make this a blocker but would be good to think about how we share this across different components (especially as this gives a bit of a horrible error screen)

CarlyG55: Joe and I were just talking this morning about making our error screens nicer 😅

CarlyG55 (approved):

PR #8587, opened 2026-07-09
ATO-2740: Add helper for making identity token requests

cearl1:
### Wider context of change

We would like to consolidate some of the identity code in the IPVCallbackHandler, as we will need it for the SISCallbackHandler. 

### What’s changed

This PR creates a new module called `identity-shared` which will be used to store code shared by `ipv-api` and `sis-api`. This PR also adds a helper class to send a token request, and adds some tests alongside it.

These changes have been essentially copied from IPVCallbackHandler L326-343

I've also added a custom `AuditEventConfiguration` class, as IPV and SIS will send different events.

### Checklist


identity-shared/src/test/java/uk/gov/di/orchestration/identity/helpers/IdentityCallbackHelperTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        assertAuditEventSent(TestAuditEvent.TEST_UNSUCCESSFUL_TOKEN_RESPONSE_RECEIVED);
+        verifyNoMoreInteractions(auditService);
+
+        assertThat(
+                redirectLogging.events(),
+                hasItem(
+                        withLevelAndMessageContaining(
+                                ERROR,
+                                "Redirecting to frontend error page: " + FRONT_END_ERROR_URI)));
```
cearl1, line 80: I'm not sure this assertion has much value, but thought i'd include it anyway. Happy to remove if people think its pointless :smile:

identity-shared/src/main/java/uk/gov/di/orchestration/identity/helpers/IdentityCallbackHelper.java
```diff
@@ -0,0 +1,51 @@
...
+                segmentedFunctionCall("getToken", () -> identityTokenService.getToken(authCode));
+        if (!tokenResponse.indicatesSuccess()) {
+            auditService.submitAuditEvent(
+                    auditEventConfiguration.unsuccessfulTokenResponseReceived(), clientId, user);
+            return Optional.of(
+                    RedirectService.redirectToFrontendErrorPageWithErrorLog(
+                            frontend.errorURI(),
+                            new Exception(
+                                    String.format(
+                                            "TokenResponse was not successful: %s",
```
Louisasa, line 44: If this is going to be used for both IPV and SIS, I think it might be good if we can add who's called it in the logs maybe? It might help with quicker debugging but if it's too awkward dw, just a thought

cearl1: We should be able to tell who's called it by the log group that the log message sits in. It will either be in IPVCallbackHandler's log group or SISCallbackHandler's log group

Louisasa (approved):

PR #8596, opened 2026-07-14
ATO-2213: Remove leftover keys

cearl1:
### Wider context of change

We recently switched to using a new external token signing key, and moved the old keys into the template as a variable so we can still support verifying refresh tokens. We now no longer need the old external token signing keys, so we can safely remove them now.

### What’s changed

This PR removes the old external token signing keys, as well as the feature flags used during the migration process.

### Checklist

### DIFF
--- a/orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/JwksServiceTest.java
+++ b/orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/JwksServiceTest.java
@@ -47,15 +47,15 @@ class JwksServiceTest {
     }
 
     @Test
-    void shouldRetrievePublicTokenSigningRsaKeyFromKmsAndParseToJwk() throws Exception {
-        var keyAlias = "25252525252525";
-        when(configurationService.getExternalTokenSigningKeyRsaAlias()).thenReturn(keyAlias);
+    void shouldRetrieveExternalTokenSigningRSAKeyFromKmsAndParseToJwk() throws Exception {
+        var keyAlias = "test-external-token-signing-key-rsa-alias";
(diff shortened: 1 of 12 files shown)


orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/JwksServiceTest.java
```diff
@@ -31,37 +31,6 @@ class JwksServiceTest {
...
-
-        JWK publicKeyJwk = jwksService.getPublicTokenJwkWithOpaqueId();
-
-        assertThat(publicKeyJwk.getKeyID(), equalTo(hashSha256String(keyAlias)));
-        assertThat(publicKeyJwk.getAlgorithm(), equalTo(JWSAlgorithm.ES256));
-        assertThat(publicKeyJwk.getKeyUse(), equalTo(KeyUse.SIGNATURE));
-    }
-
-    @Test
-    void shouldRetrievePublicTokenSigningRsaKeyFromKmsAndParseToJwk() throws Exception {
```
Ryan-Andrews99, line 50: I wonder if there is an equivalent test for the new key - because sonar is complaining the `generateRsaKey` method is unused. Mind chucking one in for this?

orchestration-shared/src/test/java/uk/gov/di/orchestration/shared/services/TokenValidationServiceTest.java
Ryan-Andrews99: Think there might be some test removing/simplification now the feature flags are removed

Ryan-Andrews99 (approved):

PR #8608, opened 2026-07-15
ATO-2749: Create EndOfJourneyService

cearl1:
### Wider context of change

We have 2 places in the code (Auth and IPV) where we check for an AIS intervention, and logout the user if the response is BLOCKED or SUSPENDED. We will be adding another place soon where this will be called too (in SIS). It would be useful to extract this into its own service so we can reference the same code in all 3 places.

### What’s changed

This PR creates a new service called AISLogoutService (i'm open to better names!) which logs the user out if they have a BLOCKED or SUSPENDED AIS response. 

This code is the same as the `IPVCallbackHandler.checkForAisIntervention` method

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/EndOfJourneyService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/EndOfJourneyService.java
@@ -1,0 +1,128 @@
+package uk.gov.di.orchestration.shared.services;
+
+import com.amazonaws.services.lambda.runtime.events.APIGatewayProxyResponseEvent;
+import com.nimbusds.oauth2.sdk.ErrorObject;
+import com.nimbusds.openid.connect.sdk.AuthenticationErrorResponse;
+import com.nimbusds.openid.connect.sdk.AuthenticationRequest;
+import com.nimbusds.openid.connect.sdk.AuthenticationSuccessResponse;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.orchestration.audit.AuditContext;
+import uk.gov.di.orchestration.audit.TxmaAuditUser;
+import uk.gov.di.orchestration.shared.entity.AccountIntervention;
+import uk.gov.di.orchestration.shared.entity.DestroySessionsRequest;
+import uk.gov.di.orchestration.shared.entity.OrchSessionItem;
+import uk.gov.di.orchestration.shared.entity.ResponseHeaders;
+
+import java.util.Map;
+import java.util.Optional;
+
+import static uk.gov.di.orchestration.shared.entity.AccountInterventionStatus.SUSPENDED_REPROVE_ID;
+import static uk.gov.di.orchestration.shared.helpers.ApiGatewayResponseHelper.generateApiGatewayProxyResponse;
+import static uk.gov.di.orchestration.shared.helpers.InstrumentationHelper.segmentedFunctionCall;
+
+public class EndOfJourneyService {
+
+    private static final Logger LOG = LogManager.getLogger(EndOfJourneyService.class);
+    private final ConfigurationService configurationService;
+    private final AccountInterventionService accountInterventionService;
+    private final LogoutService logoutService;
+    private final OrchAuthCodeService orchAuthCodeService;
+
+    public EndOfJourneyService(
+            ConfigurationService configurationService,
+            AccountInterventionService accountInterventionService,
+            LogoutService logoutService,
+            OrchAuthCodeService orchAuthCodeService) {
+        this.configurationService = configurationService;
+        this.accountInterventionService = accountInterventionService;
+        this.logoutService = logoutService;
+        this.orchAuthCodeService = orchAuthCodeService;
+    }
+
+    public AccountIntervention getIntervention(
+            String internalCommonSubjectId, AuditContext auditContext) {
+        return segmentedFunctionCall(
+                "AIS: getAccountIntervention",
+                () ->
+                        accountInterventionService.getAccountIntervention(
+                                internalCommonSubjectId, auditContext));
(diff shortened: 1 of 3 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AISLogoutService.java
```diff
@@ -0,0 +1,51 @@
...
+            TxmaAuditUser auditUser,
+            String clientId) {
+        AccountIntervention intervention =
+                segmentedFunctionCall(
+                        "AIS: getAccountIntervention",
+                        () ->
+                                accountInterventionService.getAccountIntervention(
+                                        orchSession.getInternalCommonSubjectId(), auditContext));
+        if (configurationService.isAccountInterventionServiceActionEnabled()
+                && (intervention.getBlocked() || intervention.getSuspended())) {
```
Joe-Edwards-GDS: FYI this logic might need to be slightly more detailed to work for both auth-only and identity journeys, to add an exception for reprove identity interventions (see https://govukverify.atlassian.net/wiki/spaces/Architecture/pages/6797033485/Account+Interventions+in+Orchestration#Near-future-(post-SIS))

You could also leave that until later if we know this is always called on an identity journey.

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AISLogoutService.java
Joe-Edwards-GDS, line 1: I wonder if we can/should reframe this as 'end of journey checks', or perhaps 'preparing for RP redirect'. That way it's extensible if we need to do any other activity before routing back to the RP - e.g. metrics, audit events?

You could even consider generating the auth code and redirect itself in this service - there's still a bunch of duplication between `AuthenticationCallbackHandler` and `IPVCallbackHelper`, for example. It might be too difficult if there are subtle differences, but the flow should be pretty similar!

cearl1: Oooh good shout, i'll see if i can extract some more shared code here

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/AISLogoutService.java
cearl1: Thanks! added this exception now :smile:

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/EndOfJourneyService.java
```diff
@@ -0,0 +1,111 @@
...
+                isIdentityRequired);
+    }
+
+    public Optional<APIGatewayProxyResponseEvent> checkForIntervention(
+            OrchSessionItem orchSession,
+            AccountIntervention intervention,
+            TxmaAuditUser auditUser,
+            String clientId,
+            boolean isAuthJourney,
+            boolean isIdentityRequired) {
```
Joe-Edwards-GDS: What does `isAuthJourney` mean here - when will we have `isAuthJourney = true` and `isIdentityRequired = true`?

cearl1: yeah it's a bit vague isn't it? The `isAuthJourney` checks are ones exclusive to the `AuthenticationCallbackHandler`, which also has a check for `isIdentityRequired`

I couldn't think of a better name, i'll have a rethink about it today!

Joe-Edwards-GDS: I guess it depends if we're making the change to skip the fraud check between auth/ipv - in which case we might be able to simplify slightly...

cearl1: Oh yeah good point! If we're happy to go ahead and remove that fraud check between auth and IPV i can remove the extra flags and refactor the AuthenticationCallbackHandler in a separate ticket to call this method if identity is not required (because we still want to call AIS at the end of an auth only journey right?)

Joe-Edwards-GDS: Yeah exactly (fitting perfectly with `EndOfJourney` service 😄 ).

I think you'll still want _one_ of the flags, so we can include the 'ignore reprove identity intervention for auth-only journeys' rule, but shouldn't need both.

cearl1: Ah yep thanks, removed the `isIdentityJourney` flag!

orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/EndOfJourneyService.java
```diff
@@ -0,0 +1,146 @@
...
+        var errorResponseUri =
+                new AuthenticationErrorResponse(
+                                authenticationRequest.getRedirectionURI(),
+                                error,
+                                authenticationRequest.getState(),
+                                authenticationRequest.getResponseMode())
+                        .toURI();
+
+        if (isLogoutRequired) {
+            return logoutService.handleReauthenticationFailureLogout(
```
CarlyG55: Is reauthentication failure the correct logout case here? if so could we make it a bit more clear somehow from the boolean? I think now this logic has been moved out, it's not as clear to me that this is correct

cearl1: Ah yep i've removed the logic surrounding this reauth failure from this method and will keep that in auth callback.

CarlyG55 (approved):

PR #8623, opened 2026-07-17
ATO-2755: Create IdentitySPOTService

cearl1:
### Wider context of change

This is part of the work to consolidate the identity callback code so it can be shared between IPV and SIS

### What’s changed

This PR adds a service to send SPOT requests and wait for a SPOT response. This has been copied from IPVCallbackHelper.

I will remove the copied code once we use these new shared identity methods in IPVCallbackHandler

### Checklist

### DIFF
--- a/identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
+++ b/identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
@@ -1,0 +1,9 @@
+package uk.gov.di.orchestration.identity.entity;
+
+public enum ProcessingIdentityStatus {
+    COMPLETED,
+    PROCESSING,
+    ERROR,
+    NO_ENTRY,
+    INTERVENTION
+}
--- a/identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
+++ b/identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
@@ -1,0 +1,75 @@
+package uk.gov.di.orchestration.identity.service;
+
+import com.nimbusds.oauth2.sdk.id.Subject;
+import com.nimbusds.openid.connect.sdk.claims.UserInfo;
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import uk.gov.di.orchestration.identity.entity.LogIds;
+import uk.gov.di.orchestration.identity.entity.SPOTClaims;
+import uk.gov.di.orchestration.identity.entity.SPOTRequest;
+import uk.gov.di.orchestration.shared.api.OidcAPI;
+import uk.gov.di.orchestration.shared.entity.AuthUserInfoClaims;
+import uk.gov.di.orchestration.shared.entity.IdentityClaims;
+import uk.gov.di.orchestration.shared.services.AwsSqsClient;
+import uk.gov.di.orchestration.shared.services.ConfigurationService;
+import uk.gov.di.orchestration.shared.services.SerializationService;
+
+import static uk.gov.di.orchestration.shared.entity.IdentityClaims.VOT;
+
+public class IdentitySPOTService {
+    private static final Logger LOG = LogManager.getLogger(IdentitySPOTService.class);
+    private final ConfigurationService configurationService;
+    private final AwsSqsClient spotSqsClient;
+    private final OidcAPI oidcApi;
+    private final SerializationService objectMapper;
+
+    public IdentitySPOTService(
+            ConfigurationService configurationService,
+            AwsSqsClient spotSqsClient,
+            OidcAPI oidcApi,
+            SerializationService objectMapper) {
+        this.configurationService = configurationService;
+        this.spotSqsClient = spotSqsClient;
+        this.oidcApi = oidcApi;
+        this.objectMapper = objectMapper;
+    }
+
+    public void queueSPOTRequest(
+            LogIds logIds,
+            String sectorIdentifier,
+            UserInfo authUserInfo,
+            Subject pairwiseSubject,
+            UserInfo userIdentityUserInfo,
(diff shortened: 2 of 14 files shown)


identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
```diff
@@ -0,0 +1,9 @@
+package uk.gov.di.orchestration.identity.entity;
+
+public enum ProcessingIdentityStatus {
+    COMPLETED,
+    PROCESSING,
+    ERROR,
+    NO_ENTRY,
+    INTERVENTION
+}
```
CarlyG55, line 9: can I check, what's the difference between where these two are used? how come the first one is missing INTERVENTION?

(I think this is from removal of spot spinner work right but I've forgotten 😅)

identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
```diff
@@ -72,4 +89,44 @@ public void queueSPOTRequest(
...
+                    generateApiGatewayProxyResponse(
+                            302,
+                            "",
+                            Map.of(ResponseHeaders.LOCATION, frontend.ipvCallbackURI().toString()),
+                            null));
+        }
+        return Optional.of(
+                RedirectService.redirectToFrontendErrorPageWithErrorLog(
+                        frontend.errorURI(), new Error("Failed to create redirectURI")));
+    }
```
CarlyG55, line 134: I don't think this is correct? We now hit this block if spot returns processing. originally this was only hit if `redirectUri` wasn't created

CarlyG55: Can we make sure we have a test case for this!

CarlyG55 (commented):
Added what I think is a logic issue in the waitForSpot extraction

identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
cearl1: It _should_ be impossible for `IdentityProgressService.pollForStatus` to return `PROCESSING`. There's a while loop which checks if the staus is `PROCESSING` in the service and if the max number of retries is reached `ERROR` is returned instead

identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
cearl1: Yeah so this is part of the old spinner, when the `ProcessingIdentityHandler` got a processing status of `COMPLETED` it would perform an AIS intervention check. If that returned `BLOCKED` or `SUSPENDED`, then it would return the `INTERVENTION` status.

Since we're doing this intervention check separately in IPVCallbackHandler this isn't needed for the new status enum

CarlyG55: Does the intervention check still happen right at the end of the journey, just to check?

identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
CarlyG55: Hmm - it feels like that's a bit obscured, having one case of the enum handled separately to all the other cases. Is there a way we can handle them all in one place?

identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
cearl1: With the new spinner? Yep, when we get a `COMPLETED` state from the sync wait for spot method we do an intervention check, then redirect back to the RP after.

identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
cearl1: Yeah agreed, the types make it look like PROCESSING can be returned here. I could add an extra conditional to ensure it's handled here? 
The alternative is to either create a new enum for the identity processing end states, or refactor the enum and polling code to not explicitly look for PROCESSING, instead to look for anything thats not a valid end state

identity-shared/src/main/java/uk/gov/di/orchestration/identity/entity/ProcessingIdentityStatus.java
CarlyG55: Okay cool, thanks :)

identity-shared/src/main/java/uk/gov/di/orchestration/identity/service/IdentitySPOTService.java
CarlyG55: Yeah I think we just handle it here for now

cearl1: Done! Also added a test for it 😃

CarlyG55: Sorry to be picky, but I still think I don't love this. It now looks like PROCESSING is an error state if you're looking at it fresh, it's not clear that it's only because you shouldn't have made it here! Can we make that clear somehow?

cearl1: Ahh yep ok, i can maybe throw a runtime exception and comment explicitly saying this should never happen

CarlyG55 (approved):

PR #8644, opened 2026-07-23
AUT-5528: Storing skipped passkey

ben-fernandes-sw:
## What

Added a new `SKIP_ADDING_PASSKEY` function to the `UpdateProfileHandler`. Calling this function will update the `LastSkippedAddingPasskey` field on the UserProfile with the current timestamp.

Also did some refactoring in the handler to make it cleaner to add this feature.

Have left some TODOs in the code for where we need to add the audit event-related code as part of AUT-5464 in the future.

## How to review

1. Code Review
1. Test in authdev2 and ensure that
    1. The new `SKIP_ADDING_PASSKEY` logic is setting the timestamp correctly (this will have to be done by calling the endpoint via API Gateway as the frontend isn't hooked up yet)
    2. The existing `UPDATE_TERMS_CONDS` logic is still working as intended (this can be done by performing a journey in the frontend)

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
@@ -8,7 +8,6 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.frontendapi.entity.UpdateProfileRequest;
-import uk.gov.di.authentication.shared.domain.AuditableEvent;
 import uk.gov.di.authentication.shared.entity.AuthSessionItem;
 import uk.gov.di.authentication.shared.entity.ErrorResponse;
 import uk.gov.di.authentication.shared.entity.UserProfile;
@@ -106,7 +105,6 @@ public class UpdateProfileHandler extends BaseFrontendHandler<UpdateProfileReque
                             userContext.getClientSessionId(), userContext.getTxmaAuditEncoded()));
         }
 
-        AuditableEvent auditableEvent;
         String auditablePhoneNumber =
                 userContext
                         .getUserProfile()
@@ -124,10 +122,10 @@ public class UpdateProfileHandler extends BaseFrontendHandler<UpdateProfileReque
         if (request.getUpdateProfileType().equals(UPDATE_TERMS_CONDS)) {
             authenticationService.updateTermsAndConditions(
                     request.getEmail(), configurationService.getTermsAndConditionsVersion());
-            auditableEvent = AUTH_UPDATE_PROFILE_TERMS_CONDS_ACCEPTANCE;
             LOG.info(
                     "Updated terms and conditions for Version: {}",
                     configurationService.getTermsAndConditionsVersion());
+            auditService.submitAuditEvent(AUTH_UPDATE_PROFILE_TERMS_CONDS_ACCEPTANCE, auditContext);
         } else {
             LOG.error(
                     "Encountered unexpected error while processing session: {}",
@@ -135,7 +133,6 @@ public class UpdateProfileHandler extends BaseFrontendHandler<UpdateProfileReque
             return generateErrorResponse(ErrorResponse.INVALID_UPDATE_PROFILE_TYPE, auditContext);
         }
 
-        auditService.submitAuditEvent(auditableEvent, auditContext);
         return generateEmptySuccessApiGatewayResponse();
     }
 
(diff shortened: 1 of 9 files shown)


frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandlerTest.java
```diff
@@ -117,56 +119,92 @@ void setUp() {
...
+                            auditContextWithOnlyClientSession);
+            verify(auditService)
+                    .submitAuditEvent(
+                            AUTH_UPDATE_PROFILE_TERMS_CONDS_ACCEPTANCE,
+                            auditContextWithAllUserInfo);
+        }
+
+        @Test
+        void
+                checkUpdateProfileTermsCondsAcceptanceAuditEventStillEmittedWhenTICFHeaderNotProvided() {
```
BeckaL: I'm in favour of removing these tests, they give unnecessary importance to something which has not much to do with the handler logic, and cases where the txma header is provided or not already handled at the audit service level

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
```diff
@@ -119,18 +119,23 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
...
+                authenticationService.updateTermsAndConditions(
+                        request.getEmail(), configurationService.getTermsAndConditionsVersion());
+                LOG.info(
+                        "Updated terms and conditions for Version: {}",
+                        configurationService.getTermsAndConditionsVersion());
+                auditService.submitAuditEvent(
+                        AUTH_UPDATE_PROFILE_TERMS_CONDS_ACCEPTANCE, auditContext);
+            }
+            default -> {
+                LOG.error(
```
BeckaL, line 119: While we're here, what do you think about improving this error log? Doesn't seem very descriptive of the actual case (unknown update type)

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
```diff
@@ -110,6 +110,11 @@ public APIGatewayProxyResponseEvent handleRequestWithUserContext(
                 auditService.submitAuditEvent(
                         AUTH_UPDATE_PROFILE_TERMS_CONDS_ACCEPTANCE, auditContext);
             }
+            case SKIP_ADDING_PASSKEY -> {
+                authenticationService.renewLastSkippedAddingPasskeyTimestamp(request.getEmail());
+                LOG.info("Renewed lastSkippedAddingPasskey timestamp");
```
BeckaL, line 115: Are there any failure cases we want to handle here? Acceptable answer = no, we let an exception from the database propagate

BeckaL (dismissed):
Looks sensible to me, a few comments / questions but happy for you to respond or ignore, no dealbreakers

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
ben-fernandes-sw: I think letting it propagate is fine. It's back-office stuff that I think we'd want to know about if it started failing.

On the frontend, I think we should be ignoring any errors that come out of this lambda.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandler.java
ben-fernandes-sw: Yeah, that error message isn't really meaningful any more. Will update to something better.

Side note: We won't ever fall into this branch, assuming we've defined the branches to match the enum. If you try to use a value that doesn't exist in the enum, it'll be caught at the object parsing step. The only time you'd end up here is if we added an extra option to the enum, but forgot to update the switch.

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/UpdateProfileHandlerTest.java
ben-fernandes-sw: As in, remove the `___TICFHeaderNotProvided` tests?

BeckaL: yup, think they create unnecessary noise

BeckaL (approved):

PR #8651, opened 2026-07-27
BAU: Apply new spotless changes to build.gradle files

cearl1:
### Wider context of change

We used to use `v8.5.1` of gradle spotless, however a [recent dependabot update](https://github.com/govuk-one-login/authentication-api/pull/8546) has updated spotless to `v8.7.0`. This version isn't happy with the format of these gradle files for some reason. 

### What’s changed

This PR applies the formatting changes to all build.gradle files (something that pre-commit kept doing every time i tried to commit something 🙃)


cearl1:
False alarm - I recently updated my laptop and it was defaulting to java 21 which is what was causing this issue. Java 17 works!

PR #8653, opened 2026-07-27
ATO-2728: Implement SISCallbackHandler

cearl1:
### Wider context of change

We would like an equivalent to the IPVCallbackHandler for SIS, except when we receive an `access_denied` error from SIS, we would like to redirect the user to IPV instead.

### What’s changed

This PR implements SISCallbackHandler. It's quite a big PR and there is a lot of duplication - i've tried to reduce the duplication by extracting shared identity code, but ive treated this as a first pass at refactoring, hoping that we can iteratively reduce the duplication later (else this PR would never be ready!).

I have split this PR into many commits, each implementing a section of code, alongside the tests that apply to that code. Hopefully that makes it a bit more readable!

### Manual testing

Have tested this using the SIS stub and temporarily adding a redirect to SIS from the auth callback handler. 
- Tested that configuring the stub to send back an `access_denied` error redirects the user to the IPV stub
- Tested that configuring the stub to send back an `access_denied` error with `record_update_requested` error description redirects the user to the IPV stub with the claim `update_identity`
- Tested that configuring the stub to send back a generic error (not `access_denied`) redirects the user to the generic error page
- Tested that configuring the stub to send no error back redirects the user back to the RP, completing the identity journey.

### Checklist

  - This will be done in another ticket ^


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/services/EndOfJourneyService.java
```diff
@@ -110,10 +110,16 @@ public AuthenticationSuccessResponse generateSuccessfulAuthResponse(
...
     public APIGatewayProxyResponseEvent generateAuthenticationErrorResponse(
             AuthenticationRequest authenticationRequest, ErrorObject error) {
+        return generateAuthenticationErrorResponse(authenticationRequest, error, "");
+    }
+
+    public APIGatewayProxyResponseEvent generateAuthenticationErrorResponse(
+            AuthenticationRequest authenticationRequest, ErrorObject error, String additionalLog) {
         LOG.warn(
-                "Error in Authorisation Response. ErrorCode: {}. ErrorDescription: {}.",
+                "Error in Authorisation Response. ErrorCode: {}. ErrorDescription: {}. {}",
```
cearl1, line 127: I did this because i wanted to avoid having 2 log messages when this happened, i'm not sure about it now looking back at it!

It's used for the cross browser checks in this commit: https://github.com/govuk-one-login/authentication-api/pull/8653/changes/ac46c23003fa23f735f912cf09780df6e62f860f#diff-90ac8a3bcb85ab738291ff980aae3de9af01082c4bdafec283735fdadef6d545R208

CarlyG55: It feels like we should try and add it to the error description somehow ideally, but I'm not that fussed really

sis-api/src/main/java/uk/gov/di/orchestration/sis/domain/SISAuditableEvent.java
```diff
@@ -0,0 +1,13 @@
+package uk.gov.di.orchestration.sis.domain;
+
+import uk.gov.di.orchestration.shared.domain.AuditableEvent;
+
+public enum SISAuditableEvent implements AuditableEvent {
+    ORCH_SIS_SUCCESSFUL_AUTHORISATION_RESPONSE_RECEIVED,
```
CarlyG55, line 6: All of our other audit events automatically have AUTH added to the front. Have managed to avoid this here?

cearl1: Yeah I was speaking to Joe about this, I've refactored the AuditService to add a method that doesn't hardcode the `AUTH_` prefix as its a bit clearer to match it up to the audit events in the event catalogue.

Theres a commit called `BAU: Add option to send audit event without AUTH_ prefix`

sis-api/src/main/java/uk/gov/di/orchestration/sis/lambda/SISCallbackHandler.java
```diff
@@ -107,4 +167,43 @@ private IdentityContextResponse getIdentityContext(APIGatewayProxyRequestEvent i
...
     private record IdentityContextResponse(
             IdentityContext identityContext, APIGatewayProxyResponseEvent earlyRedirect) {}
+
+    private Optional<APIGatewayProxyResponseEvent> validateAuthResponse(
+            APIGatewayProxyRequestEvent input,
+            IdentityContext identityContext,
+            AuditContext auditContext,
+            TxmaAuditUser user) {
+        var validationErrorOpt =
+                segmentedFunctionCall(
```
CarlyG55, line 177: We're pulling out x-ray now so we want to get rid of all the segmented function call references!

cearl1: Thanks, will resolve in the new PR https://github.com/govuk-one-login/orchestration-api/pull/40

cearl1:
Have recreated this in https://github.com/govuk-one-login/orchestration-api/pull/40

PR #8654, opened 2026-07-28
Bau/errors in finish assertion

BeckaL:
Narrow down the scope of things that can trigger the PASSKEY_VERIFICATION_FAILED event and metric to just verification failures, rather than anything that can cause the finish passkey assertion to fail (including unexpected errors).

This is so that our metrics and audit events better reflect things that are actually verification failures.

Also adds some logging.


rpjgds (commented):
Agree with not emitting this audit event in these conditions. This will be more useful when there are separate reasons for passkey rejection (insufficient credential strength, PQC etc).

ChristianSoftwire (approved):
LGTM

PR #8656, opened 2026-07-28
Bau: stop emitting audit event for technical error

BeckaL:
Previously, in the case of a failure to generate a valid mfa response in the mfa create handler, we were emitting the add failed audit event. This is despite:

a) it being a very very unlikely edge case that should never happen in theory (it would only happen if the method we added didn't have a known type, but we already validate this on the request validation so it shouldn't ever happen in practice); and
b) in these cases, we would have actually added the mfa method in the database, and in theory it would be useable, so it's more appropriate to emit the add created event.

This enables us to simplify the code and move the creation of the response closer to where it's used, which reduces the level of complexity in the handler.

It's a step towards making the logic in this handler in general more comprehensible, as currently it's difficult to follow.

## How to review
1. Code review


sw-mattw (approved):
Agree with behaviour changes where we've added the method already prior to failure

PR #8693, opened 2026-08-04
BAU: make generic OAuth service

Ryan-Andrews99:
### Wider context of change

We have many OAuth interfaces with different components of One Login such as Auth, IPV, DCMAW and soon to be SIS. Despite this, we have tended to duplicate a lot of the OAuth code especially around making the `/token` and `/userinfo` requests 

### What’s changed
- Adds a generic OAuth service, which is configurable via thin configuration record `OAuthConfiguration`
- Adds several public methods which aim to represent each part of the OAuth flow (making the `/authorize`, `/token` and `/userinfo` requests.
- Adds a couple of helper methods to validate  and store state
- Moves most of the callback validation into this class, with a specific interface providing our handling of the cases in which we're returned an OAuth error, in which case we delegate to our handleError method. This is to account for the fact that sometimes we need to handle specific OAuth error codes 

### Manual testing

- N/A - this is new code that is unit tested and is part of some preparatory refactoring of our existing OAuth interfaces

### Checklist

### DIFF
--- a/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/oauth/OAuthService.java
+++ b/orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/oauth/OAuthService.java
@@ -1,8 +1,11 @@
 package uk.gov.di.orchestration.shared.oauth;
 
+import com.nimbusds.jwt.JWTClaimsSet;
 import com.nimbusds.oauth2.sdk.AuthorizationCode;
 import com.nimbusds.oauth2.sdk.AuthorizationCodeGrant;
+import com.nimbusds.oauth2.sdk.AuthorizationRequest;
 import com.nimbusds.oauth2.sdk.ParseException;
+import com.nimbusds.oauth2.sdk.ResponseType;
 import com.nimbusds.oauth2.sdk.TokenRequest;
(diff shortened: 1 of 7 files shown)


orchestration-shared/src/main/java/uk/gov/di/orchestration/shared/oauth/OAuthService.java
```diff
@@ -0,0 +1,257 @@
...
+    }
+
+    protected final OAuthConfiguration clientConfig;
+    private final OrchJwtService jwtService;
+    private final NowHelper.NowClock clock;
+    private final HTTPRequestSender httpRequestSender;
+    private final StateStorageService stateStorageService;
+    private final CrossBrowserOrchestrationService crossBrowserOrchestrationService;
+    private final CallbackValidator callbackValidator;
+    private final Logger LOG = LogManager.getLogger(this.getClass());
```
cearl1: Can we make this logger static and move it to near the top of the class like we have elsewhere?

Ryan-Andrews99: Done!

cearl1 (approved):
Some very nice refactoring here! 👍

PR #8703, opened 2026-08-05
AUT-5606: migrate remove account to data api

alhcomer:
## What

Move the remove `{env}-delete-account-lambda` to call ADAPI to delete an account rather than interacting with DynamoDB directly. The manual delete and bulk user delete lambdas will follow. Until then, the ADAPI request is only made _if_ a token is present. After we've migrated those lambdas over as well, we will require the presence of a token.

## How to review

1. Code Review
2. Calculate your ICS using the authentication-api `export-ics.sh` script
3. Create the ADAPI token using the following shell commands:

```
export AUTH_TO_ACCOUNT_DATA_SIGNING_KEY="alias/authdev2-auth-to-account-data-signing-key"
export AUTH_TO_ACCOUNT_DATA_API_AUDIENCE="https://account.authdev2.dev.account.gov.uk"
export AUTH_ISSUER_CLAIM="https://signin.authdev2.dev.account.gov.uk"
export AMC_CLIENT_ID="auth"

aws sso login --profile di-authentication-development-AdministratorAccessPermission && eval "$(aws configure export-credentials --profile di-authentication-development-AdministratorAccessPermission --format env)"

python3 ./account-data-api/scripts/generate_account_data_token.py \
    --public-subject-id="<your public subject ID>" \
    --scope "account-delete" \
    --profile di-authentication-development-AdministratorAccessPermission
```
4. Make a request in the {env}-delete-account-lambda test area using the following request template:
```
  {
    "resource": "/delete-account",
    "path": "/delete-account",
    "httpMethod": "POST",
    "headers": {
      "X-ADAPI-AccessToken": "<token retrieved from step 3"
    },
    "requestContext": {
      "requestId": "test-123",
      "authorizer": {
        "principalId": "<ICS calculated in step 2>"
      },
      "identity": {
        "sourceIp": "127.0.0.1"
      }
    },
    "body": "{\"email\": \"<your email>\"}",
    "isBase64Encoded": false
  }
```
5. See the "Successfully deleted account via Data API" message in the lambda logs

## Checklist

### DIFF
--- a/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
+++ b/account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
@@ -10,6 +10,7 @@ import uk.gov.di.audit.AuditContext;
 import uk.gov.di.authentication.auditevents.entity.AuthDeleteAccount;
 import uk.gov.di.authentication.auditevents.services.StructuredAuditService;
 import uk.gov.di.authentication.shared.entity.UserProfile;
+import uk.gov.di.authentication.shared.exceptions.UnsuccessfulAccountDataApiResponseException;
 import uk.gov.di.authentication.shared.helpers.ClientSessionIdHelper;
 import uk.gov.di.authentication.shared.helpers.ClientSubjectHelper;
 import uk.gov.di.authentication.shared.helpers.IpAddressHelper;
@@ -17,6 +18,7 @@ import uk.gov.di.authentication.shared.helpers.LocaleHelper;
 import uk.gov.di.authentication.shared.helpers.PersistentIdHelper;
 import uk.gov.di.authentication.shared.helpers.RequestHeaderHelper;
 import uk.gov.di.authentication.shared.serialization.Json;
+import uk.gov.di.authentication.shared.services.AccountDataApiService;
 import uk.gov.di.authentication.shared.services.AuthenticationService;
 import uk.gov.di.authentication.shared.services.ConfigurationService;
 import uk.gov.di.authentication.shared.services.SerializationService;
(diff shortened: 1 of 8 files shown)


account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
```diff
@@ -77,10 +108,14 @@ public void removeAccount(
         var email = userProfile.getEmail();
 
         LOG.info("Deleting user account");
-        dynamoDeleteService.deleteAccount(
-                email,
-                internalCommonSubjectIdentifier.getValue(),
-                userProfile.getPublicSubjectID());
+        if (accountDataApiToken.isPresent() && accountDataApiService != null) {
+            deleteAccountViaDataApi(userProfile.getPublicSubjectID(), accountDataApiToken.get());
+        } else {
```
sw-mattw, line 113: I'm wondering if we should have an info log within this else block to indicate we're performing a deletion directly on the Dynamo tables? The deletion service itself doesn't have any logs at present

Thinking this might be useful to determine if all requests have the required header (so this code path is safe to remove)?

sw-mattw (dismissed):
Thanks, ran through deletion request both with and without the ADAPI HTTP header and both worked as expected thanks. Added a minor non-blocking comment about logging related to the approach running both deletion paths

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/AccountDeletionService.java
sw-mattw: Thanks

sw-mattw (approved):
Nice, looks good thanks

PR #8716, opened 2026-08-10
BAU: Update dlq runbook link

BeckaL:
Update runbook link for the email notification dlq alarm, as we're deprecating the old rota doc where this was previously contained.

## How to review

Check the old and new link link to equivalent information


sw-mattw (approved):
Nice, thanks

Confirmed new link leads to expected runbook in Confluence

PR #8719, opened 2026-08-10
BAU: Remove X-Ray tracing from Orchestration

Joe-Edwards-GDS:
### Wider context of change

X-Ray was in use for monitoring performance of the service, but was superseded by Dynatrace. The instrumentation code has been switched off for a while, and we see warning/error logs relating to incomplete X-Ray configuration.

We should get rid of the X-Ray specific code and readdress later once the new observability tooling is in place - likely to be OpenTelemetry based for portability across providers.

If we want to restore detailed segment tracing, it may be easier to use something like the [OpenTelemetry annotations package](https://opentelemetry.io/docs/zero-code/java/agent/annotations/) to instrument specific method calls, rather than our own wrappers that tend to mangle exceptions etc.:
```
@WithSpan
public AccessToken callTokenEndpoint() {
    ...
}
```

### What’s changed

This is split into multiple commits, and we could take a subset:
- Remove X-Ray as a dependency
- Set up OpenTelemetry spans instead (needs testing - currently traces are broken in Dynatrace :( )
- Remove all the segmented tracing


oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -393,20 +386,6 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
-                LOG.info(
-                        "Scope parameter in request object contains commas. Client ID: {}",
-                        clientId);
-            }
-        }
-        var queryParamScopes = input.getQueryStringParameters().get("scope");
-        if (queryParamScopes.contains(",")) {
-            LOG.info(
-                    "Scope parameter in query parameters contains commas. Client ID: {}", clientId);
-        }
```
Joe-Edwards-GDS, line 409: (no longer required since the nimbus update)

oidc-api/src/main/java/uk/gov/di/authentication/oidc/lambda/AuthorisationHandler.java
```diff
@@ -445,7 +424,19 @@ public APIGatewayProxyResponseEvent authoriseRequestHandler(
...
+        } catch (java.text.ParseException e) {
+            LOG.error("Failed to parse VTR list", e);
+            return generateErrorResponse(
+                    authRequest.getRedirectionURI(),
+                    authRequest.getState(),
+                    authRequest.getResponseMode(),
+                    OAuth2Error.INVALID_REQUEST,
+                    authRequest.getClientID().getValue(),
+                    user);
+        }
```
Joe-Edwards-GDS, line 439: This is a change of behaviour: previously this checked exception was caught by the instrumentation helper and re-thrown as a runtime exception :(

Joe-Edwards-GDS: Ideally we'd throw sensible exceptions and catch them in a higher level block, but didn't want to overcomplicate things in this PR!

CarlyG55, line 439: we'll need to add this change to simulator

CarlyG55: Given we don't support reauth in the simulator, and this is catching an exception only possible in the case of a reauth journey, going to close this

PR #8631, opened 2026-07-20
AUT-5658: Add initial IAD tracker entity

sw-mattw:
## What

* Introduced the `InactiveAccountTrackerItem` DynamoDB entity and added helper logic to construct it from partial database items.
* Updated the active date calculation to evaluate all timestamp attributes across user profile and user credentials to determine the most recent activity timestamp.
* Added logic to automatically calculate a deletion date (five years after the last active date) and fixed date formats in test data to match production.

NOTE: Work around `shouldNotify` boolean on this item has been split out to AUT-5691 - awaiting clarification on these changes from another team.

## How to review

1. Static code review

## Checklist

- [x] Deployment of this PR will not break active user journeys **- N/A**
- [x] Impact on orch and auth mutual dependencies has been checked. **- N/A**
- [x] No changes required or changes have been made to stub-orchestration. **- N/A**
- [ ] A UCD review has been performed. **- N/A**

PR #8720, opened 2026-08-10
AUT-5733: Use common subject ID in IAD data export

sw-mattw:
## What

The "common subject ID" value in the tracker item was previously being set to the users "subject ID", this was incorrect, the "common subject ID" should be used here.

This change follows the project convention of calculating the "common subject ID" using the `getSubjectWithSectorIdentifier` helper method.

Generates and persists salt on the UserProfile item if missing or empty.

## How to review

1. Code Review
1. Optionally, deploy the utils module to a dev environment, clear the `*-stub-inactive-account-tracker` DynamoDB table for that dev environment, then make a test request to the `*-inactive-account-data-export-lambda` Lambda to scan the dev tables. Then review the dev tracker table items, pick two items, calculate the ICS ID (e.g., using `scripts/export-ics.sh` or another method) from the users subject ID and salt and that dev envs sector URI and check it matches what was written to the dev tracker table item.
    - I have followed these steps - see "Testing" section below

## Testing

1. Deployed the utils module to authdev1
1. Cleared the `authdev1-stub-inactive-account-tracker` DynamoDB table
1. Made a test request to the `authdev1-inactive-account-data-export-lambda` Lambda to scan the dev tables
1. Once the scans had completed, reviewed the authdev1 tracker table items and picked two items:
    a. For each item, calculated the common subject ID (using the users subject ID and salt and authdev1 sector URI (`identity.authdev1.dev.account.gov.uk`)) and checked it matches what was written to the corresponding authdev1 tracker table item.
1. Confirmed the common subject IDs in the dev table all had prefixes `urn:fdc:gov.uk:2022:`

Also confirmed that an account missing a salt on its UserProfile item has one generated and persisted.

## Checklist

- [x] Assessed the impact on active user sessions and confirmed no breaking changes **- N/A, utils lambda change**

- [x] Orch and Auth mutual dependencies have been checked for breaking changes **- N/A, utils lambda change**

- [x] Stub-orchestration has been updated (or no changes required) **- N/A, utils lambda change**

## Related PRs

- PR which introduced the bug: https://github.com/govuk-one-login/authentication-api/pull/8631 ([commit](https://github.com/govuk-one-login/authentication-api/commit/7c94231f1f22a58b4302072f996d53f7d0e155f9))

### DIFF
--- a/ci/cloudformation/utils/function/inactive-account-data-export.yaml
+++ b/ci/cloudformation/utils/function/inactive-account-data-export.yaml
@@ -87,6 +87,13 @@ Resources:
               !Ref Environment,
               inactiveAccountExportMaxInvocations,
             ]
+          INTERNAl_SECTOR_URI: !If
+            - IsNotProduction
+            - !If
+              - UseSubEnvironment
+              - !Sub "https://identity.${SubEnvironment}.${Environment}.account.gov.uk"
+              - !Sub "https://identity.${Environment}.account.gov.uk"
+            - "https://identity.account.gov.uk"
 
       Policies:
         - !Ref LambdaBasicExecutionPolicy
--- a/ci/cloudformation/utils/function/inactive-account-data-export.yaml
+++ b/ci/cloudformation/utils/function/inactive-account-data-export.yaml
@@ -90,7 +90,7 @@ Resources:
 
       Policies:
         - !Ref LambdaBasicExecutionPolicy
-        - !Ref DynamoUserReadAccessPolicy
+        - !Ref DynamoUserReadWriteAccessPolicy
         - !Ref DynamoCredentialReadAccessPolicy
         - !Ref CloudWatchLogsKmsAccessPolicy
         - !Ref InactiveAccountDataExportSelfInvokePolicy
--- a/utils/src/main/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelper.java
+++ b/utils/src/main/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelper.java
@@ -10,10 +10,12 @@ import software.amazon.awssdk.services.dynamodb.model.KeysAndAttributes;
 import software.amazon.awssdk.services.dynamodb.model.UpdateItemRequest;
 import uk.gov.di.authentication.shared.entity.UserCredentials;
 import uk.gov.di.authentication.shared.entity.UserProfile;
+import uk.gov.di.authentication.shared.helpers.ClientSubjectHelper;
 import uk.gov.di.authentication.shared.helpers.NowHelper;
 import uk.gov.di.authentication.shared.helpers.SaltHelper;
 import uk.gov.di.authentication.utils.entity.InactiveAccountTrackerItem;
 
+import java.net.URI;
 import java.time.LocalDateTime;
 import java.util.ArrayList;
 import java.util.HashMap;
@@ -162,7 +164,11 @@ public class InactiveAccountDataExportHelper {
         }
 
         String email = getStringAttribute(userProfileItem, UserProfile.ATTRIBUTE_EMAIL);
-        LOG.info("Generating salt for email '{}': salt was missing or empty", email);
+        String publicSubjectId =
+                getStringAttribute(userProfileItem, UserProfile.ATTRIBUTE_PUBLIC_SUBJECT_ID);
+        LOG.info(
+                "Generating salt for public subject ID '{}': salt was missing or empty",
+                publicSubjectId);
 
         byte[] newSalt = SaltHelper.generateNewSalt();
         dynamoDbClient.updateItem(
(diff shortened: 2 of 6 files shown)


ci/cloudformation/utils/function/inactive-account-data-export.yaml
```diff
@@ -87,6 +87,13 @@ Resources:
               !Ref Environment,
               inactiveAccountExportMaxInvocations,
             ]
+          INTERNAl_SECTOR_URI: !If
```
sw-mattw, line 90: NOTE: Lowercase "l" in "INTERNAl" is not a typo in this PR, this is an existing issue

utils/src/main/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelper.java
```diff
@@ -150,12 +151,31 @@ public static String calculateDateForDeletion(String lastActiveDate) {
...
+        AttributeValue saltAttr = userProfileItem.get(UserProfile.ATTRIBUTE_SALT);
+        if (saltAttr == null || saltAttr.b() == null || saltAttr.b().asByteArray().length == 0) {
+            LOG.warn(
+                    "Skipping tracker item for public subject ID '{}': salt is missing or empty",
+                    publicSubjectId);
+            return null;
+        }
+        byte[] salt = saltAttr.b().asByteArray();
+
+        if (subjectId == null) {
```
sw-mattw: Shouldn't be the case but guarding calculatePairwiseIdentifier which assumes a string and operates on it

ci/cloudformation/utils/function/inactive-account-data-export.yaml
dbes-gds: there are sonarcloud issues for these

dbes-gds (dismissed):

utils/src/main/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelper.java
```diff
@@ -150,12 +151,31 @@ public static String calculateDateForDeletion(String lastActiveDate) {
 
     public static InactiveAccountTrackerItem buildTrackerItem(
             Map<String, AttributeValue> userProfileItem,
-            Map<String, AttributeValue> userCredentialsItem) {
+            Map<String, AttributeValue> userCredentialsItem,
+            String internalSectorUri) {
```
alhcomer, line 193: I think this internalSectorUri needs to be cast to a URI type so that the sector host can be extracted here:

```
    public static String calculatePairwiseIdentifier(String subjectID, URI uri, byte[] salt) {
        var host = returnHost(uri.toString());
        return calculatePairwiseIdentifier(subjectID, host, salt);
    }

    public static String calculatePairwiseIdentifier(
            String subjectID, String sectorHost, byte[] salt) {
        try {
            var md = MessageDigest.getInstance("SHA-256");

            md.update(sectorHost.getBytes(StandardCharsets.UTF_8));
            md.update(subjectID.getBytes(StandardCharsets.UTF_8));

            byte[] bytes = md.digest(salt);

            var sb = Base64Url.encode(bytes);

            return PAIRWISE_PREFIX + sb;
        } catch (NoSuchAlgorithmException e) {
            LOG.error("Failed to hash", e);
            throw new RuntimeException(e);
        }
    }
```

either that or change the env variable to just the sector host (not uri) `identity.${Environment}.account.gov.uk`, but that would be inconsistent with how we do it elsewhere

ci/cloudformation/utils/function/inactive-account-data-export.yaml
```diff
@@ -87,6 +87,13 @@ Resources:
...
+            - IsNotProduction
+            - !If
+              - UseSubEnvironment
+              - !Sub "https://identity.${SubEnvironment}.${Environment}.account.gov.uk"
+              - !Sub "https://identity.${Environment}.account.gov.uk"
+            - "https://identity.account.gov.uk"
```
alhcomer, line 96: these are the URIs i've got in mind that would need to be stripped for the ICS calculation

alhcomer (commented):
Looks good!

Quick thought, what's going to happen to the users without a salt? Do we need some kind of way around this before we go forward this?

ci/cloudformation/utils/function/inactive-account-data-export.yaml
sw-mattw: Thanks, fixed elsewhere

utils/src/main/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelper.java
sw-mattw: Thanks, good spot, have fixed and added a unit test (`buildTrackerItemShouldExtractHostFromUriWhenCalculatingPairwiseIdentifier`) to catch this issue

Think when I was manually testing I previously had the URI.create call and it was present on the deployed version I was testing, but then it got lost along the way (eg rebasing etc). This explains why the ICS IDs in my testing where the same, when they should differ if the protocol is present

sw-mattw: Have retested and the calculated ICS IDs match

alhcomer (approved):

PR #8725, opened 2026-08-11
BAU: Handle unknown passkey transport values

alhcomer:
## What

- Replace AuthenticatorTransport.valueOf() with AuthenticatorTransport.of() in transportStringsToSet(). valueOf() throws IllegalArgumentException for any transport string not matching one of the five known constants (USB, NFC, BLE, HYBRID, INTERNAL). of() returns a matching constant for known values and wraps unknown values in a new instance without throwing:

```
  @JsonCreator
  public static AuthenticatorTransport of(@NonNull String id) {
    return Stream.of(values())
        .filter(v -> v.getId().equals(id))
        .findAny()
        .orElseGet(() -> new AuthenticatorTransport(id));
  }
```

- The WebAuthn spec defines AuthenticatorTransport as a hint for how clients might communicate with an authenticator. The spec uses DOMString rather than a closed enum specifically to allow new values without breaking implementations. The W3C WebAuthn Level 3 spec states in section 5.2.1 that Relying Parties ["SHOULD accept and store unknown values"](https://www.w3.org/TR/webauthn-3/#enum-transport:~:text=credential%20data.-,%5B%5Btransports%5D%5D,of%20AuthenticatorTransport%20but%20Relying%20Parties%20SHOULD%20accept%20and%20store%20unknown%20values.,-5.2.1.1.%20Easily), and in section 5.8.3 that client platforms ["MUST ignore unknown values"](https://www.w3.org/TR/webauthn-3/#enum-transport:~:text=%5B%5Btransports%5D%5D,%2C%20but%20client%20platforms%20MUST%20ignore%20unknown%20values.). This means unknown transports are silently skipped during authenticator discovery but do not prevent the credential from being used.

- The immediate cause was passkey records stored with transport value "cable" (an early pre-standard identifier for what became "hybrid" in the W3C spec). Some browsers and authenticators stored "cable" during registration before the standard settled on "hybrid". When a user with such a record attempted to sign in, startAssertion called getCredentialIdsForUsername, which called transportStringsToSet, which called valueOf("CABLE") and threw. The unhandled exception crashed the Lambda and returned a 500, completely blocking that user from signing in with their passkey.

- This triggered the production-start-passkey-assertion-error-rate alarm on 10/08/2026 at 18:58 (10% error rate threshold breached within a 1-minute window, caused by 5+ failures from repeated retries by affected users).

- Impact: users whose passkeys were registered with "cable" (or any other non-standard transport string) can now sign in successfully. The unknown transport value passes through to the allowCredentials response as a transport hint. The browser ignores hints it does not recognise and discovers the authenticator through its default mechanisms. The credential ID is still present in allowCredentials, so the assertion ceremony proceeds normally.

- No impact on users with standard transport values. of() returns the same constant instances as valueOf() for known values.

See: https://www.w3.org/TR/webauthn-3/#enum-transport
See: https://www.w3.org/TR/webauthn-3/#dom-publickeycredentialdescriptor-transports
See: https://developers.yubico.com/java-webauthn-server/JavaDoc/webauthn-server-core/latest/com/yubico/webauthn/data/AuthenticatorTransport.html#of(java.lang.String)

## How to review

1. Code Review
2. Wipe your authenticators in your test environment
3. Register a passkey using the `generate_webauthn_script.py` script and running the output in dev tools console on sign in and create page, but change this line:
```
const transports = (response.getTransports ? response.getTransports() : (credential.getTransports ? credential.getTransports() : ["internal"]));

```
to: 
```
const transports = ["cable"];
```
4. Create a record in the authenticators table by copying the output of running the dev tools script
5. Log in, see you are taken to https://signin.dev.account.gov.uk/sign-in-passkey successfully. 

## Checklist

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
@@ -63,7 +63,7 @@ public class AccountDataCredentialRepository implements CredentialRepository {
 
     Set<AuthenticatorTransport> transportStringsToSet(List<String> passkeyTransports) {
         return passkeyTransports.stream()
-                .map(s -> AuthenticatorTransport.valueOf(s.toUpperCase()))
+                .map(AuthenticatorTransport::of)
                 .collect(Collectors.toSet());
     }
 
--- a/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepositoryTest.java
+++ b/frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepositoryTest.java
@@ -240,7 +240,7 @@ class AccountDataCredentialRepositoryTest {
                 "some-aaguid",
                 true,
                 5,
-                List.of("BLE", "internal"),
+                List.of("ble", "internal"),
                 true,
                 true,
                 true,
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/StartPasskeyAssertionHandlerIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/StartPasskeyAssertionHandlerIntegrationTest.java
@@ -143,7 +143,7 @@ class StartPasskeyAssertionHandlerIntegrationTest extends ApiGatewayHandlerInteg
                     passkeysResponse(
                             passkeyJson(FIRST_PASSKEY_ID),
                             passkeyJsonWithTransports(
-                                    SECOND_PASSKEY_ID, List.of("BLE", "INTERNAL"))));
+                                    SECOND_PASSKEY_ID, List.of("ble", "internal"))));
 
             var response =
                     makeRequest(


frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/services/webauthn/AccountDataCredentialRepository.java
```diff
@@ -63,7 +63,7 @@ private Optional<PublicKeyCredentialDescriptor> toCredentialDescriptor(
 
     Set<AuthenticatorTransport> transportStringsToSet(List<String> passkeyTransports) {
         return passkeyTransports.stream()
-                .map(s -> AuthenticatorTransport.valueOf(s.toUpperCase()))
+                .map(s -> AuthenticatorTransport.of(s.toLowerCase()))
```
dbes-gds: see this switched from toUpperCase to toLowerCase

alhcomer: Oops the PR was stale. Pushed now. I've removed any case manipulation. The enum values sent from the browser are inherently lower case: https://www.w3.org/TR/webauthn-3/#enum-transport:~:text=enum%20AuthenticatorTransport)-,enum%20AuthenticatorTransport%20%7B%0A%20%20%20%20%22usb%22%2C%0A%20%20%20%20%22nfc%22%2C%0A%20%20%20%20%22ble%22%2C%0A%20%20%20%20%22smart%2Dcard%22%2C%0A%20%20%20%20%22hybrid%22%2C%0A%20%20%20%20%22internal%22%0A%7D%3B,-NOTE%3A%20The

dbes-gds (approved):

PR #8794, opened 2026-08-20
BAU: DO NOT MERGE: Filter AZ data source to exclude eu-west-2d (Redis resources)

sw-mattw:
## What

Filter AZ data source to exclude eu-west-2d.

Used by the two ElastiCache replication groups - [code link](https://github.com/search?q=repo%3Agovuk-one-login%2Fauthentication-api+preferred_cache_cluster_azs&type=code).

Old account pipelines (only data stores in use there now)

My current understanding is that this is zero-downtime as it's filtering down to 3 AZs which I believe is what we're already running

## How to review

1. Code Review

## Testing

1. Deployed same "core" networking infra fix ([PR](https://github.com/govuk-one-login/infrastructure/pull/650)) to dev.
1. Then deployed this fix to dev (using the workflow)
1. Observed a passing pipeline
1. Reviewed the deploy AM action, observed `Apply complete! Resources: 0 added, 0 changed, 0 destroyed.`

My current understanding is that this is zero-downtime as it's filtering down to 3 AZs which I believe is what we're already running

Screenshot 2026-08-20 at 15 54 29

Screenshot 2026-08-20 at 15 54 51

### DIFF
--- a/ci/terraform/account-management/site.tf
+++ b/ci/terraform/account-management/site.tf
@@ -39,4 +39,9 @@ data "aws_caller_identity" "current" {}
 
 data "aws_partition" "current" {}
 
-data "aws_availability_zones" "available" {}
+data "aws_availability_zones" "available" {
+  filter {
+    name   = "zone-name"
+    values = ["eu-west-2a", "eu-west-2b", "eu-west-2c"]
+  }
+}
--- a/ci/terraform/shared/site.tf
+++ b/ci/terraform/shared/site.tf
@@ -70,4 +70,9 @@ locals {
 
 data "aws_caller_identity" "current" {}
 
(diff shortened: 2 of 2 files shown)


alhcomer (approved):

ci/terraform/account-management/site.tf
```diff
@@ -39,4 +39,9 @@ data "aws_caller_identity" "current" {}
```
sw-mattw: This issue is no longer blocking the pipeline so asking @pskushwaha1 to take a look at these changes prior to merging, thanks

sw-mattw, line 39: Comment to block merge

PR #8811, opened 2026-08-21
BAU: Populate Gradle build cache on push to main

alhcomer:
## What

- Add `workflow_dispatch` so GH cache can be populated manually
- Remove `--dry-run` from gradlew test invocations in warm-configuration-cache.yml. This will trigger a full gradle build, populating the build cache on push to main
- As this happens on push to main, this happens silently in the background _after_ a PR has gone through the merge queue successfully. This will _not_ block merge to main, even if this approach doesn't work as planned.
- This will hopefully mean that on new PRs, if given gradle submodules have not changed (and so the hashes of module contents the gradle/GH cache uses as a key have not changed), the test job will skip and refer to a cached test output
- Even though integration tests are treated as their own modules, their hash is calculated based on the contents of modules they are dependent on. As an example, if ADAPI integration tests do not change but ADAPI application code changes, the integration tests will run due to the `testImplementation project(":account-data-api")` line in the ADAPI integration tests build.gradle file.
- Also removed the run-code-analysis job as this runs in the sonar-analysis-on-merge workflow, which also runs on push to main, so this is redundant.

## How to review

1. Code Review

## Checklist

### DIFF
--- a/.github/workflows/warm-configuration-cache.yml
+++ b/.github/workflows/warm-configuration-cache.yml
@@ -5,6 +5,7 @@ env:
   GRADLE_ENCRYPTION_KEY: ${{ secrets.GRADLE_ENCRYPTION_KEY }}
 
 on:
+  workflow_dispatch:
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


csophocleous (approved):
LGTM

PR #8781, opened 2026-08-18
AUT-5609: Create initial IAD queue processor

sw-mattw:
## What

Added new SQS queue processor Lambda to consume IAD messages.

A new IAD token service mints access tokens for authenticating with the Account Data (AD) API. A new `inactive-account-deletion` client ID has been added and the AD API authoriser has been updated to allow this.

The queue processor triggers account deletion via the AD API for the public subject ID in the SQS message.

Added a stub SQS queue for the dev and build environments for testing.

A future PR under a subsequent ticket will link up with the home-owned SQS queue in staging and above. This queue processor will only be attached to the stub and home-owned queues, and will not be attached to an API gateway.

## How to review

1. Code review, suggested commit-by-commit
1. Optionally, deploy to a dev environment, identify an account you would like to delete in that dev environments `user-profile` DB table, copy the publicSubjectId value for that account, then within the AWS Console, locate the stub queue (not the DLQ!) and add a message with that ID of the format `{"publicSubjectId": "ID_HERE"}` and send, then review the Lambda logs and check the account has been deleted.

## Testing

I followed the above steps on authdev1 and confirmed the token was minted OK and the account was deleted successfully via the AD API.

deletion logs

## Checklist

- [x] Deployment of this PR will not break active user journeys

- [x] Impact on orch and auth mutual dependencies has been checked. **- N/A**

- [x] No changes required or changes have been made to stub-orchestration. **- N/A**

PR #8821, opened 2026-08-26
AUT 5607: Migrate `ManuallyDeleteAccountHandler` and `BulkRemoveAccountHandler` to Call Data API

Q-GDS:
## What
  
  Migrates the manual and bulk account deletion lambdas (`ManuallyDeleteAccountHandler`,
  `BulkRemoveAccountHandler`) to call the Data API instead of writing to DynamoDB
  directly, matching how user-initiated deletion already works.
  
  Neither lambda has a user session, so each mints its own short-lived access token
  via a new `AccountDeletionTokenService`, signed with the existing
  auth-to-account-data KMS key. Each lambda uses its own client ID
  (`manual-account-deletion`, `bulk-account-deletion`), added to the Data API
  authoriser's allow list alongside the existing AMC, Home and
  inactive-account-deletion clients.
  
  The token is passed through the existing `removeAccount` call rather than
  bypassing it, so the audit event, legacy SNS publish and (for manual deletion)
  the user notification email are all preserved.
  
  Gated behind `ACCOUNT_DELETION_DATA_API_ENABLED`, on for authdev1, authdev2,
  authdev3, dev and build, off for staging, integration and production pending
  sign-off.

  The existing DynamoDB path is kept in place, since the feature flag's fallback depends on it; 
  removing it is deferred   to the cleanup ticket that also removes the flag, as the ticket specifies.
  
  ## How to review
  
  1. Code review.
  2. Deployed and manually tested on authdev2 for both lambdas (manual delete and
     bulk delete), including the flag toggle. Logs confirm token minting, KMS
     signing, authoriser acceptance and a successful Data API response for both
     client IDs.
  3. Optionally, redeploy to a dev environment and repeat: invoke
     `manually-delete-account-lambda` with a test email, or
     `bulk-remove-account-lambda` with a `BulkUserDeleteRequest` payload, and
     check CloudWatch for `AccountDeletionService` logging
     "Successfully deleted account via Data API".
  
  ## Checklist
  
  - [x] Orch and Auth mutual dependencies have been checked for breaking changes - N/A
  - [x] Stub-orchestration has been updated (or no changes required) - no changes required
  - [x] A UCD review has been performed (or no review required) - no review required, no user-facing or frontend change
  
  ## Related PRs
  
  Builds on #8781 (AUT-5609), which added the `inactive-account-deletion` client
  ID and the `kms:Sign` policy pattern this PR follows for its own two client IDs.

### DIFF
--- a/account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandlerTest.java
+++ b/account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandlerTest.java
@@ -144,48 +144,17 @@ class AuthorizeHandlerTest {
                     JsonParser.parseString(new Gson().toJson(result)));
         }
 
-        @Test
-        void authorizeHandlerShouldAllowInactiveAccountDeletionClientId() throws JOSEException {
-            var handler = new AuthorizeHandler(configurationService, remoteJwksService);
-
-            var bearerAccessToken =
-                    createBearerAccessTokenWithExpiry(
-                            expiryDateFiveMinutesFromNow,
-                            ecSigningKey,
-                            AccountDataScope.ACCOUNT_DELETE.getValue(),
-                            INACTIVE_ACCOUNT_DELETION_CLIENT_ID);
-
-            event.setAuthorizationToken(bearerAccessToken.toAuthorizationHeader());
-
-            var result = handler.handleRequest(event, context);
-
-            var expectedPolicyDocument =
-                    """
-                            {
-                                "principalId": "%s",
-                                "context": {"scope": "%s"},
-                                "policyDocument": {
-                                    "version": "2012-10-17",
-                                    "statement": [{
-                                        "action": "execute-api:Invoke",
-                                        "effect": "Allow",
-                                        "resource": ["%s"]
-                                    }]
-                                }
-                            }
-                            """
-                            .formatted(
-                                    SUBJECT,
-                                    AccountDataScope.ACCOUNT_DELETE.getValue(),
-                                    METHOD_ARN);
-
-            assertEquals(
-                    JsonParser.parseString(expectedPolicyDocument),
-                    JsonParser.parseString(new Gson().toJson(result)));
+        static Stream<Arguments> deletionClientIds() {
+            return Stream.of(
+                    Arguments.of(INACTIVE_ACCOUNT_DELETION_CLIENT_ID),
+                    Arguments.of(MANUAL_ACCOUNT_DELETION_CLIENT_ID),
+                    Arguments.of(BULK_ACCOUNT_DELETION_CLIENT_ID));
         }
 
-        @Test
-        void authorizeHandlerShouldAllowManualAccountDeletionClientId() throws JOSEException {
+        @ParameterizedTest
+        @MethodSource("deletionClientIds")
+        void authorizeHandlerShouldAllowAccountDeletionClientIds(String clientId)
+                throws JOSEException {
             var handler = new AuthorizeHandler(configurationService, remoteJwksService);
 
             var bearerAccessToken =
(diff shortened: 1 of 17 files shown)


account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandlerTest.java
```diff
@@ -177,6 +183,86 @@ void authorizeHandlerShouldAllowInactiveAccountDeletionClientId() throws JOSEExc
...
+
+            assertEquals(
+                    JsonParser.parseString(expectedPolicyDocument),
+                    JsonParser.parseString(new Gson().toJson(result)));
+        }
+
+        @Test
+        void authorizeHandlerShouldAllowBulkAccountDeletionClientId() throws JOSEException {
+            var handler = new AuthorizeHandler(configurationService, remoteJwksService);
+
```
alhcomer: Could this test be merged to be a parameterized test with `authorizeHandlerShouldAllowManualAccountDeletionClientId()` and `authorizeHandlerShouldAllowInactiveAccountDeletionClientId()`? 

There's an example of one [here](https://github.com/govuk-one-login/authentication-api/blob/85fd06f7731cd379c6e3790d0439d8c353ce37fe/account-management-api/src/test/java/uk/gov/di/accountmanagement/lambda/AuthenticateHandlerTest.java#L256-L291)

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
```diff
@@ -55,19 +72,34 @@ public DeletedAccountIdentifiers manuallyDeleteAccount(
...
+        var tokenResult =
+                accountDeletionTokenService.createAccountDataApiAccessToken(
+                        userProfile.getPublicSubjectID(), clientId);
+        if (tokenResult.isFailure()) {
+            throw new RuntimeException(
+                    "Failed to mint account-delete token for publicSubjectId: "
+                            + userProfile.getPublicSubjectID());
         }
+        return Optional.of(tokenResult.getSuccess().getValue());
     }
```
alhcomer, line 110: Can we pass through Result here rather than retrieving the success/failure and re-wrapping in an Optional? Can talk about this further if you're unfamiliar with the Result pattern

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
```diff
@@ -55,19 +72,34 @@ public DeletedAccountIdentifiers manuallyDeleteAccount(
...
+        return accountIdentifiers;
+    }
+
+    private Optional<String> getAccountDataApiToken(UserProfile userProfile) {
+        if (accountDeletionTokenService == null) {
+            return Optional.empty();
+        }
+        var tokenResult =
+                accountDeletionTokenService.createAccountDataApiAccessToken(
+                        userProfile.getPublicSubjectID(), clientId);
```
alhcomer: nitpick: imo we should avoid using var to type if it's not clear what the method returns (`toString()` obviously returns `String`, as a trivial example). I think typing as the explicit `Result<JwtFailureReason, BearerAccessToken>` is clearer and easier to navigate in the IDE

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
```diff
@@ -55,19 +72,34 @@ public DeletedAccountIdentifiers manuallyDeleteAccount(
...
-                    sendNotification);
-            var deletedAccountPayload =
-                    SerializationService.getInstance()
-                            .writeValueAsString(legacyAccountDeletionMessage);
-            legacyAccountDeletionSnsClient.publish(deletedAccountPayload);
-            return accountIdentifiers;
-        } catch (Json.JsonException e) {
+                    sendNotification,
+                    getAccountDataApiToken(userProfile));
+        } catch (RuntimeException e) {
```
alhcomer, line 80: This is my fault as I wrote `deleteAccountViaDataApi()` where this RuntimeException is thrown, but it feels unclear as to where this `RuntimeException` is coming from. Might be worth defining our own exception. Not gonna block the PR on this though as it's my doing and not in scope for this ticket specifically

ci/cloudformation/account-management/parent.yaml
```diff
@@ -496,6 +500,7 @@ Mappings:
       inactiveAccountDeletionClientId: inactive-account-deletion
       authToAccountDataSigningKeyArn: arn:aws:kms:eu-west-2:058264536367:key/9a0c6ef7-ce45-42fe-b55e-48b09f9ff53e
       manualAccountDeletionClientId: manual-account-deletion
+      accountDeletionDataApiEnabled: true
```
alhcomer: Should this be false for the time being?

semi-separately, as a general rule build should mimic production. staging is where we have a more "in-dev" state. Doesn't matter so much for this ticket as it deletes the account either way but for future reference

alhcomer (commented):
Really good PR btw. Great commit messages and really well structured. Nice

account-data-api/src/test/java/uk/gov/di/authentication/accountdata/lambda/AuthorizeHandlerTest.java
Q-GDS: Done, merged into one parameterised test over the three client ids, matching `validScopes` in the same file. Thanks for the pointer.

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
Q-GDS: Using `fold` now instead of unwrapping. Also folded in the `JwtFailureReason` to the exception message while I was in there since it wasn't being surfaced before.

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
Q-GDS: Agreed, same commit as the `fold` change.

account-management-api/src/main/java/uk/gov/di/accountmanagement/services/ManualAccountDeletionService.java
Q-GDS: agreed this would be good to fix and it lines up with what we found while checking error handling, nothing in this path is a typed exception. Happy to raise a follow up ticket covering both rather than fix it in this PR since it touches behaviour beyond just these two lambdas.

ci/cloudformation/account-management/parent.yaml
Q-GDS: Flipped build to false. On the general rule, checked testClientsEnabled/testSigningKeyEnabled in this file, they're inconsistent with it (both true for build and staging), so worth a separate chat if we want to standardise.

alhcomer: These ones are an exception for tests specifically which run in build and staging

alhcomer (approved):

PR #8854, opened 2026-09-03
BAU: Removed reference to old Pager duty sns and SSM client id

pskushwaha1:
## What

BAU: Removed reference to old Pager duty sns and SSM client id

OIDC code is not functional so removing alerting and reference to  SSM parameter , we have a tidy PR to follow to remove Legacy resource 

## How to review

Code review


sw-mattw (approved):
Looks sensible, old account, thanks

PR #8859, opened 2026-09-07
BAU: Auto-start Docker Desktop when running composeUp locally

sw-mattw:
## What

Dev quality-of-life improvement. Checks if Docker is running before running composeUp locally (e.g., when running integration tests), and attempts to start Docker Desktop (via the bash script added in this PR) if not.

Mirrors existing pattern added in the frontend - https://github.com/govuk-one-login/authentication-frontend/pull/3442

Tried to add guards to stop this ever running in CI / GitHub workflows/actions / CodeBuild jobs as a safeguard, my current understanding of each "env" is:
- Locally: composeUp runs as a dependency of integration tests. This is where I want the new "Docker running check" to run.
- GitHub Actions:
    - In pre-merge-checks integration tests are run with `-x composeUp` to skip `composeUp`.
    - Deployment workflows only run buildZip or assemble, which have no dependencies on composeUp
- CodeBuild: These only deal with the artifact uploaded in GHA

<details>

<summary>Before</summary>

Before

</details>

<details>

<summary>After</summary>

After

</details>

## How to review

1. Code Review
1. Checkout main locally, quit Docker Desktop (or ensure stopped), try to run an integration test, should fail to run due to Docker not running. Checkout this branch, ensure Docker Desktop stopped, repeat prior steps, should automatically start Docker Desktop and the integration tests should pass.
    - I've ran through this - see the "Testing" section below.

## Testing

Ran through the above testing steps locally.

Also deployed this branch to dev to check everything looks OK there.

## Checklist

## Related PRs

- Mirrors existing pattern added in the frontend - https://github.com/govuk-one-login/authentication-frontend/pull/3442

### DIFF
--- a/scripts/ensure-docker-running.sh
+++ b/scripts/ensure-docker-running.sh
@@ -1,0 +1,39 @@
+#!/usr/bin/env bash
+set -eu
+
+# Guard: only run locally, not in CI or deployed environments
+if [ -n "${CI:-}" ] || [ -n "${GITHUB_ACTIONS:-}" ] || [ -n "${CODEBUILD_BUILD_ID:-}" ]; then
+  echo "CI environment detected, skipping Docker check."
+  exit 0
+fi
+
+if docker info > /dev/null 2>&1; then
+  echo "Docker daemon is already running."
+  exit 0
+fi
+
+echo "Docker daemon is not running. Attempting to start Docker Desktop..."
+
+if [ "$(uname -s)" != "Darwin" ]; then
+  echo "Error: Docker is not running. Please start Docker manually." >&2
+  exit 1
+fi
+
+open -a Docker
+
+DOCKER_STARTUP_TIMEOUT=60
+DOCKER_POLL_INTERVAL=3
+elapsed=0
+
+echo "Waiting for Docker to start..."
+while ! docker info > /dev/null 2>&1; do
+  sleep ${DOCKER_POLL_INTERVAL}
+  elapsed=$((elapsed + DOCKER_POLL_INTERVAL))
+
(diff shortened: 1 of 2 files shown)


Q-GDS (approved):

scripts/ensure-docker-running.sh
```diff
@@ -0,0 +1,39 @@
...
+echo "Waiting for Docker to start..."
+while ! docker info > /dev/null 2>&1; do
+  sleep ${DOCKER_POLL_INTERVAL}
+  elapsed=$((elapsed + DOCKER_POLL_INTERVAL))
+
+  if [ ${elapsed} -ge ${DOCKER_STARTUP_TIMEOUT} ]; then
+    echo "Error: Docker failed to start within ${DOCKER_STARTUP_TIMEOUT}s" >&2
+    exit 1
+  fi
+done
```
sw-mattw, line 37: Same retry behaviour as added to the frontend - keep checking it's running

scripts/ensure-docker-running.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env bash
+set -eu
+
+# Guard: only run locally, not in CI or deployed environments
+if [ -n "${CI:-}" ] || [ -n "${GITHUB_ACTIONS:-}" ] || [ -n "${CODEBUILD_BUILD_ID:-}" ]; then
+  echo "CI environment detected, skipping Docker check."
+  exit 0
+fi
```
sw-mattw, line 8: Added two sets of guards: here and in the Gradle file itself. See PR desc for why I don't think we should hit these (although I think we should still have them in place)

PR #8867, opened 2026-09-08
Aut 5809/allow retrieval of session with previous session

BeckaL:
## What

If we receive a double request for an authenticated user to the start handler in quick succession, this often means the user is not silently logged in.

This happens because our logic for handling requests for sessions with previous sessions is to look up the previous session from the database. If it exists, we copy it over to the new session and delete the previous session. If it doesn't exist, we create a new session.

In the double request scenario this plays out as follows:

* Two requests are received for session `b` with previous authenticatied session `a`
* In the first request, we look up session `a` from the database. It exists, so we create a new session `b` and copy over the details from `a`. `a` is no longer needed, so we delete it.
* In the second request, we look up session `a` from the database. It no longer exists. We then create an entirely new session `b`, overwriting what we did in the first request.
* This means that the first request will return authenticated = false to the frontend, since the session `b` does not have an email address on it, meaning we need to make the user fully log in.

This PR fixes this, by allowing the second request to see if session `b` can be retrieved from the database already, before creating a new one.

We allow reuse of session `b` if:
* it has a previous session `a`; and
* it was created less than 1 second ago

These measures provide additional assurance that we are dealing with duplicate requests. The time limit may be reviewed in future as it's possible we can be more flexible.

### DIFF
--- a/integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
+++ b/integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
@@ -283,6 +283,66 @@ class StartIntegrationTest extends ApiGatewayHandlerIntegrationTest {
     }
 
     @Test
+    void shouldHandleMultipleDuplicateRequestsWhenThereIsAnExistingPreviousSession()
+            throws Json.JsonException {
+        var userEmail = "joe.bloggs+3@digital.cabinet-office.gov.uk";
+        var isAuthenticated = true;
+        var sessionId = IdGenerator.generate();
+        authSessionExtension.addSession(PREVIOUS_SESSION_ID);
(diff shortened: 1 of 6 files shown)


integration-tests/src/test/java/uk/gov/di/authentication/api/StartIntegrationTest.java
```diff
@@ -282,6 +282,66 @@ void shouldReturn200WithCorrectMfaMethodTypeWhenTheseIsAnExistingSession(
                 txmaAuditQueue, List.of(AUTH_START_INFO_FOUND));
     }
 
+    @Test
+    void shouldHandleMultipleDuplicateRequestsWhenTheseIsAnExistingPreviousSession()
```
csophocleous: Small typo in test name

csophocleous (approved):
LGTM - happy to ignore sonar as the code it is saying isn't covered is

PR #8869, opened 2026-09-09
AUT-5795: Stop overwriting entire session when doing session updates in StartPasskeyAssertionHandler

csophocleous:
## What

We were seeing issues in the StartPasskeyAssertionHandler where a duplicate request was causing the `hasVerifiedWithPasskey` attribute to be overwritten in the session even though the user successfully signed in with their passkey. This means they were not issued an auth code when they should have been. 

To fix this, I am only updating the relevant property on the session in the StartPasskeyAssertionHandler (rather than updating the full session) to prevent stale sessions causing issues.

## How to review

1. Code review

### DIFF
--- a/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartPasskeyAssertionHandler.java
+++ b/frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartPasskeyAssertionHandler.java
@@ -130,10 +130,8 @@ public class StartPasskeyAssertionHandler extends BaseFrontendHandler<StartPassk
                     500, ErrorResponse.UNEXPECTED_INTERNAL_API_ERROR);
         }
 
-        authSessionService.updateSession(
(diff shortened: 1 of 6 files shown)


shared/src/test/java/uk/gov/di/authentication/shared/services/AuthSessionServiceTest.java
```diff
@@ -91,6 +94,31 @@ void updateSessionThrowsAnyDynamoExceptions() {
                 () -> authSessionService.updateSession(sessionToBeUpdated));
     }
 
+    @Test
+    void shouldBuildUpdateItemRequestAndCorrectlyUpdatePasskeyAssertionRequest() {
```
BeckaL: Minor: I don't think this test is worth having. I largely think unit tests for a database layer are not valuable except for cases where the database throws an exception

BeckaL (approved):

PR #8877, opened 2026-09-10
AUT-5797: Make user perform login when a user orch say is authenticated will not pass auth code protection

Q-GDS:
## What

When orch reports a user as authenticated but auth's own session state disagrees (`canIssueAuthCode` returns false), `StartHandler` previously only logged a warning and returned `authenticated=true` anyway. This sent users into a silent login that failed later with an auth code protection 500, no way back to the RP.

Override `authenticated` to `false` in the Start response instead, so the user is sent to log in again rather than hitting that failure. Uplift journeys stay excluded as before. Reauth journeys are excluded too, because the branch above clears the verified flags and would otherwise trip the override on every request.

The flag is worked out before `startService.buildUserStartInfo(...)` is called and passed in, so `UserStartInfo` is built once rather than built and then rebuilt with one field changed.

Clear `hasVerifiedWithPassword`, `hasVerifiedWithMfa` and `hasVerifiedWithPasskey` on the stored session when the override fires, so the stale claim does not survive into the next journey. The session is written a second time rather than moving the existing `addSession` below the decision, because `canIssueAuthCode` has to read the flags while they are still set. The reauth branch above already cleared the same three fields, so both now call one method.

Replaced the existing disagreement log with one that says the override fired. It does not repeat the credential strengths or the `hasVerifiedWithX` flags, because `PermissionDecisionManager.canIssueAuthCode` already logs all five of those, plus the specific reason it blocked, on the lines immediately before.

Added a new `CloudwatchMetrics` entry, emitted when the override fires, to track how often this happens in production.

Split `handleRequest` into named methods per phase (session setup, audit context, reauth lockout, the new override). It was 234 lines and already over the cognitive complexity and method length thresholds the Sonar quality gate enforces before this change, and adding the override made it worse. Its own commit, no behaviour change.

Also fixed three `StartHandlerIntegrationTest` cases that built sessions with no verification state and asserted `authenticated=true` came back regardless, an assumption that only held because of this bug. Gave all three a genuinely verified session instead of changing what they assert, since none of them is about the disagreement. One of those, `shouldHandleMultipleDuplicateRequestsWhenThereIsAnExistingPreviousSession`, arrived with AUT-5809 after this branch was written, which is why someone else's test appears in the diff.

Out of scope: suppressing `canIssueAuthCode`'s diagnostic logging.

## How to review

1. Code review.
2. Deploy to a dev environment using the appropriate [GitHub dev deployment workflow](https://github.com/govuk-one-login/authentication-api/actions) (Deploy Internal API, External API, IPV API and Delivery Receipts API, since `StartHandler` lives in `frontend-api`).
3. Reproduce the ticket's manual simulation: log in once, clear the verified fields on that session (or use the orch stub's "authenticated" option if testing against an orch stub rather than an RP stub, since silent login only fires automatically with an RP stub), start a new journey marked as authenticated.
4. Confirm you're sent to log in again instead of hitting an auth code failure.
5. Check the new `AuthStartOverriddenAuthenticated` CloudWatch metric fires, and the new log line (`"Auth code protection did not pass, overriding authenticated to false"`) appears, along with `canIssueAuthCode`'s own log lines just before it.

The wipe will not be visible in that simulation. Clearing the verified fields by hand is what sets the scenario up, so they are already false and clearing them again changes nothing on screen. To see the wipe do something, use a session that is partially verified, password verified and MFA not, on a `Cl.Cm` request. The integration test covers that case.

## Checklist


integration-tests/src/test/java/uk/gov/di/authentication/api/StartHandlerIntegrationTest.java
```diff
@@ -289,6 +340,9 @@ void shouldHandleMultipleDuplicateRequestsWhenThereIsAnExistingPreviousSession()
         var sessionId = IdGenerator.generate();
         authSessionExtension.addSession(PREVIOUS_SESSION_ID);
         authSessionExtension.addEmailToSession(PREVIOUS_SESSION_ID, userEmail);
+        authSessionExtension.addAchievedCredentialTrustToSession(PREVIOUS_SESSION_ID, MEDIUM_LEVEL);
+        authSessionExtension.addHasVerifiedWithPasswordAndMfaToSession(
+                PREVIOUS_SESSION_ID, true, true);
```
Q-GDS, line 342: This test arrived after this branch was written. It built a session with no verification state and asserted authenticated=true, which only held because of the bug this PR fixes. Given it a verified session rather than changing what it asserts, since it is about duplicate request handling.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -341,6 +165,226 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+        }
+
+        var authenticated = isUserAuthenticatedWithValidProfile;
+        if (authenticated
+                && !reauthenticate
+                && !upliftRequired
+                && !permissionDecisionManager.canIssueAuthCode(authSession)) {
+            authenticated = false;
+            clearVerifiedState(authSession);
+            authSessionService.updateSession(authSession);
```
Q-GDS, line 214: canIssueAuthCode has to read the flags while they are still set and addSession at 178 runs before that so the wipe cannot ride on the existing write.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -341,6 +165,226 @@ public APIGatewayProxyResponseEvent handleRequest(
...
+                        requestedCredentialTrustLevel, authSession.getAchievedCredentialStrength());
+        authSessionService.addSession(authSession.withUpliftRequired(upliftRequired));
+
+        var userContext = startService.buildUserContext(authSession);
+        var clientStartInfo = buildClientStartInfo(startRequest, authSession);
+
+        var reauthenticate = isReauthenticateRequest(input);
+        if (reauthenticate) {
+            LOG.info("Reauthentication - clearing verified state");
+            clearVerifiedState(authSession);
```
Q-GDS, line 186: Same three fields were being cleared here and in the override so both call one method now.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -205,6 +205,15 @@ private APIGatewayProxyResponseEvent buildStartResponse(
                             internalSubjectId);
         }
 
+        var authenticated = isUserAuthenticatedWithValidProfile;
+        if (authenticated
+                && !reauthenticate
+                && !upliftRequired
+                && !permissionDecisionManager.canIssueAuthCode(authSession)) {
```
BeckaL: Think here would be good to have a test case for the upliftRequired cases, that prove that even if canIssueAuthCode returns false, we still return authenticated = true (with uplift = true)

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartHandlerTest.java
```diff
@@ -329,21 +333,37 @@ void considersUserAuthenticatedButLogsWhenCannotIssueAuthCode(boolean canIssueAu
                 apiRequestEventWithHeadersAndBody(
                         VALID_HEADERS, makeRequestBodyWithAuthenticatedField(true));
 
-        var result = handler.handleRequest(event, context);
+        handler.handleRequest(event, context);
 
-        StartResponse response = objectMapper.readValue(result.getBody(), StartResponse.class);
-        assertTrue(response.user().isAuthenticated());
+        verify(startService)
```
BeckaL, line 342: Oh I see why this had to assert on the start service stub. I'm not a massive fan of these assertions, it doesn't massively express what I care about here - which is the response returned, not the calls to a stub. If you have time afterwards, maybe a nice follow up would be a refactor of these tests to stop stubbing out the start service, and to allow assertions on the start response.

shared/src/main/java/uk/gov/di/authentication/shared/domain/CloudwatchMetrics.java
```diff
@@ -45,7 +45,8 @@ public enum CloudwatchMetrics {
     PASSKEY_DELETION_FAILED("PasskeyDeletionFailed"),
     PASSWORD_REHASH_COMPLETED("PasswordRehashCompleted"),
     AUTH_CODE_ISSUED("AuthCodeIssued"),
-    ENHANCED_AUTH_CODE_BLOCKED("EnhancedAuthCodeBlocked");
+    ENHANCED_AUTH_CODE_BLOCKED("EnhancedAuthCodeBlocked"),
+    AUTH_START_OVERRIDDEN_AUTHENTICATED("AuthStartOverriddenAuthenticated");
```
BeckaL: I wonder if we should be more specific here. There are other cases where we override authenticated (e.g. reauth). AuthenticatedOverridenDueToAuthCodeProtectionProblems or something more pithy?

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartHandlerTest.java
BeckaL: Or really, we can refactor the method out of start service entirely. All it does is override authenticated in the case of reauth, which we already do in some cases in the start handler.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -211,6 +208,8 @@ private APIGatewayProxyResponseEvent buildStartResponse(
                 && !upliftRequired
                 && !permissionDecisionManager.canIssueAuthCode(authSession)) {
             authenticated = false;
+            clearVerifiedState(authSession);
```
BeckaL, line 213: Worth also logging here what we're doing? (ie clearing the state like we do in the reauth case?

BeckaL (commented):
Looking great!  A few comments and a suggestion for a follow up refactor PR.

shared/src/main/java/uk/gov/di/authentication/shared/domain/CloudwatchMetrics.java
Q-GDS: Thanks renamed to AuthCodeProtectionOverride.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
Q-GDS: Thanks it now states: "Auth code protection did not pass, clearing verified state and overriding authenticated to false".

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
Q-GDS: Put it in the Uplift integration tests rather than StartHandlerTest because a session that reached Cl on a request for Cl.Cm makes uplift required and canIssueAuthCode false at the same time so it needs no stubbing and can assert on the response rather than on a call to the mocked StartService.

frontend-api/src/test/java/uk/gov/di/authentication/frontendapi/lambda/StartHandlerTest.java
Q-GDS: Agreed. We will follow this up.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -211,6 +208,8 @@ private APIGatewayProxyResponseEvent buildStartResponse(
                 && !upliftRequired
                 && !permissionDecisionManager.canIssueAuthCode(authSession)) {
```
csophocleous: Could we pull this boolean expression into a variable? Maybe `orchAndAuthDisagreeOnAuthenticated` or something?

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
```diff
@@ -211,6 +208,8 @@ private APIGatewayProxyResponseEvent buildStartResponse(
...
             authenticated = false;
+            clearVerifiedState(authSession);
+            authSessionService.updateSession(authSession);
```
csophocleous, line 214: Should this just live at the end of the handler? Is there a specific reason it's in this if block and not elsewhere?

csophocleous (commented):
Looking great! Really like the refactor of StartHandler. Some small comments to just to increase clarity :)

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
Q-GDS: It's in the if because it's only needed when the override fires. At the end of the handler it'd write the session on every Start request, on top of the addSession at 178.

Plus the reauth branch above clears the same three flags in memory and never saves them so a write at the end would start persisting that. It was spotted that's actually a bug in reauth rather than intentional but it's a separate fix.

frontend-api/src/main/java/uk/gov/di/authentication/frontendapi/lambda/StartHandler.java
Q-GDS: Done. Called it shouldOverrideAuthenticated rather than orchAndAuthDisagreeOnAuthenticated

integration-tests/src/test/java/uk/gov/di/authentication/api/StartHandlerIntegrationTest.java
```diff
@@ -224,6 +231,44 @@ void shouldReturn200AndStartResponseWithAuthenticatedFalseWhenReauthenticationIs
...
+    @Test
+    void shouldReturnAuthenticatedFalseWhenAuthCodeProtectionDoesNotPass()
+            throws Json.JsonException {
+        var sessionId = IdGenerator.generate();
+        userStore.signUp(EMAIL, "password");
+        authSessionExtension.addSession(PREVIOUS_SESSION_ID);
+        authSessionExtension.addEmailToSession(PREVIOUS_SESSION_ID, EMAIL);
+        authSessionExtension.addAchievedCredentialTrustToSession(PREVIOUS_SESSION_ID, null);
+        authSessionExtension.addHasVerifiedWithPasswordAndMfaToSession(
+                PREVIOUS_SESSION_ID, false, false);
```
Q-GDS, line 243: Flags are false now so it no longer looks like they agree. Also set the credential strength to null explicitly, since that's what canIssueAuthCode blocks on first, before it reads any flag. It doesn't change anything the field was already null but I'd rather it was stated than left as an absence.

Renamed it and dropped the wipe assertions StartHandlerTest covers the clearing already.

csophocleous (dismissed):
LGTM - great stuff on the refactor!

csophocleous (approved):

PR #8880, opened 2026-09-11
AUT-5821: Add lastSignedIn to IAD export timestamp candidates

sw-mattw:
## What

Since the IAD export Lambda was developed, the `lastSignedIn` attribute has been added to the user profile, this should be considered alongside the other timestamps when determining the initial* last active date of the account.

[* Initial as home will need to replay the audit events on top of this data whenever this is run (rarely).]

## How to review

1. Code Review

## Checklist

### DIFF
--- a/utils/src/test/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelperTest.java
+++ b/utils/src/test/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelperTest.java
@@ -476,6 +476,107 @@ class InactiveAccountDataExportHelperTest {
     }
 
     @Test
+    void calculateLastActiveDateShouldReturnLastSignedInWhenMostRecent() {
+        Map<String, AttributeValue> userProfileItem =
+                Map.of(
+                        UserProfile.ATTRIBUTE_CREATED,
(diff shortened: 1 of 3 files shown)


utils/src/test/java/uk/gov/di/authentication/utils/helpers/InactiveAccountDataExportHelperTest.java
```diff
@@ -475,6 +475,107 @@ void calculateLastActiveDateShouldHandleOnlyCredentialsItemProvided() {
...
+                        AttributeValue.builder().s("2024-09-15T11:30:00.123456").build());
+
+        LastActiveDate result = calculateLastActiveDate(userProfileItem, null);
+
+        assertEquals("2024-09-15T11:30:00.123456", result.timestamp());
+        assertEquals("UserProfile.LastSignedIn", result.source());
+    }
+
+    @Test
+    void calculateLastActiveDateShouldHandleLastSignedInNotPresent() {
```
sw-mattw, line 549: Added an explicit test for this case as the vast majority of user profiles at present won't have lastSignedIn set. The intention is to backfill this, but there is the possibility that the export may be rerun prior to this.

Q-GDS (approved):

PR #8885, opened 2026-09-14
BAU: Persist reauth verified state reset

Q-GDS:
## What

On a reauthentication journey `StartHandler` clears `hasVerifiedWithPassword`, `hasVerifiedWithMfa` and `hasVerifiedWithPasskey`, but it did that after the only write to the session. The reset never left the Lambda, so the stored session kept claiming whatever it claimed before the reauth started.

The fix moves the clear above the write so the reset is saved.

## How to review

1. Code review.
1. The integration test is the proof of the fix. To see it fail, move `authSessionService.addSession(...)` back above the reauth block and run `./gradlew :integration-tests:test --tests "uk.gov.di.authentication.api.StartHandlerIntegrationTest"`.

## Checklist

## Related PRs

- #8877, the PR review surfaced this.


csophocleous (approved):
LGTM

PR #8901, opened 2026-09-16
BAU: Make auth session integration test use auth session service not extension

BeckaL:
The auth session service integration tests were a bit odd, as they acted on the auth session extension, not directly on the auth session service.

This was a confusing bit of indirection - the auth session extension normally proxied straight through to the equivalent methods on the auth service service, but this is confusing and non transparent testing.

Probably the auth session extension should not contain any proxy methods at all, but for now I've limited changes to just making the tests for the auth session service less confusing.

I've done this by exposing the auth session service from the auth session extension. This was the easiest way of doing this as it makes use of all the test setup in the auth session extension for creating the service. A nicer way in future would be to pass the auth session service into the extension constructor. But this a first step.


sw-mattw (approved):
Nice cleanup, tests read clearer now, thanks

